PR #13635, opened 2020-09-21
[magnum] add `angle` feature to support build with ANGLE

linuxaged:
we can now build magnum with ANGLE: `vcpkg install --head corrade[*] magnum[sdl2application,glfwapplication,angle]`
have tested on Win64 and passed.


NancyLi1013 (commented):
Could you please also update the version in `CONTROL `file?
```
Version: 2020.06
Port-Version: 1
```

NancyLi1013:
Have you tested the feature? Can those work fine?

linuxaged:
> Have you tested the feature? Can those work fine?

ANGLE support is supposed to be Windows only currently, have tested on Windows and works as expected.
How to make the `angle` feature in CONTROL platform constrained? this seems can be done easily in new manifest `vcpkg.json`
We have some discussion [here](https://github.com/mosra/magnum/pull/470).

Squareys:
@linuxaged Huh, this is interesting, specifying the target as feature feels a little weird 🤔 
There are a couple of issues I see with this:
 - If you install `magnum[*]` you will unexpectedly build for GLES instead of desktop GL
 - Assuming android or iOS triplets, `angle` would mean `gles3`, which doesn't feel right
 - Assuming emscripten triplets, you would be forcing `WebGL2`, but that is not supported everywhere, so using WebGL1 might be preferred.

There should definitely be a way to specify this (and the GLES version) while building _somehow_. That would make this a lot more flexible.

As I use these packages a lot myself, I can imagine a use case, where I want both Angle and Desktop versions installed.
With vcpkg that would therefor map well to a triplet properly.

`vcpkg install magnum[*]:x64-windows-gles3` vs `vcpkg install magnum[*]:x64-windows`.

linuxaged:
> * If you install `magnum[*]` you will unexpectedly build for GLES instead of desktop GL
> * Assuming android or iOS triplets, `angle` would mean explicitly setting the GLES version to 3, which doesn't feel right
> * Assuming emscripten triplets, you would be forcing `WebGL2`, but that is not supported everywhere, so using WebGL1 might be preferred.

before, i thought these could be solved with `vcpkg.json`:
```
    "angle-gles2": {
      "description": "use GLES2 on desktop with ANGLE",
      "dependencies": [
        "angle"
      ],
      "platform": "windows"
    },
    "angle-gles3": {
      "description": "use GLES3 on desktop with ANGLE",
      "dependencies": [
        "angle"
      ],
      "platform": "windows"
    },
```
but unfortunately the field 'platform' is not supported in `features`

> There should definitely be a way to specify this (and the GLES version) while building _somehow_. That would make this a lot more flexible.

we could separate the `angle` feature into `angle-gles2` and `angle-gles3`, but using `feature` is not a good idea as mentioned above.
 
> I want both Angle and Desktop versions installed.
> `vcpkg install magnum[*]:x64-windows-gles3` vs `vcpkg install magnum[*]:x64-windows`.

If so, how does the building tools know which one to build target?

Summary: this patch is just a quick hack to get Magnum work with ANGLE easily on Windows. More discussion and work need to be done make it acceptable to all users!

NancyLi1013:
Hi @linuxaged 

Sorry for the long delay.

In fact, feature has supported for `platform `field in vcpkg.json. 
It seems that the format is not correct in your vcpkg.json file.

It should be like this:

 ```
"features": {
  "angle-gles2": {
      "description": "use GLES2 on desktop with ANGLE",
      "dependencies": [
         {
           "name" "angle",
           "platform": "windows"
         }
      ],
    },
    "angle-gles3": {
      "description": "use GLES3 on desktop with ANGLE",
      "dependencies": [
        {
          "angle"
          "platform": "windows"
        }
      ],
    },
}
```

Please refer to this port:
https://github.com/microsoft/vcpkg/blob/master/ports/kenlm/vcpkg.json

So would like to try this again?

**Comment by NancyLi1013** [comment]
@linuxaged 

I noticed you have reverted `angle `feature. 

So in this PR, you just update `CONTROL` file as` vcpkg.json` file and add these two options  `${GLES_OPTION}`
        `${GLES2_OPTION}` in `vcpkg_configure_cmake()`. Is this right?

**Comment by linuxaged** [comment]
convert `CONTROL` into `vcpkg.json` is step zero, i'm still working on it && doing some testings locally. Have not pushed to remote yet.

**Comment by NancyLi1013** [comment]
Please let me know if this PR is ready for review now. If not, you can convert it to draft.

**Comment by linuxaged** [comment]
blocked by `EGL_NO_DISPLAY` error. 
ANGLE(gles2) + Magnum seems not work on both macOS and Windows:

```
INFO: egl::Debug::insertMessage(462): EGL ERROR: eglQueryString: display is EGL_NO_DISPLAY.
Renderer: ANGLE (NVIDIA GeForce GTX 1050 Ti Direct3D11 vs_5_0 ps_5_0) by Google Inc.
OpenGL version: OpenGL ES 3.0.0 (ANGLE 2.1.0.invalid-hash)
Using optional features:
    GL_ANGLE_framebuffer_blit
    GL_ANGLE_framebuffer_multisample
    GL_ANGLE_instanced_arrays
    GL_EXT_discard_framebuffer
    GL_EXT_draw_buffers
    GL_EXT_robustness
    GL_EXT_texture_filter_anisotropic
    GL_EXT_texture_storage
    GL_KHR_debug
    GL_OES_vertex_array_object
INFO: gl::Debug::insertMessage(183): GL error: HIGH: Uniform size does not match uniform method.
```
I'm using the Magnum built with `TARGET_DESKTOP_GLES` which does not depend on ANGLE.

NancyLi1013:
Please ping me if this PR is ready for review @linuxaged.

NancyLi1013:
@linuxaged 

linuxaged:
Magnum can be built with `TARGET_DESKTOP_GLES`, this is the simpler way to use Magnum with GLES on desktop.
This PR could be closed.

NancyLi1013:
Thanks for your feedback @linuxaged.

PR #13644, opened 2020-09-21
Update the latest g++-10 version if exists

cngzhnp:
The latest g++/gcc version(10.0) has been already released for different Linux distros, so it would be great to use it if exists on the environment.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -219,7 +219,9 @@ fetchTool()
 selectCXX()
 {
     if [ "x$CXX" = "x" ]; then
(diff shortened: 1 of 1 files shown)


jmdejoanelli:
Homebrew installs GCC 10.2.0 by default, so following the README quickstart for Unix on MacOS, `bootstrap-vcpkg.sh` fails. 

This PR fixed my issue of not being able to build vcpkg on Majave 10.14.6 with Xcode 10.2.1 and clang 10.0.1.

ras0219-msft (approved):
This LGTM, thanks for the PR!

JackBoosY:
Blocked by #13722.

PR #13645, opened 2020-09-22
[polyhook2] Update to new version

stevemk14ebr:
Updates polyhook2 to latest. Supports the triples x86-windows, x86-windows-static, x64-windows, x64-windows-static. PR tested locally on my system and all tests pass.

### DIFF
--- a/ports/polyhook2/portfile.cmake
+++ b/ports/polyhook2/portfile.cmake
@@ -3,8 +3,8 @@ vcpkg_fail_port_install(ON_ARCH "arm" "arm64" ON_TARGET "Linux" "OSX" "UWP")
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO stevemk14ebr/PolyHook_2_0
-    REF  b3c0baf639614c4b4ca56e5df2e9da688986917b
-    SHA512 a1fb531fcbd7d6ac084efc7a67b7a1d9e661f0e770df02db7458a311c54ca1a1f11b8be0ae438a40259603414bfb2e64bab69d45754bc910363daa912e995550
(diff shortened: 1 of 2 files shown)


ports/polyhook2/portfile.cmake
```diff
@@ -18,12 +18,6 @@ vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     virtuals  POLYHOOK_FEATURE_VIRTUALS
 )
 
-if (VCPKG_LIBRARY_LINKAGE STREQUAL "static")
-    set(BUILD_SHARED_LIB OFF)
-else()
-    set(BUILD_SHARED_LIB ON)
-endif()
```
ras0219-msft, line 25: Could you also remove `-DPOLYHOOK_BUILD_SHARED_LIB=${BUILD_SHARED_LIB}` from the `vcpkg_configure_cmake()` directive below?

ras0219-msft (changes requested):
LGTM once the additional line is removed! Thanks for the PR!

ports/polyhook2/portfile.cmake
LilyWangL: The files in source are using `-DPOLYHOOK_BUILD_SHARED_LIB`, it can't be removed.

LilyWangL (approved):

ports/polyhook2/portfile.cmake
ras0219-msft: Then we cannot remove this `if/else/endif` block because it is defining `POLYHOOK_BUILD_SHARED_LIB`.

LilyWangL: Thanks for your reply. I got it.

LilyWangL:
Blocked by #13722.

PR #13659, opened 2020-09-22
[protobuf] Add vcpkg-cmake-wrapper.cmake

### LINKED ISSUE #13543: opencv_dnn link error
**Describe the bug**
Using the `opencv_dnn` module of the opencv package, with the `x64-windows-static` target triplet, results in a fatal linking error (LNK1120) when building using MSVC.
```
main.exe : fatal error LNK1120: 7 unresolved externals
```

**Environment**
- OS: Windows 10 64-bit
- Compiler: MSVC 19.27.29111.0
- vcpkg: f1bef4aa7ca7e2a6ea4f5dfe4850d95fce60b431

**To Reproduce**
1.  `.\vcpkg.exe install opencv:x64-windows-static`
2.  `mkdir dnn_project`
3.  Create CMakeLists.txt, and main.cpp (see below)
4.  Run the next commands in the x64 Native Tools Command Prompt for VS 2019
5.  `mkdir cmake-build-debug && cd cmake-build-debug`
6.  `cmake .. -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static -GNinja`
7.  `ninja`

**Expected behavior**
The project should compile and link successfully.

**Failure logs** 
-(please attached failure logs)
[error.log](https://github.com/microsoft/vcpkg/files/5225210/error.log)

**Additional context**
- main.cpp (I know that the below code results in a failed assertion at runtime, that's not the problem)
```
#include <iostream>
#include <opencv2/opencv.hpp>

int main()
{
	std::cout << "Hello, world!" << std::endl;
	cv::dnn::Net n;
	std::cout << n.dump() << std::endl;
}
```

- CMakeLists.txt
```
cmake_minimum_required(VERSION 3.0)
project(opencv_dnn_linkage_error)

find_package(OpenCV CONFIG REQUIRED)

cmake_policy(SET CMP0054 NEW)
if (POLICY CMP0091)
	cmake_policy(SET CMP0091 NEW)
endif()
# Statically link against the CRT (/MTd for Debug, else /MT)
set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MT")
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MTd")
add_compile_options(
		/permissive-
)

add_executable(main main.cpp)
target_link_libraries(main PRIVATE opencv_dnn)
```
^^ The above CMakeLists.txt can be reduced to the following without changing the error;
```
cmake_minimum_required(VERSION 3.0)
project(opencv_dnn_linkage_error)

find_package(OpenCV CONFIG REQUIRED)

add_executable(main main.cpp)
target_link_libraries(main PRIVATE opencv_dnn)
```

**Workarounds**
Building the OpenCV package with an older version of CMake seems to fix the issue.
E.g. you can use the 2020.04.zip vcpkg release (which downloads and uses CMake 3.14), and it works fine.
(If you have a later version of CMake installed on your system, vcpkg uses that while building OpenCV and you're stuck with the same problem again)

I've found the following CMake versions seems to work with the 2020.04.zip vcpkg release (newer releases require 3.17, which causes the linking error)
```
3.14.7  -- ✓
3.15.7  -- ✓
3.16.8  -- ✓
3.17.4  -- ✗
3.18.2  -- ✗
```

using the `x64-windows` triplet, instead of the `x64-windows-static` triplet seems to also work fine.
Also note that the any version of CMake seems to work when building the project in main.cpp - it only matters which version was used to build the OpenCV package.

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/13543

The FindProtobuf.cmake from cmake module need set'Protobuf_USE_STATIC_LIBS' to ON to force use of the static libs, however, this is not defined in opencv, so 'PROTOBUF_USE_DLLS' defined in static link, that cause the issue.

The protubuf already set to use the vcpkg installed one in opencv4, add vcpkg-cmake-wrapper.cmake for protobuf to set the Protobuf_USE_STATIC_LIBS to ON when find the package via Module mode.


Neumann-A:
> The FindProtobuf.cmake from cmake module need set'Protobuf_USE_STATIC_LIBS' to ON to force use of the static libs, however, this is not defined in opencv, so 'PROTOBUF_USE_DLLS' defined in static link, that cause the issue.

Directly says this does not need a change in opencv but in protobuf.
Namely protobuf seems to need a `vcpkg-cmake-wrapper.cmake` to take `VCPKG_LIBRARY_LINKAGE` into account and set the appropiate define. 

Furthermore it should be checked wether or not `${name}_LIBRARY_RELEASE` and `${name}_LIBRARY_DEBUG` are always correctly set. (with name being Protobuf;Protobuf_LITE;Protobuf_PROTOC). If not these also requires correction

PhoebeHui:
@Neumann-A, thanks for your review! 

The vcpkg-cmake-wrapper for protobuf has been added, and the ${name}_LIBRARY_RELEASE and ${name}_LIBRARY_DEBUG are correctly set in FindProtobuf.cmake.

JackBoosY (approved):

ras0219-msft:
LGTM, thanks everyone!

PR #13668, opened 2020-09-22
[polyclipping] Fix cmake patch for finding include file

FabienPean:
Usage for port polyclipping given by vcpkg is incorrect. Currently it says
```cmake
find_package(polyclipping CONFIG REQUIRED)
target_link_libraries(main PRIVATE polyclipping::polyclipping)
```
However the include file needed to use it cannot be found.
This PR fixes the cmake patch so that the include file can be found by the end user as `#include<polyclipping/clipper.hpp>`

Note that besides the config export there is a findCLIPPER.cmake installed with the library. Alternatively to this fix, a usage file could be created with the following:
```cmake
find_package(CLIPPER CONFIG REQUIRED)
target_include_directories(main PRIVATE ${CLIPPER_INCLUDE_DIRS})
target_link_libraries(main PRIVATE ${CLIPPER_LIBRARIES})
```
It depends on how you would prefer to resolve the issue.

### DIFF
--- a/ports/polyclipping/CONTROL
+++ b/ports/polyclipping/CONTROL
@@ -1,5 +1,5 @@
 Source: polyclipping
 Version: 6.4.2
-Port-Version: 5
+Port-Version: 6
 Homepage: https://sourceforge.net/projects/polyclipping/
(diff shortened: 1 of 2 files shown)


ras0219-msft:
This looks like the right fix to me, thanks!

We prefer promoting target-based CMake as much as possible.

PR #13681, opened 2020-09-23
[libtins] Update to version 4.3

cnbatch:
Update `libtins` to version 4.3 and enables C++11 support.

### DIFF
--- a/ports/libtins/portfile.cmake
+++ b/ports/libtins/portfile.cmake
@@ -3,8 +3,8 @@ include(vcpkg_common_functions)
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO mfontanini/libtins
-    REF v4.2
-    SHA512 46d07712604c780e418135c996f195046fd85a9e1411962c9bcee3c8d0fc64f494aa50164236ffd1e77ff8a398e9617bbf040b3e01a5771c5621c0faa1ce312f
+    REF v4.3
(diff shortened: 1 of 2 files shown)


ports/libtins/portfile.cmake
```diff
@@ -21,6 +21,7 @@ vcpkg_configure_cmake(
     OPTIONS
         -DLIBTINS_BUILD_SHARED=${LIBTINS_BUILD_SHARED}
         -DLIBTINS_ENABLE_PCAP=${ENABLE_PCAP}
+        -DLIBTINS_ENABLE_CXX11=1
```
JackBoosY, line 24: This requires a judgment, if the current compiler does not support c11, we should not add this parameter.
See https://github.com/microsoft/vcpkg/pull/12612/commits/a11b055c3a1989ee87b1f4a4ee5046d37d1fc105

JackBoosY (changes requested):

ports/libtins/portfile.cmake
cnbatch: It's not C11.

cnbatch: As I mentioned in comment, this option is for C++11 (ISO/IEC 14882:2011) only. If a compiler supports vcpkg (Visual Studio 2015 Update 3 or greater / g++ >= 6), it's also supports this option too.
C11 (ISO/IEC 9899:2011) is irrelevant to this option here.

cnbatch (commented):
C++11 is not C11. They're not the same.

ras0219-msft:
LGTM, thanks for the PR!

PR #13687, opened 2020-09-23
[usd] fix dll path and update to 20.08

ousttrue:
**Describe the pull request**
- fix dll path in cmake for find_package
- fix dll path in for runtime plug load
- version up to 20.08(usd-20.05 fixed release build issue: https://github.com/PixarAnimationStudios/USD/issues/960)
- add usage( find_package(usd) => find_package(pxr) )

The following Debug and Release work fine.

```c++
#define NOMINMAX
#include <pxr/usd/usd/stage.h>
#include <pxr/usd/sdf/path.h>
#include <pxr/usd/usdGeom/sphere.h>

int main()
{
    auto stage = pxr::UsdStage::CreateInMemory();
    auto sphere = pxr::UsdGeomSphere::Define(stage, pxr::SdfPath("/TestSphere"));
    auto root = stage->GetRootLayer();
    root->Export("test_sphere.usda");
    return 0;
}
```


LilyWangL:
Thanks for your PR. Now `usd` build failed on Linux and MACOS with the following error:
```
CMake Error at /mnt/vcpkg-ci/downloads/tools/cmake-3.17.2-linux/cmake-3.17.2-Linux-x86_64/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:164 (message):
  Could NOT find TBB (missing: tbb) (found version "2020.1")
  /mnt/vcpkg-ci/downloads/tools/cmake-3.17.2-linux/cmake-3.17.2-Linux-x86_64/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:445 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindTBB.cmake:277 (find_package_handle_standard_args)
  /agent/_work/1/s/scripts/buildsystems/vcpkg.cmake:493 (_find_package)
  cmake/defaults/Packages.cmake:112 (find_package)
  CMakeLists.txt:23 (include)
```

ousttrue:
All green 👍 
TBB handling is difficult.

ports/usd/usage
```diff
@@ -0,0 +1,5 @@
+The package usd:x64-windows provides CMake targets:
+
+    find_package(pxr CONFIG REQUIRED)
+    # Note: 22 target(s) were omitted.
+    target_link_libraries(main PRIVATE ar gf js tf)
```
ras0219-msft, line 5: How does this usage differ from the generated usage?

If it is listing out all required targets, we should remove the comment about targets being omitted.

ports/usd/portfile.cmake
```diff
@@ -6,17 +6,22 @@ SET(VCPKG_POLICY_EMPTY_PACKAGE enabled)
...
-        fix-build-error.patch
+        fix_build-location.patch
 )
 
 vcpkg_find_acquire_program(PYTHON2)
 get_filename_component(PYTHON2_DIR "${PYTHON2}" DIRECTORY)
 vcpkg_add_to_path("${PYTHON2_DIR}")
 
+IF (WIN32)
+ELSE()
```
ras0219-msft, line 21: Removing source files based on platform is problematic; if it is safe to remove it on Windows, it would be preferable to always do so.

Otherwise, we must use `VCPKG_TARGET_IS_WINDOWS` instead of `WIN32`.

ports/usd/portfile.cmake
```diff
@@ -46,14 +54,34 @@ vcpkg_copy_pdbs()
...
 file(GLOB DEBUG_DLL ${CURRENT_PACKAGES_DIR}/debug/lib/*.dll)
+file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/debug/bin)
 foreach(CURRENT_FROM ${RELEASE_DLL} ${DEBUG_DLL})
     string(REPLACE "/lib/" "/bin/" CURRENT_TO ${CURRENT_FROM})
     file(RENAME ${CURRENT_FROM} ${CURRENT_TO})
 endforeach()
+
+file(COPY ${CMAKE_CURRENT_LIST_DIR}/usage DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT})
+
+function(file_replace_regex filename match_string replace_string)
```
ras0219-msft, line 71: Instead of these fixups, `vcpkg_fixup_cmake_targets()` should probably be supplied with a `TARGET_PATH` of `share/pxr`.

ras0219-msft (changes requested):
Thanks for the PR!

ports/usd/portfile.cmake
ousttrue: I fixed

ports/usd/usage
ousttrue: I fixed. share/pxr

ports/usd/portfile.cmake
ousttrue: I fixed. use VCPKG_TARGET_IS_WINDOWS

ousttrue:
cmake TARGET_PATH fixed.
CI x86_windows #13722

LilyWangL (approved):

ports/usd/portfile.cmake
dan-shaw: Do we need the function `file_replace_regex` anymore? The change to `vcpkg_fixup_cmake_targets` should fix this.

LilyWangL: Pinging @ousttrue for response.

LilyWangL: I test this locally, function `file_replace_regex` is needed.

strega-nil:
Gonna let this run through CI after merging with master, then merge :)

PR #13713, opened 2020-09-24
[vcpkg/manifest/toolchain] Add install/bootstrap options & fix CACHE variables usage

Neumann-A:
cc @strega-nil @ras0219


Neumann-A:
For fun just try running:
```
set(MY_MESSAGE "Reconfiguring!")
set(MY_MESSAGE "First configure!" CACHE STRING "Message to print!")
message(STATUS "MY_MESSAGE:${MY_MESSAGE}")
```
and marvel at the output ;)

strega-nil (approved):

strega-nil:
@Neumann-A thanks I hate it haha

Neumann-A:
@strega-nil The fun part is that it is actual documented:
> If the cache entry does not exist prior to the call or the FORCE option is given then the cache entry will be set to the given value. Furthermore, any normal variable binding in the current scope will be removed to expose the newly cached value to any immediately following evaluation.

(but really that last sentence should be big and in red)

Issue #13548, opened 2020-09-15
vcpkg_copy_pdbs() does not copy PDB files for static libraries


alek-d:
****

**Is your feature request related to a problem? Please describe.**
After I added static linking to Scintialla port, I noticed that PDB file is not copied alongside LIB file, like it is with DLL file in case of dynamic linkage.

**Proposed solution**
`vcpkg_copy_pdbs.cmake` uses `dumpbin` to find the path to PDB file for the specified DLL, but this method doesn't work for LIB files.

In case of static libraries, `vcpkg_copy_pdbs.cmake` should try to find PDB file with the same name as the library, and copy it alongside LIB file.

**Additional context**
I'm not sure if this will work for all libraries, but I think that we can assume that the pattern `lib_name.lib` -> `lib_name.pdb` in the same directory, is the most common.

PhoebeHui:
Related to https://github.com/microsoft/vcpkg/issues/10564

alek-d:
@PhoebeHui I only briefly checked, and I'm not 100% sure, but it looks to me that in #10564 generating PDBs is not set in compilation settings. So, probably, there is nothing to copy there.

NancyLi1013:
Considering the discussion in #13721, there is no need to copy PDB files for static libraries. So I think this issue can be closed now.

PR #13721, opened 2020-09-24
Copy pdb for static

alek-d:
**Describe the pull request**

- What does your PR fix?
Fixes "vcpkg_copy_pdbs() does not copy PDB files for static libraries" #13548

All `*-windows-*` triplets where tested successfully. 

I hope so.

This PR is my attempt to solve the problem described in #13548.

It works only for static libraries which are compiled with debug symbols in PDB format. Of course.
In most cases PDB files are generated in the same folder as LIB file and also with the same same.
So I assumed that at least for some libraries this solution will work.

I tested it with Scintilla static build, and in this case it works perfectly.
I couldn't find any other library which is configured to produce PDBs, but I tried only a few libs.

As you can see, to make it work, I also needed to modify `vcpkg_install_msbuild.cmake` script.
I did it for two reasons:

- `vcpkg_install_msbuild.cmake` passes a variable called `LIBS` to `vcpkg_copy_pdbs.cmake` with the path/s to currently compiled LIB file/s. So you don't have to search for LIB files in a build tree, and as it happens `LIBS` contain the right path to just compiled LIB file.

- Looks like there is no tool like `dumpbin` which will allow us to extract the path to the PDB file from the LIB file. So searching for LIBs in a similar way as it's done for DLLs – in `CURRENT_PACKAGES_DIR` – would not work, because we won't be able to find the path for specified PDB. There is a path to PDB embedded in LIB file, but I couldn't find a method to extract it as it can be done using `dumpbin`. Maybe someone knows such a tool, or maybe it would be worth to create one for this purpose.

Ideally it would be nice if some higher level build machinery (cmake?) will always copy PDB alongside LIB like it does with LIBs for DLLs. I'm new to `vcpkg` internals, so maybe I'm missing some obvious solution.

Summary:

- It works only for PDB with the same name as LIB, and when they are located in the same directory. Which is BTW default `MSVC` setting for PDBs.

- It only works when called from `vcpkg_install_msbuild.cmake`, or called with the same set of needed parameters, which are `BUILD_TYPE` and `LIBS` - which is somehow passed to `vcpkg_copy_pdbs.cmake`. How it works, that some parameters are passed from script to script, and some are not, it is still a mystery to me, but I'm new to `cmake` as well.

- Modifications done to  `vcpkg_install_msbuild.cmake` are not affecting, "normal" use of  `vcpkg_copy_pdbs.cmake`. As you can see in the diff, they only work when `BUILD_TYPE` parameter is set and at the same time `VCPKG_LIBRARY_LINKAGE` is static.


JackBoosY:
I think we don't need pdbs when building static.

Hoikas:
PDBs for static library dependencies are incredibly useful when debugging my own applications.

alek-d:
@JackBoosY please just google "pdb for static" or similar. PDBs are very useful not only with dynamically linked libraries even if you don't plan to debug those libs. Lack of PDB files for 3-rd party libs can generate a lot of annoying warnings.

@Hoikas for sure this isn't the ultimate solution. It just works for those libraries which are using default settings. When you turn on generating debug information for your library in MSVC properties dialog, the default output parameter for PDB file is `$(IntDir)$(ProjectName).pdb` which translates to `library_name.lib -> library_name.pdb` in the same output directory. So, in my opinion this is quite reliable solution.

Hoikas:
Right - I'm just thinking of some specific cases that I'm aware of where this might not work as expected. IIRC in python3, the pythoncore project produces python38.lib and python38_d.lib... but I seem to recall some problems with the pdb being misnamed. But that's based off potentially faulty recollections. I am very glad to see that someone is looking into this :)

ras0219:
Note that we pass /Z7 by default, which results in the debugging information being embedded in the object files (thus inside the `.lib` for static libs).

PR #13556, opened 2020-09-16
[python3] Improve support for embedding and add tools

Hoikas:
This pull request improves the downstream experience when working with the python3 port by building a functioning interpreter with support for pip and improving support when cmake's FindPython3 finder is used. Previously, when using FindPython3, if both the `DEVELOPMENT` and `INTERPRETER` components were requested, you would have to have a local python install matching the vcpkg python. Further, on Windows, this module preferred registry python over any slightly misconfigured python artifacts. Fixing these issues allows us to simplify ports whose cmake build systems need the python3 development libraries.

Anyone who intends to do further work on the python3 port after this might consider fixing `vcpkg_find_acquire_program(PYTHON3)` to use the interpreter installed by this port.

Note that on Windows a patch is applied in static python interpreters to disable extension modules-due to the way DLLs function, it is not possible to load them without crashing the interpreter. Further, support for the static CRT has been removed both to avoid surprises in that regard and due to the nature of `PyObject`s.

- What does your PR fix?
This should address #9026 and #10903.

- Which triplets are supported/not supported? Have you updated the CI baseline?
Support for x64-windows-static has been dropped.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
I think so 😄

PR #13727, opened 2020-09-25
Python3 offer tool feature

marekr:
This provides the feature for the python3 port to build the executable and make a working python install inside /tools.

Why?
There is a build process I have that requires the python executable to match the python library version being built against, I also need the python modules that don't exist in the python-embed packages that vcpkg_acquire_program retrieves.

This only adds the tool for Windows.

Now I expect problems and have questions for the vcpkg maintainers....

1. Python3 has important modules that used external libraries.

Currently it wants these exact versions of libraries which are tracked by in source script get_externals.bat
```
bzip2-1.0.6
libffi-3.3.0-rc0-r1
openssl-1.1.1f
 sqlite-3.31.1.0
tcl-core-8.6.9.0
tk-8.6.9.0
tix-8.4.3.6
xz-5.2.2 
zlib-1.2.11
```

of which it builds directly with the module source (not static or shared) and others it links shared. 

While I understand vcpkg policy is to use the split out ports. I do not think it's the best plan for python which calls out ~explicit~ versions which the vcpkg port system cannot respect as far I know.
 These shared libraries only get placed in the /share/python3/DLLs folder next to the /share/python3/Lib folder and not /bin which is the standard placement for these shared modules on windows python.

2. I made it install the core python modules in both /share/python3/DLLs and /tools/python3/DLLs. python.exe could be made to use /share/python3/DLLs in theory manually by users invoke it and setting to /share/python3

### DIFF
--- a/ports/python3/CONTROL
+++ b/ports/python3/CONTROL
@@ -7,3 +7,6 @@ Build-Depends: libffi, openssl, zlib (!uwp&!windows)
 
 Feature: enable-shared
 Description: Build shared libraries in addition to static ones built by default
+
+Feature: tools
+Description: Build python executable
--- a/ports/python3/portfile.cmake
+++ b/ports/python3/portfile.cmake
@@ -53,9 +53,15 @@ if (VCPKG_TARGET_IS_WINDOWS)
 		message(FATAL_ERROR "Unsupported architecture: ${VCPKG_TARGET_ARCHITECTURE}")
 	endif()
 
+	vcpkg_execute_build_process( 
+		COMMAND ${SOURCE_PATH}/PCBuild/get_externals.bat 
+		WORKING_DIRECTORY ${SOURCE_PATH}/PCBuild 
+	   )
+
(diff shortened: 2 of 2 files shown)


ports/python3/CONTROL
```diff
@@ -7,3 +7,6 @@ Build-Depends: libffi, openssl, zlib (!uwp&!windows)
 
 Feature: enable-shared
 Description: Build shared libraries in addition to static ones built by default
+
```
NancyLi1013, line 10: Could you please update the `Port-Version`?
`Port-Version: 2`

NancyLi1013 (changes requested):

Hoikas:
Duplicates part of #13556.

marekr:
Defering to the other PR that's better and hopefully gets merged to make the python3 port usable ;)

Hoikas:
I'm glad to see that I wasn't the only one frustrated by some of the aspects of the python3 port :smile:

PR #13732, opened 2020-09-25
[vcpkg_build_cmake] Set PARALLEL_ARG when using Ninja

### LINKED ISSUE #10631: Ninja invocation should respect  VCPKG_MAX_CONCURRENCY
**Problem**

Currently, ninja invocation by vcpkg does not seem to respect VCPKG_MAX_CONCURRENCY. As a result, ninja will use it's default parallelism, which can result in out of memory error if the RAM/core ratio is low, e.g., on my 6 core system with 32GB RAM, ninja defaults to 14 parallel jobs and run out of memory when compiling PCL from the master branch. When that happens , vcpkg prints "Restarting Build without parallelism because memory exceeded" and restarts the build with **NO** parallelism, which is very SLOW. 

**Proposed solution**

It would be much better, if vcpkg invokes ninja with `-j ` followed by value specified by VCPKG_MAX_CONCURRENCY. 

**Alternatives**
Ninja doesn't allow the number of max parallel jobs to be controlled by an environment variable.

JackBoosY:
Fixes #10631.

### DIFF
--- a/scripts/cmake/vcpkg_build_cmake.cmake
+++ b/scripts/cmake/vcpkg_build_cmake.cmake
@@ -41,6 +41,11 @@ function(vcpkg_build_cmake)
 
     if(_VCPKG_CMAKE_GENERATOR MATCHES "Ninja")
         set(BUILD_ARGS "-v") # verbose output
+        if (VCPKG_MAX_CONCURRENCY)
+            set(PARALLEL_ARG "-j${VCPKG_MAX_CONCURRENCY}")
+        else()
+            set(PARALLEL_ARG "-j${VCPKG_CONCURRENCY}")
+        endif()
         set(NO_PARALLEL_ARG "-j1")
     elseif(_VCPKG_CMAKE_GENERATOR MATCHES "Visual Studio")
         set(BUILD_ARGS
--- a/scripts/cmake/vcpkg_build_cmake.cmake
+++ b/scripts/cmake/vcpkg_build_cmake.cmake
@@ -41,11 +41,7 @@ function(vcpkg_build_cmake)
 
     if(_VCPKG_CMAKE_GENERATOR MATCHES "Ninja")
         set(BUILD_ARGS "-v") # verbose output
-        if (VCPKG_MAX_CONCURRENCY)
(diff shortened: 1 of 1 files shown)


scripts/cmake/vcpkg_build_cmake.cmake
```diff
@@ -41,6 +41,11 @@ function(vcpkg_build_cmake)
 
     if(_VCPKG_CMAKE_GENERATOR MATCHES "Ninja")
         set(BUILD_ARGS "-v") # verbose output
+        if (VCPKG_MAX_CONCURRENCY)
+            set(PARALLEL_ARG "-j${VCPKG_MAX_CONCURRENCY}")
+        else()
+            set(PARALLEL_ARG "-j${VCPKG_CONCURRENCY}")
+        endif()
```
Neumann-A, line 48: shouldn't `VCPKG_CONCURRENCY` be enough? as far as i remember `VCPKG_MAX_CONCURRENCY` is a env variable used to set that value of `VCPKG_CONCURRENCY`

JackBoosY: @Neumann-A Nope:
https://github.com/microsoft/vcpkg/blob/5d331ed5538b36d040e90fdd96a70fa606fe597a/toolsrc/src/vcpkg/build.cpp#L432
Only set it here.

Neumann-A: Yeah but that sets `VCPKG_CONCURRENCY` so you do not need to use `VCPKG_MAX_CONCURRENCY`. `VCPKG_MAX_CONCURRENCY` is an environment variable and not a normal cmake variable!

JackBoosY: @Neumann-A Hmm.. make sence.

JackBoosY, line 48:
```suggestion
        set(PARALLEL_ARG "-j${VCPKG_CONCURRENCY}")
```

ras0219 (approved):

ras0219-msft:
LGTM

Issue #13027, opened 2020-08-20
Discourage use of unofficial:: targets if a CMake findModule is available.


Neumann-A:
`unofficial::` targets are just an anoyance if there are CMake findModules available. The usage of the `unofficial::` is probably done because the corresponding findModules was not doing the right thing. So instead of fixing that a shortcut is used to get the desired behavior which basically requires patching a lot of downstream projects. 
As such the use of `unofficial::` targets should be discouraged and fixing the findModule via a `vcpkg-cmake-wrapper.cmake` should be done. 

See the changeset of #11776 why using `unofficial::` targets is bad.

JackBoosY:
I think it is not encouraged to **directly** use `unofficial::` better.

`FindABC.cmake` provided by cmake is not necessarily suitable for the port built in vcpkg, because as the port is updated, its header file name and library name may **change**, which is why cmake has always maintained `FindABC.cmake`. Unfortunately, the cmake in vcpkg will not be updated in time to update these cmake files.

The cmake files we exported is different. They will be generated synchronously by cmake as the port is updated, which largely guarantees its correctness. The only disadvantage is that they are NOT official, and other ports do not use them by default.

JackBoosY:
List:
<details>
- brotli
- libcroco
- librsvg
- libiconv
- tinyxml
- pixman
- nana
- libvpx
- parson
- sajson
- secp256k1
- glib
- cctz
- tinyxml
- concurrentqueue
- libuv
- gettext
- mathc
- tiny-aes-c
- tgc
- tiny-regex-c
- cello
- tiny-bignum-c
- mongoose
- sqlite3
- iniparser
- secp256k1
- concurrentqueue
- libsvm
- nativefiledialog
- kissfft
- http-parser
- breakpad
- tiny-aes-c
- tiny-bignum-c
- tiny-regex-c
- cello
- portable-snippets
- mathc
- tgc
- brotli
- portable-snippets
- sodium
- libconfuse
- ampl-mp
- nativefiledialog
- curlpp
- libsvm
- libpff
- pixman
- vulkan-memory-allocator
- cairo
- vulkan-memory-allocator
- sajson
- tiny-process-library
- parson
- mongoose
- box2d
- http-parser
- iniparser
- kissfft
- libstk
- libcroco
- gainput
- graphicsmagick
- libsodium
- fontconfig
- libstk
- tensorflow-cc
- libintl
- angle
- secp256k1
- concurrentqueue
- libui
- gl3w
[...]

PR #13762, opened 2020-09-27
[atk][exiv2][gdk-pixbuf][gettext][glib][libcroco][tesseract] Replace uses of unofficial-gettext with FindIntl

ras0219:
See title. See also #13027.

We prefer supporting built-in cmake modules whenever possible instead of providing our own replacements.


strega-nil:
I think this might've been fixed already by @Neumann-A in #11776, no?

PhoebeHui:
After compared with PR https://github.com/microsoft/vcpkg/pull/11776, the changes for atk and gettext still needed, I will help to undo changes for other ports and resolve the conflicts.

ports/gettext/portfile.cmake
```diff
@@ -59,24 +59,21 @@ else()
...
-file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
-
-set(GNU_DLL_PATHS lib/ debug/lib/)
-set(GNU_DLL_NAME GNU.Gettext.dll) #C# dll?
-foreach(DLL_PATH IN LISTS GNU_DLL_PATHS)
-    if(EXISTS "${CURRENT_PACKAGES_DIR}/${DLL_PATH}${GNU_DLL_NAME}")
-       file(REMOVE "${CURRENT_PACKAGES_DIR}/${DLL_PATH}${GNU_DLL_NAME}")
-    endif()
+# These targets are provided for backwards compatibility with previous vcpkg versions
+vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-gettext TARGET_PATH share/unofficial-gettext)
```
PhoebeHui, line 63: The CI failed with following errors:

```
Make Error at scripts/cmake/vcpkg_fixup_cmake_targets.cmake:126 (message):
  'D:/packages/gettext_x64-windows-static/debug/share/unofficial-gettext'
  does not exist.
  ports/gettext/portfile.cmake:63 (vcpkg_fixup_cmake_targets)
```

PhoebeHui (changes requested):
@ras0219, I think we need check if unofficial-gettext exist.

ports/gettext/portfile.cmake
Neumann-A: it does not exists anymore since gettext was changed to native build which is make based

PR #13770, opened 2020-09-27
[trantor] Update to 1.0.0

an-tao:


NancyLi1013 (approved):

ras0219 (approved):

ports/trantor/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: trantor
-Version: v1.0.0-rc16
+Version: v1.0.0
```
ras0219, line 2:
```suggestion
Version: 1.0.0
```

In vcpkg, we remove preceeding `v`'s from library versions.

PR #13777, opened 2020-09-28
[vcpkg] Implement versions db generator

vicroms:
Reduced scope of this PR to only generating version files from our Git history

- [X] Modify `x-history` to allow JSON output and add more version related information
- [X] Make `x-history` actually parse `CONTROL` and `vcpkg.json` files to fix several edge cases
- [X] Create script that generates version database files for each port
- [X] Detect version scheme when loading CONTROL files
- [X] Separate version string from port version in old CONTROL files


vicroms:
I'm not sure whether to check in the db files, there are too many of them. 
Right now the DB files are one per port (located inside `scripts/port_versions_db` and use JSON format. Generating them took about 1 hour on my computer.

Also, it seems like I forgot to clean up some unrelated code.

toolsrc/src/vcpkg/versions.cpp
```diff
@@ -0,0 +1,252 @@
...
+                                   const ComputedVersions& versions,
+                                   const std::string& baseline)
+{
+    (void)baseline;
+    (void)versions;
+
+    Checks::check_exit(VCPKG_LINE_INFO, versions.conflicts.empty(), "There are conflicts in the computed versions.");
+
+    auto& fs = paths.get_filesystem();
+    const auto working_dir = paths.root / "vcpkg-temp";
```
ras0219-msft, line 214:
```suggestion
    const auto working_dir = paths.buildtrees / "versioning_tmp";
```
because underscores are not valid port names, this will avoid creating a new root folder and won't collide with ports.

toolsrc/src/vcpkg/versions.cpp
```diff
@@ -0,0 +1,252 @@
...
+    Checks::exit_with_message(VCPKG_LINE_INFO, "Couldn't find version '%s' of %s", requested_version, package_name);
+}
+
+void Versions::fetch_port_versions(const VcpkgPaths& paths,
+                                   const ComputedVersions& versions,
+                                   const std::string& baseline)
+{
+    (void)baseline;
+    (void)versions;
+
```
ras0219-msft, line 210:
```suggestion
```

ras0219-msft (changes requested):
Thanks for opening this PR!

I haven't finished reading more deeply into the algorithm, but here are some more trivial comments.

toolsrc/src/vcpkg/versions.cpp
```diff
@@ -0,0 +1,252 @@
...
+    auto& fs = paths.get_filesystem();
+    const auto working_dir = paths.root / "vcpkg-temp";
+
+    if (fs.exists(working_dir))
+    {
+        fs.remove_all(working_dir, VCPKG_LINE_INFO);
+    }
+
+    auto output = run_git_command(paths, working_dir, "clone https://github.com/microsoft/vcpkg vcpkg-temp");
```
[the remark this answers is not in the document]
vicroms: Implemented `--no-checkout` but I'm not sure what you mean about how to use the local `.git` dir.

toolsrc/src/vcpkg/commands.porthistory.cpp
```diff
@@ -7,90 +8,206 @@
...
+            const fs::path dot_git_dir = paths.root / ".git";
 
-    static std::string get_version_from_commit(const VcpkgPaths& paths,
-                                               const std::string& commit_id,
-                                               const std::string& port_name)
-    {
-        const std::string cmd = Strings::format(R"(show %s:ports/%s/CONTROL)", commit_id, port_name);
-        auto output = run_git_command(paths, cmd);
+            const std::string full_cmd =
+                Strings::format(R"("%s" --git-dir="%s" %s)", fs::u8string(git_exe), fs::u8string(dot_git_dir), cmd);
```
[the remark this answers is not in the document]
vicroms: I'll look into this next

toolsrc/src/vcpkg/versions.cpp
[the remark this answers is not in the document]
ras0219-msft: We could clone from `paths.root` instead of github, possibly running a `fetch` from github afterwards if a commit is missing.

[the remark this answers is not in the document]
vicroms: Got it, I'll take a look at the `git clone` docs.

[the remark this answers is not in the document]
vicroms: Done

toolsrc/src/vcpkg/sourceparagraph.cpp
```diff
@@ -275,10 +273,20 @@ namespace vcpkg
         TextRowCol textrowcol;
         std::string buf;
         parser.optional_field(SourceParagraphFields::BUILD_DEPENDS, {buf, textrowcol});
-        spgh->dependencies = parse_dependencies_list(buf, origin, textrowcol).value_or_exit(VCPKG_LINE_INFO);
+
+        auto maybe_dependencies = parse_dependencies_list(buf, origin, textrowcol);
+        if (maybe_dependencies.has_value())
```
ras0219, line 278: This change causes failures to result in a "successful" return with no dependencies.

Instead, this needs to return the failure.

ras0219 (changes requested):

toolsrc/src/vcpkg/paragraphs.cpp
```diff
@@ -261,36 +266,88 @@ namespace vcpkg::Paragraphs
         return fs.exists(path / fs::u8path("CONTROL")) || fs.exists(path / fs::u8path("vcpkg.json"));
     }
 
-    static ParseExpected<SourceControlFile> try_load_manifest(const Files::Filesystem& fs,
-                                                              const std::string& port_name,
-                                                              const fs::path& path_to_manifest,
-                                                              std::error_code& ec)
+    static ParseExpected<SourceControlFile> try_load_manifest_object(
+        const std::string& origin,
+        const ExpectedT<std::pair<vcpkg::Json::Value, vcpkg::Json::JsonStyle>, std::unique_ptr<Parse::IParseError>>&
```
ras0219, line 271: It's very strange to pass an `ExpectedT` as an argument; it would be better for the caller to have handled any errors first and just pass extracted `Json::Value`.

toolsrc/src/vcpkg/sourceparagraph.cpp
```diff
@@ -275,10 +273,20 @@ namespace vcpkg
...
+            spgh->dependencies = maybe_dependencies.value_or_exit(VCPKG_LINE_INFO);
+        }
         buf.clear();
         parser.optional_field(SourceParagraphFields::DEFAULT_FEATURES, {buf, textrowcol});
-        spgh->default_features = parse_default_features_list(buf, origin, textrowcol).value_or_exit(VCPKG_LINE_INFO);
+
+        auto maybe_default_features = parse_default_features_list(buf, origin, textrowcol);
+        if (maybe_default_features.has_value())
+        {
+            spgh->default_features = maybe_default_features.value_or_exit(VCPKG_LINE_INFO);
```
ras0219, line 288: This shouldn't drop the error information (see above).

ras0219 (changes requested):
I've lightly reviewed the changes to port history since that's an `x-` command.

toolsrc/src/vcpkg/paragraphs.cpp
vicroms: It is strange, but this function is made to avoid repeating the error handling code 😅

toolsrc/src/vcpkg/paragraphs.cpp
```diff
@@ -261,36 +266,88 @@ namespace vcpkg::Paragraphs
...
     }
 
+    static ParseExpected<SourceControlFile> try_load_manifest_text(const std::string& text,
+                                                                   const std::string& origin,
+                                                                   std::error_code& ec)
+    {
+        auto error_info = std::make_unique<ParseControlErrorInfo>();
+        auto res = Json::parse(text);
+        if (ec) return error_info;
+        // TODO: Handle Json::parse(text) errors
```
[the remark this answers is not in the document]
vicroms: Forgot to remove my comment but `try_load_manifest_object(origin, res)` does exactly that.
However, the `error_info` declaration and `if (ec) return error_info;` lines are useless and should be removed.

toolsrc/src/vcpkg/paragraphs.cpp
vicroms: If you feel strongly about it, I can refactor it.

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+import json
+import time
+
+from subprocess import CalledProcessError
+from json.decoder import JSONDecodeError
+from pathlib import Path
+
+def make_dir(dir_path):
+    if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
```
strega-nil, line 14: Could be replaced with `os.makedirs(dir_path, exist_ok=True)`

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+    Path(rev_file).touch()
+    elapsed_time = time.time() - start_time
+    print('Processed {} total ports. Elapsed time: {:.2f} seconds'.format(total_count, elapsed_time))
+    
+
+def main(ports_path, db_path):
+    revision = get_current_git_ref()
+
+    if not revision:
+        print('Couldn\'t fetch current Git revision')
```
strega-nil, line 64:
```suggestion
        print('Couldn\'t fetch current Git revision.', file=sys.stderr)
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+        quit()
+
+    rev_file = os.path.join(db_path, revision)
+    if os.path.exists(rev_file):
+        print('Database files already exist for commit {}'.format(revision))
```
strega-nil, line 69:
```suggestion
        print(f'Database files already exist for commit {revision}')
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+        quit()
```
strega-nil, line 70:
```suggestion
        sys.exit(0)
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+    
+    generate_port_versions_db(ports_path, db_path, revision)
+
+
+if __name__ == "__main__":
+    main('../ports', '../port_versions')
```
strega-nil, line 76:
```suggestion
    main(ports_path=path.join(SCRIPT_DIRECTORY, '../ports'), db_path=path.join(SCRIPT_DIRECTORY, '../port_versions'))

```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+    if not revision:
+        print('Couldn\'t fetch current Git revision')
+        quit()
+
+    rev_file = os.path.join(db_path, revision)
+    if os.path.exists(rev_file):
+        print('Database files already exist for commit {}'.format(revision))
+        quit()
+    
+    generate_port_versions_db(ports_path, db_path, revision)
```
strega-nil, line 72:
```suggestion
    generate_port_versions_db(ports_path=ports_path, db_path=db_path, revision=revision)

```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+    elapsed_time = time.time() - start_time
+    print('Processed {} total ports. Elapsed time: {:.2f} seconds'.format(total_count, elapsed_time))
+    
+
+def main(ports_path, db_path):
+    revision = get_current_git_ref()
+
+    if not revision:
+        print('Couldn\'t fetch current Git revision')
+        quit()
```
strega-nil, line 65:
```suggestion
        sys.exit(1)
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_current_git_ref():
+    try: 
+        output = subprocess.run(args=['git.exe', 'rev-parse', '--verify', 'HEAD'], capture_output=True)
+        if output.returncode == 0:
+            return output.stdout.decode('utf-8').strip()
+        return ''
+    except CalledProcessError as err:
+        print("Failed to get git ref")
```
strega-nil, line 24:
```suggestion
    output = subprocess.run(['git', '-C', SCRIPT_DIRECTORY, 'rev-parse', '--verify', 'HEAD'], capture_output=True, encoding='utf-8')
    if output.returncode == 0:
        return output.stdout.strip()
    print('Failed to get current git ref:', output.stderr.strip(), file=sys.stderr)
    return None
```

Few things:

* `subprocess.run` doesn't raise a `CalledProcessError` unless you pass `check=True`
* Correctly outputting to stderr, and outputting the actual error message
* Use `git -C <path>` to make sure that we're using the right git repository
* (minor) returning None if there's an error is better than empty string
* (minor) adding the encoding in `run` is easier.

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+            try:
+                output = subprocess.run(args=['../vcpkg.exe', 'x-history', port_name, '--x-json'], capture_output=True)
+                try:
+                    versions_object = json.loads(output.stdout)
+                    with open(output_filepath, 'w') as output_file:
+                        json.dump(versions_object, output_file)
+                except JSONDecodeError as json_err:
+                    print('Failed to load JSON for {}'.format(port_name))
+            except CalledProcessError as err:
+                print("Failed to run {}".format(err.cmd))
```
strega-nil, line 47:
```suggestion
            output = subprocess.run([path.join(SCRIPT_DIRECTORY, '../vcpkg'), 'x-history', port_name, '--x-json'], capture_output=True)
            if output.returncode == 0:
                try:
                    versions_object = json.loads(output.stdout)
                    with open(output_filepath, 'w') as output_file:
                        json.dump(versions_object, output_file)
                except JSONDecodeError as json_err:
                    print(f'Malformed JSON from vcpkg x-history {port_name}:', output.stdout, file=sys.stderr))
            else:
                print(f'x-history {port_name} failed:', output.stdout, file=sys.stderr)
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+import os.path
+import sys
+import subprocess
+import json
+import time
+
+from subprocess import CalledProcessError
+from json.decoder import JSONDecodeError
+from pathlib import Path
+
```
strega-nil, line 11:
```suggestion

SCRIPT_DIRECTORY = path.dirname(path.abspath(__file__))

```

allows us to use file-relative paths later, rather than CWD-relative.

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+        print("Failed to get git ref")
+
+
+def generate_port_versions_db(ports_path, db_path, revision):
+    start_time = time.time()
+
+    port_names = [item for item in os.listdir(ports_path) if os.path.isdir(os.path.join(ports_path, item))]
+    total_count = len(port_names)
+    for counter, port_name in enumerate(port_names):
+        containing_dir = os.path.join(db_path, port_name[0])
```
strega-nil, line 33:
```suggestion
        containing_dir = os.path.join(db_path, f'{port_name[0]}-')
```

don't forget the `-` :)

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+        make_dir(containing_dir)
+        
+        output_filepath = os.path.join(containing_dir, '{}.json'.format(port_name))
```
strega-nil, line 36:
```suggestion
        output_filepath = os.path.join(containing_dir, f'{port_name}.json')

```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+                        json.dump(versions_object, output_file)
+                except JSONDecodeError as json_err:
+                    print('Failed to load JSON for {}'.format(port_name))
+            except CalledProcessError as err:
+                print("Failed to run {}".format(err.cmd)) 
+
+        # This should be replaced by a progress bar
+        if counter > 0 and counter % 100 == 0:
+            elapsed_time = time.time() - start_time
+            print('Processed {} out of {}. Elapsed time: {:.2f} seconds'.format(counter, total_count, elapsed_time))
```
strega-nil, line 52:
```suggestion
            print(f'Processed {counter} out of {total_count}. Elapsed time: {elapsed_time:.2f} seconds'.)
```

scripts/generatePortVersionsDb.py
```diff
@@ -0,0 +1,76 @@
...
+    
+    rev_file = os.path.join(db_path, revision)
+    Path(rev_file).touch()
+    elapsed_time = time.time() - start_time
+    print('Processed {} total ports. Elapsed time: {:.2f} seconds'.format(total_count, elapsed_time))
```
strega-nil, line 57:
```suggestion
    print(f'Processed {total_count} total ports. Elapsed time: {elapsed_time:.2f} seconds')
```

strega-nil (changes requested):
I haven't looked at the rest of the code, but here are the things I see that should be corrected with the python.

ras0219 (approved):

strega-nil (commented):
Thanks for the changes victor!

PR #13783, opened 2020-09-28
[libtheora] add unofficial cmake config, remove FindOGG.cmake

mcmtroffaes:
- What does your PR fix? This adds cmake config file for unofficial::theora targets, to use by other ports and consumers of vcpkg. It also removes the FindOGG.cmake since libogg in vcpkg already provides a config file. Tested on x64-linux, x64-windows, and x64-windows-static.


ports/libtheora/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ file(GLOB HEADERS
 )
 
 include_directories("include")
-include_directories(${OGG_INCLUDE_DIR})
+include_directories(${Ogg_INCLUDE_DIR})
```
ras0219, line 15:
```suggestion
```

This should be handled automatically by the `Ogg::ogg` target.

ras0219 (changes requested):
Thanks for the PR!

ports/libtheora/CMakeLists.txt
mcmtroffaes: Thanks, and yes, I realised that. I tried deleting it but for some reason it does not find the headers for some files that are compiled. I'm going to see if I can get to the root of this.

mcmtroffaes: Adding a target_link_libraries against Ogg for the object libraries fixed it. I've just pushed another commit to this effect.

PhoebeHui: The targets may need have target_include_directories(${target} PUBLIC $<INSTALL_INTERFACE:include>)

mcmtroffaes: Ah, good point. Might as well fix that too.

mcmtroffaes:
Thanks @ras0219 @PhoebeHui for your comments. I think I've now done everything and done another series of tests to ensure it's all compiling as it should. Let me know if you still spot some issues.

I'll try to get this cmake file be included and maintained upstream. Not that theora still gets a lot of development, but it can't hurt trying, and it feels like the right thing to do.

PhoebeHui (approved):

mcmtroffaes:
You're welcome. I've submitted the cmake work to upstream here:

https://github.com/xiph/theora/pull/14

Whilst I was at it, I've also submitted the patch by @Mixaill (which fixes the debug build) to upstream:

https://github.com/xiph/theora/pull/13

dan-shaw:
Thanks for the PR!

PR #13786, opened 2020-09-28
[sqlcipher] Add new port

SvenPStarFinanz:
This is a sqlite edition with encryption, partially copied from the sqlite3 port.
The amalgamation files are built by the tclsh port.
Currently only Windows DLL support.


ras0219:
Thanks for the PR!

Note to reviewers: this library is a _replacement_ for sqlite3, so we cannot build it in CI.

ports/sqlcipher/portfile.cmake
```diff
@@ -0,0 +1,84 @@
+vcpkg_fail_port_install(ON_LIBRARY_LINKAGE "static" ON_TARGET "UWP" "OSX" "Linux")
```
NancyLi1013, line 1:
```suggestion
vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)

vcpkg_fail_port_install( ON_TARGET "UWP" "OSX" "Linux")
```
We suggest to use `vcpkg_check_linkage()` to express unsupported for static or dynamic library.

I also noticed that you add `!arm` in vcpkg.json file. If it doesn't support arm, please also add `arm `in unsupported messages.

scripts/ci.baseline.txt
```diff
@@ -1623,6 +1623,11 @@ speexdsp:x64-linux=fail
 speexdsp:x64-osx=fail
 spirv-tools:arm-uwp=fail
 spirv-tools:x64-uwp=fail
+sqlcipher:arm-uwp=fail
+sqlcipher:x64-linux=fail
+sqlcipher:x64-osx=fail
+sqlcipher:x64-uwp=fail
+sqlcipher:x64-windows-static=fail
```
NancyLi1013, line 1630:
```suggestion
# sqlcipher is a replacement for sqlite3
sqlcipher:x86-windows=skip
sqlcipher:x64-windows=skip
```
You have added `supports `field in vcpkg.json file, there is no need to add these unsupported triplets here.

For the supported triplets, since  `sqlcipher` is a replacement for `sqlite3`, we set them as "skip" to disable detection in CI.

NancyLi1013 (changes requested):

ports/sqlcipher/portfile.cmake
SvenPStarFinanz: Optimistically I'm hoping arm is working, so not excluded any more.

scripts/ci.baseline.txt
SvenPStarFinanz: sqlcipher is a replacement from app perspective, but from vcpkg view it may exist in parallel to sqlite3, because of different dll name and include directory. So i will remove from baseline.

NancyLi1013 (approved):

NancyLi1013:
I have installed `sqlcipher` on my local and the headers (`sqlite3.h sqlite3xt.h`) are installed in `/include/sqlcipher/` directory. There is only one library named `sqlcipher.lib`. 

I also installed `sqlite3` and the headers (`sqlite3.h sqlite3xt.h) are installed in `include/` directory.

So I think these two points have no conflicts with each other. There is no need to skip the build on our CI.

jonboh:
@SvenPStarFinanz do you remember why this is not valid for static builds?

PR #13815, opened 2020-09-29
[fmt] Add usage file

dgellow:
**Describe the pull request**

### What does your PR fix?

When `fmt` is installed from `vcpkg`, you get the output:

```
The package fmt:x64-windows provides CMake targets:

    find_package(fmt CONFIG REQUIRED)
    target_link_libraries(main PRIVATE fmt::fmt fmt::fmt-header-only)
```

Unfortunately that's not correct, we shouldn't link both `fmt::fmt` and the header only version, `fmt::fmt-header-only`.

This PR adds a usage file to instead show to the user the CMake statements from `fmt` own documentation: https://fmt.dev/latest/usage.html#usage-with-cmake.

I don't think anything has to be updated. Just tell me if I have to change something.

I believe so, yes.

### DIFF
--- a/ports/fmt/usage
+++ b/ports/fmt/usage
@@ -1,0 +1,7 @@
+The package fmt provides CMake targets:
+
+    find_package(fmt CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE fmt::fmt)
+
+    # Or use the header-only version
+    target_link_libraries(main PRIVATE fmt::fmt-header-only)
--- a/ports/fmt/CONTROL
+++ b/ports/fmt/CONTROL
@@ -1,5 +1,5 @@
 Source: fmt
 Version: 7.0.3
-Port-Version: 2
+Port-Version: 3
 Homepage: https://github.com/fmtlib/fmt
 Description: Formatting library for C++. It can be used as a safe alternative to printf or as a fast alternative to IOStreams.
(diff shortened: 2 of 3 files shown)


ports/fmt/usage
```diff
@@ -0,0 +1,7 @@
+The package fmt provides CMake targets:
+
+    find_package(fmt CONFIG REQUIRED)
+    target_link_libraries(main fmt::fmt)
```
PhoebeHui, line 4:
```suggestion
    target_link_libraries(main PRIVATE fmt::fmt)
```

PhoebeHui (changes requested):
Could you also update the 'Port-Version: 2' to ''Port-Version: 3' in CONTROL file?

ports/fmt/usage
dgellow: I didn't specify the `PRIVATE` because it is missing from fmt's own documentation: https://fmt.dev/latest/usage.html#usage-with-cmake

> You can detect and use an installed version of {fmt} as follows:
> find_package(fmt)
> target_link_libraries(<your-target> fmt::fmt)

Do you still think that we should add it? Just to be sure because I commit.

ras0219-msft (changes requested):
Hi @dgellow, thanks for the PR!

Did you verify that this actually achieves the change you wanted? The PR does not appear to deploy the `usage` file, so it will not affect the actual printed message in vcpkg.

The file needs to be deployed to `share/fmt/usage`.

dgellow:
Thanks for the review. The usage file should be copied to the correct location now, though I have issues testing it locally because of network timeouts errors when fetching mingw from MSYS2. They seem to face some problems with their servers.

I will try again later, hopefully that will be fixed.

image
 
image

BillyONeal (approved):

BillyONeal:
Thanks for your contribution!

PR #13816, opened 2020-09-29
Docs: use the minimum CMake version supported by vcpkg

dgellow:
**Describe the pull request**

### What does your PR fix?

vcpkg examples use CMake minimum versions that aren't compatible with the one specified in https://github.com/microsoft/vcpkg/blob/master/scripts/vcpkgTools.xml.

This PR instead uses the smallest CMake version number present in `vcpkgTools.xml`, which is version `3.12.4` at the current time.

I don't believe that's relevant.

Yes, I think that's the case.

### DIFF
--- a/docs/examples/installing-and-using-packages.md
+++ b/docs/examples/installing-and-using-packages.md
@@ -115,7 +115,7 @@ If you are using CMake through Open Folder with Visual Studio you can define `CM
 Now let's make a simple CMake project with a main file.
 ```cmake
 # CMakeLists.txt
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.12.4)
 project(test)
 
 find_package(unofficial-sqlite3 CONFIG REQUIRED)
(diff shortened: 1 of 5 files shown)


Neumann-A:
Personal opinion since I dont give anything about linux distros. It should be 3.17 or even better 3.18 since this will be the minimum vcpkg internally uses.

ras0219-msft:
Thanks for the PR!

The version of CMake used in our examples is explicitly decoupled from the version in `vcpkgTools.xml` -- users are not required to use the version of CMake that vcpkg builds packages with.

This also makes a bunch of unrelated changes: every markdown bullet is changed from `*` to `-` and the manifest file example has been formatted away from our required canonical format.

Given this, I'm going to close this PR for now. If you'd like to open new documentation improvement PRs in the future, we're happy to review them.

dgellow:
You're right, my editor auto formatted on save and I didn't notice. Could you clarify if you would accept a PR that would only do the change of CMake version, or if that's something you would be against? At minimum that would make it more consistent and suggest to use a more recent version of CMake, but I don't know if that's valuable from your point of view.

PR #13824, opened 2020-09-30
[PCL] Remove -fext-numeric-literals when building x64-osx

fran6co:
`-fext-numeric-literals` does not exist in clang and it builds just fine without it

### DIFF
--- a/ports/pcl/boost-1.70.patch
+++ b/ports/pcl/boost-1.70.patch
@@ -6,7 +6,7 @@ index b301cf1..789b660 100644
  
  # Boost (required)
  include("${PCL_SOURCE_DIR}/cmake/pcl_find_boost.cmake")
-+if (NOT APPLE AND NOT WIN32)
++if (NOT APPLE AND NOT WIN32 AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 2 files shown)


JackBoosY (changes requested):

ports/pcl/boost-1.70.patch
```diff
@@ -6,7 +6,7 @@ index b301cf1..789b660 100644
  
  # Boost (required)
  include("${PCL_SOURCE_DIR}/cmake/pcl_find_boost.cmake")
-+if (NOT APPLE AND NOT WIN32)
++if (NOT APPLE AND NOT WIN32 AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
JackBoosY, line 9: The macro `APPLE` should be `ON` in osx, can you confirm that?

JackBoosY (changes requested):

ports/pcl/boost-1.70.patch
fran6co: I think the APPLE flag should be removed as well because the default compiler is clang and not gcc. But I don't have MacOs to test it. This flag it's only valid for gcc regardless of the Os

JackBoosY (approved):

PR #13847, opened 2020-10-01
[prometheus-cpp] Update to version 0.10.0

gjasny:
Tested locally on MacOS.


ras0219-msft:
LGTM, thanks for the update!

PR #13861, opened 2020-10-03
Add Baseee support

moegodot:
This adds support for [Baseee](https://github.com/chhdao/baseee).A modern C++ library use Mit License. 
Add base features for c++ like boosts.


sfhacker:
image

chhdao:
The lib doean't  supports x86 and arm.So I add  "supports": "!(x86 | arm)" to the vcpkg.json.  
But the IC still running in error port and  checks were not successful.

PR #13867, opened 2020-10-03
[tensorflow-c-gpu] Add new port

ianormy:
**TensorFlow C API (GPU)**

- This PR adds a new port for the TensorFlow C API (GPU) for Windows. It downloads and builds a package using the prebuilt files provided by the TensorFlow team. [https://www.tensorflow.org/install/lang_c](https://www.tensorflow.org/install/lang_c). This works for the latest version, 2.3.0.

- The only triplets supported are x64-windows


ianormy:
My support string in the vcpkg.json file clearly tells the CI not to build x86 for Windows, but your CI seems to ignore it. Here's my string:

`"supports": "windows & !x86 & !uwp & !static & !arm"`


Is there something I am missing? Any help would be gratefully appreciated. It should only build **x64-windows**.

Thanks

Ian

ianormy:
I have checked the logs for the x86_windows build and its failing when trying to build zlib:x86-windows-static. It doesn't appear to be related to my code at all.

RDCH106:
Try to exclude from the [CI baseline](https://github.com/microsoft/vcpkg/blob/master/scripts/ci.baseline.txt) `tensorflow-c-gpu:x64-uwp`

scripts/ci.baseline.txt
```diff
@@ -1665,6 +1665,14 @@ tensorflow-cc:x64-osx=skip
...
 tensorflow-cc:x64-windows-static=fail
 tensorflow-cc:x86-windows=fail
+tensorflow-c-gpu:arm-uwp=fail
+tensorflow-c-gpu:arm64-windows=fail
+tensorflow-c-gpu:x64-linux=fail
+tensorflow-c-gpu:x64-osx=fail
+tensorflow-c-gpu:x64-windows-static=fail
+tensorflow-c-gpu:x64-uwp=fail
+tensorflow-c-gpu:x86-windows-static=fail
+tensorflow-c-gpu:x86-windows=fail
```
JackBoosY, line 1675: Since we have `supports` on manifest, we don't need add this.

JackBoosY (changes requested):
What's the different between this port with `tensorflow-cc`?

ports/tensorflow-c-gpu/portfile.cmake
```diff
@@ -0,0 +1,44 @@
+# # Specifies if the port install should fail immediately given a condition
+vcpkg_fail_port_install(MESSAGE "tensorflow-c-gpu currently only supports Windows platforms" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")
+vcpkg_fail_port_install(MESSAGE "tensorflow-c-gpu currently only supports x64 architecture" ON_ARCH "x86" "arm" "arm64")
+vcpkg_fail_port_install(MESSAGE "tensorflow-c-gpu currently does not support static libraries" ON_LIBRARY_LINKAGE "static")
```
NancyLi1013, line 4:
```suggestion
vcpkg_fail_port_install(ON_ARCH "x86" "arm" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")

vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
```
We use `vcpkg_check_linkage()` to judge if the port supports static or dynamic build.

NancyLi1013 (changes requested):

ports/tensorflow-c-gpu/portfile.cmake
ianormy: @NancyLi1013 thanks for the suggested change. I will commit as you have suggested.

scripts/ci.baseline.txt
ianormy: @JackBoosY I only added this because the *supports* wasn't apparently being parsed correctly. This was suggested previously on this PR as a fix for that.

ianormy: @JackBoosY I started with **tensorflow-cc** but this doesn't build on Windows. That's why I went for using the TensorFlow C API. It's more hard-core, and there's no documentation for it, but it's very fast. I can now load and run inference on my model even quicker than the equivalent Python code!

NancyLi1013:
Hi @ianormy 

> I have checked the logs for the x86_windows build and its failing when trying to build zlib:x86-windows-static. It doesn't appear to be related to my code at all.

This failure is not related to your PR. You can ignore it. So this port might support x86, you should remove it from `Supports` field and portfile,cmake. Then let's see the new CI tests. Sorry, I didn't notice your comment before.

Also you can remove the codes added in ci.baseline.txt, since you have added `Supports `field.

ports/tensorflow-c-gpu/vcpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "tensorflow-c-gpu",
+  "version-string": "2.3.0",
+  "description": "tensorflow C-API (GPU)",
+  "homepage": "https://www.tensorflow.org/install/lang_c",
+  "supports": "windows & !x86 & !uwp & !static & !arm"
```
NancyLi1013, line 6:
```suggestion
  "supports": "windows & !uwp & !static & !arm"
```

NancyLi1013 (changes requested):

ianormy:
I've applied the recommended changes by @NancyLi1013  and answered @JackBoosY question. Both of these comments have been resolved but the status still says that both of you require changes that I need to respond to. Is there something else I need to do to resolve this? I can't see any other buttons to press - unless I'm missing something. Please could you let me know if there is something else I should do. Thanks.

ports/tensorflow-c-gpu/portfile.cmake
```diff
@@ -0,0 +1,43 @@
...
+        ${CURRENT_PACKAGES_DIR}/debug/lib
+        ${CURRENT_PACKAGES_DIR}/debug/bin
+    )
+
+file(COPY 
+        ${SOURCE_PATH}/include 
+        DESTINATION ${CURRENT_PACKAGES_DIR}
+    )
+
+file(COPY ${SOURCE_PATH}/lib/tensorflow.lib
```
JackBoosY, line 33: Can we build it instead of using the binaries?

JackBoosY (changes requested):

ports/tensorflow-c-gpu/portfile.cmake
ianormy: @JackBoosY Building it would have the same issues as the tensorflow-cc port under Windows. 

I can see this port isn't going to get through this process. It's a great solution for me locally (I've used the same approach for using the onnxruntime on my machine) so I'll leave it at that and abandon this PR. I appreciate your time.

PR #13886, opened 2020-10-05
[sltbench] Add new port

### LINKED ISSUE #13106: [New Port Request] sltbench
Library name: sltbench

Library description: C++ microbench framework

Source repository URL: https://github.com/ivafanas/sltbench

brukted:
- What does your PR fix? Fixes #13106

### DIFF
--- a/ports/sltbench/portfile.cmake
+++ b/ports/sltbench/portfile.cmake
@@ -1,0 +1,19 @@
+set(VCPKG_LIBRARY_LINKAGE static)
+
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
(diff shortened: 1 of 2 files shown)


ports/sltbench/portfile.cmake
```diff
@@ -0,0 +1,19 @@
+set(VCPKG_LIBRARY_LINKAGE static)
```
PhoebeHui, line 1: Does this mean the sltbench port can only be built with static library linkage?  if so , you should use vcpkg_check_linkage(ONLY_STATIC_LIBRARY) instead.

PhoebeHui (changes requested):

PhoebeHui (approved):

PR #13889, opened 2020-10-05
[vcpkg_acquire_msys] Use Mirrors

ras0219-msft:
This PR adds mirrors to `vcpkg_acquire_msys()` to handle when the primary source is down.

The sourceforge mirror is not used because it only keeps the latest packages, whereas the other mirrors appear to have the same contents as the primary source.

### DIFF
--- a/scripts/cmake/vcpkg_find_acquire_program.cmake
+++ b/scripts/cmake/vcpkg_find_acquire_program.cmake
@@ -425,41 +425,25 @@ function(vcpkg_find_acquire_program VAR)
     set(HASH 2a5480d503ac6e8203040c7e516a3395028520da05d0ebf3a2d56d5d24ba5d17630e8f318dd4e3cc2094cc4668b90108fb58e8b986b1ffebd429995058063c27)
   elseif(VAR MATCHES "PKGCONFIG")
     set(PROGNAME pkg-config)
-    set(VERSION 0.29.2-1)
-    set(LIBWINPTHREAD_VERSION git-8.0.0.5906.c9a21571-1)
     if(ENV{PKG_CONFIG})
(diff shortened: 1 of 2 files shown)


scripts/cmake/vcpkg_find_acquire_program.cmake
```diff
@@ -425,41 +425,25 @@ function(vcpkg_find_acquire_program VAR)
...
-          FILENAME mingw-w64-i686-pkg-config-${VERSION}-any.pkg.tar.xz
+        set(VERSION 0.29.2-1)
+        set(LIBWINPTHREAD_VERSION git-8.0.0.5906.c9a21571-1)
+        vcpkg_acquire_msys(
+          PKGCONFIG_ROOT
+          NO_DEFAULT_PACKAGES
+          DIRECT_PACKAGES
+            "https://repo.msys2.org/mingw/i686/mingw-w64-i686-pkg-config-${VERSION}-any.pkg.tar.xz"
+            3b1b706a24d9aef7bbdf3ce4427aaa813ba6fbd292ed9dda181b4300e117c3d59a159ddcca8b013fd01ce76da2d95d590314ff9628c0d68a6966bac4842540f0
+            "https://repo.msys2.org/mingw/i686/mingw-w64-i686-libwinpthread-${LIBWINPTHREAD_VERSION}-any.pkg.tar.zst"
```
dan-shaw, line 442: Do we need to extract this archive, since this is different file format from the first package?

dan-shaw (approved):

scripts/cmake/vcpkg_find_acquire_program.cmake
ras0219: All of the links are assumed to be archives that can be extracted with `cmake -E tar`

ionutcosminmihaipfa:
Still not working for protobuf:arm64-windows

PR #13903, opened 2020-10-06
[vcpkg] run both cxx and c compiler id.

Neumann-A:
fixes: https://discord.com/channels/400588936151433218/687365466422902841/762996065850621953

```
  if ( DEFINED CMAKE_C_COMPILER_ID AND DEFINED CMAKE_CXX_COMPILER_ID
       AND NOT "x${CMAKE_C_COMPILER_ID}" STREQUAL "x${CMAKE_CXX_COMPILER_ID}")
    message(FATAL_ERROR "The current configuration mixes Clang and MSVC or "
            "some other CL compatible compiler tool. This is not supported. "
            "Use either clang or MSVC as both C and C++ compilers.")
  endif()
```

CXX run was added in:
https://github.com/microsoft/vcpkg/commit/741c8cbecde77f771f8c05f00f59a436a2b4213b


BillyONeal (approved):

ras0219-msft:
Thanks!

PR #13926, opened 2020-10-07
[physfs]Add suffix to debug library so that it can be indepently found

malfmalf:
The title describes all of the change.
physfs had the same name for debug and release, and it didn't provide a cmake config file
trying to make a find_library to link the correct version was error prone and difficult
Simply adding a suffix to the library in debug solves this and I cannot foresee any problems with this


Hoikas:
Per the [maintainer's guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-rename-binaries-outside-the-names-given-by-upstream), libraries should not be renamed from their upstream names. You should consider writing a cmake config to work around this issue.

NancyLi1013:
Hi @malfmalf 
Thanks for your PR.
I think the suggestion from @Hoikas is right, we should not rename the binaries outside the names given by upstream.  If you have the demand, it might be better to post this to upstream.

Thanks.

strega-nil:
Given that this is not the way we like to solve this problem, closing the PR. Thanks so much for your contribution @malfmalf :)

PR #13932, opened 2020-10-08
[ffmpeg:x64-linux] install tools also on non-windows platforms

mcmtroffaes:
- What does your PR fix? Non-windows builds of ffmpeg do not install the tools. This patch ensures tools are installed on all platforms. Tested x64-linux, x86-windows, and x86-windows-static. Note the use of ``file(COPY ...)`` instead of ``file(INSTALL ...)`` to ensure executable permissions are retained.


ports/ffmpeg/portfile.cmake
```diff
@@ -467,24 +467,26 @@ if(VCPKG_TARGET_IS_WINDOWS)
...
-        if (EXE_FILES_DBG)
-            file(INSTALL ${EXE_FILES_DBG} DESTINATION ${CURRENT_PACKAGES_DIR}/debug/tools/${PORT})
-            vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/debug/tools/${PORT})
-        endif()
-        file(REMOVE ${FILES_TO_REMOVE})
-    endif()
+    list(APPEND FILES_TO_REMOVE ${EXP_FILES} ${LIB_FILES} ${DEF_FILES})
+endif()
+
+# Handle tools
```
Neumann-A, line 475: There exists a function called `vcpkg_copy_tools`

mcmtroffaes: Thanks for pointing out. I assumed there was a reason that vcpkg_copy_tools wasn't used since the original code doesn't use it either, however I'll see if I can migrate it to use the proper function.

mcmtroffaes:
As suggested, I've added a commit that updates the code to use vcpkg_copy_tools. Tested on x64-linux and x86-windows. Works as before on windows except that the debug version of the tools are now no longer installed.

PhoebeHui (approved):
LGTM, thanks for the PR!

mcmtroffaes:
Thanks for merging!

PR #13969, opened 2020-10-10
[directxtex] Use CMake build system, add OpenEXR support and dx12 feature

RT2Code:
Use the provided CMake file instead of the VS solutions to build the library, as recommended in the maintainer guidelines. DirectXTex can be built without DirectX12 (default for desktop) for Windows 7 Service Pack 1 or later, or with DirectX12 (default for ARM or UWP) for Windows 10.

Also add the openexr feature to enable OpenEXR format support in the library and its tools : https://github.com/Microsoft/DirectXTex/wiki/Adding-OpenEXR


RT222:
Regressions indeed, otherwise it would have been too easy. :) I'll try to fix that as soon as possible.

ports/directxtex/enable_openexr_support.patch
```diff
@@ -0,0 +1,162 @@
...
++//                 (void)SetFileInformationByHandle(m_handle, FileDispositionInfo, &info, sizeof(info));
++//             }
++//         }
++
++//         void clear() { m_handle = nullptr; }
++
++//     private:
++//         HANDLE m_handle;
++//     };
++// }
```
Hoikas, line 159: The patch could be smaller if you use `#if 0`

ports/directxtex/portfile.cmake
```diff
@@ -8,63 +8,57 @@ vcpkg_from_github(
...
+if("openexr" IN_LIST FEATURES)
+    file(INSTALL ${SOURCE_PATH}/DirectXTex/DirectXTexEXR.h DESTINATION ${CURRENT_PACKAGES_DIR}/include)
+endif()
+
+file(INSTALL ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-dbg/bin/CMake/Debug/DirectXTex.lib DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib)
+file(INSTALL ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel/bin/CMake/Release/DirectXTex.lib DESTINATION ${CURRENT_PACKAGES_DIR}/lib)
+if(NOT VCPKG_TARGET_IS_UWP AND NOT TRIPLET_SYSTEM_ARCH STREQUAL "arm64")
+    file(INSTALL ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel/bin/CMake/Release/texassemble.exe DESTINATION ${CURRENT_PACKAGES_DIR}/tools/${PORT})
+    file(INSTALL ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel/bin/CMake/Release/texconv.exe DESTINATION ${CURRENT_PACKAGES_DIR}/tools/${PORT})
+    file(INSTALL ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel/bin/CMake/Release/texdiag.exe DESTINATION ${CURRENT_PACKAGES_DIR}/tools/${PORT})
```
Hoikas, line 61:
```suggestion
    vcpkg_copy_tools(
        TOOL_NAMES texassemble texconv texdiag
        SEARCH_DIR ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel/bin/CMake/Release
    )
```

ports/directxtex/portfile.cmake
```diff
@@ -8,63 +8,57 @@ vcpkg_from_github(
...
 endif()
 
-file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
```
Hoikas, line 64: Trailing newline please :smile:

ports/directxtex/portfile.cmake
RT222: I didn't know about vcpkg_copy_tools, thanks.

ports/directxtex/portfile.cmake
RT222: Whoopsie. :D

ports/directxtex/enable_openexr_support.patch
RT222: You're right, but I did that because I wanted to follow the instructions very closely and they ask to comment this section out. Do you really think I should change it? I have to admit that I fail to see the importance of gaining a few lines on a patch.

Hoikas: I think the goal with commenting sections out is to use block-style comments. While we could use `/* */`, those could easily be broken by upstream changes. If the size of the patch is decreased, we reduce the chance that trivial upstream changes will cause the patch to fail when the port is upgraded.

ports/directxtex/portfile.cmake
```diff
@@ -30,15 +30,16 @@ endif()
 
 vcpkg_check_features(
     OUT_FEATURE_OPTIONS FEATURE_OPTIONS
-    openexr ENABLE_OPENEXR_SUPPORT
+    FEATURES openexr ENABLE_OPENEXR_SUPPORT
+    INVERTED_FEATURES win7 BUILD_DX12
```
Hoikas, line 34: I thought DirectX 12 was backported to Windows 7 by Microsoft? Regardless, It seems awkward to require opting-into a feature to get Windows 7 support. I realize that Windows 7 is no longer supported by Microsoft - but my personal opinion is that it would make more sense to have a feature for opting-in to DX12.

RT222: I'm not entirely sure how the DX12 on Windows 7 works, but DirectXTex only enable it for Windows 10 as explained here : https://github.com/microsoft/DirectXTex/wiki/DirectXTex#using-project-to-project-references

I think you're right about the feature, and the previous port was building without DX12 support anyway so there's less chances to break things for users if we keep it that way. I'm changing it.

ports/directxtex/enable_openexr_support.patch
RT222: Good point, I'm changing it.

LilyWangL (approved):

PR #13975, opened 2020-10-11
[directxtk12] Use CMake build system

RT2Code:
Use the provided CMake file instead of the VS solutions to build the library, as recommended in the maintainer guidelines.


ports/directxtk12/CONTROL
```diff
@@ -1,4 +1,5 @@
 Source: directxtk12
-Version: aug2020
+Version: sept2020
+Port-Version: 2
```
Hoikas, line 3:
```suggestion
```

RT222: Fixed, thanks. ;)

ports/directxtk12/portfile.cmake
```diff
@@ -8,46 +8,16 @@ vcpkg_from_github(
...
-    set(SLN_NAME "Windows10_${VS_VERSION}")
-else()
-    set(SLN_NAME "Desktop_${VS_VERSION}_Win10")
-endif()
-
-vcpkg_build_msbuild(
-    PROJECT_PATH ${SOURCE_PATH}/DirectXTK_${SLN_NAME}.sln
-    PLATFORM ${TRIPLET_SYSTEM_ARCH}
+vcpkg_configure_cmake(
+    SOURCE_PATH ${SOURCE_PATH}
```
NancyLi1013, line 12:
```suggestion
    SOURCE_PATH ${SOURCE_PATH}
    PREFER_NINJA
```
Can this use `NINJA `to build?

RT222: It required some changes because ninja doesn't export the libraries to the same path depending on platforms, but it works.

NancyLi1013 (approved):

PR #13991, opened 2020-10-12
[icu] Disable layoutex feature

### LINKED ISSUE #13527: icu[core]:x64-linux build failure
**Host Environment**
- OS: Ubuntu 18.04

**To Reproduce**

`./vcpkg install icu`

**Failure logs**
 
```
-- pkg-config call failed with error code:1
-- pkg-config output:
CMake Error at scripts/cmake/vcpkg_fixup_pkgconfig.cmake:76 (message):
  pkg-config error output:Package 'harfbuzz' requires 'glib-2.0 >= 2.19.1'
  but version of glib-2.0 is
  scripts/cmake/vcpkg_fixup_pkgconfig.cmake:297 (vcpkg_fixup_pkgconfig_check_files)
  ports/icu/portfile.cmake:236 (vcpkg_fixup_pkgconfig)
```

**Additionally**

```
$ ./vcpkg list | grep glib
glib:x64-linux                                     2.52.3-14-7      Portable, general-purpose utility library.
glibmm:x64-linux                                   2.52.1-11        This is glibmm, a C++ API for parts of glib that...
harfbuzz[glib]:x64-linux                                            Glib Unicode callbacks support
```

gergondet:
**Describe the pull request**

icu tries to detect the presence of `icu-le-hb` to select the default value for the `layoutex` feature.

This package is not provided by vcpkg but when this package is otherwise available on the system the build will fail during `vcpkg_fixup_pkgconfig` as the pkg-config file pulls the dependency from the system.

- What does your PR fix? This is the same issue as reported by @phoyd in #13527 but not the same issue as what was reported originally so I am not sure it should be closed.

This PR does not change the triplets supported by this port.

Yes. (I've followed what seems to be the historical convention on this port to increment the version rather than introducing a port-version)
Fixes #13527

### DIFF
--- a/ports/icu/CONTROL
+++ b/ports/icu/CONTROL
@@ -1,6 +1,6 @@
 Source: icu
 Version: 67.1
-Port-Version: 4
+Port-Version: 5
 Homepage: http://icu-project.org/apiref/icu4c/
 Description: Mature and widely used Unicode and localization library.
 Supports: !(arm|uwp)
--- a/ports/icu/portfile.cmake
+++ b/ports/icu/portfile.cmake
@@ -25,7 +25,7 @@ vcpkg_extract_source_archive_ex(
 vcpkg_find_acquire_program(PYTHON3)
 set(ENV{PYTHON} "${PYTHON3}")
 
-set(CONFIGURE_OPTIONS "--disable-samples --disable-tests")
+set(CONFIGURE_OPTIONS "--disable-samples --disable-tests --disable-layoutex")
 
 if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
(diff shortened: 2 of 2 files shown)


phoyd:
--disable-layoutex fixes my issue in #13527, thanks!

ports/icu/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: icu
-Version: 67.1-3
+Version: 67.1-4
```
Hoikas, line 2:
```suggestion
Version: 67.1
Port-Version: 4
```

gergondet: Thanks. I've rebased on master since #13980 made the change and incremented the `Port-version`

LilyWangL (approved):

PR #14002, opened 2020-10-12
[vcpkg/configure_cmake] add LOGNAME option

Neumann-A:
preparation for #12936


PhoebeHui (approved):
Thanks for the improvements!

PhoebeHui:
The openxr_loader failed with the following failures:

CMake Error: install(EXPORT "openxr_loader_export" ...) includes target "openxr_loader" which requires target "jsoncpp_interface" that is not in any export set.
CMake Error in src/loader/CMakeLists.txt:
  export called with target "openxr_loader" which requires target
  "jsoncpp_interface" that is not in any export set.

strega-nil:
I'll merge this after all the non-osx stuff passes, since we just deleted the osx archives

PhoebeHui:
@strega-nil , all the non-osx triplet passed. ready to merge.

PR #14037, opened 2020-10-14
[cppwinrt] Update to version 2.0.201008.2

Link1J:
Updates C++/WinRT (cppwinrt) to version 2.0.201008.2. This required a new way of installing it because the repository no longer includes pre-built header files. This put more restrictions on what triplets are supported, limiting it to Windows only. It also now requires a version of the Windows SDK to be installed. With the large change needed to get the header files, a CMakeLists.txt was added, as to allow easier usage to with cmake. 

x86-windows and x64-windows are supported. arm-windows and arm64-windows may work on arm computers, I have no way to test. It may be possible to support all triplets as the cppwinrt.exe app only needs a header only library to build, but we also need the Windows' SDK for the new process.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? Yes, including a manifest file.


scripts/ci.baseline.txt
```diff
@@ -1854,3 +1854,11 @@ clapack:x86-windows=skip
...
 lapack-reference:arm64-windows=skip
 lapack-reference:arm-uwp=skip
+
+# cppwinrt.exe does not build on arm - this is known by cppwinrt and is even shown in its build files
+cppwinrt:arm64-windows = fail
+cppwinrt:arm-uwp       = fail
+
+# cppwinrt will not work on non windows platforms
+cppwinrt:x64-linux = skip
+cppwinrt:x64-osx   = skip
```
NancyLi1013, line 1864:
```suggestion
```
If this port cannot support these triplets, you can add `Supports: !(arm|osx|linux) ` to vcpkg.json file. There is no need to add these codes in ci.baseline.txt.

NancyLi1013 (changes requested):

Link1J:
Okay, I have made the recommended changes and a few more.

PR #14064, opened 2020-10-15
[cfitsio] Update to 3.49 and fix dependency zlib

JackBoosY:
Related: #13449.


JackBoosY:
@strega-nil Seems your patch doesn't work now, can you please take a look?

Thanks.

ports/cfitsio/CONTROL
```diff
@@ -1,5 +1,13 @@
 Source: cfitsio
-Version: 3.410
-Port-Version: 4
+Version: 3.49
```
strega-nil, line 2: I believe this should be `3.490`.

strega-nil (changes requested):

ports/cfitsio/CONTROL
strega-nil: Nope.

strega-nil (approved):

PR #14079, opened 2020-10-15
[vcpkg] Add `versions` feature flag and version field manifest parsing

ras0219:
* Introduce FeatureFlagSettings struct to more easily access feature flags throughout the program
* To avoid users accidentally starting to write "version" instead of "version-string" in their manifests, vcpkg explicitly detects and prevents usage of ports with schemes other than "String"

Example output when changing `abseil`'s field to just `"version"`:
```
PS C:\src\vcpkg> ./vcpkg install abseil
Error: while loading port `abseil`: Port definition rejected because the `versions` feature flag is disabled.
This can be fixed by using the "version-string" field.
```

* Drive-by fix of copiable SourceControlFileLocation and an exposed use-after-move bug

This code is largely extracted from PR #13777

### DIFF
--- a/toolsrc/include/vcpkg/versions.h
+++ b/toolsrc/include/vcpkg/versions.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <vcpkg/fwd/vcpkgpaths.h>
+
+#include <set>
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace vcpkg::Versions
+{
+    enum class Scheme
+    {
+        String,
+        Relaxed,
+        Semver,
+        Date
+    };
+}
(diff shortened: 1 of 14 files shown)


toolsrc/include/vcpkg/versions.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <vcpkg/fwd/vcpkgpaths.h>
+
+#include <set>
+#include <string>
+#include <tuple>
+#include <vector>
```
vicroms, line 8: Don't need all these `#include`

vicroms (approved):

vicroms:
Are you adding versioning scheme validation in this PR? 
E.g.: `version-dates` follows the date format described in the spec, etc.

ras0219:
I omitted it from this PR for simplicity

PR #14123, opened 2020-10-20
[vcpkg] Implement VersionedPortfileProvider and BaselineProvider

vicroms:
This PR implements two interfaces that are to be consumed by a versioned dependency resolver.
This is an important step in implementing package versioning in vcpkg.

```cpp
IVersionedPortfileProvider {
    get_port_versions(port_name) -> Version[]
    get_control_file(port_name, version) -> ControlFile
}

IBaselineProvider {
    get_baseline_version(port_name) -> Version
}
```

This is an MVP implementation that works with the default ports registry.


toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -242,4 +242,119 @@ namespace vcpkg::PortFileProvider
...
+                    auto version_obj = std::move(version.object());
+                    Versions::VersionSpec spec = extract_version_spec(package_spec, version_obj);
+
+                    auto& package_versions = versions_cache[spec.package_spec];
+                    package_versions.push_back(spec);
+
+                    auto&& git_tree = version_obj.get("git-tree")->string().to_string();
+                    git_tree_cache.emplace(std::move(spec), git_tree);
+                }
+            }
```
strega-nil, line 328:
```suggestion
            const auto& versions_object = versions_json.first.object();
            auto maybe_versions_array = versions_object.get("versions");

            if (maybe_versions_array && maybe_versions_array->is_array())
            {
                const auto& versions_array = maybe_versions_array->array();
                for (const auto& version : versions_array)
                {
                    const auto& version_obj = version.object();
                    Versions::VersionSpec spec = extract_version_spec(package_spec, version_obj);

                    auto& package_versions = versions_cache[spec.package_spec];
                    package_versions.push_back(spec);

                    auto git_tree = version_obj.get("git-tree")->string().to_string();
                    git_tree_cache.emplace(std::move(spec), git_tree);
                }
            }
```

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+        {
+            // TODO: Handle db version not existing
+            Checks::exit_fail(VCPKG_LINE_INFO);
+        }
+        auto versions_json_path = maybe_versions_json_path.value_or_exit(VCPKG_LINE_INFO);
+
+        auto versions_json = Json::parse_file(VCPKG_LINE_INFO, paths.get_filesystem(), versions_json_path);
+
+        // NOTE: A dictionary would be the best way to store this, for now we use a vector
+        if (versions_json.first.is_object())
```
ras0219-msft, line 333: Fail if not object?

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+        {
+            const auto& versions_object = versions_json.first.object();
+            auto maybe_versions_array = versions_object.get("versions");
```
ras0219-msft, line 336: This maybe should be `const auto&` -- `Json::Value` and friends are unfortunately copiable so it's easy to accidentally copy huge trees of objects. @strega-nil

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+
+            if (maybe_versions_array && maybe_versions_array->is_array())
```
ras0219-msft, line 338: Fail if not array?

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
 
         return ret;
     }
+
+    VersionedPortfileProvider::VersionedPortfileProvider(const vcpkg::VcpkgPaths& paths) : paths(paths) { }
+
+    const std::vector<vcpkg::Versions::VersionSpec>& VersionedPortfileProvider::get_port_versions(
+        const std::string& package_spec) const
```
ras0219-msft, line 314:
```suggestion
        const std::string& name) const
```

toolsrc/src/vcpkg/portfileprovider.cpp
ras0219-msft: This looks sufficiently "complex" that it would be good to use `JsonReader`/`IDeserializer` for proper error handling and messages.

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+            return cache_it->second;
+        }
+
+        // Pre-populate cache.
+        get_port_versions(version_spec.package_spec);
+
+        auto git_tree_cache_it = git_tree_cache.find(version_spec);
+        if (git_tree_cache_it == git_tree_cache.end())
+        {
+            // TODO: Try to load port from database
```
ras0219-msft, line 373: TODO

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+                for (const auto& version : versions_array)
+                {
+                    const auto& version_obj = version.object();
+                    Versions::VersionSpec spec = extract_version_spec(package_spec, version_obj);
+
+                    auto& package_versions = versions_cache[spec.package_spec];
+                    package_versions.push_back(spec);
+
+                    auto git_tree = version_obj.get("git-tree")->string().to_string();
+                    git_tree_cache.emplace(std::move(spec), git_tree);
```
ras0219-msft, line 350:
```suggestion
                    git_tree_cache.emplace(std::move(spec), std::move(git_tree));
```

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+
+        auto git_tree_cache_it = git_tree_cache.find(version_spec);
+        if (git_tree_cache_it == git_tree_cache.end())
+        {
+            // TODO: Try to load port from database
+            Checks::exit_fail(VCPKG_LINE_INFO);
+        }
+
+        const std::string git_tree = git_tree_cache_it->second;
+        const auto port_directory = paths.git_checkout_port(version_spec.package_spec, git_tree);
```
ras0219-msft, line 378:
```suggestion
        auto port_directory = paths.git_checkout_port(version_spec.package_spec, git_tree);
```

This can't be const if you want to move it below on line 385

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +456,70 @@ If you wish to silence this error and use classic mode, you can:
         return m_pimpl->m_tool_cache->get_tool_version(*this, tool);
     }
 
+    System::ExitCodeAndOutput VcpkgPaths::run_git_command(const std::string& dot_git_dir,
```
ras0219-msft, line 459: I think this would be more ideal if structured as:
```c++
System::CmdLineBuilder git_cmd_builder(const VcpkgPaths&, const std::string& git_dir, const std::string& work_tree);
```
Then the consumer can use the CmdLineBuilder functions to correctly append other arguments and pass the result to any of the `System::cmd_execute_xyz` functions.

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +456,70 @@ If you wish to silence this error and use classic mode, you can:
...
+                                                          const std::string& work_tree,
+                                                          const std::string& cmd_line) const
+    {
+        const fs::path& git_exe = get_tool_exe(Tools::GIT);
+
+        System::CmdLineBuilder builder;
+        builder.path_arg(git_exe)
+            .string_arg(Strings::concat("--git-dir=", fs::u8string(dot_git_dir)))
```
ras0219-msft, line 467:
```suggestion
            .string_arg(Strings::concat("--git-dir=", dot_git_dir))
```
`dot_git_dir` and `work_tree` aren't `fs::path`s.

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -263,4 +307,110 @@ namespace vcpkg::PortFileProvider
...
+                const auto& versions_array = maybe_versions_array->array();
+                for (const auto& version : versions_array)
+                {
+                    const auto& version_obj = version.object();
+                    Versions::VersionSpec spec = extract_version_spec(package_spec, version_obj);
+
+                    auto& package_versions = versions_cache[spec.package_spec];
+                    package_versions.push_back(spec);
+
+                    auto git_tree = version_obj.get("git-tree")->string().to_string();
```
ras0219-msft, line 349: We may want to sanitize this input and ensure it is a full commit sha (40 hex characters)

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +456,70 @@ If you wish to silence this error and use classic mode, you can:
...
+         * The `--git-dir` directory will store all the Git files,
+         * and the  `--work-tree` is the directory where checked out files will be stored.
+         *
+         * Since we are checking a git tree object, all files will be checked out to the root of `work-tree`.
+         * Because of that, it makes sense to use the git hash as the name for the directory.
+         */
+        const auto local_repo = this->root;
+        // TODO: Decide final location for these two
+        const auto dot_git_dir = this->buildtrees / "versioning_tmp" / ".git";
+        const auto working_dir = this->buildtrees / "versioning_tmp" / "versions" / port_name / git_tree;
```
ras0219-msft, line 486: These prefixes probably become members of `VcpkgPathsImpl` (e.g. `working_dir_prefix = this->buildtrees / "versioning_tmp" / "versions"`)

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +456,70 @@ If you wish to silence this error and use classic mode, you can:
...
+
+        auto& fs = get_filesystem();
+        if (!fs.exists(working_dir / "CONTROL") && !fs.exists(working_dir / "vcpkg.json"))
```
ras0219-msft, line 489: We should instead check for a stamp outside the directory (`working_dir / ".." / (git_tree+".stamp")`) that we expressly mint _after_ the git checkout has succeeded. This ensures that if git crashes halfway through checkout we don't have a corrupt directory.

toolsrc/include/vcpkg/vcpkgpaths.h
```diff
@@ -133,5 +137,9 @@ namespace vcpkg
 
     private:
         std::unique_ptr<details::VcpkgPathsImpl> m_pimpl;
+
+        System::ExitCodeAndOutput run_git_command(const std::string& dot_git_dir,
```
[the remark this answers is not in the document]
vicroms: That's how I originally implemented it, but I wasn't sure what our design philosophy is regarding VcpkgPaths 😅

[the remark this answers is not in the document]
vicroms: Removed this function in favor of using a CmdLineBuilder helper.

toolsrc/src/vcpkg/vcpkgpaths.cpp
vicroms: Done

toolsrc/src/vcpkg/vcpkgpaths.cpp
vicroms: Fixed, made parameters `fs::path`

toolsrc/src/vcpkg/vcpkgpaths.cpp
vicroms: Done

toolsrc/src/vcpkg/portfileprovider.cpp
vicroms: This is next on my to-do list

toolsrc/src/vcpkg/portfileprovider.cpp
vicroms: We do a `get_port_versions(version_spec.port_name)` just before this part to pre-populate the cache. This should now return an error.

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +496,183 @@ If you wish to silence this error and use classic mode, you can:
...
     }
 
+    void VcpkgPaths::git_checkout_subpath(const VcpkgPaths& paths,
+                                          const std::string& commit_sha,
+                                          const fs::path& subpath,
+                                          const fs::path& local_repo,
+                                          const fs::path& destination,
+                                          const fs::path& dot_git_dir,
+                                          const fs::path& work_tree)
+    {
```
ras0219, line 506: After a bit of experimenting, I found this interesting procedure:
```
C:\src\vcpkg>git archive 9ed51699b767864712ca696dbe2670fe79899e41 | (cd docsfoo && tar xf -)
```
This accepts a treeish object and writes it out in tar format (uncompressed) to stdout. I then pipe it into `C:\Windows\System32\tar.exe` which extracts to the current working directory. This allows us to perform a "checkout" entirely read-only.

Notes:
- `git.exe` must be run from a "tracked" directory (or the `.git` folder)
- `tar.exe` is available as part of the system on win10 17063+
- `cmake -E tar` can't read from standard in
- `tar.exe` is also available in our git tools suite: `downloads\tools\git-2.26.2-1-windows\usr\bin\tar.exe`

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -7,6 +8,56 @@
 #include <vcpkg/sourceparagraph.h>
 #include <vcpkg/vcpkgcmdarguments.h>
 #include <vcpkg/vcpkgpaths.h>
+#include <vcpkg/versiondeserializers.h>
+
+using namespace vcpkg;
+
+namespace
+{
+    vcpkg::Versions::VersionSpec extract_version_spec(const std::string& port_name, const Json::Object& version_obj)
```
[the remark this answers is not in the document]
vicroms: I think this is no longer getting used anywhere.

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -257,4 +308,154 @@ namespace vcpkg::PortFileProvider
...
+            const auto& versions_object = versions_json.first.object();
+            auto maybe_versions_array = versions_object.get("versions");
+
+            if (maybe_versions_array && maybe_versions_array->is_array())
+            {
+                const auto& versions_array = maybe_versions_array->array();
+                for (const auto& version : versions_array)
+                {
+                    const auto& version_obj = version.object();
+                    Versions::VersionSpec spec = extract_version_spec(port_name, version_obj);
```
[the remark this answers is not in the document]
vicroms: This was removed in favor of a deserializer.

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -257,4 +287,188 @@ namespace vcpkg::PortFileProvider
...
+
+        const std::string git_tree = git_tree_cache_it->second;
+        auto port_directory = m_impl->paths.git_checkout_port(version_spec.port_name, git_tree);
+
+        auto maybe_control_file = Paragraphs::try_load_port(m_impl->paths.get_filesystem(), port_directory);
+        if (auto scf = maybe_control_file.get())
+        {
+            if (scf->get()->core_paragraph->name == version_spec.port_name)
+            {
+                return SourceControlFileLocation{std::move(*scf), std::move(port_directory)};
```
ras0219-msft, line 406: This is returning a reference to a temporary instead of inserting into the map.
```cpp
            if (scf->get()->core_paragraph->name == version_spec.name)
            {
                return m_impl->control_cache
                    .emplace(version_spec, SourceControlFileLocation{std::move(*scf), std::move(port_directory)})
                    .first->second;
            }
```

vicroms:
@strega-nil @BillyONeal @ras0219
 
I think this is ready for merge now. I will investigate the `git archive` method in a separate PR.

scripts/generatePortVersionsDb.py
```diff
@@ -4,6 +4,7 @@
 import subprocess
```
strega-nil, line 4: Did this go through pep8?

toolsrc/include/vcpkg/portfileprovider.h
```diff
@@ -36,4 +37,51 @@ namespace vcpkg::PortFileProvider
         std::vector<fs::path> overlay_ports;
         mutable std::unordered_map<std::string, SourceControlFileLocation> cache;
     };
+
+    struct IVersionedPortfileProvider
+    {
+        virtual const std::vector<vcpkg::Versions::VersionSpec>& get_port_versions(
+            const std::string& port_spec) const = 0;
```
strega-nil, line 44:
```suggestion
            StringView port_spec) const = 0;
```

best to change the rest as well.

toolsrc/include/vcpkg/vcpkgpaths.h
```diff
@@ -107,6 +107,10 @@ namespace vcpkg
         const fs::path& get_tool_exe(const std::string& tool) const;
         const std::string& get_tool_version(const std::string& tool) const;
 
+        // Git manipulation
+        fs::path git_checkout_baseline(const std::string& commit_sha) const;
```
strega-nil, line 111:
```suggestion
        fs::path git_checkout_baseline(StringView commit_sha) const;
```

same for the rest

toolsrc/include/vcpkg/vcpkgpaths.h
```diff
@@ -133,5 +137,20 @@ namespace vcpkg
 
     private:
         std::unique_ptr<details::VcpkgPathsImpl> m_pimpl;
+
+        static void git_checkout_subpath(const VcpkgPaths& paths,
+                                         const std::string& commit_sha,
```
strega-nil, line 142:
```suggestion
                                         StringView commit_sha,
```

same with the rest.

toolsrc/include/vcpkg/versiondeserializers.h
```diff
@@ -0,0 +1,152 @@
...
+#include <vcpkg/versions.h>
+#include <vcpkg/versiont.h>
+
+namespace vcpkg
+{
+    struct VersionDbEntry
+    {
+        const VersionT version;
+        const Versions::Scheme scheme;
+        const std::string git_tree;
```
strega-nil, line 17:
```suggestion
        VersionT version;
        Versions::Scheme scheme;
        std::string git_tree;
```

toolsrc/include/vcpkg/versions.h
```diff
@@ -10,6 +12,25 @@ namespace vcpkg::Versions
         String
     };
 
+    struct VersionSpec
+    {
+        const std::string port_name;
+        const VersionT version;
+        const Scheme scheme;
```
strega-nil, line 19:
```suggestion
        std::string port_name;
        VersionT version;
        Scheme scheme;
```

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -257,4 +287,190 @@ namespace vcpkg::PortFileProvider
...
+        // Avoid warning treated as error.
+        if (maybe_versions_array != nullptr)
+        {
+            Json::Reader r;
+            std::vector<VersionDbEntry> db_entries;
+            r.visit_in_key(*maybe_versions_array, "versions", db_entries, VersionDbEntryArrayDeserializer::instance);
+
+            for (const auto& version : db_entries)
+            {
+                std::regex is_commit_sha("^[\\da-f]{40}$");
```
strega-nil, line 362: should this allow uppercase as well?

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -257,4 +287,190 @@ namespace vcpkg::PortFileProvider
 
         return ret;
     }
+
+    namespace details
+    {
+        struct BaselineProviderImpl
+        {
+            const VcpkgPaths& paths;
+            const std::string baseline;
```
strega-nil, line 296: I don't trust these; why are they non-static member variables of reference/const type?

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -257,4 +287,190 @@ namespace vcpkg::PortFileProvider
...
+            BaselineProviderImpl(const VcpkgPaths& paths, const std::string& baseline)
+                : paths(paths), baseline(baseline)
+            {
+            }
+            ~BaselineProviderImpl() { }
+        };
+
+        struct VersionedPortfileProviderImpl
+        {
+            const VcpkgPaths& paths;
```
strega-nil, line 309: Same here

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +496,183 @@ If you wish to silence this error and use classic mode, you can:
...
+        }
+    }
+
+    fs::path VcpkgPaths::git_checkout_baseline(const std::string& commit_sha) const
+    {
+        const fs::path local_repo = this->root;
+        const fs::path destination = m_pimpl->baselines_output(*this) / commit_sha / fs::u8path("baseline.json");
+        const fs::path baseline_subpath = fs::u8path("port_versions") / fs::u8path("baseline.json");
+
+        if (!get_filesystem().exists(destination))
```
strega-nil, line 637: I don't like this; this function should take a `Filesystem&`

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -456,6 +496,183 @@ If you wish to silence this error and use classic mode, you can:
...
+         * The `--git-dir` directory will store all the Git metadata files,
+         * and the  `--work-tree` is the directory where files will be checked out.
+         *
+         * Since we are checking a git tree object, all files will be checked out to the root of `work-tree`.
+         * Because of that, it makes sense to use the git hash as the name for the directory.
+         */
+        const fs::path local_repo = this->root;
+        const fs::path destination = m_pimpl->versions_output(*this) / fs::u8path(git_tree) / fs::u8path(port_name);
+
+        Files::Filesystem& fs = get_filesystem();
```
strega-nil, line 663: same here

toolsrc/src/vcpkg/versions.cpp
```diff
@@ -0,0 +1,34 @@
...
+    bool operator!=(const VersionSpec& lhs, const VersionSpec& rhs) { return !(lhs == rhs); }
+
+    std::size_t VersionSpecHasher::operator()(const VersionSpec& key) const
+    {
+        using std::hash;
+        using std::size_t;
+        using std::string;
+
+        return ((hash<string>()(key.port_name) ^ (hash<string>()(key.version.to_string()) << 1)) >> 1) ^
+               (hash<int>()(static_cast<int>(key.scheme)) << 1);
```
strega-nil, line 32: Reminder to self to add hashing to the vcpkg lib.

strega-nil (changes requested):

scripts/generatePortVersionsDb.py
vicroms: It went through `Alt+ Shift + F` in VS Code.

toolsrc/src/vcpkg/portfileprovider.cpp
vicroms: Sorry, I don't follow. What's the problem with them being non-static?

ras0219: non-static _reference/const_ members prevent the type from being used like an `int`; for example,
```cpp
struct CFoo {
  const int x;
  const int y;
};
CFoo a{1, 2};
CFoo b{3, 4};

b = a; // doesn't compile
```
In most cases where you'd use const members, you really just want non-const members and to declare the actual variable / member functions as const:
```cpp
struct Foo {
  int x, y;
};
const Foo a{1,2};
Foo b{3,4};

b = a; // Works!
a = b; // Doesn't compile, as expected
```

Reference members in this context function similarly to a constant pointer: `int * const x` and cause similar problems.

Reference members _can_ make sense for Resource-types that you know you'll never assign to, such as proper Objects participating in inheritance hierarchies that are heap allocated and immediately "hidden" behind a pointer:
```cpp
// hidden in an implementation file
struct Derived : IBase {
    Derived(Other& o) : m_other(o) {}

private:
    Other& m_other; // ok because Derived is not ever intended for copying, assigning, or moving.
};

// this is the only exposed public interface
std::unique_ptr<IBase> make_default_frobnicator(Other& o) {
  return std::make_unique<Derived>(o);
}
```

Compiler Explorer link: https://godbolt.org/z/ThqxG5

ras0219: In the above context, I think that these uses are actually fine because these _are_ non-movable/copiable resources that will probably change to directly inherit from the relevant interfaces (`IBaselineProvider`, etc).

PR #14174, opened 2020-10-23
[vcpkg] Include "feature-packages" functionality in user help

vmiheer:
I know it might be bit controversial that this link already exist in specifications but as a user of vcpkg I would definitely ignore "specifications" section while trying to find info about how to enable/disable features.

**Describe the pull request**


BillyONeal (changes requested):
I think this needs to be actual user documentation rather than just linking to an implementation spec

PR #11758, opened 2020-06-03
[vcpkg] RFC: Versioning

vicroms:
We are currently working on implementing package versioning, the model is based on Go's module management implementation with some modifications.

This feature allows users to specify package versions to be installed as well as allowing packages to specify the versions of their dependencies.

There are some rough edges that need to be smoothed out, we appreciate any feedback that helps improve the current design.

P.S.: [is easier to read the spec directly from the branch](https://github.com/vicroms/vcpkg/blob/versioning-spec/docs/specifications/versioning.md).

Closes #1681 #10674

PR #14183, opened 2020-10-23
[vcpkg] Implement constraints and overrides in manifests

ras0219:
This PR implements parsing of constraints and overrides for manifest files as specified in #11758.

It includes guard logic to prevent accidental usage of these new features in manifest files while the experimental `versioning` feature flag is not specified; this ensures users will not be broken as we continue to develop.

See especially the tests in `toolsrc/src/vcpkg-test/manifests.cpp` for exactly compiling examples.


toolsrc/include/vcpkg/versions.h
```diff
@@ -11,4 +11,14 @@ namespace vcpkg::Versions
...
+
+    struct Constraint
+    {
+        enum class Type
+        {
+            None,
+            Minimum,
+            Exact
+        };
+    };
```
vicroms, line 23: Why have `Constraing` and `DependencyConstraint`? Can't we move the enum to `DependencyConstraint` and get rid of this one?

toolsrc/src/vcpkg/packagespec.cpp
```diff
@@ -269,4 +277,14 @@ namespace vcpkg
...
     }
     bool operator!=(const Dependency& lhs, const Dependency& rhs);
+
+    bool operator==(const DependencyOverride& lhs, const DependencyOverride& rhs)
+    {
+        if (lhs.version_scheme != lhs.version_scheme) return false;
+        if (lhs.port_version != lhs.port_version) return false;
+        if (lhs.name != rhs.name) return false;
+        if (lhs.version != lhs.version) return false;
+        return lhs.extra_info == rhs.extra_info;
```
vicroms, line 287:
```suggestion
        if (lhs.version_scheme != rhs.version_scheme) return false;
        if (lhs.port_version != rhs.port_version) return false;
        if (lhs.name != rhs.name) return false;
        if (lhs.version != rhs.version) return false;
        return lhs.extra_info == rhs.extra_info;
```

toolsrc/include/vcpkg/versions.h
ras0219: I wanted to minimize the delta with your outstanding separate PR, however your `Constraint` type included a field for port name and so wasn't appropriate to use as a member of `Dependency`. Thus I added the minimum shared subset of `Constraint` (`Constraint::Type`).

Once both PRs are in, it will be easy to consolidate where appropriate.

toolsrc/src/vcpkg/packagespec.cpp
ras0219: oof, thanks

vicroms (approved):

vicroms:
@ras0219 is there any work left here?

PR #14186, opened 2020-10-23
[icu] Add MSVC ARM64 support

janisozaur:
- What does your PR fix?

This PR adds support for building icu using MSVC for ARM64.

ARM64, yes.

Yes

### DIFF
--- a/ports/icu/portfile.cmake
+++ b/ports/icu/portfile.cmake
@@ -1,4 +1,4 @@
-vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp")
+vcpkg_fail_port_install(ON_TARGET "uwp")
 
 set(ICU_VERSION_MAJOR 67)
 set(ICU_VERSION_MINOR 1)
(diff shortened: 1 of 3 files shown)


strega-nil:
Cool, thanks @janisozaur :)

This should be fixed in the future by tool dependencies :)

ports/icu/portfile.cmake
```diff
@@ -1,239 +1,251 @@
...
+        set(ICU_RUNTIME "-MD")
+    endif()
+
+    if(CMAKE_HOST_WIN32 AND (VCPKG_TARGET_ARCHITECTURE STREQUAL "arm" OR VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64") AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "MinGW")
+        set(ICU_MSVC_CROSS_COMPILE_TO_ARM ON)
+        # Need the buildtrees dir, as the required files (e.g. icucross.mk) are not part of the installed package
+        get_filename_component(ICU_HOST_PATH "${BUILDTREES_DIR}/icu/x86-windows-rel" ABSOLUTE)
+        if(NOT EXISTS "${ICU_HOST_PATH}")
+            message(FATAL_ERROR "The x86 icu must be be built locally to build for non-x86/x64 platforms. Please run `vcpkg install icu:x86-windows`.")
+        endif()
```
davidebeatrici, line 103: Why is this required?

janisozaur: There's some data generated on the host. It could be improved a tiny bit to skip this requirement if host==target==arm, but I suspect it's a niche use case at this time

davidebeatrici: That's what I thought, thanks for confirming.

The problem with this approach is that it only works when `icu` is installed for `x86-windows`. In my case it was installed for `x86-windows-static-md`.

janisozaur: The triplet used we based on similar issue with boost, where every `boost-*` depends on `boost-build:x86-windows`

davidebeatrici: Oh, I see.

janisozaur: It seems there is a better way to address that kind of problem now: https://devblogs.microsoft.com/cppblog/vcpkg-host-dependencies/

davidebeatrici: Awesome!

autoantwort: Would be fixed by #16577 :) Then you can set your host triplet to `x86-windows-static-md` and icu will use that instead of `x86-windows`

davidebeatrici: Thank you very much!

PR #14202, opened 2020-10-24
[vcpkg] Remove hard-coded ApiKey from nuget push

### LINKED ISSUE #12984: nugetconfig binary source ignores API key in config file
For `nugetconfig` binary caching sources, vcpkg passes `-ApiKey AzureDevOps` as an option to the `nuget push` command, which overrides the API key specified in the nuget config file. This prevents pushing to corporate nuget repositories that require specific credentials.

**Proposed solution**
Remove `-ApiKey AzureDevOps` from the `nuget push` command for `nugetconfig` sources, or provide a vcpkg flag that causes it to be left off or overwritten with the real API key.

Add a `write`-only `nuget` binary source pointing at a temporary local directory, and then, after running `vcpkg install`, run a separate `nuget` command manually without the `-ApiKey` flag. This has at least a few disadvantages: more complicated config for separate `read` and `write` sources, requiring an extra step to upload and then cleanup files vs. it happening automatically, and not being able to rely on the copy of nuget vcpkg downloads for itself to use.

ekilmer:
**Describe the pull request**
After experimenting with vcpkg binary caching feature and trying to upload the built binaries to GitHub Packages using NuGet, I ran into a similar authentication issue as #12984.

This PR removes the hard-coded NuGet API key so that a user's specified key can be used in a `VCPKG_BINARY_SOURCES` environment variable.

This was my first time using NuGet and setting everything up with GitHub Packages, so if this hard-coded key name is somehow intentional, I'd like to know how to work around this or what a better binary caching setup would look like.

- What does your PR fix?
Fixes #12984 

N/A

Yes.

### DIFF
--- a/toolsrc/src/vcpkg/binarycaching.cpp
+++ b/toolsrc/src/vcpkg/binarycaching.cpp
@@ -393,8 +393,6 @@ namespace
                     cmd.path_arg(nuget_exe)
                         .string_arg("push")
                         .path_arg(nupkg_path)
-                        .string_arg("-ApiKey")
-                        .string_arg("AzureDevOps")
                         .string_arg("-ForceEnglishOutput")
                         .string_arg("-Source")
                         .string_arg(write_src);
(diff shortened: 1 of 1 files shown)


ras0219 (approved):
Thanks for using binary caching and extra thanks the fix!

I have tested this against ADO artifacts feeds and confirmed that it no longer needs an ApiKey parameter, so this LGTM.

PR #14243, opened 2020-10-26
[vcpkg] X_VCPKG_APPINSTALL_DEPS_INSTALL optionally install dependencies on install

sandercox:
Continuing on the work from #14129 #13011 this adds an install command override again as proposed initially but behind an `if` check for the variable `X_VCPKG_APPLOCAL_DEPS_INSTALL` marking it as experimental and allowing the feature to _just work_ as I initially intended.

I was already using the install function as used here in my own projects but I think this allows everybody to really use the functionality introduced by the `x_vcpkg_install_local_dependencies` function.

As originally requested in #1653

### DIFF
--- a/scripts/buildsystems/vcpkg.cmake
+++ b/scripts/buildsystems/vcpkg.cmake
@@ -469,6 +469,46 @@ function(x_vcpkg_install_local_dependencies)
     endif()
 endfunction()
 
+set(X_VCPKG_APPLOCAL_DEPS_INSTALL ${X_VCPKG_APPLOCAL_DEPS_INSTALL} CACHE BOOL "(experimental) Automatically copy dependencies into the install target directory for executables.")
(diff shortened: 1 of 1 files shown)


Shatur (changes requested):
Works as expected for me with install command and CPack.
One small suggestion: you don't need to expand variables in `if` expressions, this will be done automatically. Elsewhere in this file, the same approach is used, so it's best to do the same for consistency. :) I also tested with this changes, everything works fine for me.

Shatur (approved):

JackBoosY:
cc @BillyONeal for review this PR.

strega-nil:
Given that this is an experimental feature, I'm up for merging it, with the understanding that this very well may be removed or modified, especially as we get closer to stabilization :)

PR #14303, opened 2020-10-30
[vcpkg] Add support to more paths for vcpkg usage hint

### LINKED ISSUE #10881: vcpkg install: Usage hint for CMake find_package, incomplete support for possible paths
**Describe the bug**
Vcpkg does not provide the right usage hint for  `<pkg_name>Config.cmake` files installed into a path ending in `cmake`.
E.g. `/share/Microsoft.GSL/cmake/Microsoft.GSLConfig.cmake`, used in [Microsoft/GSL/CMakeLists.txt](https://github.com/microsoft/GSL/blob/b43855631afdb9f7ccd4b56ed05330a8e3817af1/CMakeLists.txt#L92).

Without the work-around added in the portfile for #10831 `vcpkg install ms-gsl` returns the following usage hint:
````CMake
The package ms-gsl:x86-windows provides CMake targets:

    find_package(cmake CONFIG REQUIRED)
    target_link_libraries(main PRIVATE Microsoft.GSL::GSL)
````
Note the `cmake`  in `find_package`, this should be `Microsoft.GSL`.

The current implementation shows a package name taken from the wrong folder.

**Proposed solution**
An update of the algorithm used to determine the package name.
There might be more valid paths that are not matched. (See the table below.)

Fixes, such as added to #10831, moving the files to a recognised path, are easy to forget and hard to maintain.
And they should not be necessary for projects using a valid path.

**Additional context**
> The tables below show the directories searched. Each entry is meant for installation trees following Windows (W), UNIX (U), or Apple (A) conventions:
> ````
> <prefix>/                                                       (W)
> <prefix>/(cmake|CMake)/                                         (W)
> <prefix>/<name>*/                                               (W)
> <prefix>/<name>*/(cmake|CMake)/                                 (W)
> <prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         (W/U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               (W/U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/ (W/U)
> ````
Source: [CMake documentation - `find_package` Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure)
****

schultetwin1:
Currently vcpkg will auto-generate a usage hint for installed packages.
These usage hints support the following paths under `/share/`:

* `<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/`
* `<prefix>/(lib/<arch>|lib*|share)/<name>*/`
* `<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/`
* `<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/`

This PR adds support for:

* `<prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/`
* `<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/`

You can read more about these paths here:
https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

Fixes #10881

### DIFF
--- a/toolsrc/src/vcpkg/install.cpp
+++ b/toolsrc/src/vcpkg/install.cpp
@@ -625,6 +625,11 @@ namespace vcpkg::Install
                     auto path = paths.installed / suffix;
                     auto maybe_contents = fs.read_contents(path);
                     auto find_package_name = fs::u8string(path.parent_path().filename());
+                    if (Strings::case_insensitive_ascii_equals("cmake", find_package_name) &&
+                        path.parent_path().has_parent_path())
+                    {
+                        find_package_name = fs::u8string(path.parent_path().parent_path().filename());
+                    }
                     if (auto p_contents = maybe_contents.get())
                     {
                         std::sregex_iterator next(p_contents->begin(), p_contents->end(), cmake_library_regex);


JackBoosY:
Thanks for this PR! Sadly, since all cmake files are placed in `<prefix>/share/<name>` in vcpkg, this PR is invalid.

schultetwin1:
@JackBoosY 

> since all cmake files should be placed in <prefix>/share/<name> in vcpkg

Is this documented somewhere? I looked but couldn't find anything. `<prefix>/share/<name>/CMake` is a valid install path for a CMake module, why does vcpkg care where it gets installed?

PR #14310, opened 2020-10-30
[vcpkg] Add korean version Readme

AkiaCode:
This PR add korean readme and I made it with @jiyuPk


strega-nil:
Thanks so much @AkiaCode 😄  I love all the readmes we're starting to collect 😄

Issue #2662, opened 2018-01-27
Add pkg-config like options


galexite:
`pkg-config --cflags --libs somepackage` on a Linux distribution would return compiler flags and linker options for linking with an installed library, allowing pretty much any build system to find those and link.

vcpkg seems to only support CMake as an alternative build system. If something analogous to `pkg-config`  was created, providing locations of libraries, include directories, and MSVC flags, other build systems could be made aware of libraries installed using vcpkg.

purell:
I played a little bit with `vcpkg`+`pkg-config-lite`+`meson` so I will resume my experience...

Some package like `protobuf` install pkg-config file and produce the `*.pc` file required by pkg-config program to find library. 

Ex :
```bash
D:.
├───vcpkg
└───x86-windows
    ├───bin
    │       libprotobuf-lite.dll
    │       libprotobuf-lite.pdb
    │       libprotobuf.dll
    │       libprotobuf.pdb
    │       libprotoc.dll
    │       libprotoc.pdb
    │
    ├───lib
    │   │   libprotobuf-lite.lib
    │   │   libprotobuf.lib
    │   │   libprotoc.lib
    │   │
    │   └───pkgconfig
    │           protobuf-lite.pc
    │           protobuf.pc
```

Install and configure vc-pkg
```bash
> D:\
> git clone https://github.com/Microsoft/vcpkg.git
> cd vcpkg
> .\bootstrap-vcpkg.bat
> .\vcpkg integrate install
> .\vcpkg install zlib libsodium protobuf
```

Now install [pkg-config-lite](https://sourceforge.net/projects/pkgconfiglite/) and configure it.
```bash
> set PKG_CONFIG_PATH=%PKG_CONFIG_PATH%;D:\Data\vcpkg\packages\protobuf_x86-windows\lib\pkgconfig
```

Use [meson](https://mesonbuild.com/) to build a c++ program that link to protobuf
```meson
project('myprogram', 'cpp', version : '1.0')

src = ['main.cpp'] 

protobuf = dependency('protobuf', version: '>=3.0')

executable('myprogram',
           sources: src,
           dependencies: [protobuf])
```

This work perfectly as pkg-config-lite find the `libprotobuf.pc` file by looking at the `%PKG_CONFIG_PATH%` path.

[...]

PR #14143, opened 2020-10-21
[Proj4] Update to 7.1.1

Matioupi:
Bump proj4 version to 7.1.1

- What does your PR fix? Fixes #

- Which triplets are supported/not supported? Have you updated the CI baseline?
x64-windows tested locally

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

it should. I've seen PR #13428 on the same subject.

PR #14328, opened 2020-10-31
[vcpkg] vcpkg_pkgconfig helper function

Matioupi:
This PR adds a new helper function : vcpkg_pkgconfig in order to easily retrieve the pkgconfig link information from installed packages.
I found that it helps fixing some ports when building tool feature in static configurations.
I also added wldap32 and crypt32 to VCPKG_SYSTEM_LIBRARIES for windows target. I don't know if there is a rationale for not having them there by default.

Some issues e.g. https://github.com/microsoft/vcpkg/issues/2662 seems to request such function and I had the same feeling.

It should, also I have not seen much regarding PR that are not about ports.


JackBoosY:

Neumann-A:
Should probably use the functions defined in the FindPkgConfig module https://cmake.org/cmake/help/v3.19/module/FindPkgConfig.html

But I am missing an application for this within VCPKG. In general the build scripts of a port need to correctly discover dependencies not the portfiles.

Matioupi:
I'll try to modify and use the FindPkgConfig function instead of what I've tried.
As for the use case, I agree that  the build scripts should be responsible in first place, but I found cases where building tools with static configuration triplets would fail.
On example I have it the build of proj 7.1.1 (which is also a PR I authored : https://github.com/microsoft/vcpkg/pull/14143)
When building this package with x64-windows-static, and utilities enabled, then build fail in some cases (e.g. when curl is built with additional features) and the failure is a link stage because the additional link libraries associated to curl features are not added to linker.
I understand it may be considered (and probably is) a dirty fix, but adding those libs thought the VCPKG_LINKER_FLAGS_RELEASE/DEBUG variable at port file level provides an easy way to get around this.
the helper function I proposed was developed in that exact use case frame.

Matioupi:
I see now that the FindPkgConfig in addition to finding the executable path also has a pkg_get_variable that probably deprecates the need for this PR...

PR #14376, opened 2020-11-03
[libigl] Fix dynamic build and enable header-only mode

### LINKED ISSUE #14371: [libigl] Enable header-only mode
The libigl port is now only available as a (huge) static library (>1GB). The library is heavily template based thus it would be preferable to keep the header-only mode.

I was attempting to use the port but encountered linker error problems because the types passed as arguments, despite being valid, were not explicitly instantiated by the library during vcpkg build.

**Proposed solution**
Removing the build only static constraint and fix what breaks 😄 

The explicit instantiations could be added on the libigl side, but this is not a long-term solution due to the combinatorial nature of templates. One may always reach a point where the needed template was not instantiated, and also it raises the problem of binary size (which is already quite extreme here)

FabienPean:
This is an attempt at getting rid of libigl huge static .lib file and getting the header-library mode. I could make it work on my machine (Windows VS2019 Clang 10) with the current set of modifications. Not necessarily optimal nor "clean", but hopefully a good start to resolve the issue. 

Let me know if anything is missing or feel free to take over from here.

- What does your PR fix? Fixes #14371 

Dynamic library version works on windows but not the static, and other platform untested. I don't know what is the CI baseline.

I think so.

### DIFF
--- a/ports/libigl/CONTROL
+++ b/ports/libigl/CONTROL
@@ -1,6 +1,6 @@
 Source: libigl
 Version: 2.2.0
-Port-Version: 2
+Port-Version: 3
 Homepage: https://github.com/libigl/libigl
 Description: libigl is a simple C++ geometry processing library. We have a wide functionality including construction of sparse discrete differential geometry operators and finite-elements matrices such as the cotangent Laplacian and diagonalized mass matrix, simple facet and edge-based topology data structures, mesh-viewing utilities for OpenGL and GLSL, and many core functions for matrix manipulation which make Eigen feel a lot more like MATLAB.
 Build-Depends: eigen3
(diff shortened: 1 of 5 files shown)


FabienPean:
Alright so, despite compiling successfully, the baseline x64-windows-static is not actually usable. It leads to `/failifmismatch: mismatch detected for '_ITERATOR_DEBUG_LEVEL'`. Current changes also successfully pass the CI and leads to same problem.

However my question is, should `vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)` be imposed and `ci.baseline.txt` updated to `libigl:x64-windows-static=fail` or stick to "if it compiles it is fine"

Otherwise, I am out of idea to make the static build actually be usable. Since it does not alter the original behavior (on x64-windows) the PR should be ok, but could be simplified if dynamic only is decided (I most probably added too many `if(LIBIGL_USE_STATIC_LIBRARY)` initially hoping to solve the static build issue).

Let me know your opinion on the subject

FabienPean:
The CI x86_windows instance is hogging.

For convenience, I made the gist with the test files here (requires building igl from head for fixes on writeDAE): https://gist.github.com/FabienPean/9c1a6c5eea6b304d0e1241024e0387b1

LilyWangL (approved):

strega-nil:
Thanks @FabienPean :)

PR #14392, opened 2020-11-04
[tensorflow] Add C port

jgehw:
This PR adds a C port for tensorflow. As large parts are identical to the C++ port [tensorflow-cc], I moved them into a common file.

I suggest to wait with reviewing until #13028 is merged into master, because this PR builds upon #13028, and once #13028 is merged the file diffs of this PR will collapse to just a few simple changes.


jgehw:
@NancyLi1013 dependent PR is now merged into master, so this PR is ready for review and merge

While renaming
ports/tensorflow-cc/tensorflow-cc-config-shared.cmake.in -> ports/tensorflow-cc/tensorflow-config-shared.cmake.in
ports/tensorflow-cc/tensorflow-cc-config-static.cmake.in -> ports/tensorflow-cc/tensorflow-config-static.cmake.in
is correctly displayed in the code diff view, it didn't work for
ports/tensorflow-cc/tensorflow-cc-config-windows-dll.cmake.in -> ports/tensorflow-cc/tensorflow-config-windows-dll.cmake.in
ports/tensorflow-cc/tensorflow-cc-config-windows-lib.cmake.in -> ports/tensorflow-cc/tensorflow-config-windows-lib.cmake.in
so you may want to manually diff them to ease the review (it's just a few changes). Especially, you may want to diff:
ports/tensorflow-cc/portfile.cmake -> ports/tensorflow-cc/tensorflow-common.cmake

ports/tensorflow-cc/CONTROL
```diff
@@ -1,4 +1,4 @@
 Source: tensorflow-cc
 Version: 2.3.1
```
NancyLi1013, line 2:
```suggestion
Version: 2.3.1
Port-Version: 1
```

ports/tensorflow/portfile.cmake
```diff
@@ -0,0 +1,16 @@
+set(TF_LIB_SUFFIX "")
+set(TF_PORT_SUFFIX "")
+set(TF_PATCHES_PREFIX "../tensorflow-cc/")
+set(TF_INCLUDE_DIRS "${INSTALL_PREFIX}/${TARGET_TRIPLET}/include")
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}-cc)
+include(tensorflow-common)
```
NancyLi1013, line 6: `tensorflow-common` seems not exist on master branch 
https://github.com/microsoft/vcpkg/tree/master/ports/tensorflow-cc

ports/tensorflow/portfile.cmake
```diff
@@ -0,0 +1,16 @@
+set(TF_LIB_SUFFIX "")
```
NancyLi1013, line 1:
```suggestion
vcpkg_fail_port_install(ON_ARCH "x86" "arm" ON_TARGET "uwp")

set(TF_LIB_SUFFIX "")
```
It would be better to add this unsupported messages here.

NancyLi1013 (changes requested):

ports/tensorflow/portfile.cmake
jgehw: Yes, on master branch everything is in portfile.cmake. Because 95% of the [tensorflow-cc] port code can be shared with the [tensorflow] port, I moved the common code from portfile.cmake into tensorflow-common.cmake.

ports/tensorflow/CONTROL
```diff
@@ -0,0 +1,6 @@
+Source: tensorflow
+Version: 2.3.1
+Port-Version: 1
```
NancyLi1013, line 3:
```suggestion
Port-Version: 0
```
Since this is a new port, there is no need to update this field or add this filed.

JackBoosY (approved):

ports/tensorflow-cc/tensorflow-common.cmake
```diff
@@ -0,0 +1,439 @@
+set(TF_VERSION 2.3.1)
```
JackBoosY, line 1: Maybe we should separate this to a new port, like `boost-build`.

jgehw: `tensorflow-common.cmake` is not runnable on its own; it's included by portfile.cmake. If we'd create a new port, say `tensorflow-build`, this port wouldn't do anything but just copy its contents to `installed/share/tensorflow-build`, from where it would be included by the portfile then instead. If you'd like me to move all the common part elsewhere, how about adding a port `tensorflow-common` which is not installable (portfile calls vcpkg_fail) but just holds the common files which can then be referenced by `../tensorflow-common/` in the tensorflow[-cc] portfiles?

JackBoosY: @jgehw Yes, that's what I think.

jgehw: ok, I'll do it.

strega-nil (commented):
Needs a few minor things in the python, but that's it.

strega-nil:
@jgehw have you tried out tensorflow at it's latest version on your local machine?

jgehw:
> @jgehw have you tried out tensorflow at it's latest version on your local machine?

yes, I verified it on my local machine

strega-nil:
Alright, thanks so much @jgehw :D

Issue #14252, opened 2020-10-27
[tensorflow-cc] add support for x64-uwp


ianier:
Feature request: add support for the x64-uwp triplet to the 2.x tensorflow-cc port.

The expected benefit is to be able to use TF, at least for inference, on Store apps.

It looks like this is realistically feasible (see PR #13028). The resulting TF library may not support all features, though (e.g. those relying on file system access).

ianier:
I'm looking myself into this. There are no major issues with the code itself that prevent it from running under WinRT, and the missing features won't be showstoppers. I can already share a list of impacted code files and the possible workarounds.

However, the challenge is that some external dependencies (protoc.exe, grpc_cpp_plugin.exe, nasm.exe, def_file_filter.exe), which are a byproduct of the TF build process, are used during the build process itself. This means that these dependencies shouldn't target uwp, but the host config (say x64). .

What I don't know, though, is whether some of the objects built as part of these dependencies are also linked into the final output. If so, we'd have to find a way around this issue. If not, we need to figure out how to pass the right build switches only to the final output (tensorflow_cc.dll) and not the external deps.

ianier:
> What I don't know, though, is whether some of the objects built as part of these dependencies are also linked into the final output.

They are, but luckily the TF build framework already addresses this issue. The switch --distinct_host_configuration does exactly what we want (see https://docs.bazel.build/versions/master/guide.html). Setting it to true allows us to use different configs for the compiler and linker, via --host_copt and --host_linkopt when building the tools that are used in the build process itself. Note that the default value for --distinct_host_configuration is true, but TF sets it to false for Windows builds in the .bazelrc config file, as it assumes that there’s never a need for cross-compilation on Windows.
Now, regarding the source code. As mentioned in #13028, we have to compile with -DWINAPI_FAMILY=WINAPI_FAMILY_APP and -D_WIN32_WINNT=0x0A00, and link with /APPCONTAINER.
It’s worth noting that a couple of TF dependencies already support UWP. So, if we assume that these dependencies are used only to build the main library and not the build tools themselves, we can just configure them, specifically:
SQLite (.bzl\_bazel_devtools\ozkcue5a\execroot\org_tensorflow\external\org_sqlite): in BUILD.bazel, set "-DSQLITE_OS_WINRT=1" 
[...]

PR #14394, opened 2020-11-04
[tensorflow-cc/tensorflow-common] Add uwp port

jgehw:
Implementation of patches discussed in #14252.

I suggest to wait with reviewing until #14392 is merged into master, because this PR builds upon #14392, and once #14392 is merged the file diffs of this PR will collapse to just a few simple changes.

Current state: work-in-progress


ianier:
@jgehw : Attached is an updated [uwppatch.h](https://github.com/microsoft/vcpkg/files/5505969/uwppatch.zip) (removed macros that are defined in UWP like MAX_PATH)

PhoebeHui:
@jgehw, thanks for the PR! if it's ready for review, feel free to ping us.

jgehw:
This PR was on hold for quite some time as I was hoping for a solution of https://stackoverflow.com/questions/65184913/how-do-i-add-an-additional-system-include-directory-using-bazel but as there isn't any solution yet, I applied a work-around. Let's merge it with the work-around applied and update it later when (if ever) a cleaner solution is available.

jgehw:
@ianier if you have time, please verify the build on your machine and in your application

ianier:
> @ianier if you have time, please verify the build on your machine and in your application

@jgehw: You're the man! 🥇 

Yes, it builds fine, provided that:
-Your Windows user login has no spaces, because it results on temp dirs with spaces in it, which TF or Bazel itself don't like. This issue is not specific to the UWP build and was there before.
-You're logged on as an admin user, otherwise you get access denied on a certain path during build. This is new to me, as the last time I compiled TF this wasn't a requirement.

The app and models run fine with the resulting Release DLL (I didn't try the Debug one). I'll use this port in an upcoming release and then we'll know for sure if the certification process goes through, though I'd be surprise if it didn't.

BTW, in the release notes for the UWP triplet it should be mentioned that that the 'Code Generation' capability needs to be checked in the app manifest.

jgehw:
> -You're logged on as an admin user, otherwise you get access denied on a certain path during build. This is new to me, as the last time I compiled TF this wasn't a requirement.

This is new. It's because of the work-around that I copy the `uwppatch.h` as `__vcpkg_uwppatch.h` into the system includes folder until someone has a solution for bazel. (I prefixed it with `__vcpkg_` to avoid side effects if someone else eventually comes up with another `uwppatch.h` file.)

> BTW, in the release notes for the UWP triplet it should be mentioned that that the 'Code Generation' capability needs to be checked in the app manifest.

Good hint, I added this note, too.

jgehw:
> @jgehw, thanks for the PR! if it's ready for review, feel free to ping us.

@PhoebeHui ready for review now :-)

jgehw:
@PhoebeHui I fixed version files, so also x86 CI errors are solved now.

JackBoosY:
I think these changes are not official and we should wait for upstream approval.

PhoebeHui:
@jgehw, for the patch, could you submit an issue or PR to Upstream?

jgehw:
> @jgehw, for the patch, could you submit an issue or PR to Upstream?

@PhoebeHui An according issue was already filed as https://github.com/tensorflow/tensorflow/issues/44463 by @ianier several months ago. Upstream said that they don't have enough resources to officially support this, and suggested that we support it via a community build. So, we created this port as a community build.

ianier:
@PhoebeHui: Come on. If Microsoft shows no commitment to UWP (not even upstream), who will?

PhoebeHui:
@jgehw, thanks for the information!

@strega-nil @ras0219-msft, could you help futher review?

ports/tensorflow-common/tensorflow-common.cmake
```diff
@@ -211,19 +216,39 @@ foreach(BUILD_TYPE dbg rel)
...
 
+	set(ADDITIONAL_TARGETS_PRE)
+	if(VCPKG_TARGET_IS_UWP)
+		# the following doesn't work yet, use work-around for the time being
+		# details: https://stackoverflow.com/questions/65184913/how-do-i-add-an-additional-system-include-directory-using-bazel
+		#set(AT_CHAR_FOR_CONFIGURE_FILE "@")
+		#configure_file(${CMAKE_CURRENT_LIST_DIR}/uwp_genrule.BUILD.in ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${BUILD_TYPE}/patching_target/BUILD @ONLY)
+		#set(ADDITIONAL_TARGETS_PRE "///patching_target:patch_include_dirs")
+		message(STATUS "UWP build requires a patched header to be force-included. As this currently is not possible with bazel, we will copy '__vcpkg_uwppatch.h' to '$ENV{VCToolsInstallDir}/include/'. This might require vcpkg to be run as Administrator if the next command fails.")
+		configure_file(${CMAKE_CURRENT_LIST_DIR}/uwppatch.h $ENV{VCToolsInstallDir}/include/__vcpkg_uwppatch.h COPYONLY)
```
ras0219, line 227: This is unfortunately against vcpkg policy and cannot be merged as-is due to modifying global system state.

ras0219 (changes requested):
Thanks for this PR!

>and suggested that we support it via a community build. So, we created this port as a community build.

I assume that by "community build", they mean that they will accept PRs upstream but won't provide testing resources.

Given the extreme complexity of tensorflow, we're very hesitant to accept large patches like this without some amount of upstream support. To be extra-clear about what we would want to see: we don't need tensorflow to provide builds for arm64 or uwp, but we _do_ need upstream to commit to reviewing and accepting PRs to make it work. Specifically, we need to see most of the patches and changes in this PR made against upstream and merged.

We'd love to see tensorflow on UWP, but we (vcpkg) cannot commit to the complexity required by this PR without upstream willingness to accept contributions.

ianier:
> We'd love to see tensorflow on UWP, but we (vcpkg) cannot commit to the complexity required by this PR without upstream willingness to accept contributions.

The unfortunate fact is that upstream support for Windows features is limited, and it won’t get any better without at least some involvement from Microsoft. This is a chicken-and-egg problem: the platform won’t gain users if the features aren’t there, and developers won’t add features for platforms where there are no users.
For example, where’s upstream support for Windows on ARM? In comparison, support for MacOS on the M1 CPU is already being worked on. The situation for UWP is even worse; just look at how few vcpkg libs support the -uwp triplets.
Ranting aside, I understand this is not the ideal situation and I’m grateful to @jgehw for this solution. @ras0219 do you happen to know if Microsoft has contributors upstream we could talk to?

ports/tensorflow-common/tensorflow-common.cmake
jgehw: Ok, I understand. So, as all other approaches I tried didn't work in the end, we need to wait until someone else has a solution or somebody answers the according question I posted for this: https://stackoverflow.com/questions/65184913/how-do-i-add-an-additional-system-include-directory-using-bazel

jgehw:
As my [question](https://stackoverflow.com/questions/65184913/how-do-i-add-an-additional-system-include-directory-using-bazel) on StackOverflow for a clean solution of the bazel issue has continuously been viewed but not answered yet after more than two months, and as my work-around is against vcpkg policy, we can put this PR on hold from my side since there's nothing more I can do about it. At least UWP users can check out this branch and build for the time being.

ras0219-msft:
@ianier I don't currently know of an appropriate contact to put you in touch with unfortunately.

Let me restate how I feel we can make progress with this PR:
1. @jgehw or @ianier opens a PR against tensorflow/tensorflow containing the source changes in this PR (https://github.com/microsoft/vcpkg/pull/14394/files#diff-c359b7d2a4231436bf48bf6475452d7c4fe52fef17c57ea7ee7d7fa91554e5e7 et al)
2. After back-and-forth, upstream merges the PR. While they did say that they won't be adding UWP to their build matrix in https://github.com/tensorflow/tensorflow/issues/44463, they did not say they will refuse PRs -- in fact, I take the statement about "community builds" as an invitation for PRs.
3. We then do one of:
   1. Wait for the next release that contains the changes
   2. Download the relevant commit diffs on-demand via a url like https://github.com/tensorflow/tensorflow/commit/e86be327976e8cca447072562f2fae049148f74f.diff
   3. Backport the merged changes into local patches (least preferable)

ianier:
@ras0219-msft : As a developer, my incentive here is to get features implemented in my UWP app. I felt vcpkg was a reasonable way to do that and I can’t thank @jgehw enough for getting me there. 
The unfortunate fact is that the UWP developer community is small, and this won’t change without additional efforts to promote the technology. This is why I’ve been (naively) hoping that some kind of UWP evangelist would help us upstream.
So, right now, it’s pretty clear to me that the safest way for us app developers to go in the long run is to steer away from UWP.

jgehw:
@ras0219-msft Some of the changes of this PR won't get merged upstream as they are quick hacks to work around compilation errors, forsaking some minor tensorflow features most people never use. I'm pretty sure, upstream would require clean removal of these un-compilable features by adding compilation switches etc. This involves quite some work. As I'm not a UWP user, and never will be, I can't donate enough time to cover this. After @ianier found a way to make tensorflow compile on UWP, I just incorporated his findings into patches and a vcpkg port.
For this reason, and because a vcpkg compliant solution not modifying global system state is still not availble (see discussion two weeks ago), I think it's best to put this PR on hold until eventually someone volunteers to do the required upstream work.

ianier:
> As I'm not a UWP user, and never will be,

And the way things are going, I'm afraid eventually nobody will :-( 
The main motivations to use UWP right now are its modern UI and Store support, so I wouldn't be surprised if UWP got deprecated once WinUI for Win32/desktop gains traction.

PhoebeHui:
@jgehw @ianier, can we close this PR? We would help reopen if someone want to continue this PR.

ianier:
> @jgehw @ianier, can we close this PR? We would help reopen if someone wants to continue this PR.

I guess you may close this one on the account of the WinRT debacle. BTW, I just realized that the other 'official' UWP triplet is arm-uwp (32-bit ARM is today a truly dead platform), while arm64-uwp is a community triplet. Go figure.

PR #14397, opened 2020-11-04
[vcpkg] Add x-ignore-lock-failures

### LINKED ISSUE #14281: [vcpkg] Locking `.vcpkg-root` fails on NFS drives
**Describe the bug**
When using a vcpkg in manifestmode with CMAKE_TOOLCHAIN_FILE cmake is unable to select a build program that is corresponding to "Unix Makefiles".

GNU Make is installed and in the PATH

**Environment**
- OS: Linux Debian stable
- Compiler: gcc8

**To Reproduce**
1. add vcpkg.json file
2. run cmake with CMAKE_TOOLCHAIN_FILE set to the vcpkg toolchain file

**Expected behavior**
installation of the pakages declared in the vcpkg.json file and configuration of the project

**Failure logs** 
```
16:26 $ cmake -B build/
-- Running vcpkg install
CMake Error at share/vcpkg/scripts/buildsystems/vcpkg.cmake:386 (message):
  Running vcpkg install - failed
  /usr/share/cmake-3.13/Modules/CMakeDetermineSystem.cmake:94 (include)
  CMakeLists.txt:8 (project)


CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```

**Additional context**
the toolchain file is selected in the CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.10)

set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/share/vcpkg/scripts/buildsystems/vcpkg.cmake"
    CACHE STRING "Vcpkg toolchain file")

set(CMAKE_CXX_STANDARD 17)

project(librethfeld VERSION 0.0.1)

file(GLOB_RECURSE SOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "src/*.cpp")

add_library(librethfeld ${SOURCES})

target_include_directories(librethfeld PRIVATE include/librethfeld)
target_include_directories(librethfeld PUBLIC "${PROJECT_BINARY_DIR}/include")

find_package(Boost REQUIRED mpi serialization)
include_directories(${Boost_INCLUDE_DIRS})
target_link_libraries(librethfeld PRIVATE Boost::mpi Boost::serialization)

find_package(spdlog REQUIRED)
target_link_libraries(librethfeld PRIVATE spdlog::spdlog)
```

strega-nil:
This should fix #14281 at least as a stopgap.

Additionally, adds better errors.


scripts/buildsystems/vcpkg.cmake
```diff
@@ -322,75 +334,83 @@ else()
...
             COMMAND "${_VCPKG_BOOTSTRAP_SCRIPT}" ${VCPKG_BOOTSTRAP_OPTIONS}
+            OUTPUT_FILE "${_VCPKG_BOOTSTRAP_LOG}"
+            ERROR_FILE "${_VCPKG_BOOTSTRAP_LOG}"
             RESULT_VARIABLE _VCPKG_BOOTSTRAP_RESULT)
 
         if (NOT _VCPKG_BOOTSTRAP_RESULT EQUAL 0)
-            message(FATAL_ERROR "Bootstrapping vcpkg before install - failed")
+            message(STATUS "Bootstrapping vcpkg before install - failed")
+            _vcpkg_add_fatal_error("vcpkg install failed. See logs for more information: ${_VCPKG_BOOTSTRAP_LOG}")
+        else()
```
BillyONeal, line 354: Since you're adding an else can you flip the blocks and drop the not?

scripts/buildsystems/vcpkg.cmake
```diff
@@ -322,75 +334,83 @@ else()
...
 
-    set(VCPKG_OVERLAY_PORTS "${VCPKG_OVERLAY_PORTS}" CACHE STRING "Overlay ports to use for vcpkg install in manifest mode" FORCE)
-    mark_as_advanced(VCPKG_OVERLAY_PORTS)
-    set(VCPKG_OVERLAY_TRIPLETS "${VCPKG_OVERLAY_TRIPLETS}" CACHE STRING "Overlay triplets to use for vcpkg install in manifest mode" FORCE)
-    mark_as_advanced(VCPKG_OVERLAY_TRIPLETS)
-    set(VCPKG_INSTALL_OPTIONS "${VCPKG_INSTALL_OPTIONS}" CACHE STRING "Additional install options to pass to vcpkg" FORCE)
-    mark_as_advanced(VCPKG_INSTALL_OPTIONS)
-
-    message(STATUS "Running vcpkg install")
+    if (NOT _VCPKG_HAS_FATAL_ERROR)
```
BillyONeal, line 359: Why not just emit the `FATAL_ERROR` here which gives you effectively an early exit?

(I'm more to better understand than to argue to change what you wrote)

scripts/buildsystems/vcpkg.cmake
```diff
@@ -322,75 +334,83 @@ else()
...
-            ${_VCPKG_ADDITIONAL_MANIFEST_PARAMS}
-            ${VCPKG_INSTALL_OPTIONS}
-        OUTPUT_FILE "${CMAKE_BINARY_DIR}/vcpkg-manifest-install-out.log"
-        ERROR_FILE "${CMAKE_BINARY_DIR}/vcpkg-manifest-install-err.log"
-        RESULT_VARIABLE _VCPKG_INSTALL_RESULT
-    )
+        if (NOT _VCPKG_INSTALL_RESULT EQUAL 0)
+            message(STATUS "Running vcpkg install - failed")
+            _vcpkg_add_fatal_error("vcpkg install failed. See logs for more information: ${_VCPKG_MANIFEST_INSTALL_LOG}")
+        else()
```
BillyONeal, line 406: Ditto added else, flip the sense?

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -259,21 +259,24 @@ namespace vcpkg
...
-                m_pimpl->file_lock_handle = filesystem.try_take_exclusive_file_lock(vcpkg_lock, ec);
-            }
-            if (ec)
-            {
-                System::printf(
-                    System::Color::error, "Failed to take the filesystem lock on %s:\n", fs::u8string(vcpkg_lock));
-                System::printf(System::Color::error, "    %s\n", ec.message());
-                Checks::exit_fail(VCPKG_LINE_INFO);
+                const auto vcpkg_lock = root / ".vcpkg-root";
+                if (args.wait_for_lock.value_or(false))
```
BillyONeal, line 265: Should we diagnose explicit `--disable-lock --wait-for-lock`?

BillyONeal (changes requested):
Does this do the right thing? `x-wait-for-lock` is already off by default...

BillyONeal:
(To clarify, I thought we only set `--x-wait-for-lock` in the msbuild integration pieces which should be unaffected by this NFS stuff since that's Windows?)

scripts/buildsystems/vcpkg.cmake
strega-nil: If you emit a `FATAL_ERROR` in a toolchain file before you get `MAKE_PROGRAM` and the like set up, you get lots of fun extra errors.

strega-nil:
@BillyONeal yeah, `disable-lock` allows a user to just deal with "locking a file fails for a reason that isn't someone else owning the lock"

toolsrc/src/vcpkg/vcpkgpaths.cpp
strega-nil: Eh, I don't think it's necessary. There are legitimate reasons that might happen.

BillyONeal:
> @BillyONeal yeah, `disable-lock` allows a user to just deal with "locking a file fails for a reason that isn't someone else owning the lock"

That's not what this does though; it's making `disable-lock` unconditionally skip everything related to locking

strega-nil:
yeah, exactly; if someone is dealing with "Weird Things", they can use this to disable the lock to just get things working.

BillyONeal:
> yeah, exactly; if someone is dealing with "Weird Things", they can use this to disable the lock to just get things working.

Why can't they just not pass `--wait-for-lock` then? That's what I meant by

>> (To clarify, I thought we only set `--x-wait-for-lock` in the msbuild integration pieces which should be unaffected by this NFS stuff since that's Windows?)

Can't we distinguish between "failed to take the lock because of an error" and "failed to take the lock because there's another vcpkg alive here"?

strega-nil:
@BillyONeal yes, we can, that's why `--wait-for-lock` exists. The errors on the original thread were errors that _weren't_ caused by another vcpkg. Are you asking if we should add a `--allow-lock-to-fail` or something?

BillyONeal:
I'm saying that it is unclear here why instead of adding a switch that says "don't do X", we aren't just asking the user to not pass the switch that says "do X". If the user didn't pass `--wait-for-lock`, why are we treating any management of the lock as a fatal error?

strega-nil:
@BillyONeal you need a lock in order to do anything? I'm confused where you're confused.

If you have a `vcpkg install` going, it needs to access the `buildtrees` and the like in a mutable way. Therefore, you need mutual exclusion. We do that via the lock file; if `--x-wait-for-lock` is passed, then we wait until the lock is free, else we try a few times and then print an error if we can't get the lock.

This solves an entirely different issue, which is where the lock isn't get-able for reasons that aren't "it's already taken".

BillyONeal:
I discussed this with @strega-nil over IM and my confusion was because I was thinking the world was like this:
image
and so I was wondering why were adding a switch that just undid what another switch did. The actual effects post this PR are:
image

BillyONeal (approved):

BillyONeal:
This is "info:reviewed" but I'm not merging it because Nicole has the power to do that herself and might want further feedback from others on the team

toolsrc/include/vcpkg/vcpkgcmdarguments.h
```diff
@@ -163,9 +163,9 @@ namespace vcpkg
         constexpr static StringLiteral WAIT_FOR_LOCK_SWITCH = "x-wait-for-lock";
         Optional<bool> wait_for_lock = nullopt;
 
-        constexpr static StringLiteral DISABLE_LOCK_SWITCH = "x-disable-lock";
-        constexpr static StringLiteral DISABLE_LOCK_ENV = "X_VCPKG_DISABLE_LOCK";
-        Optional<bool> disable_lock = nullopt;
+        constexpr static StringLiteral ALLOW_SPURIOUS_LOCK_FAILURES_SWITCH = "x-allow-spurious-lock-failures";
```
BillyONeal, line 166: I don't think this should be called "SPURIOUS" lock failures; that sounds like it's a mitigation for a nondeterministic thing but this is entirely deterministic.

BillyONeal (approved):

BillyONeal (approved):

PR #14290, opened 2020-10-29
[libmysql/libmariadb] Export unofficial cmake targets

JackBoosY:
In order to use libmysql and libmariadb easily, export the unofficial targets.
For libmysql, since the library and the target generated by its dynamic build and static build is inconsistent, a wrapper is added to unify the output interface.

When the executable in the bin is cleaned up, the executable related pdb also need to be cleaned up.

This PR prepares for poco update.

PR #14426, opened 2020-11-06
[vcpkg baseline][rtabmap] Fix dependency issue when build tools

PhoebeHui:
Related https://github.com/microsoft/vcpkg/pull/14290

1.Update to build tools as a feature
2. When build tools, it requires yaml-cpp to build rtabmap-euroc_dataset, add the dependency.

Check all tools in https://github.com/introlab/rtabmap/blob/master/tools/CMakeLists.txt
- [X] ImagesJoiner and StereoEval tools not installed by the Upstream
- [X] The rtabmap_gui not built since we disabled QT, 
- [X] OPENCV_NONFREE not found when I installed opencv[nonfree] feature, I didn't find related macro exported. So I think we have coverred all tools here.

Failrues:
```
- Performing post-build validation
The following EXEs were found in /bin or /debug/bin. EXEs are not valid distribution targets.

    D:/packages/rtabmap_x64-windows/bin/rtabmap-euroc_dataset.exe

The following EXEs were found in /bin or /debug/bin. EXEs are not valid distribution targets.

    D:/packages/rtabmap_x64-windows/debug/bin/rtabmap-euroc_dataset.exe

```

### DIFF
--- a/ports/rtabmap/portfile.cmake
+++ b/ports/rtabmap/portfile.cmake
@@ -10,10 +10,17 @@ vcpkg_from_github(
         001_opencv.patch
 )
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    tools BUILD_TOOLS
+)
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
     OPTIONS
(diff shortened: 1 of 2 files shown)


JackBoosY (approved):

BillyONeal (approved):
Thanks! This port really needed "with all the other ports installed" testing :(

Someday!

Issue #6077, opened 2019-04-13
help add repo onnxruntime


bitnick10:
https://github.com/Microsoft/onnxruntime

JackBoosY:
We can’t reopen this issue until a PR is opened, unfortunately, consider making a pull request!

PR #14481, opened 2020-11-09
[onnxruntime-gpu] Add port for onnxruntime (GPU)

ianormy:
**Describe the pull request**

This port uses the binaries created by Microsoft for the onnxruntime which will run with a GPU. 

It is currently used for an internal build but I thought it would be good to offer it to more people, which is why I've raised this PR. Whilst I appreciate this doesn't actually build the binaries, it does show how to include binaries and libraries from elsewhere that can then be included in your projects.

- currently only supports the triplets x64-windows
- PR follows the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)

### DIFF
--- a/ports/onnxruntime-gpu/portfile.cmake
+++ b/ports/onnxruntime-gpu/portfile.cmake
@@ -1,6 +1,6 @@
-vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports Windows platforms" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")
-vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports x64 architecture" ON_ARCH "x86" "arm" "arm64")
-vcpkg_fail_port_install(MESSAGE "onnxruntime currently does not support static libraries" ON_LIBRARY_LINKAGE "static")
+vcpkg_fail_port_install(ON_ARCH "x86" "arm" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")
+
+vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
 
 set(VERSION 1.5.1)
 
--- a/ports/onnxruntime-gpu/portfile.cmake
+++ b/ports/onnxruntime-gpu/portfile.cmake
@@ -1,0 +1,43 @@
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports Windows platforms" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports x64 architecture" ON_ARCH "x86" "arm" "arm64")
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently does not support static libraries" ON_LIBRARY_LINKAGE "static")
+
+set(VERSION 1.5.1)
+
+vcpkg_download_distfile(ARCHIVE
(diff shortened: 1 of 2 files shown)


ports/onnxruntime-gpu/portfile.cmake
```diff
@@ -0,0 +1,43 @@
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports Windows platforms" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently only supports x64 architecture" ON_ARCH "x86" "arm" "arm64")
+vcpkg_fail_port_install(MESSAGE "onnxruntime currently does not support static libraries" ON_LIBRARY_LINKAGE "static")
```
NancyLi1013, line 3:
```suggestion
vcpkg_fail_port_install(ON_ARCH "x86" "arm" ON_TARGET "UWP" "LINUX" "ANDROID" "FREEBSD" "OSX")

vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
```

NancyLi1013 (changes requested):

ports/onnxruntime-gpu/portfile.cmake
ianormy: Thanks @NancyLi1013 for this change. I thought the error messages might be helpful, but your solution is a lot simpler.

NancyLi1013:
Since this PR just copies those binaries to the related folders and they are not generated by building the sources.

I add "Required: discussion" label for it to have a further review from you. @strega-nil

ianormy:
@NancyLi1013 on a completely different note, is the **1013** at the end of your username CR+LF - 0x0A0D? If it is, that's a clever way to get a unique username :-)

ports/onnxruntime-gpu/portfile.cmake
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+file(COPY
+        ${SOURCE_PATH}/onnxruntime-win-x64-gpu-1.5.1/include
+        DESTINATION ${CURRENT_PACKAGES_DIR}
+    )
+
+file(COPY ${SOURCE_PATH}/onnxruntime-win-x64-gpu-1.5.1/lib/onnxruntime.lib
+    DESTINATION ${CURRENT_PACKAGES_DIR}/lib)
+file(COPY ${SOURCE_PATH}/onnxruntime-win-x64-gpu-1.5.1/lib/onnxruntime.lib
```
strega-nil, line 35: This seems incorrect; onnxruntime is a C++ library, so it has different ABI between debug and release (on Windows). This shouldn't be linkable..?

ianormy: I've used this vcpkg port with a release and debug build and it picks up the difference and links in correctly. I've then used it to load an ONNX format machine learning model and successfully run inference. It's super quick!

strega-nil:
Also, is there a reason that we aren't just building it?

ianormy:
@strega-nil if it was that easy to build surely someone else would have done it already :) 

I know others have tried and failed:

https://github.com/microsoft/onnxruntime/issues/931

And also that there is a need for it:

https://github.com/microsoft/vcpkg/issues/6077

I added the suffix "gpu" to this port with a view to possibly one day putting together a proper build of the onnxruntime and calling the port "onnxruntime". Creating this port provides something right now that people (including me) can use in their projects which is GPU enabled. That's why I mentioned it at the very beginning of this PR. I'm not looking to set a precedent here of having non-building ports, I just thought others could benefit from this now whilst we wait for a proper port of the onnxruntime.

PR #14499, opened 2020-11-10
[freetype] Fix INTERFACE_LINK_LIBRARIES in exported cmake file

### LINKED ISSUE #9952: [freetype] (Linux) Wrong INTERFACE_LINK_LIBRARIES in config
```
set_target_properties(freetype PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/freetype"
  INTERFACE_LINK_LIBRARIES "/mnt/d/xlinux/installed/x64-linux/debug/lib/libz.a;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/d/xlinux/installed/x64-linux/lib/libbz2.a>;\$<\$<CONFIG:DEBUG>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libbz2d.a>;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libpng.a>;\$<\$<CONFIG:DEBUG>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libpng16d.a>;/mnt/d/xlinux/installed/x64-linux/debug/lib/libz.a"
)
```
entry: /mnt/d/xlinux/installed/x64-linux/debug/lib/libz.a missing generator expression for debug/release. Links only debug

JackBoosY:
Fix wrong INTERFACE_LINK_LIBRARIES in _freetype-config.cmake_:
```
set_target_properties(freetype PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/freetype"
  INTERFACE_LINK_LIBRARIES "/mnt/d/xlinux/installed/x64-linux/debug/lib/libz.a;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/d/xlinux/installed/x64-linux/lib/libbz2.a>;\$<\$<CONFIG:DEBUG>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libbz2d.a>;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libpng.a>;\$<\$<CONFIG:DEBUG>:/mnt/d/xlinux/installed/x64-linux/debug/lib/libpng16d.a>;/mnt/d/xlinux/installed/x64-linux/debug/lib/libz.a"
)
```

Fixes #9952.

### DIFF
--- a/ports/freetype/fix-exports.patch
+++ b/ports/freetype/fix-exports.patch
@@ -1,0 +1,70 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 04ce73a..cda36a5 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -454,31 +454,31 @@ endif ()
+ set(PKG_CONFIG_REQUIRED_PRIVATE "")
+ 
+ if (ZLIB_FOUND)
+-  target_link_libraries(freetype PRIVATE ${ZLIB_LIBRARIES})
+-  target_include_directories(freetype PRIVATE ${ZLIB_INCLUDE_DIRS})
+-  list(APPEND PKG_CONFIG_REQUIRED_PRIVATE "zlib")
++  target_link_libraries(freetype PUBLIC ZLIB::ZLIB)
++  target_include_directories(freetype PUBLIC ${ZLIB_INCLUDE_DIRS})
++  list(APPEND PKG_CONFIG_REQUIRED_PUBLIC "zlib")
+ endif ()
+ if (BZIP2_FOUND)
+-  target_link_libraries(freetype PRIVATE ${BZIP2_LIBRARIES})
+-  target_include_directories(freetype PRIVATE ${BZIP2_INCLUDE_DIR}) # not BZIP2_INCLUDE_DIRS
+-  list(APPEND PKG_CONFIG_REQUIRED_PRIVATE "bzip2")
++  target_link_libraries(freetype PUBLIC ${BZIP2_LIBRARIES})
++  target_include_directories(freetype PUBLIC ${BZIP2_INCLUDE_DIR}) # not BZIP2_INCLUDE_DIRS
++  list(APPEND PKG_CONFIG_REQUIRED_PUBLIC "bzip2")
+ endif ()
+ if (PNG_FOUND)
+-  target_link_libraries(freetype PRIVATE ${PNG_LIBRARIES})
+-  target_compile_definitions(freetype PRIVATE ${PNG_DEFINITIONS})
+-  target_include_directories(freetype PRIVATE ${PNG_INCLUDE_DIRS})
+-  list(APPEND PKG_CONFIG_REQUIRED_PRIVATE "libpng")
++  target_link_libraries(freetype PUBLIC ${PNG_LIBRARIES})
++  target_compile_definitions(freetype PUBLIC ${PNG_DEFINITIONS})
++  target_include_directories(freetype PUBLIC ${PNG_INCLUDE_DIRS})
++  list(APPEND PKG_CONFIG_REQUIRED_PUBLIC "libpng")
+ endif ()
+ if (HARFBUZZ_FOUND)
+-  target_link_libraries(freetype PRIVATE ${HARFBUZZ_LIBRARIES})
+-  target_include_directories(freetype PRIVATE ${HARFBUZZ_INCLUDE_DIRS})
+-  list(APPEND PKG_CONFIG_REQUIRED_PRIVATE "harfbuzz >= ${HARFBUZZ_MIN_VERSION}")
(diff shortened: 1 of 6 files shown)


ports/freetype/fix-exports.patch
```diff
@@ -0,0 +1,70 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -454,31 +454,31 @@ endif ()
+ set(PKG_CONFIG_REQUIRED_PRIVATE "")
+ 
+ if (ZLIB_FOUND)
+-  target_link_libraries(freetype PRIVATE ${ZLIB_LIBRARIES})
+-  target_include_directories(freetype PRIVATE ${ZLIB_INCLUDE_DIRS})
+-  list(APPEND PKG_CONFIG_REQUIRED_PRIVATE "zlib")
++  target_link_libraries(freetype PUBLIC ZLIB::ZLIB)
```
Neumann-A, line 12: why all those PUBLIC/PRIVATE changes?
I think you are mixing up static/dynamics builds here. 

You should rather find out why ${ZLIB_LIBRARIES} is incorrect.

ports/freetype/fix-exports.patch
```diff
@@ -0,0 +1,70 @@
...
++  target_include_directories(freetype PUBLIC ${ZLIB_INCLUDE_DIRS})
++  list(APPEND PKG_CONFIG_REQUIRED_PUBLIC "zlib")
```
Neumann-A, line 14: The PUBLIC/PRIVATE in pc files will be handled by ras0219 pkgconfig PR. So no need to change it here.

ports/freetype/fix-exports.patch
JackBoosY: @Neumann-A `${ZLIB_LIBRARIES}` is correct, but it's value is not `optimized;...;debug;...`, so different values are generated in `INTERFACE_LINK_LIBRARIES` in `freetype-config.cmake`, and `vcpkg_fixup_cmake_targets` deletes the debug `freetype-config.cmake`, which causes this problem.
We should avoid using the macros provided by cmake's module.

PhoebeHui (approved):

ports/freetype/fix-exports.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 04ce73a..ee3cc05 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -454,7 +454,7 @@ endif ()
+ set(PKG_CONFIG_REQUIRED_PRIVATE "")
+ 
+ if (ZLIB_FOUND)
+-  target_link_libraries(freetype PRIVATE ${ZLIB_LIBRARIES})
```
Neumann-A, line 9: >${ZLIB_LIBRARIES} is correct, but it's value is not optimized;...;debug;..., so different values are generated in INTERFACE_LINK_LIBRARIES in freetype-config.cmake, and vcpkg_fixup_cmake_targets deletes the debug freetype-config.cmake, which causes this problem.
We should avoid using the macros provided by cmake's module.

That means ZLIB requires a wrapper to correctly set that value on Linux. You are just dropping a droplet of water on a hot stone with that patch. You might have fixed freetype but all other ports using `ZLIB_LIBRARIES` still have the same problem. The root cause ist that `ZLIB_LIBRARIES` is not correctly set with debug/optimized keywords and needs fixing on Linux/OSX.

JackBoosY: > > ${ZLIB_LIBRARIES} is correct, but it's value is not optimized;...;debug;..., so different values are generated in INTERFACE_LINK_LIBRARIES in freetype-config.cmake, and vcpkg_fixup_cmake_targets deletes the debug freetype-config.cmake, which causes this problem.
> > We should avoid using the macros provided by cmake's module.
> 
> That means ZLIB requires a wrapper to correctly set that value on Linux. You are just dropping a droplet of water on a hot stone with that patch. You might have fixed freetype but all other ports using `ZLIB_LIBRARIES` still have the same problem. The root cause ist that `ZLIB_LIBRARIES` is not correctly set with debug/optimized keywords and needs fixing on Linux/OSX.

This means that all module modes(`Find*.cmake`) that call cmake by default have a high probability of this problem. So I think we should deal with it uniformly (by vcpkg instead of port).

vicroms (approved):

strega-nil:
Merged as "makes the experience better, but it'd be great to fix the underlying issues"

JackBoosY:
In order to prevent this from happening, we at least need to find in POST_CHECK whether there is a release version of the library in `*config.cmake` / `*targets.cmake`.

PR #14245, opened 2020-10-27
[vcpkg baseline] Clean up baseline, use Keyword Supports Part 1

JackBoosY:
Separate PR #13384.

PR #14528, opened 2020-11-11
[3fd] Update to 2.6.3, fix port

strega-nil:
Instead of saying that it supports only uwp (see PR #14245), fix the port and update to 2.6.3


JackBoosY (changes requested):
Please also remove the codes on baseline.txt

ports/3fd/vcpkg.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "3fd",
-  "version-string": "2.6.2",
-  "port-version": 3,
+  "version-string": "2.6.3",
   "description": "C++ Framework For Fast Development",
+  "supports": "windows",
```
JackBoosY, line 5:
```suggestion
  "supports": "windows&!static&!arm64",
```
Reason:
```
CMake Error at scripts/cmake/vcpkg_check_linkage.cmake:48 (message):
  3fd only supports dynamic crt linkage
```
and
```
CMake Error at ports/3fd/portfile.cmake:11 (message):
  Unsupported architecture: arm64
```

JackBoosY (changes requested):

Issue #14511, opened 2020-11-11
[libvpx[core]:x64-windows] build failure


ivanovmp:
**Host Environment**
- OS: Windows 10 Home Single Language (Russian) x64
- Compiler: Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29110

**To Reproduce**
`./vcpkg install libvpx[core]:x64-windows`

**Failure logs** 
```
C:\misc\vcpkg>vcpkg install libvpx[core]:x64-windows
Computing installation plan...
The following packages will be built and installed:
    libvpx[core]:x64-windows
Detecting compiler hash for triplet x64-windows...
Starting package 1/1: libvpx:x64-windows
Building package libvpx[core]:x64-windows...
Could not locate cached archive: C:\Users\user\AppData\Local\vcpkg\archives\3f\3f5f6426b4370fcf9fd7812ce359cb51167ac16c.zip
-- Note: libvpx only supports static library linkage. Building static library.
-- Using cached C:/misc/vcpkg/downloads/webmproject-libvpx-v1.9.0.tar.gz
-- Cleaning sources at C:/misc/vcpkg/buildtrees/libvpx/src/v1.9.0-8ebe882649.clean. Use --editable to skip cleaning for the packages you specify.
-- Extracting source C:/misc/vcpkg/downloads/webmproject-libvpx-v1.9.0.tar.gz
-- Using source at C:/misc/vcpkg/buildtrees/libvpx/src/v1.9.0-8ebe882649.clean
-- Using msys root at C:/misc/vcpkg/downloads/tools/msys2/b1b4e6089ab84adb
CMake Error at installed/x64-windows/share/yasm-tool-helper/yasm-tool-helper.cmake:16 (message):
  Cross-targetting and x64 ports requiring yasm require the x86-windows
  yasm-tool to be available.  Please install yasm-tool:x86-windows first.
Call Stack (most recent call first):
  ports/libvpx/portfile.cmake:27 (yasm_tool_helper)
  scripts/ports.cmake:136 (include)


Error: Building package libvpx:x64-windows failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: libvpx:x64-windows
  Vcpkg version: 2020.06.15-nohash

Additionally, attach any relevant sections from the log files above.
```

**Additional context**
[...]

PR #14545, opened 2020-11-12
[libvpx] switch from yasm to nasm

mcmtroffaes:
- What does your PR fix? Since upstream supports nasm, this patch switches libvpx from yasm to nasm, as nasm is better maintained. It also avoids issues with the use of the yasm helper on 64 bit systems (see for instance #14511). Minor patch is needed to ensure the libvpx msbuild system works with nasm. Has been tested through a custom ffmpeg build against a nasm build of libvpx, and both vp8 and vp9 work as expected, producing identical files as the yasm build, and at the same speed.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? To the best of my knowledge, yes.


mcmtroffaes:
Looks like the x64_osx test is failing due to nasm not being installed on the test system. Not sure how to fix this. For the record, my own osx build with this patch is passing (with nasm installed): https://travis-ci.org/github/mcmtroffaes/ffmpeg-msvc-build/jobs/743162386

mcmtroffaes:
I've tried to fix it in f65ad1b to no avail. Could someone with a better understanding of how azure works with vcpkg suggest how to fix the regression test so osx gets nasm installed?

JackBoosY:
```
CMake Error at scripts/cmake/vcpkg_find_acquire_program.cmake:516 (message):
  Could not find nasm.  Please install it via your package manager:

      brew install nasm
```

mcmtroffaes:
Thanks @JackBoosY !

JackBoosY:
I will handle this regression next week.

JackBoosY:
@BillyONeal Please add `nasm` to the brew install list.

JackBoosY (approved):

mcmtroffaes:
Just resolved a conflict against master.

@longnguyen2004 Could you check if the mingw build of libvpx still works with this patch? I noticed you exported the yasm path to ENV{AS} to fix the mingw build of libvpx. I cannot easily test here locally if something similar is also necessary for nasm under mingw.

longnguyen2004:
@mcmtroffaes Alright I'll check

longnguyen2004:
Yes the ENV{AS} export is still required for `mingw`, not sure about MSVC. Can you give me write access so I can commit directly to your fork?

mcmtroffaes:
Many thanks for looking into it, @longnguyen2004! I'm not sure how to set up write access specifically for you (it seems only official maintainers can do that), but I think it's possible to make a suggested change via the github review functionality, which I can then directly apply via the github interface. See https://haacked.com/archive/2019/06/03/suggested-changes/ Does that work?

longnguyen2004:
> Many thanks for looking into it, @longnguyen2004! I'm not sure how to set up write access specifically for you (it seems only official maintainers can do that), but I think it's possible to make a suggested change via the github review functionality, which I can then directly apply via the github interface. See https://haacked.com/archive/2019/06/03/suggested-changes/ Does that work?

Sure, although I've discovered another weird error, not sure if it's because of my toolchain or something else

longnguyen2004:
Also GitHub Review doesn't seem to work outside your diff. You can invite me as a collaborator to give me write access.

mcmtroffaes:
Alas, too bad it doesn't work via github review. I think I've sent you an invite now.

longnguyen2004:
Can someone get me the log for `x64-osx`? I'll see if it's the same regression as I'm getting on `x64-mingw-dynamic`

longnguyen2004:
@mcmtroffaes I'm wondering if the lack of DWARF debug info for Windows would be a problem for mingw toolchain, since that's the standard debug format there. But since libvpx also defaults to CodeView for mingw toolchain, I'm gonna keep it that way,

longnguyen2004:
Looking good!

mcmtroffaes:
Excellent. Many thanks @longnguyen2004 for helping out with this! Looks like we got some new regressions but seems unrelated to this patch (hash error on vtk-m).

longnguyen2004:
@mcmtroffaes You're welcome! Now we just need to wait for VM update on `x64-osx`

JackBoosY:
LGTM. @BillyONeal Do you have any other questions?

vicroms:
Thanks for the PR!

PR #14546, opened 2020-11-12
[aws-sdk-cpp] Make aws-sdk-cpp rpath relocatable

### LINKED ISSUE #14544: Port aws-sdk-cpp has a non relocatable rpath, which becomes problematic when used with (remote) vcpkg binary caching
**Describe the bug**
`aws-sdk-cpp` sets `CMAKE_INSTALL_RPATH_USE_LINK_PATH` to true in its **CMakeLists.txt** file. The consequence is that when we install it with vcpkg (with a triplet that uses dynamic libraries, like the triplet `x64-linux-dynamic` described [in the documentation](https://vcpkg.readthedocs.io/en/latest/examples/overlay-triplets-linux-dynamic/)), the `RPATH` of the installed libraries contains the full path where it was installed. This becomes problematic when the binary caching feature of vcpkg is used.

For instance, assume that I install `aws-sdk-cpp` with vcpkg in `/home/klalumiere/Projects/vcpkg/`. The library `/home/klalumiere/Projects/vcpkg/installed/x64-linux-dynamic/lib/libaws-cpp-sdk-core.so` now contains the RPATH `/home/klalumiere/Projects/vcpkg/installed/x64-linux-dynamic/lib`. If `VCPKG_BINARY_SOURCES` contains a Nuget server, vcpkg uploads the library on this server.

Now, my colleague, which use exactly the same "toolchain" as I do, also installs the library with vcpkg. vcpkg downloads the library from the Nuget server since it (correctly) determines that the "toolchains" (the hash) are the same. However, my colleague now has a library with a RPATH that doesn't exist on its machine. Moreover, the library is unlikely to work because it won't found its dependencies (e.g. `libcurl.so`).

**Environment**
- OS: Ubuntu 20.04.1 LTS
- Compiler: clang version 10.0.0-4ubuntu1

**To Reproduce**
1. `./vcpkg install aws-sdk-cpp:x64-linux-dynamic --overlay-triplets=../vcpkg-overlays/triplets`
2. `objdump -x installed/x64-linux-dynamic/lib/libaws-cpp-sdk-core.so | grep PATH`
3. A `RUNPATH` particular to your machine is observed.

**Expected behavior**
`RUNPATH` should be `$ORIGIN`, which is relocatable. I'll submit a pull request in a few minutes that fix the issue.

klalumiere:
This is important since vcpkg offers binary caching which could be
shared between different machines (as long as both machines use the same
"toolchains" [more technically, as long as the vcpkg hash is the same])

**Describe the pull request**

- What does your PR fix? Fixes #14544

- Which triplets are supported/not supported? Have you updated the CI baseline? No changes related to triplets support or to CI baseline.

### DIFF
--- a/ports/aws-sdk-cpp/CONTROL
+++ b/ports/aws-sdk-cpp/CONTROL
@@ -1,6 +1,6 @@
 Source: aws-sdk-cpp
 Version: 1.8.83
-Port-Version: 0
+Port-Version: 1
 Homepage: https://github.com/aws/aws-sdk-cpp
 Description: AWS SDK for C++
 Build-Depends: openssl (!uwp&!windows), curl (!uwp&!windows), aws-c-event-stream
--- a/ports/aws-sdk-cpp/patch-relocatable-rpath.patch
+++ b/ports/aws-sdk-cpp/patch-relocatable-rpath.patch
@@ -1,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index a653e38628..76d6067646 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -142,7 +142,6 @@ else()
+ endif()
+ 
+ # Add Linker search paths to RPATH so as to fix the problem where some linkers can't find cross-compiled dependent libraries in customer paths when linking executables.
+-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH true)
+ 
+ # build third-party targets
+ if (BUILD_DEPS)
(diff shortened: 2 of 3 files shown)


PhoebeHui (changes requested):
@klalumiere, thanks for the PR!

Could you also update the Port-Version from 0 to 1 in CONTROL file?

klalumiere:
> @klalumiere, thanks for the PR!
> 
> Could you also update the Port-Version from 0 to 1 in CONTROL file?

I always forget about this :flushed: :laughing:. Done :heavy_check_mark:.

PhoebeHui:

PhoebeHui (approved):

strega-nil:
Thanks @klalumiere 👍

PR #14547, opened 2020-11-12
[ffmpeg] use nasm instead of yasm

mcmtroffaes:
- What does your PR fix? Yasm is no longer being actively maintained, so it seems like a good idea to switch to nasm since upstream supports it. Patch also replaces the deprecated --enable-yasm option with --enable-x86asm.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? Yes, to the best of my knowledge.

### DIFF
--- a/ports/ffmpeg/portfile.cmake
+++ b/ports/ffmpeg/portfile.cmake
@@ -23,9 +23,16 @@ if (SOURCE_PATH MATCHES " ")
     message(FATAL_ERROR "Error: ffmpeg will not build with spaces in the path. Please use a directory with no spaces")
 endif()
 
-vcpkg_find_acquire_program(NASM)
-get_filename_component(NASM_EXE_PATH ${NASM} DIRECTORY)
-vcpkg_add_to_path(${NASM_EXE_PATH})
+
+if(${VCPKG_TARGET_ARCHITECTURE} STREQUAL x86)
+    vcpkg_find_acquire_program(YASM)
+    get_filename_component(YASM_EXE_PATH ${YASM} DIRECTORY)
+    vcpkg_add_to_path(${YASM_EXE_PATH})
+else()
+    vcpkg_find_acquire_program(NASM)
+    get_filename_component(NASM_EXE_PATH ${NASM} DIRECTORY)
+    vcpkg_add_to_path(${NASM_EXE_PATH})
+endif()
 
 if(VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_MINGW)
     #We're assuming that if we're building for Windows we're using MSVC
--- a/ports/ffmpeg/portfile.cmake
+++ b/ports/ffmpeg/portfile.cmake
@@ -25,6 +25,7 @@ endif()
 
 
 if(${VCPKG_TARGET_ARCHITECTURE} STREQUAL x86)
+    # ffmpeg nasm build gives link error on x86, so fall back to yasm
     vcpkg_find_acquire_program(YASM)
     get_filename_component(YASM_EXE_PATH ${YASM} DIRECTORY)
     vcpkg_add_to_path(${YASM_EXE_PATH})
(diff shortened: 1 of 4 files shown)


JackBoosY:
I did the same job as https://github.com/microsoft/vcpkg/pull/14545/commits/a42d9e74b42c39133b846f328027158719dc00d5, so when we merged one of these 2 PRs and the other should resolve the file conflict.

mcmtroffaes:
Yep, that's swell, thanks @JackBoosY!

BillyONeal (approved):

JackBoosY:
I will check the `86-windows` manually later.

JackBoosY:
When building `ffmpeg:x86-windows`:
```
LD	libavcodec/avcodec-58.dll
   Creating library libavcodec/avcodec.lib and object libavcodec/avcodec.exp
LINK : fatal error LNK1000: unknown error at 00007FF6F720AC26; consult documentation for technical support options
```

mcmtroffaes:
I've done a fresh rebase and spent some time trying to diagnose this problem. For me, on the 32 bit build, the linker just hangs for a really long time (like 20 minutes) at the avcodec linking stage, and then causes the powershell terminal to crash, but I suppose it's the same error. Not sure where to go from here, except to fall back on yasm for x86... 😕 I will add a commit to this effect.

mcmtroffaes:
Looks like this has fixed the x86 build. 🎉

ports/ffmpeg/portfile.cmake
```diff
@@ -35,8 +44,7 @@ else()
     set(LIB_PATH_VAR "LIBRARY_PATH")
 endif()
 
-set(ENV{PATH} "$ENV{PATH}${VCPKG_HOST_PATH_SEPARATOR}${YASM_EXE_PATH}")
-set(OPTIONS "--enable-asm --enable-yasm --disable-doc --enable-debug --enable-runtime-cpudetect")
+set(OPTIONS "--enable-asm --enable-x86asm --disable-doc --enable-debug --enable-runtime-cpudetect")
```
JackBoosY, line 47: Since we still use yasm on x86 build, should we also change this line?

JackBoosY (changes requested):

ports/ffmpeg/portfile.cmake
mcmtroffaes: That's good question. But we do not: the --enable-yasm option is deprecated, and is equivalent to --enable-x86asm. See https://github.com/FFmpeg/FFmpeg/blob/master/configure#L4070:

```
        --enable-yasm|--disable-yasm)
            warn "The ${opt} option is only provided for compatibility and will be\n"\
                 "removed in the future. Use --enable-x86asm / --disable-x86asm instead."
            test $opt = --enable-yasm && x86asm=yes || x86asm=no
 ```

JackBoosY (approved):

JackBoosY:
Need test features, should be good.

mcmtroffaes:
Sure, FWIW here's my build matrix on appveyor (against a slightly more recent than version of ffmpeg compared to the one on vcpkg, but it also includes the nasm patch).

https://ci.appveyor.com/project/mcmtroffaes/ffmpeg-msvc-build/builds/37449560

Baklap4:
@mcmtroffaes did you ever find out why it hangs on `nasm` in the linker phase? I was wondering if [this](https://github.com/microsoft/vcpkg/pull/14547/commits/06c622e07a7e5dc300a75fe5a14079f56d0f1c1c) commit was still needed in latest master. But It seems i'm running in the same problem that it just 'hangs' and doesn't do anything with `nasm`.  I came here because of #19121 as `yasm` cannot be retrieved due to certificate error.

mcmtroffaes:
@Baklap4 That's an excellent question. Unfortunately, I never found out why nasm fails with x86 on ffmpeg... I agree it would be best if we could simply phase out yasm for ffmpeg completely.

PR #14552, opened 2020-11-13
[vcpkg] Replace APPEND to PREPEND on vcpkg.cmake

### LINKED ISSUE #13457: How do I use my own lib instead of system libs
Hi,
I'd like to use my own built libs instead of the system's one by default.
I saw in scripts/buildsystems/vcpkg.cmake(L221) this:
```
...
if(CMAKE_BUILD_TYPE MATCHES "^[Dd][Ee][Bb][Uu][Gg]$" OR NOT DEFINED CMAKE_BUILD_TYPE) #Debug build: Put Debug paths before Release paths.
    list(APPEND CMAKE_PREFIX_PATH
        ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}
    )
    list(APPEND CMAKE_LIBRARY_PATH
        ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug/lib/manual-link ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib/manual-link
    )
    list(APPEND CMAKE_FIND_ROOT_PATH
        ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug ${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}
    )
...
```

The append directive bothers me because the system lib is found before my own on MacOS. (ZLib)
(System lib): /usr/lib/libz.tbd
(My lib): /tmp/test/vcpkg/x86-darwin-release/lib/libz.a

Is there any way to change this behavior ?

For now, my workaround is to change vcpkg's APPEND to PREPEND.

JackBoosY:
Change `APPEND` to `PREPEND` to avoid finding the system library first.

Fixes #13457.


JackBoosY:
@Nemirtingas Can you please test this PR?

Thanks.

JackBoosY:
@strega-nil Should I also need to modify line 293 to `list(PREPEND CMAKE_PREFIX_PATH "/")` ?

Nemirtingas:
> 
> 
> @Nemirtingas Can you please test this PR?
> 
> Thanks.

Hi, I will test it asap, thanks.

Neumann-A:
@strega-nil @JackBoosY 
I disagree with this change.
Reason:
a) It wasn't disclosed in #13457 why system libraries are found. `CMAKE_LIBRARY_PATH`, `CMAKE_PREFIX_PATH` are normally empty according to the cmake docs and I highly suspect `CMAKE_FIND_ROOT_PATH` to be also empty. As such the root cause for the issue has not yet been discovered.
b) PREPEND was only added in newer CMake versions.
c) if it would have been a problem an issue would have bubbled up much earlier
d) It changes the default behavior. 
e) and most importantly you are screwing up the users. Due to d) it becomes virtually impossible to overwrite vcpkg by setting `CMAKE_PREFIX_PATH` `CMAKE_LIBRARY_PATH` in a user toolchain or before the first `project()` call

Nemirtingas:
> 
> 
> @strega-nil @JackBoosY
> I disagree with this change.
> Reason:
> a) It wasn't disclosed in #13457 why system libraries are found. `CMAKE_LIBRARY_PATH`, `CMAKE_PREFIX_PATH` are normally empty according to the cmake docs and I highly suspect `CMAKE_FIND_ROOT_PATH` to be also empty. As such the root cause for the issue has not yet been discovered.
> b) PREPEND was only added in newer CMake versions.
> c) if it would have been a problem an issue would have bubbled up much earlier
> d) It changes the default behavior.
> e) and most importantly you are screwing up the users. Due to d) it becomes virtually impossible to overwrite vcpkg by setting `CMAKE_PREFIX_PATH` `CMAKE_LIBRARY_PATH` in a user toolchain or before the first `project()` call

Hi, I use https://github.com/tpoechtrager/osxcross/blob/master/tools/toolchain.cmake, could this be the cause ?

Neumann-A:
Yes the toolchain is loaded before vcpkg so it affects everything. I strongly assume that:

```
# where is the target environment
set(CMAKE_FIND_ROOT_PATH
  "${OSXCROSS_SDK}"
  "${OSXCROSS_TARGET_DIR}/macports/pkgs/opt/local")
```
is the reason since: 
> The CMake variable CMAKE_FIND_ROOT_PATH specifies one or more directories to be prepended to all other search directories. This effectively “re-roots” the entire search under given locations. Paths which are descendants of the CMAKE_STAGING_PREFIX are excluded from this re-rooting, because that variable is always a path on the host system. By default the CMAKE_FIND_ROOT_PATH is empty.

I also don't know how the PKGCONFIG environment will interact with VCPKG.

strega-nil:
Closing as @Neumann-A makes very good points.

PR #14558, opened 2020-11-13
Fix libtommath linking with MSVC

andrew-boyarshin:
- _What does your PR fix?_ Fixes linking with MSVC-built libtommath, [see](https://github.com/libtom/libtommath/blob/c18817cc229f9c60a0a70b76cf49077e91983f41/s_mp_rand_platform.c#L120-L138).

- _Which triplets are supported/not supported?_ *-windows[-static], uwp not tested, but this can only improve, not worsen.

- _Have you updated the CI baseline?_ No need.

- _Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?_ Hopefully.

Should affect only MSVC builds.

### DIFF
--- a/ports/libtommath/portfile.cmake
+++ b/ports/libtommath/portfile.cmake
@@ -15,6 +15,44 @@ if(VCPKG_TARGET_IS_WINDOWS)
         set(CRTFLAG "/MT")
     endif()
 
+    set(LIBTOMMATH_RAND_PLATFORM ${SOURCE_PATH}/bn_s_mp_rand_platform.c)
+
+    file(READ ${LIBTOMMATH_RAND_PLATFORM} _contents)
+
+    string(
+        REPLACE
+        "if ((err != MP_OKAY) && MP_HAS(S_READ_ARC4RANDOM))"
+        "// if ((err != MP_OKAY) && MP_HAS(S_READ_ARC4RANDOM))"
+        _contents
+        "${_contents}"
+    )
+
+    string(
+        REPLACE
+        "if ((err != MP_OKAY) && MP_HAS(S_READ_GETRANDOM))"
+        "// if ((err != MP_OKAY) && MP_HAS(S_READ_GETRANDOM))"
+        _contents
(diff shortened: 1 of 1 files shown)


ports/libtommath/portfile.cmake
```diff
@@ -15,6 +15,44 @@ if(VCPKG_TARGET_IS_WINDOWS)
         set(CRTFLAG "/MT")
     endif()
 
+    set(LIBTOMMATH_RAND_PLATFORM ${SOURCE_PATH}/bn_s_mp_rand_platform.c)
```
Neumann-A, line 18: This should really be a patch instead.

andrew-boyarshin: Then it would be impossible to build HEAD_REF and current (very old) release. The patch will never apply cleanly.

JackBoosY: I agree with @Neumann-A. We must explicitly state what needs to be fixed (using patch).

JackBoosY (changes requested):

andrew-boyarshin:

But 1.2.0 _is_ the latest. And it _is_ very old, releases are rare for `libtom*` libs.

JackBoosY:
@andrew-boyarshin Add `Port-Version: 2` and remove `-1` here:
https://github.com/microsoft/vcpkg/blob/5dc53211caedebf4387d590155ed53ee44161f10/ports/libtommath/CONTROL#L2

JackBoosY:
Pinging @andrew-boyarshin  for response. Is work still being done for this PR?

andrew-boyarshin:
Not really. I can patch vcpkg port file myself when I need to, creating different patches for latest release and HEADREF + adding relevant CMake logic is too much of a hassle to do when you don't really need that. I don't mind if someone picks this PR up, of course.

JackBoosY:
Close this PR and wait for other contributors to take over.

PR #14564, opened 2020-11-13
[vcpkg] fix cmake paths in tags (alternative)

Neumann-A:
alternative to #13710 (after discussion on slack or discord.... cannot remember anymore)

cc @ras0219-msft @ras0219

### DIFF
--- a/toolsrc/src/vcpkg/cmakevars.cpp
+++ b/toolsrc/src/vcpkg/cmakevars.cpp
@@ -88,12 +88,28 @@ namespace vcpkg::CMakeVars
         Strings::append(extraction_file, "macro(vcpkg_triplet_file VCPKG_TRIPLET_ID)\n");
         for (auto& p : emitted_triplets)
         {
-            Strings::append(extraction_file,
-                            "if(VCPKG_TRIPLET_ID EQUAL ",
-                            p.second,
-                            ")\n",
-                            fs.read_contents(paths.get_triplet_file_path(p.first), VCPKG_LINE_INFO),
-                            "\nendif()\n");
+            const auto path_to_triplet = paths.get_triplet_file_path(p.first).string();
+            auto triplet_contents = fs.read_contents(path_to_triplet, VCPKG_LINE_INFO);
+
+            const std::string cmake_current_list_dir{"${CMAKE_CURRENT_LIST_DIR}"};
+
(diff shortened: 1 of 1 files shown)


toolsrc/src/vcpkg/cmakevars.cpp
```diff
@@ -88,12 +88,28 @@ namespace vcpkg::CMakeVars
...
+            {
+                if (cmake_current_list_dir_pos != 0)
+                {
+                    if (triplet_contents.compare(cmake_current_list_dir_pos - 1, std::string::npos, R"(\)"))
+                    {
+                        cmake_current_list_dir_pos += cmake_current_list_dir.length();
+                        continue;
+                    }
+                }
+                triplet_contents.replace(cmake_current_list_dir_pos, cmake_current_list_dir.length(), path_to_triplet);
```
strega-nil, line 107:
```suggestion
                triplet_contents.replace(cmake_current_list_dir_pos, cmake_current_list_dir.size(), path_to_triplet);
```

Neumann-A: because it is shorter?

strega-nil: It's the one we use as standard.

toolsrc/src/vcpkg/cmakevars.cpp
```diff
@@ -88,12 +88,28 @@ namespace vcpkg::CMakeVars
...
+            const std::string cmake_current_list_dir{"${CMAKE_CURRENT_LIST_DIR}"};
+
+            auto cmake_current_list_dir_pos = triplet_contents.find(cmake_current_list_dir, 0);
+            while (cmake_current_list_dir_pos != std::string::npos)
+            {
+                if (cmake_current_list_dir_pos != 0)
+                {
+                    if (triplet_contents.compare(cmake_current_list_dir_pos - 1, std::string::npos, R"(\)"))
+                    {
+                        cmake_current_list_dir_pos += cmake_current_list_dir.length();
```
Neumann-A, line 103:
```suggestion
                        cmake_current_list_dir_pos += cmake_current_list_dir.size();
```

strega-nil: mmh, thanks

PR #14535, opened 2020-11-12
[libarchive] Fix dependencies, combine patches

JackBoosY:
- Add `REQUIRED` on `find_package`.
- Combine all fix dependecy related patches
- Use `OpenSSL::Crypto` instead of `OPENSSL_CRYPTO_LIBRARY` to avoid missing openssl's dependencies.

Fixes #12750.

All features are default.

PR #14571, opened 2020-11-14
Revert "[libarchive] Fix dependencies, combine patches"

Neumann-A:
Reverts microsoft/vcpkg#14535

That was not the right fix for the mentioned issue. 

The correct fix is to use `OPENSSL_CRYPTO_LIBRARIES`  and fix it in a vcpkg-cmake-wrapper.cmake. Same for `OPENSSL_SSL_LIBRARIES`

Furthermore zstd linkage is completly wrong.


Neumann-A:
closing in favor of #14579

PR #14575, opened 2020-11-15
[vcpkg/scripts] Make cmake-vars file OPTIONAL

ekilmer:
This fixes the case where a custom triplet specifies only one build type.

ping @Neumann-A , author of https://github.com/microsoft/vcpkg/pull/12936 where this line was introduced.

There's almost definitely a more robust solution than this, like one that might warn if the file was missing, but I feel like this PR is a good enough fix, since my custom triplet (Release-builds only) is broken without it.


ekilmer:
Seems `libarchive` is failing. It shouldn't be from my changes... should it?

Neumann-A:
please close. The include should not be optional -> see #14580

ekilmer:
@Neumann-A Thank you for implementing a better solution!

PR #14576, opened 2020-11-15
[seal] Update to version 3.6.0. Changed default features.

WeiDaiWD:
**Describe the pull request**

Updated ports/seal to version 3.6.0. A new feature zstd is added. Default features include all optional features, because these features are all recommended. Both dynamic linking and static linking work except for Windows where only static linking is supported. If on Windows dynamic linking is detected, a CMake fatal error will be thrown with an explanation. Previous patch file is removed since SEAL 3.6.0 includes necessary changes to support vcpkg.

Updated ports/seal to version 3.6.0. Changed default features to enable all.

Supported: x64-linux, x64-windows-static, x86-windows-static.

Not supported: x64-windows, x86-windows, arm64-windows.

Not tested but should be supported: x64-osx and other linux/osx community triplets.

I have not updated the CI baseline.

Yes.


ports/seal/portfile.cmake
```diff
@@ -1,43 +1,58 @@
...
+set(USE_ZSTD OFF)
+if("ms-gsl" IN_LIST FEATURES)
+    set(USE_MSGSL ON)
+endif()
 if("zlib" IN_LIST FEATURES)
-    message("SEAL currently does not support non-vendored zlib -- ignoring feature 'zlib'")
+    set(USE_ZLIB ON)
+endif()
+if("zstd" IN_LIST FEATURES)
+    set(USE_ZSTD ON)
```
NancyLi1013, line 23: Could you please use `vcpkg_check_features()` to handle these features here?

ports/seal/portfile.cmake
```diff
@@ -1,43 +1,58 @@
-vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+if(NOT VCPKG_CMAKE_SYSTEM_NAME AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic") # Win32 and dynamic
+    message(FATAL_ERROR "Only static linking is supported on Windows. Choose '*-windows-static' triplets.")
+endif()
```
NancyLi1013, line 3:
```suggestion
if(VCPKG_TARGET_IS_WINDOWS) # Win32 and dynamic
    vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
endif()
```

NancyLi1013 (changes requested):

ports/seal/portfile.cmake
WeiDaiWD: Fixed in the latest commit.

ports/seal/portfile.cmake
WeiDaiWD: Fixed in the latest commit. Thx!

WeiDaiWD:
Windows triplets failed randomly. Why is that?

NancyLi1013:
```
CMake Error at CMakeLists.txt:578 (configure_file):
  configure_file Problem configuring file


CMake Error at CMakeLists.txt:609 (configure_file):
  configure_file Problem configuring file
```

You can try to add `DISABLE_PARALLEL_CONFIGURE ` to `vcpkg_configure_cmake()`.

WeiDaiWD:
Is this hanging test normal?

WeiDaiWD:
> ```
> CMake Error at CMakeLists.txt:578 (configure_file):
>   configure_file Problem configuring file
> 
> 
> CMake Error at CMakeLists.txt:609 (configure_file):
>   configure_file Problem configuring file
> ```
> 
> You can try to add `DISABLE_PARALLEL_CONFIGURE ` to `vcpkg_configure_cmake()`.

This is resolved. Thanks!

WeiDaiWD:
I'm waiting for an update from SEAL. Please don't merge yet. Is there a way to point to a repository's branch name rather than a commit hash? So that we can always get an up-to-date version.

NancyLi1013:
> Is there a way to point to a repository's branch name rather than a commit hash? So that we can always get an up-to-date version.

Currently, there is no way to do this. Actually, we don't recommend to do like this. We prefer to use the commit for a stable version rather than a branch or a tag. Since the commit is the only one, which is more stable to be maintained.

WeiDaiWD:
> > Is there a way to point to a repository's branch name rather than a commit hash? So that we can always get an up-to-date version.
> 
> 
> 
> Currently, there is no way to do this. Actually, we don't recommend to do like this. We prefer to use the commit for a stable version rather than a branch or a tag. Since the commit is the only one, which is more stable to be maintained.

Got it! Thanks! Sorry for the confusion, my latest changes to commit hash do not affect the library built. Those updates do not affect vcpkg. Is there anything else I should do for this PR?

ports/seal/portfile.cmake
```diff
@@ -1,43 +1,49 @@
-vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+if(VCPKG_TARGET_IS_WINDOWS) # Win32 and dynamic
+    vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+endif()
```
NancyLi1013, line 3:
```suggestion
    vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
This change seems to be unnecessary. Since vcpkg only supports static build on Unix platform. Sorry, I didn't notice this point before.

NancyLi1013 (changes requested):

NancyLi1013:
There is nothing else needs to be done for this PR now except for the last review suggestion.

ports/seal/portfile.cmake
```diff
@@ -1,43 +1,47 @@
-vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+    vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
strega-nil, line 1:
```suggestion
vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```

strega-nil:
Thanks @WeiDaiWD!

PR #14605, opened 2020-11-17
CGAL: Update to 5.1.1

maxGimeno:
Upgrade CGAL to 5.1.1


ports/cgal/portfile.cmake
```diff
@@ -2,9 +2,9 @@ vcpkg_buildpath_length_warning(37)
 
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
-    REPO CGAL/cgal
-    REF b271ec54df4547dd63db6b84efc5b5ecd57750a7 # v5.1
-    SHA512 c1a9166eaa3b437f0f164fc04b76d2b3805a284f333a3c40b4045160210fc331f6c5d2a9fa73e3959d8e3bedbe88abc38786552e2ce5a121b3e6dda27818e103
+    REPO CGAL
```
NancyLi1013, line 5:
```suggestion
    REPO CGAL/cgal
```
Why do you update the repo?

NancyLi1013 (changes requested):

ports/cgal/portfile.cmake
maxGimeno: I'd rather use a litteral ref than a commit number, so I used a more specific repo to be able to use the v5.1.1 REF. It will make it easier to maintain later, I think.

lrineau: The is that `CGAL` is not a repo, on Github. That is the CGAL organization. I do not see how it can work at all.

maxGimeno: Ok it works with CGAL/cgal, it must have been a problem of branch not up to date when I tested it yesterday.

NancyLi1013:
Have you tested the feature `qt`?

NancyLi1013 (approved):

maxGimeno:
Yes

PR #14613, opened 2020-11-17
[triplets] Add v141 triplets

rleigh-codelibre:
To permit the building of packages using the *v141* platform toolset on systems with both VS2019 and VS2017.

- Fixes building on systems with both VS2019 and VS2017 installed.
- Adds community triplets `x64-windows-v141` and `x86-windows-v141`
- Copies convention of existing `x86-windows-v120` community triplet

I haven't updated the CI baseline, it didn't seem necessary.  I'm not sure what your policy is for testing community triplets, but adding a trivial package to test each wouldn't hurt.

To further explain the necessity of these triplets, [this build log](https://gitlab.com/rleigh/ome-files-cpp/-/jobs/855640802) is an example of a build failure caused by platform toolset incompatibility:

```
spdlogd.lib(spdlog.cpp.obj) : error LNK2001: unresolved external symbol __CxxFrameHandler4
spdlogd.lib(async.cpp.obj) : error LNK2001: unresolved external symbol __CxxFrameHandler4
spdlogd.lib(color_sinks.cpp.obj) : error LNK2001: unresolved external symbol __CxxFrameHandler4
spdlogd.lib(spdlog.cpp.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck_EH4
spdlogd.lib(async.cpp.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck_EH4
spdlogd.lib(color_sinks.cpp.obj) : error LNK2001: unresolved external symbol __GSHandlerCheck_EH4
test\minimaltiffreader.exe : fatal error LNK1120: 2 unresolved externals
```

The build system has both VS2019 and VS2017.  vcpkg has built all the ports with `x64-windows` which defaulted to *v142*, and this is the first time I've seen an incompatibility.  The above error is with a VS2017 build.  Having a triplet to force building against *v141* will permit both Visual Studio versions to be supported on the same CI node with both variants built with the same vcpkg repository clone.


JackBoosY:
I think this is not necessary. The user can manually set the toolset to achieve this, and when using the same toolset, there will be no incompatibility issues.
@strega-nil What do you think about?

strega-nil:
@JackBoosY I think that you have the right idea, yeah. This is how we expect people to deal with older macOS versions, for example

rleigh-codelibre:
"The user can manually set the toolset"

Where would you manually set it if not in the triplet file?  Bear in mind I need to build arbitrary stuff for both VS2019 and VS2017 and I need vcpkg to build everything twice for both toolsets.

JackBoosY:
@rleigh-codelibre What scenarios require you to use v140 and v141 at the same time?
Usually, the users only use one toolset, either v140, or v141, or v142.

rleigh-codelibre:
I am building a collection of libraries with vcpkg for use with VS2017 (with this `x64-windows-v141` triplet) and then the default set for VS2019 (with the standard `x64-windows` triplet).  Builds will then use `VCPKG_TARGET_TRIPLET` to select the set to use as part of the CI [machine definition](https://gitlab.com/codelibre/ome/ome-gitlab-ci/-/blob/master/machines.yml#L54) and then used [here](https://gitlab.com/codelibre/ome/ome-gitlab-ci/-/blob/master/scripts.yml#L133).

The actual software being built by the CI build node using the vcpkg-built libraries has no idea about any of the vcpkg details.  The bit I need to control is the platform toolset used when building the collection, and it looked like adding a triplet file as here was the best option for that.  If there's a better alternative to achieve the same effect, I'll be happy to consider it.

strega-nil:
@rleigh-codelibre what we expect is that one adds a custom triplet for their own use, since we don't wish to have it in the mainline repo. We expect that people use `--overlay-triplets` to do this; put the custom `x64-windows-v141` into a `custom_triplets` folder, and pass `--overlay_triplets=C:/absolute/path/to/custom_triplets` or add `OVERLAY_TRIPLETS=C:/absolute/path/to/custom_triplets` to the environment variables.

rleigh-codelibre:
Thanks for the clarification.

It looks like overlay triplets could be used to solve the problem, so thanks for the suggestion.

I think your policies and suggestions here would be useful to document, because the existence of the "community" triplets implies that additions here would be accepted if there was a need, and existing community triplets like `x86-windows-v120` are clearly doing pretty much the same thing as this PR and have been accepted in the past.  The documentation regarding triplets does not explain any of this.  Looking at [stable/users/triplets](https://vcpkg.readthedocs.io/en/stable/users/triplets/) I see no mention at all of "overlay triplets" on the page documenting triplets.  Now I know what to look for, a search took me to [this example page](https://vcpkg.readthedocs.io/en/stable/examples/overlay-triplets-linux-dynamic/), but I don't see a cross-reference from anywhere else to hint that it exists.  It might be worth considering moving the documentation of the option from the example page onto the main page, and then linking to the example.

Kind regards,
Roger

strega-nil:
@rleigh-codelibre yes, we absolutely need to document this and discuss the best way forward among the team. I'll open an issue.

Thank you so much,
Nicole

PR #14549, opened 2020-11-12
[vcpkg] initial openbsd (community) support

jgilje:
**Describe the pull request**
This PR adds OpenBSD support for vcpkg. I've tested a couple of packages so far, which compile fine (libpq, botan, brotli, sdl2 with dependencies).The triplet-definition is added in the community folder. The patch consists mostly of a set of defines, I was helped in large part by looking at the previous FreeBSD support.

I've defaulted the port to use clang as the gcc included with OpenBSD is quite old (the last GPL2 release). In vcpkg_configure_make(), I've changed the behaviour when running configure to lookup bash in path first.

Quick instructions (included in the triplet file)
Bootstrap using
`CC=/usr/bin/clang CXX=/usr/bin/clang++ ./bootstrap-vcpkg.sh`
Use with
`VCPKG_FORCE_SYSTEM_BINARIES=1 ./vcpkg install brotli `

- What does your PR fix?
OpenBSD support for vcpkg

- Which triplets are supported/not supported? Have you updated the CI baseline?
x64-openbsd

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
Yes

PR #14614, opened 2020-11-17
[vcpkg docs] Check for documentation generation in CI

strega-nil:
Also, regenerate the documentation yay!


strega-nil:
Depends on #14549

ras0219 (approved):

PR #14615, opened 2020-11-17
[liblzma] use upstream CMakeLists.txt + Add pkgconfig.

Neumann-A:
lets see what will break


NancyLi1013 (changes requested):

NancyLi1013:
Seems you forget to remove CMakeLists.txt from `ports/liblzma` folder.

NancyLi1013 (approved):

ports/liblzma/win_output_name.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 120dc7548..728ca3591 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -398,11 +398,12 @@ else()
+ endif()
+ 
+ if(WIN32)
++    set_target_properties(liblzma PROPERTIES OUTPUT_NAME lzma)
```
JackBoosY, line 9: This is not the official behavor, should we apply this?

JackBoosY (changes requested):

ports/liblzma/win_output_name.patch
Neumann-A: i think that is a difference between CMake based builds and autoconf based builds. If I would switch to autoconf here it is highly likely that the name of the library would simply be `lzma.lib`. In cmake it is simply the target name. I just added this so I don't have to deal with the fallout in downstream ports not using cmake. This is also the reason why you should use find_library in the portfiles instead of hardcoding library names......

JackBoosY (approved):

ports/liblzma/vcpkg-cmake-wrapper.cmake
```diff
@@ -1,4 +1,8 @@
 list(REMOVE_ITEM ARGS "NO_MODULE")
 list(REMOVE_ITEM ARGS "CONFIG")
-list(REMOVE_ITEM ARGS "MODULE")
+list(REMOVE_ITEM ARGS "MODULE") # Always call CONFIG.
 _find_package(${ARGS} CONFIG)
+if(NOT TARGET LibLZMA::LibLZMA AND TARGET liblzma::liblzma)
+    add_library(LibLZMA::LibLZMA INTERFACE IMPORTED) # Too lazy to fix wrong target usage all over vcpkg.
```
strega-nil, line 6: (* snrk *)

strega-nil:
Awesome, thanks @Neumann-A :)

PR #14629, opened 2020-11-18
[proj] Added support for optionally building the tools

SNiLD:
**Describe the pull request**

- What does your PR fix?
  - Added support for building PROJ tools which are generally needed when developing with the library (but not actual mandatory dependency).

  - No changes to how the main library is built.

  - Yes. Tools are installed in tools directory and debug versions of tools are removed. `Port-Version` is not increased since library itself is not modified.

### DIFF
--- a/ports/proj4/portfile.cmake
+++ b/ports/proj4/portfile.cmake
@@ -22,6 +22,12 @@ endif()
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     database BUILD_PROJ_DATABASE
+    tools BUILD_CCT
+    tools BUILD_CS2CS
+    tools BUILD_GEOD
+    tools BUILD_GIE
+    tools BUILD_PROJ
+    tools BUILD_PROJINFO
 )
 if ("database" IN_LIST FEATURES)
     if (VCPKG_TARGET_IS_WINDOWS)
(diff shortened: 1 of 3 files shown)


NancyLi1013 (commented):
Could you please also update `Port-Version`?

NancyLi1013:
Have you tested these features? Can they work fine?

ports/proj4/portfile.cmake
```diff
@@ -53,18 +59,17 @@ vcpkg_configure_cmake(
...
     -DPROJ_TESTS=OFF
     -DEXE_SQLITE3=${SQLITE3_BIN_PATH}/sqlite3${BIN_SUFFIX}
 )
 
 vcpkg_install_cmake()
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/proj4)
+if ("tools" IN_LIST FEATURES)
+    vcpkg_copy_tools(TOOL_NAMES cct cs2cs geod gie proj projinfo AUTO_CLEAN)
+    file(GLOB debug_tools ${CURRENT_PACKAGES_DIR}/debug/bin/*_d${BIN_SUFFIX})
+    file(REMOVE ${debug_tools})
```
NancyLi1013, line 71: Can you rename debug tools name as the same name with release? If so, we only use `vcpkg_copy_tools()` is enough.

NancyLi1013 (changes requested):

ports/proj4/portfile.cmake
SNiLD: Sure, but that will complicate the code a lot.

SNiLD: It's probably better to patch the src\CMakeLists.txt so that it doesn't add the `CMAKE_DEBUG_POSTFIX` to binary targets.

SNiLD:
> Have you tested these features? Can they work fine?

Yeah I have used the tools and built the library without the tools as well. vcpkg copies dependencies (proj.dll) correctly to `tools/proj4` with the executables. When using PROJ tools you have to set `PROJ_LIB` environment variable to point to the `share/proj4` (with full path) yourself.

NancyLi1013 (approved):

strega-nil:
Awesome, thanks @SNiLD

PR #14633, opened 2020-11-18
[vcpkg] Introduce `create_versioned_install_plan()`

ras0219:
This PR introduces a consumer for the interfaces implemented in #14123 that implements the resolution algorithm described in #11758.

It is not currently accessible via a user-facing command.


toolsrc/include/vcpkg/dependencies.h
```diff
@@ -172,5 +172,16 @@ namespace vcpkg::Dependencies
                                                            std::vector<std::string> features,
                                                            CMakeVars::CMakeVarProvider& var_provider);
 
+    /// <param name="provider">Contains the ports of the current environment.</param>
+    /// <param name="specs">Feature specifications to resolve dependencies for.</param>
+    /// <param name="status_db">Status of installed packages in the current environment.</param>
+    ExpectedS<ActionPlan> create_versioned_install_plan(PortFileProvider::IVersionedPortfileProvider& vprovider,
+                                                        PortFileProvider::IBaselineProvider& bprovider,
```
vicroms, line 179: As implemented in #14123  these can be made `const`.

toolsrc/src/vcpkg/dependencies.cpp
```diff
@@ -1151,4 +1151,663 @@ namespace vcpkg::Dependencies
...
+    {
+        struct VersionedPackageGraph
+        {
+        private:
+            using IVersionedPortfileProvider = PortFileProvider::IVersionedPortfileProvider;
+            using IBaselineProvider = PortFileProvider::IBaselineProvider;
+
+        public:
+            VersionedPackageGraph(IVersionedPortfileProvider& ver_provider,
+                                  IBaselineProvider& base_provider,
```
vicroms, line 1165: These can be made const references.

PR #14650, opened 2020-11-19
[tesseract] Enable dynamic build, format portfile.cmake

### LINKED ISSUE #10174: [tesseract] Add port for creating dynamic library instead of command line
Installing tesseract will produce the command line version static linked with library. It would be nice to be able to build the dynamic link library of tesseract to be used from another application.
I know it's possible with CMake but that defeats the purpose of vcpkg.

JackBoosY:
- Enable dynamic build
- Since some libraries required by training tools doesn't export any symbols, set them build as a static libraries
- Format `portfile.cmake`

Fixes #10174.

All features are tested succesfully on the following triplets:
- `x86-windows`
- `x64-windows-static` (since dependency glib doesn's support this triplet, skip)

### DIFF
--- a/ports/tesseract/CONTROL
+++ b/ports/tesseract/CONTROL
@@ -1,6 +1,6 @@
 Source: tesseract
 Version: 4.1.1
-Port-Version: 2
+Port-Version: 3
 Homepage: https://github.com/tesseract-ocr/tesseract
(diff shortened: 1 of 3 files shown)


NancyLi1013 (approved):

strega-nil:
Thanks @JackBoosY

PR #14653, opened 2020-11-19
[python3] fix compatibility with macOS 11 big sur

cenit:
**Describe the pull request**
fix compatibility with macOS Big Sur

### DIFF
--- a/ports/python3/0003-big-sur.patch
+++ b/ports/python3/0003-big-sur.patch
@@ -1,0 +1,45 @@
+From 8ea6353f60625c96ce96588c70ff24a77f8c71f9 Mon Sep 17 00:00:00 2001
+From: Ronald Oussoren <ronaldoussoren@mac.com>
+Date: Thu, 25 Jun 2020 16:55:48 +0200
+Subject: [PATCH] BPO-41100: Support macOS 11 when building (GH-21113)
+
+---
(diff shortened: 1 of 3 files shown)


cenit:
Almost sure that qt5-location is broken not because of this PR

Hoikas:
Big Sur is fixed in 3.9.0 according to the Python release notes. See #14510.

JackBoosY:
@cenit what do you think? should we wait for #14510 merge?

cenit:
This PR should be much lighter and unblock macOS 11, while waiting for the very nice update to python 3.9. The problem was especially in the new number. Nothing very serious beside that is applied in this patch, which is already in use in the homebrew port.

JackBoosY (approved):

Hoikas:
I disagree. The patch included here was introduced into python 3.8.4 and does not in fact completely support big sur, as mentioned in the python release notes. It merely fixes the error in the configure script. At bare minimum, the port should be bumped to 3.8.4. The correct fix is to adopt #14510, which is ready (except for false negatives in the baseline).

Hoikas:
See also: [Python 3.8 Changelog](https://docs.python.org/3.8/whatsnew/changelog.html#id31)

strega-nil:
Closed in favor of #14510

PR #10030, opened 2020-02-12
[duktape] x64-osx fail

strega-nil:
brew no longer contains python2, since it's EOL, and duktape requires that for now

see https://github.com/svaarala/duktape/issues/1794

PR #14662, opened 2020-11-20
Use OS-included Python 2.7 on macOS

LRFLEW:
This fixes building `duktape`\*, as well as other ports that depend on python2 (I haven't done a full scan for what else would be included). I didn't find a Github issue for this, but it does relate to PR #10030.

This updates `vcpkg_find_acquire_program(PYTHON2)` to change `PROGNAME` from `python2` to `python` on macOS. Apple includes Python 2.7 under this name, but not as `python2`, in all (supported) OS releases. It *is* deprecated, but it is included in 11.0 Big Sur from what I've read, and because Python 2 was removed from Homebrew, this will allow it to work for now at least.

\* `duktape` still has a minor issue caused by `easy_install` needing to be executed as root, but installing `pyyaml` ahead of time allows the build to succeed, even with `easy_install` failing.


ras0219 (approved):
LGTM; could you also consider adding a message to duktape about it's pyyaml dependency? Something like 
```
if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
  message(STATUS "Duktape requires pyyaml to build -- this is available with `easy_install pyyaml`")
endif()
```
or somesuch

LRFLEW:
I was planning on making an issue for the pyyaml duktape issue once this was merged. (I was trying to keep this change in the tooling separate from the changes in the port). There seems to be some complexity to it, because the fix could be changing `easy_install pyyaml` to **either** `sudo easy_install pyyaml` or `easy_install --user pyyaml`. There's also the fact that the `easy_install` command requires having run `brew install libyaml` ahead of time (from what I can tell at least).

JackBoosY (approved):

strega-nil:
@LRFLEW thanks!

PR #14666, opened 2020-11-20
[duktape] Add VERSION and SOVERSION properties CMake build

LRFLEW:
This is a fairly simple change that should have minimal impact. The biggest change is that it will add the version numbers to the dynamic libraries built for macOS and Linux using custom or community triplets. The `SOVERSION` is set to the major version, as the project's [semantic versioning](https://duktape.org/guide.html#versioning) indicates this is the API compatibility value.


NancyLi1013 (commented):
Could you please also add `Port-Version` to the CONTROL file?

NancyLi1013 (approved):

strega-nil:
LGTM, thanks @LRFLEW

PR #14669, opened 2020-11-20
[curl] Correct feature non-http description and add warning message

### LINKED ISSUE #14030: Unable to compile curl with LDAPS
Host Environment:
Windows 10 ver 1909, compiler : VS 2019

I tried to compiled Curl with vcpkg , and was unable to get the protocol LDAPS to be enabled in curl (I tried many options). The configuration seems to enable LDAP but not LDAPS. 

Steps to reproduce : 
Compile curl with tool (and openssl but this does not seem to matter)(in x64 Native commands tools): 
vcpkg install curl[tool,openssl] --triplet x64-windows

Then goes to package packages\curl_x64-windows\tools\curl and do:
curl ldaps://localhost
curl ldap://localhost
Result :
curl: (1) Protocol "ldaps" not supported or disabled in libcurl
curl: (7) Failed to connect to localhost port 389: Connection refused

I expect the first message to be somewhat the same as the second (here LDAPS is not activated).

When I compile LibCurl with nmake (going in the winbuild directory and doing something like :
"nmake /f Makefile.vc mode=dll WITH_SSL=dll SSL_PATH=[YOUR_SSL_PATH] VC=16 MACHINE=x64 GEN_PDB=yes ENABLE_UNICODE=yes") I am able to get the LDAPS support. 

At this point I am not sure if this is vcpkg related or just a bad makefile from LibCurl, but as nmake in winbuild works I guess that this may be related to vcpkg. 

config-x64-windows-out.log parts that can be relevant : 
-- Looking for cldap_open in wldap32;winmm;ws2_32;advapi32;crypt32;OpenSSL::SSL;OpenSSL::Crypto
-- Looking for cldap_open in wldap32;winmm;ws2_32;advapi32;crypt32;OpenSSL::SSL;OpenSSL::Crypto - found
-- Looking for include file winldap.h
-- Looking for include file winldap.h - not found
-- Looking for include file winber.h
-- Looking for include file winber.h - not found
-- Looking for include file ldap_ssl.h
-- Looking for include file ldap_ssl.h - not found
-- Looking for include file ldapssl.h
-- Looking for include file ldapssl.h - not found
[...]
-- Enabled protocols: DICT FILE FTP FTPS GOPHER HTTP HTTPS IMAP IMAPS LDAP POP3 POP3S RTSP SMB SMBS SMTP SMTPS TELNET TFTP
(no LDAPS here, it indicates it does not find winldap but I think it findsit because LDAP is activated at the end)

JackBoosY:
Since `non-http` is a default feature of curl and it's description is incorrect, fix the description and add warning message.

Fixes #14030.

Note: notify user to update this port is not needed, so don't update the port version.

### DIFF
--- a/ports/curl/CONTROL
+++ b/ports/curl/CONTROL
@@ -10,7 +10,7 @@ Feature: tool
 Description: Builds curl executable
 
 Feature: non-http
-Description: Enables protocols beyond HTTP/HTTPS/HTTP2
+Description: Enable all protocols except HTTP
 
 Feature: http2
 Build-Depends: nghttp2, curl[core,ssl]
--- a/ports/curl/portfile.cmake
+++ b/ports/curl/portfile.cmake
@@ -15,6 +15,10 @@ vcpkg_from_github(
         0011_fix_static_build.patch
 )
 
+if ("non-http" IN_LIST FEATURES)
+    message(WARNING "Using feature \"non-http\", if you wish to enable http, please install curl using command `./vcpkg install curl[core]:${TARGET_TRIPLET}`")
+endif()
+
 string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" CURL_STATICLIB)
 
 # schannel will enable sspi, but sspi do not support uwp


PhoebeHui (approved):

ras0219:
From the CMake sources[1]:
```cmake
option(HTTP_ONLY "disables all protocols except HTTP (This overrides all CURL_DISABLE_* options)" OFF)
```

This means _enabling_ `HTTP_ONLY` _disables_ ftp, ldap, etc. This does not fit the model of additive features (_enabling_ X _enables_ Y). Therefore, in vcpkg, it is an inverted feature -- _enabling_ `non-http` _enables_ ftp, ldap, etc. In this form, our description is correct:

>Feature: non-http
>Description: Enables protocols beyond HTTP/HTTPS/HTTP2

Enabling `non-http` enables ftp & friends, exactly as the current description states.

Additionally, as the poster comments here[2] this does not fix their issue. Their problem is that even when passing `non-http`, LDAPS is still not enabled.

[1] https://github.com/curl/curl/blob/c353207057b45c0e814823b5fe5eabeb4cc2d707/CMakeLists.txt#L159
[2] https://github.com/microsoft/vcpkg/issues/14030#issuecomment-730226850

JackBoosY:
Close this PR because the changes cannot solve the issue.

PR #14674, opened 2020-11-20
[vcpkg] Output the filepath on hash error. 

Neumann-A:
to have a simple and meaninful error message instead of a simple file not found error without any information about which file it actually tried to read.

### DIFF
--- a/toolsrc/include/vcpkg/base/hash.h
+++ b/toolsrc/include/vcpkg/base/hash.h
@@ -43,7 +43,8 @@ namespace vcpkg::Hash
         const auto result = get_file_hash(fs, path, algo, ec);
         if (ec)
         {
-            Checks::exit_with_message(li, "Failure to read file for hashing: %s", ec.message());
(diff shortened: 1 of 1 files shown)


toolsrc/include/vcpkg/base/hash.h
```diff
@@ -43,7 +43,8 @@ namespace vcpkg::Hash
         const auto result = get_file_hash(fs, path, algo, ec);
         if (ec)
         {
-            Checks::exit_with_message(li, "Failure to read file for hashing: %s", ec.message());
+            Checks::exit_with_message(
+                li, "Failure to read file '%s' for hashing: %s", fs::generic_u8string(path), ec.message());
```
strega-nil, line 47:
```suggestion
                li, "Failure to read file '%s' for hashing: %s", fs::u8string(path), ec.message());
```

no reason to use `generic` if it's not necessary.

strega-nil:
Cools, thanks @Neumann-A :)

PR #14676, opened 2020-11-20
[flashlight-cuda] New Port

jacobkahn:
Adds a new port for [flashlight](https://github.com/facebookresearch/flashlight), an ML library written in C++. I'm one of the project authors.

A few things to note:
- This port only contains support for the CUDA backend in flashlight. OpenCL/CPU backend support will be added in a separate port later when they are complete.
- The library depends on NCCL for which there doesn't currently exist a port ([I've just created a PR, ideally, we *commit that port first*](https://github.com/microsoft/vcpkg/pull/14683)) - it uses `find_package` natively with a `FindNCCL.cmake` module. `NCCL` and `cuDNN` are very similar in their installations next to CUDA, so the NCCL port works exactly like [the vcpkg cuDNN port](https://github.com/microsoft/vcpkg/tree/master/ports/cudnn).
  - CI will probably fail until the NCCL port is committed
- The features in the port depend on other features of the port:
```
lib
 ^
 |
 fl <— imgclass
 ^
 |
asr
```

Supports `linux` for now. I've tested `x64-linux`. The port has `!(windows|osx)`, but I will test the port on macOS shortly. I don't expect Windows integration to be complete for some time, but I will update the port if/when it is.

Yes.

cc @tlikhomanenko @vineelpratap @xuqiantong @avidov @andresy @syhw @padentomasello


JackBoosY (changes requested):
Can you revert changes about nccl and depends on #14683?

jacobkahn:
@JackBoosY done — I'll rebase on top of the NCCL port once it's done though.

jacobkahn:
@JackBoosY how can I get detailed logs of this failure in CI?

JackBoosY:
https://dev.azure.com/vcpkg/public/_build/results?buildId=45830&view=artifacts&pathAsName=false&type=publishedArtifacts
Do you have permission?

jacobkahn:
@JackBoosY I have permission. How do I get to this page for future CI runs?

JackBoosY:
@jacobkahn 
`Details`-> `View more details on Azure Pipelines` -> `* artifacts produced`

@strega-nil Maybe we need to write documentation to the contributors.

jacobkahn:
@JackBoosY is there some way I can know what command CI ran when this error happened? I'm unable to reproduce it.

JackBoosY:
@jacobkahn Maybe the environment of our CI machine is different from yours. You can only use the log content to solve the regression.

jacobkahn:
@JackBoosY fixed the issue. Needed a default feature.

strega-nil:
Thanks @jacobkahn

PR #14683, opened 2020-11-20
[nccl] New Port

jacobkahn:
Adds a port for [NCCL](https://developer.nvidia.com/nccl) ([repo](https://github.com/NVIDIA/nccl)), NVIDIA's collective multi-GPU communication library.

This port almost exactly mirrors the [cuDNN port](https://github.com/microsoft/vcpkg/tree/master/ports/cudnn):
- Checks if NCCL is available on the system. This is because the installation is somewhat specialized and new versions have improved performance but can still be compiled with CUDA 10.1 (what the [CUDA port](https://github.com/microsoft/vcpkg/tree/master/ports/cuda) is frozen at). The NCCL version minimum is `2.4.6`, so if a later version is found, that satisfies the install requirement.
- Downloads `NCCL` from [Anaconda](https://anaconda.org/nvidia/nccl) if not. Downloads a version consistent with CUDA 10.1.
- If there's a better way to call `find_package` inside the portfile to simulate finding the package, let me know/feel free to commit.

`linux` only, which is what NCCL supports. I've tested in `x64` linux linking and building with other packages.

Yes


ports/nccl/portfile.cmake
```diff
@@ -0,0 +1,62 @@
+if(VCPKG_TARGET_IS_UWP OR VCPKG_TARGET_IS_OSX OR VCPKG_TARGET_IS_WINDOWS)
+  message(FATAL_ERROR "This port is only for Linux")
+endif()
+
+if(NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "x64")
+  message(FATAL_ERROR "This port is only for x64 architectures")
+endif()
```
JackBoosY, line 7:
```suggestion
vcpkg_fail_port_install(ON_TARGET "Windows" "OSX" ON_ARCH "x86" "arm")
```

JackBoosY (changes requested):

JackBoosY:
Will test the usage locally.

JackBoosY (changes requested):
Since you provided the cmake wrapper, you should also provide a usage.

jacobkahn:
@JackBoosY added some usage in the wrapper. Is that what you meant?

jacobkahn:
I can also create an imported target (i.e. `add_library(NCCL::NCCL UNKNOWN IMPORTED)` in `FindNCCL.cmake` -- the only issue there is that `add_library` can't be called in script mode which is what will happen if `find_package(NCCL)` is called in the portfile. In that case, I can copy code from FindNCCL directly into the portfile...

Any thoughts on this/is there a best practice? Can also gate the creation of the imported target by `(NOT DEFINED ${CMAKE_SCRIPT_MODE_FILE})` which should only call `add_library` if in script mode, which means things won't break.

ports/nccl/portfile.cmake
```diff
@@ -0,0 +1,56 @@
...
+  file(INSTALL "${SOURCE_PATH}/lib/libnccl.so.${NCCL_VERSION_MAJOR}.${NCCL_VERSION_MINOR}.${NCCL_VERSION_PATCH}" DESTINATION ${CURRENT_PACKAGES_DIR}/lib)
+  file(INSTALL "${SOURCE_PATH}/lib/libnccl.so" DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib)
+  file(INSTALL "${SOURCE_PATH}/lib/libnccl.so.${NCCL_VERSION_MAJOR}" DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib)
+  file(INSTALL "${SOURCE_PATH}/lib/libnccl.so.${NCCL_VERSION_MAJOR}.${NCCL_VERSION_MINOR}.${NCCL_VERSION_PATCH}" DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib)
+
+  file(INSTALL "${SOURCE_PATH}/info/licenses/LICENSE.txt" DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+  file(INSTALL "${CMAKE_CURRENT_LIST_DIR}/FindNCCL.cmake" DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT})
+  file(INSTALL "${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake" DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT})
+
+endif()
```
JackBoosY, line 56:
```suggestion
endif()

file(INSTALL "${CURRENT_PORT_DIR}/usage" DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT})
```

JackBoosY (changes requested):

JackBoosY:
The `NCCL_LIBRARIES` always only contain the debug library, can you confirm this?

jacobkahn:
@JackBoosY — only one library ships with NCCL when it's downloaded, and it's a release library. There isn't a debug library as far as I know.

JackBoosY:
```
NCCL_INCLUDE_DIRS: /home/work/14683/vcpkg/installed/x64-linux/include
NCCL_LIBRARIES: /home/work/14683/vcpkg/installed/x64-linux/debug/lib/libnccl.so

```

jacobkahn:
@JackBoosY — nope, that's a mistake. Fixing.

jacobkahn:
@JackBoosY is it viable to not copy the library to `${CURRENT_PACKAGES_DIR}/debug`, or does it need to be there? If it has to be copied, I'll have to change `FindNCCL.cmake` to not find it there.

JackBoosY:
@jacobkahn Different types of libraries should be copied to the corresponding path, if it does not provide a debug version, you do not need to copy.

jacobkahn:
@JackBoosY sounds good. I won't copy. I will also send a PR fixing the cuDNN port.

JackBoosY (approved):

jacobkahn:
@JackBoosY  this don't actually have a dependency on a different PR. It can be merged as is.

jacobkahn:
@JackBoosY @strega-nil - looks like three remaining CI jobs have all OOMed and have been stuck for several hours - can we restart them?

jacobkahn:
👍  — is this ready to merge @JackBoosY @strega-nil?

strega-nil:
Thanks @jacobkahn :)

PR #14686, opened 2020-11-21
[cudnn] Try to find CUDNN locally before downloading

jacobkahn:
Don't download cuDNN unless it's not found on the local system.

This allows using newer versions of cuDNN (esp those compiled with newer versions of CUDA) as dependencies in packages.

Same as before. Tested `x64-linux`.

Yes

### DIFF
--- a/ports/cudnn/CONTROL
+++ b/ports/cudnn/CONTROL
@@ -1,6 +1,6 @@
 Source: cudnn
 Version: 7.6
-Port-Version: 2
+Port-Version: 3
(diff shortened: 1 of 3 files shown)


PhoebeHui (changes requested):
Thanks for the PR!

Could you update the Port-Version to 3 in CONTROL file?

jacobkahn:
@PhoebeHui done!

PhoebeHui (approved):
LGTM, @strega-nil, could you help further review?

strega-nil:
Awesome, thanks @jacobkahn :)

PR #14711, opened 2020-11-22
[geographiclib] Update to version 1.51

cffk:
This is just a routine upgrade.

- What does your PR fix? Routine upgrade to 1.51

- Which triplets are supported/not supported? Have you updated the CI baseline? No changes

### DIFF
--- a/ports/geographiclib/cxx-library-only.patch
+++ b/ports/geographiclib/cxx-library-only.patch
@@ -1,8 +1,6 @@
-diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 30875ddb..34aa6515 100644
---- a/CMakeLists.txt
-+++ b/CMakeLists.txt
-@@ -464,12 +464,12 @@ endif ()
+--- CMakeLists.txt.orig	2020-11-22 09:00:22.000000000 -0500
++++ CMakeLists.txt	2020-11-22 12:12:35.314649815 -0500
+@@ -434,12 +434,12 @@
  # documentation files into the source tree.  Skip Apple here because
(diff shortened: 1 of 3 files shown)


NancyLi1013 (approved):

NancyLi1013:
Have you tested the feature tool? @cffk

cffk:
> Have you tested the feature tool? @cffk

Yes, I have.  However I've moved this to draft so I can verify whether that it still works with shared libraries.  (This may be an issue for Linux, since the rpath is set to ../lib, which won't work once the tools are copied to tools/geographiclib.)

Also, should man pages be installed for the tools?  They aren't currently but they could easily be.

cffk:
OK, I'm done with my changes.  The main one was to fix the RPATH for the tools so that the shared libraries could be found from tools/geographiclib.  (I realize that this is moot because the linux builds are for a static library.  But I hope that at some point vcpkg supports dynamic libraries on linux.)

I've tested the tools feature on linux + windows + macos.

Should I worry about the failed checks?  They look like timeouts on downloading the source and other things (like geotrans) that I have no control over.

NancyLi1013:
Thanks for your update.

For downloading `geotrans`, it might be something with the network. I checked this on my machine and it can be downloaded successfully. I will try to rerun this PR.

For dynamic support on Linux, vcpkg might implement this point in future.

NancyLi1013:
LGTM now, thanks for your PR @cffk.

cffk:
Thanks @NancyLi1013   This PR is ready for merging!

strega-nil:
Thanks @cffk :)

PR #14718, opened 2020-11-23
[osg] Enable /MP option

### LINKED ISSUE #14695: [osg] Please add WIN32_USE_MP define for faster windows build
OSG windows build does not have a "WIN32_USE_MP" define.

https://github.com/openscenegraph/OpenSceneGraph/blob/50de0c25f70d251a000d7972d95e6c1d4249c7dd/CMakeLists.txt#L351

The build of osg take more than an hour on github actions.

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/14695

Build with multiple process for saving build time.

See https://github.com/openscenegraph/OpenSceneGraph/blob/50de0c25f70d251a000d7972d95e6c1d4249c7dd/CMakeLists.txt#L351

Doesn't need to test features.


JackBoosY (approved):

ports/osg/portfile.cmake
```diff
@@ -53,6 +53,10 @@ if (VCPKG_TARGET_IS_WINDOWS AND OSG_DYNAMIC)
 endif()
 list(APPEND OPTIONS -DENABLE_NVTT=${ENABLE_NVTT})
 
+if(VCPKG_TARGET_IS_WINDOWS)
+    list(APPEND OPTIONS -DWIN32_USE_MP=ON)
+endif()
+
```
Neumann-A, line 59: Is this really necessary? osg is a cmake based port and as far as I remember `/MP` is already part of the toolchain. 
https://github.com/microsoft/vcpkg/blob/42456b785a4508386f080ae8fad2642e056b747d/scripts/toolchains/windows.cmake#L22-L23
so please first check the buildlogs if /MP is really not used!

PhoebeHui: @Neumann-A, thanks for your review!

You're right, the /MP enabled in toolchain file, and already applied to osg. So this is not necessary.

PhoebeHui:
Close this PR since the /MP enabled in toolchain file, and already applied to osg. So this is not necessary.

PR #14723, opened 2020-11-23
[toml11] fixes installation

fran6co:
No need to build the tests and fixes the CMake integration

### DIFF
--- a/ports/toml11/CONTROL
+++ b/ports/toml11/CONTROL
@@ -1,4 +1,5 @@
 Source: toml11
 Version: 3.5.0
+Port-Version: 2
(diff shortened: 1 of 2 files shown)


ports/toml11/CONTROL
```diff
@@ -1,4 +1,5 @@
 Source: toml11
 Version: 3.5.0
+Port-Version: 2
```
Hoikas, line 3:
```suggestion
Port-Version: 1
```

JackBoosY (approved):

strega-nil:
Thanks @fran6co :)

PR #14749, opened 2020-11-24
[libuvc] added

fran6co:


ports/libuvc/portfile.cmake
```diff
@@ -0,0 +1,26 @@
...
+    REPO libuvc/libuvc
+    REF c612d4509eb0ff19ce414abc3dca18d0f6263a84
+    SHA512 df3f23463728e8ffd69dc52e251ea2610ea8df32b02f6d26dd2a6910cf217650245bb1a11e67be61df875c6992d592c9cb17675d914997bd72c9fe7eb5b65c32
+    HEAD_REF master
+    PATCHES
+        build_fix.patch
+)
+
+# Handle copyright
+file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
```
NancyLi1013, line 12: We prefer to handle copyright at last.

ports/libuvc/portfile.cmake
```diff
@@ -0,0 +1,26 @@
...
+
+vcpkg_configure_cmake(
+        SOURCE_PATH ${SOURCE_PATH}
+        PREFER_NINJA # Disable this option if project cannot be built with Ninja
```
NancyLi1013, line 16:
```suggestion
        PREFER_NINJA
```

ports/libuvc/portfile.cmake
```diff
@@ -0,0 +1,26 @@
...
+        OPTIONS -DBUILD_EXAMPLE=OFF
+)
+vcpkg_install_cmake()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib${LIB_SUFFIX}/cmake/libuvc)
```
NancyLi1013, line 21: Why do you add `LIB_SUFFIX` here?

ports/libuvc/CONTROL
```diff
@@ -0,0 +1,5 @@
+Source: libuvc
+Version: 0.0.7
```
NancyLi1013, line 2: It seems there is no 0.0.7 version for `libuvc`. Where do you get this version?

ports/libuvc/portfile.cmake
```diff
@@ -0,0 +1,26 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO libuvc/libuvc
+    REF c612d4509eb0ff19ce414abc3dca18d0f6263a84
```
NancyLi1013, line 4: The REF seems to be the latest commit instead of a stable version.

NancyLi1013 (changes requested):

ports/libuvc/portfile.cmake
fran6co: no reason, I think I just copy&pasted from another portfile

ports/libuvc/CONTROL
fran6co: The 0.0.6 is too old (2017) and it's missing important changes but no stable version is out yet. What version should I put here in this case?

NancyLi1013: In this case, we can consider to use a date that you accept the commit as the version. such as 2020-11-24.

It would be better to ask for help from upstream and let them make a new release for `libuvc`.

NancyLi1013 (approved):

NancyLi1013:
The failures on windows:

```
CMake Error:
  Running

   'D:/downloads/tools/ninja/1.10.1-windows/ninja.exe' '-C' 'D:/buildtrees/libuvc/x86-windows-rel' '-t' 'cleandead'

  failed with:

   ninja: error: build.ninja:229: multiple rules generate uvc.lib [-w dupbuild=err]
```
Needs to judge the value of `CMAKE_BUILD_TARGET` to choose which targets should be built.

Also the file ` #include <sys/time.h>` included in `include/libuvc/libuvc.h` doesn't exist on Windows platform, which also needs to fix.

fran6co:
@NancyLi1013 is it possible to mark this package as linux only? According to https://github.com/libuvc/libuvc/issues/12 it seems that Windows support is a bit uncomplete. The main reason I had to add this package is for getting azure-kinect-sensor-sdk to work on linux. I'm guessing that package already works on Windows without libuvc, right? So I could just make this dependency linux only

NancyLi1013:
> @NancyLi1013 is it possible to mark this package as linux only? According to [libuvc/libuvc#12](https://github.com/libuvc/libuvc/issues/12) it seems that Windows support is a bit uncomplete. The main reason I had to add this package is for getting azure-kinect-sensor-sdk to work on linux. I'm guessing that package already works on Windows without libuvc, right? So I could just make this dependency linux only

Sorry for the delay. Sure, we can only support Linux for `linuvc`. It might also support osx. You can try this. 

For the dependency, you can set it as` libuvc (!windows)` or `libuvc (linux)`. Please help confirm this point.

NancyLi1013:
@fran6co 

fran6co:
@NancyLi1013 not sure what else should I do for this, this package will not work for Windows without upstream changes

NancyLi1013:
You can add `Supports: ! windows` in CONTROL file until upstream supports it you can remove this.

NancyLi1013:
@fran6co 

Could you please address the review suggestion? Please let me know if you have any questions about this. Thanks.

fran6co:
@NancyLi1013 just added the `Supports` option

NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your PR @fran6co.

JackBoosY:
Ping @ras0219-msft for review and merge this PR since it block another PR.

ras0219-msft:
Thanks for the new port!

PR #14771, opened 2020-11-25
[elfutils] add new port

Neumann-A:
new port; required for mesa on linux


ports/elfutils/configure.ac.patch
```diff
@@ -0,0 +1,49 @@
...
+-eu_ZIPLIB(lzma,LZMA,lzma,lzma_auto_decoder,[LZMA (xz)])
++eu_ZIPLIB(lzma,LZMA,[lzma lzmad],lzma_auto_decoder,[LZMA (xz)])
+ AS_IF([test "x$with_lzma" = xyes], [LIBLZMA="liblzma"], [LIBLZMA=""])
+ AC_SUBST([LIBLZMA])
+-eu_ZIPLIB(zstd,ZSTD,zstd,ZSTD_decompress,[ZSTD (zst)])
++eu_ZIPLIB(zstd,ZSTD,[zstd zstdd zstd_static zstd_staticd],ZSTD_decompress,[ZSTD (zst)])
+ AS_IF([test "x$with_zstd" = xyes], [LIBZSTD="libzstd"], [LIBLZSTD=""])
+ AC_SUBST([LIBZSTD])
+-zip_LIBS="$LIBS"
++zip_LIBS="$LIBS $ac_cv_search_lzma_auto_decoder $ac_cv_search_BZ2_bzdopen $ac_cv_search_gzdirect"
```
Neumann-A, line 47:
```suggestion
+zip_LIBS="$LIBS $ac_cv_search_lzma_auto_decoder $ac_cv_search_BZ2_bzdopen $ac_cv_search_gzdirect $ac_cv_search_ZSTD_decompress"
```

PhoebeHui (approved):

PhoebeHui:
elfutils:x64-osx failed in git fetch steps, this issue could repro on local machine, @Neumann-A , could you help have a look?

git-fetch-x64-osx-err.log
```
error: Server does not allow request for unadvertised object b503c358dde835d8a1ae3ebd4968755ff396f814
```

```
-- Fetching https://sourceware.org/git/elfutils...
CMake Error at scripts/cmake/vcpkg_execute_required_process.cmake:106 (message):
    Command failed: /usr/bin/git fetch https://sourceware.org/git/elfutils b503c358dde835d8a1ae3ebd4968755ff396f814 --depth 1 -n
    Working Directory: /Users/vagrant/Data/downloads/git-tmp
    Error code: 128
      /Users/vagrant/Data/buildtrees/elfutils/git-fetch-x64-osx-err.log

  scripts/cmake/vcpkg_from_git.cmake:78 (vcpkg_execute_required_process)
  ports/elfutils/portfile.cmake:1 (vcpkg_from_git)

```

ports/elfutils/portfile.cmake
```diff
@@ -0,0 +1,51 @@
+vcpkg_from_git(
+    OUT_SOURCE_PATH SOURCE_PATH
+    URL https://sourceware.org/git/elfutils
+    REF b503c358dde835d8a1ae3ebd4968755ff396f814 #elfutils-0.182
```
PhoebeHui, line 4:
```suggestion
    REF 25d048684a82f9ba701c6939b7f28c3543bb7991 #elfutils-0.182

```

PhoebeHui (commented):
I tried the hash for the tag https://sourceware.org/git/?p=elfutils.git;a=tag;h=25d048684a82f9ba701c6939b7f28c3543bb7991, the git fetch works, it failed in config steps:

```
configure: WARNING: unrecognized options: --disable-shared, --enable-static
rm: conftest.dSYM: is a directory
configure: error: gcc with GNU99 support required
```

ports/elfutils/portfile.cmake
Neumann-A: Hmm okay. Which git version is used here? Maybe an update is needed. 
`b503c358dde835d8a1ae3ebd4968755ff396f814` worked for me.
I assume `25d048684a82f9ba701c6939b7f28c3543bb7991` will also work,
`vcpkg_from_git` sometimes a mystery to me.

Neumann-A:
> configure: error: gcc with GNU99 support required

which version of gcc? Or what is the output of the `config.log` in the subdir `<triplet>-dbg/`?

PhoebeHui:
Here is config log:
[failure logs for x64-osx (1).zip](https://github.com/microsoft/vcpkg/files/5606860/failure.logs.for.x64-osx.1.zip)

Neumann-A:
@PhoebeHui: That are not the logs I need. I need the literally called `config.log` from `<vcpkgroot>/buildtrees/<port>/<triplet>-<shortbuildtype>/config.log` Unfortunally `vcpkg_configure_make` is not able to copy that file into `buildtrees/<port>` since `vcpkg_execute_process` does error before I can inject the code for copying that file one folder down.

PhoebeHui:
Sorry, here is correct one:
[config.log](https://github.com/microsoft/vcpkg/files/5607030/config.log)

Neumann-A:
The error is:
```
configure:5165: gcc -c -fPIC -g -std=gnu99  conftest.c >&5
conftest.c:20:28: error: function definition is not allowed here
  double square (double z) { return z * z; }
                           ^
conftest.c:21:10: error: implicit declaration of function 'square' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  return square (a) + square (b);
         ^
conftest.c:26:18: error: fields must have a constant size: 'variable length array in structure' extension will never be supported
  struct S { int x[n]; };
                 ^
3 errors generated.
configure:5165: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "elfutils"
| #define PACKAGE_TARNAME "elfutils"
| #define PACKAGE_VERSION "0.182"
| #define PACKAGE_STRING "elfutils 0.182"
| #define PACKAGE_BUGREPORT "https://sourceware.org/bugzilla"
| #define PACKAGE_URL "http://elfutils.org/"
| #define PACKAGE "elfutils"
| #define VERSION "0.182"
| #define DEFAULT_AR_DETERMINISTIC false
| #define YYTEXT_POINTER 1
| /* end confdefs.h.  */
| int foo (int a)
| {
|   for (int i = 0; i < a; ++i) if (i % 4) break; int s = a; return s;
| }
|
| double bar (double a, double b)
| {
|   double square (double z) { return z * z; }
|   return square (a) + square (b);
| }
|
| void baz (int n)
| {
|   struct S { int x[n]; };
| }
configure:5173: result: no
configure:5176: error: gcc with GNU99 support required
```

which basically means that the compiler:

```
configure:3750: gcc --version >&5
Apple clang version 12.0.0 (clang-1200.0.32.2)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

does not support gnu extensions (-std=gnu99). As such I should probably put it on the CI skiplist on OSX? If a true gcc would be used on OSX it would probably compile.

PhoebeHui:
Could you please add it to 'Supports" instead?

Neumann-A:
@PhoebeHui: Supports seems wrong here. A custom toolchain switching to gcc on apple could compile this correctly.

PhoebeHui:
@Neumann-A, thanks for the information!

 I'm fine to put it on the CI skiplist on OSX, could add comments as well in ci.baseline.txt file?

PhoebeHui (approved):

strega-nil:
Thanks @Neumann-A :)

PR #14774, opened 2020-11-25
[rtaudio] Add new port

### LINKED ISSUE #11183: [New Port Request] RtAudio
Library name: RtAudio

Library description: RtAudio is a set of C++ classes that provide a common API for realtime audio input/output across Linux, Macintosh OS-X and Windows operating systems. 

Source repository URL: https://github.com/thestk/rtaudio

Project homepage (if different from the source repository): http://www.music.mcgill.ca/~gary/rtaudio/index.html

Anything else that is useful to know when adding (such as optional features the library may have that should be included): -

chausner:
**Describe the pull request**

- What does your PR fix?

This adds a new port for the latest release of [RtAudio](https://github.com/thestk/rtaudio). Fixes #11183.

Triplets x86-windows, x86-windows-static, x64-windows and x64-windows-static have been tested. Other platforms are untested.

Yes

### DIFF
--- a/ports/rtaudio/LICENSE
+++ b/ports/rtaudio/LICENSE
@@ -1,0 +1,26 @@
+RtAudio: a set of realtime audio i/o C++ classes
+Copyright (c) 2001-2019 Gary P. Scavone
+
+Permission is hereby granted, free of charge, to any person
+obtaining a copy of this software and associated documentation files
+(the "Software"), to deal in the Software without restriction,
+including without limitation the rights to use, copy, modify, merge,
+publish, distribute, sublicense, and/or sell copies of the Software,
+and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+Any person wishing to distribute modifications to the Software is
+asked to send the modifications to the original developer so that
+they can be incorporated into the canonical version.  This is,
+however, not a binding provision of this license.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR
+ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF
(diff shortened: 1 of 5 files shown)


chausner:
UWP builds seem to fail due to missing APIs. Even though I think WASAPI should be supported by UWP in theory, at least the code is currently not compatible (COM functions like `CoInitialize` are not found). So I marked UWP as not supported in the manifest for now.

Builds on Linux fail in the CI because RtAudio by default builds against ALSA on Linux for which the dev libraries seem to be missing. I could build the package successfully on Ubuntu 20.04 after installing:
```
sudo apt-get install libasound2-dev
```
So I guess this is fine but we probably still want to disable Linux CI builds since the build is expected to fail there. Shall I add `supports: "!linux"` to the manifest?

ports/rtaudio/LICENSE
```diff
@@ -0,0 +1,26 @@
...
+they can be incorporated into the canonical version.  This is,
+however, not a binding provision of this license.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR
+ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF
+CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
PhoebeHui, line 26: Does this needed? I noticed rtaudio already has the license file https://github.com/thestk/rtaudio/blob/master/LICENSE

PhoebeHui (changes requested):
Thanks for the PR!

For the package that required on Linux, I will confirm when to deploy the package on Linux CI testing machines.

ports/rtaudio/LICENSE
chausner: RtAudio has a LICENSE file in the master branch but this port is for the latest stable release and it was not present there yet. See also my comment in the portfile [here](https://github.com/microsoft/vcpkg/pull/14774/files#diff-a6e2583f338b18ad389c28f2347042ae1cd69a2187dc45a01c096238e0d9915eR24).

strega-nil:
This means that we should add `rtaudio:x64-linux=fail` to ci.baseline.txt.

chausner:
> This means that we should add `rtaudio:x64-linux=fail` to ci.baseline.txt.

Done in https://github.com/microsoft/vcpkg/pull/14774/commits/f107104c045bd5594e8f312b56e69e26ae9044f1

strega-nil:
Thanks @chausner !

PR #14782, opened 2020-11-26
[xbyak] update to 5.991

Ryan-rsm-McKenzie:


ports/xbyak/portfile.cmake
```diff
@@ -1,14 +1,23 @@
 vcpkg_from_github(
```
NancyLi1013, line 1:
```suggestion
vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp")

vcpkg_from_github(
```
Since it doesn't support arm and uwp, it would be better to add this supported message here.

ports/xbyak/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "xbyak",
-  "version-string": "5.97",
+  "version-string": "5.991",
```
NancyLi1013, line 3:
```suggestion
  "version-string": "2020-11-26",
```
Since you used the latest commit id in master branch. it is more suitable to use a date as version.

NancyLi1013 (changes requested):

NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your PR @Ryan-rsm-McKenzie.

ports/xbyak/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "xbyak",
-  "version-string": "5.97",
+  "version-string": "2020-11-26",
```
strega-nil, line 3:
```suggestion
  "version-string": "5.991",
```

this should be this right?

strega-nil (changes requested):

ports/xbyak/vcpkg.json
Ryan-rsm-McKenzie: Nancy suggested I use the date as a version since I'm using the latest commit id

strega-nil: I think we should still use the version number, tbh

NancyLi1013: @strega-nil  I just referred to our doc https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#version to review this point.

strega-nil:
Thanks @Ryan-rsm-McKenzie !

PR #4477, opened 2018-10-15
Create macOS configuration variables for triplets

LRFLEW:
This PR makes a few related changes. Firstly, I set `CMAKE_MACOSX_RPATH` to `ON` in the OSX toolchain file. This changes the default behavior for the install name on all dynamic library targets on macOS. Specifically, it sets it so the dylib id's start with `@rpath/`. This change closes #4376.

In addition, I added a new variable that can be used in the triplets, `VCPKG_INSTALL_NAME_DIR`. This correlates to the CMake variable `CMAKE_INSTALL_NAME_DIR`, and sets the default install name for all targets. Using this, users can control the path in the resulting dylib id's if the default of using `@rpath` won't work for their specific use case.

Lastly, I went ahead and created the variables `VCPKG_OSX_DEPLOYMENT_TARGET` and `VCPKG_OSX_SYSROOT`, which correlate to the CMake variables `CMAKE_OSX_DEPLOYMENT_TARGET` and `CMAKE_OSX_SYSROOT` respectively. `VCPKG_OSX_DEPLOYMENT_TARGET` does two things. Firstly, it sets the minimum macOS target version, so libraries built on a newer version of macOS can run on older versions. (This is functionally equivalent to adding `-mmacosx-version-min=` to the `VCPKG_*_FLAGS` variables). The second thing it does is change how different installed versions of the macOS SDK are considered. (While SDK's are mostly backwards-compatible, there are reasons to want to use an old SDK version). In addition, `VCPKG_OSX_SYSROOT` allows you to explicitly set the path to the macOS SDK if you want to specify it directly.

PR #14785, opened 2020-11-26
[openssl] Handle INSTALL_NAME_DIR and OSX_DEPLOYMENT_TARGET for macOS shared builds

LRFLEW:
A few years ago, I made a PR here involving the dylib-id's of shared library builds for macOS and iOS (https://github.com/microsoft/vcpkg/pull/4477). This updates the OpenSSL port to properly handle this functionality.

dylib-id's are *weird*, but put simply, they are strings stored inside of dylib files that are used when linking projects to it to say where the executable should look for the library at runtime. The default behavior generally is to set the dylib-id to the installation absolute path, as that's how it works for system-installed libraries. The default behavior for CMake and vcpkg (since my old PR), however, is to have the dylib-id be `@rpath/<library>.dylib`, which is generally what's used for user distributions (namely .app bundles). The old PR also included a way of specifying the dylib-id directory via a custom triplet.

Since the OpenSSL port (for "Unix") just wraps the automake system, this setting is not properly handled, and the dylib-id ends up being set to the absolute installation path. This PR uses `install_name_tool` to update the dylib-id's based on the CMake settings. (The other way of setting the dylib-id is to use the `-install_name` argument in the linker command, but doing this would require patching the autoconf setup, as the OpenSSL project doesn't include a way of changing the argument value.)

-----

So I realized that another change I made in that old PR, the one for the minimum macOS target, also wasn't being handled by the OpenSSL port. The solution for this one was simpler, as it's the same as for `OSX_SYSROOT`. I decided to combine the changes into the same PR, as I see little reason to have them reviewed separately.


PhoebeHui (changes requested):
Thanks for the PR!

Could you update the Port-Version to 2 in CONTROL file?

LRFLEW:
> Could you update the Port-Version to 2 in CONTROL file?

I *knew* I forgot something. Fixed and Thanks

PhoebeHui:
@LRFLEW, here is gdal failures with x64-linux and x64-osx from CI testing:

```
jpeg2000dataset.cpp: In member function ‘int JPEG2000Dataset::DecodeImage()’:
jpeg2000dataset.cpp:516:32: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             if (poBand->iDepth != jas_image_cmptprec( psImage, iBand ) ||
                                ^
jpeg2000_vsil_io.cpp:154:1: error: invalid conversion from ‘int (*)(jas_stream_obj_t*, char*, unsigned int) {aka int (*)(void*, char*, unsigned int)}’ to ‘int (*)(jas_stream_obj_t*, const char*, unsigned int) {aka int (*)(void*, const char*, unsigned int)}’ [-fpermissive]
 };
 ^
make[2]: *** [../o/jpeg2000_vsil_io.lo] Error 1
make[1]: *** [jpeg2000-install-obj] Error 2
make[1]: *** Waiting for unfinished jobs....
make: *** [frmts-target] Error 2
```
[failure logs for x64-linux (7).zip](https://github.com/microsoft/vcpkg/files/5605746/failure.logs.for.x64-linux.7.zip)

LRFLEW:
@PhoebeHui I saw that error, but I can't figure out what to make of it. It's an error in an unrelated port in a file that should have no relation to any of the changes made here. Since this error appeared in the CI for an unrelated PR as well (#14799), I suspect this might be a problem from the master branch. I can try rebasing the PR against any changes made to master in case there's a fix there.

PhoebeHui:
@LRFLEW, gdal depends on libpq, and libpq depends on openssl, so it might relate to this changes, I verified gdal:x64-linux on master, it installed successfully.

```
Starting package 9/9: gdal:x64-linux
Building package gdal[core]:x64-linux...
-- Downloading http://download.osgeo.org/gdal/3.1.3/gdal313.zip...
-- Extracting source /home/phoebe/vcpkg/downloads/gdal313.zip
-- Applying patch 0001-Fix-debug-crt-flags.patch
-- Applying patch 0002-Fix-build.patch
-- Applying patch 0003-Fix-static-build.patch
-- Applying patch 0004-Fix-std-fabs.patch
-- Applying patch 0005-Fix-cfitsio.patch
-- Using source at /home/phoebe/vcpkg/buildtrees/gdal/src/gdal313-b1dafdc415.cle                                                                                                 an
-- Configuring x64-linux-dbg
-- Configuring x64-linux-rel
-- Generating configure for x64-linux
-- Finished generating configure for x64-linux
-- Configuring x64-linux-dbg
-- Configuring x64-linux-rel
-- Building x64-linux-dbg
-- Installing x64-linux-dbg
-- Building x64-linux-rel
-- Installing x64-linux-rel
-- Installing: /home/phoebe/vcpkg/packages/gdal_x64-linux/share/gdal/usage
-- Installing: /home/phoebe/vcpkg/packages/gdal_x64-linux/share/gdal/copyright
Building package gdal[core]:x64-linux... done
Installing package gdal[core]:x64-linux...
```


```
./vcpkg depend-info gdal
libjpeg-turbo:
liblzma:
sqlite3[tool]:
szip:
zlib:
curl[non-http, ssl, schannel, sspi, winssl]: zlib
hdf5[szip, zlib]: szip, zlib
libiconv:
openssl:
proj4[database]: sqlite3
tiff: libjpeg-turbo, liblzma, zlib
cfitsio: zlib
expat:
geos:
libgeotiff: libjpeg-turbo, proj4, tiff, zlib
libpng: zlib
libpq[openssl, zlib]: openssl, zlib
libwebp[nearlossless, unicode, simd]:
libxml2: libiconv, liblzma, zlib
netcdf-c: curl, hdf5
openjpeg:
gdal: cfitsio, curl, expat, geos, hdf5, libgeotiff, liblzma, libpng, libpq, libwebp, libxml2, netcdf-c, openjpeg, proj4, sqlite3, zlib
```

LRFLEW:
@PhoebeHui I setup a Docker instance and tried building gdal on master, and I was able to confirm that it did succeed for me (once I worked around Docker killing the linker tasks). I then tried building my branch and could not reproduce the issue that the CI is seeing. I don't know what's different about my setup to the CI's exactly, but it's hard for me to debug it when I can't see the issue for myself.

Looking into the error message itself, it still seems to me like it's unrelated to this PR. The type error is happening when constructing an object of type `jas_stream_ops_t`, which appears to be defined in a header from another library called Jasper (`#include <jasper/jasper.h>` in jpeg2000_vsil_io.h). However, I'm not seeing Jasper in the output of the `depend-info` call you gave me. Where is it getting the header for it from? I suspect the problem is coming from having an inconsistent source for this header.

LRFLEW:
Well, rebasing against master did seem to "fix" the error. This could just have been a matter of re-running the test. Perhaps the Jasper thing I mentioned should be looked into, but that would be a separate issue from this one.

PhoebeHui:
@LRFLEW, thanks for looking into this issue!

It might due to Jasper, I will confirm this issue locally, anyway, it should be another issue, should not block this PR.

PhoebeHui:
Confirmed this issue on master, and found when jasper installed before gdal, gdal will fail on linux and osx. I will look into the failures. It is not relate to this changes. please ignore it.

```
jpeg2000dataset.cpp: In member function ‘int JPEG2000Dataset::DecodeImage()’:
jpeg2000dataset.cpp:516:32: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             if (poBand->iDepth != jas_image_cmptprec( psImage, iBand ) ||
                                ^
jpeg2000_vsil_io.cpp:154:1: error: invalid conversion from ‘int (*)(jas_stream_obj_t*, char*, unsigned int) {aka int (*)(void*, char*, unsigned int)}’ to ‘int (*)(jas_stream_obj_t*, const char*, unsigned int) {aka int (*)(void*, const char*, unsigned int)}’ [-fpermissive]
 };
 ^
make[2]: *** [../o/jpeg2000_vsil_io.lo] Error 1
make[1]: *** [jpeg2000-install-obj] Error 2
make[1]: *** Waiting for unfinished jobs....
make: *** [frmts-target] Error 2
```

PhoebeHui (approved):

PhoebeHui:
@strega-nil, could you help further review?

strega-nil:
LGTM! Thanks @LRFLEW

Issue #14663, opened 2020-11-20
nvtt[core]:x86-windows build failure


stevegbarry:
**Host Environment**
- OS: Windows
- Compiler: VS2019

**To Reproduce**
`./vcpkg install osgearth:x86-windows`

**Failure logs** 
Computing installation plan...
The following packages will be built and installed:
  * nvtt[core]:x86-windows
  * osg[core,plugins]:x86-windows
    osgearth[core]:x86-windows
Additional packages (*) will be modified to complete this operation.
Detecting compiler hash for triplet x86-windows...
Starting package 1/3: nvtt:x86-windows
Building package nvtt[core]:x86-windows...
Could not locate cached archive: xxx\770b711b12a6eee038e2814204ad5d22e8588627.zip
CMake Error at scripts/cmake/vcpkg_fail_port_install.cmake:94 (message):
  Architecture 'x86' not supported by nvtt!

Call Stack (most recent call first):
  ports/nvtt/portfile.cmake:1 (vcpkg_fail_port_install)
  scripts/ports.cmake:136 (include)

**Additional context**
NVTT is an optional dependent for osg and, according to the osgearth folks, not required for osgearth.  However, osgearth will not build for x86 without it.  NVTT and osgearth build successfully for x64.

NancyLi1013:
Hi @stevegbarry 

Thanks for posting this issue.

Of course, `nvtt `is an optional dependent for `osg`. But we still cannot remove it from the dependency lists for feature `plugin`.

We cannot make sure other users don't need this feature.

So we can update the dependency as `nvtt (!x86) `to solve this issue.

NancyLi1013:
@stevegbarry 

I have updated the dependency of `osg`. Please update vcpkg to the latest version to rebuild.

PR #14790, opened 2020-11-26
[osg] Update dependency

### LINKED ISSUE #14756: [osg] remove of boost
****

Try to avoid include "boost" for all builds. This is NOT required for default build. Or please add this as a feature and not in base build. 

Needed for: osgPlugins/RestHttpDevice, osgPlugins/COLLADA and ASIO ?

**Proposed solution**
Remove "boost" and/or add it as a feature.

Remove "boost" at all

**Additional context**
OSG takes currently a lot of time to build on Github Actions and it need a lot of packages from "boost" in Linux build (a lot of storage as well).

NancyLi1013:
**Describe the pull request**

- What does your PR fix? Fixes #14756 #14663

Remove `boost `from base dependency lists.

Add `boost `as the dependency lists of feature `plugins`

Update the dependencies of feature `plugins`(Change `nvtt `as `nvtt (!x86)`).

Note:

Feature plugins has passed with the following triplets:

- x86-windows

### DIFF
--- a/ports/osg/CONTROL
+++ b/ports/osg/CONTROL
@@ -1,9 +1,9 @@
 Source: osg
 Version: 3.6.5
-Port-Version: 3
+Port-Version: 4
 Homepage: https://github.com/openscenegraph/OpenSceneGraph
 Description:  The OpenSceneGraph is an open source high performance 3D graphics toolkit.
-Build-Depends: zlib, fontconfig, boost-asio (!windows), boost-core (!windows), boost-logic (!windows), boost-lexical-cast (!windows), boost-smart-ptr (!windows), boost-tuple (!windows), boost-bind (!windows), freeglut (windows), expat (windows), openimageio (osx)
+Build-Depends: zlib, fontconfig, freeglut (windows), expat (windows), openimageio (osx)
 
 Feature: collada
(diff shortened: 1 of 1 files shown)


JackBoosY (approved):

remoe:
This is a lot better ;) But it still use boost for *all* plugins, right? This is not needed. Only "RestHttpDevice" "COLLADA" and ASIO need it?

NancyLi1013:
@remoe 

Thanks for your attention. You're right that `boost `is not required for all plugins. But since `plugin` has been added as a feature for `osg` and `RestHttpDevice` is one part of `osgPlugins`. We cannot spilt this part separately. If removing `boost `from the dependency lists of `osg[plguins]`, `RestHttpDevice` cannot be built anymore. 

In a word, I think it would be better to keep `boost `as its dependency.

remoe:
@NancyLi1013 ok, never used this boost based plugins :) 
Is it possible to split it to more features? Most users doesn't need this plugins?

strega-nil:
It's likely possible to split plugins up into multiple features; we should consider doing it, but that shouldn't stop this from being merged.

Thanks @NancyLi1013 !

PR #14793, opened 2020-11-26
[devil] Fix pkgconfig on Windows

### LINKED ISSUE #14775: [devil] pkg-config files specify non-existent IL.lib
****

DevIL distributes pkg-config (.pc) files, however they specify `-lIL` which doesn't exist. `DevIL.lib` exists instead. Because of that using DevIL with pkg-config is limited. I'm able to use its `CFLAGS`, but to link against it I use a direct path to `DevIL.lib` like that:

```
pkg_check_modules(IL REQUIRED IL)

target_compile_options(main PRIVATE ${IL_CFLAGS})
target_include_directories(main PRIVATE ${IL_INCLUDE_DIRS})
target_link_libraries(main ${IL_LIBDIR}/DevIL.lib)
```

**Proposed solution**
Distribute CMake configs. If not possible, fix the .pc files to specify correct lib names.

Link directly to DevIL.lib.

**Additional context**
-

JackBoosY:
In _SOURCE/DevIL/src-IL/CMakeLists.txt_:
```cmake
if(WIN32)
	# Linux uses libIL.so, libILU.so, and libILUT.so, but Windows uses
	#  DevIL.dll, ILU.dll, and ILUT.dll.
	set_target_properties(IL PROPERTIES OUTPUT_NAME "DevIL")
endif(WIN32)
```
This will caused the wrong name in its pkgconfig files.

Fixes #14775.

### DIFF
--- a/ports/devil/0005-fix-pkgconfig.patch
+++ b/ports/devil/0005-fix-pkgconfig.patch
@@ -1,0 +1,52 @@
+diff --git a/DevIL/src-IL/CMakeLists.txt b/DevIL/src-IL/CMakeLists.txt
+index 4accaa1..c3a1874 100644
+--- a/DevIL/src-IL/CMakeLists.txt
++++ b/DevIL/src-IL/CMakeLists.txt
+@@ -160,6 +160,27 @@ if(WIN32)
+ 	# Linux uses libIL.so, libILU.so, and libILUT.so, but Windows uses
+ 	#  DevIL.dll, ILU.dll, and ILUT.dll.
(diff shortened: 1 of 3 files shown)


NancyLi1013 (approved):

Neumann-A:
the issue also indicates that a `vcpkg_fixup_pkgconfig()` call is missing from the portfile.

JackBoosY:
@Neumann-A  Yes, I've already added it.

strega-nil:
Thanks @JackBoosY

PR #14797, opened 2020-11-26
[vcpkg] Fix issue in vcpkg.targets that introduced in 13755

### LINKED ISSUE #14729: How do I add vcpkg to Visual Studio 2019
Hey,
I changed "use vcpkg" to no and the whole vcpkg option is gone from my project! It's still showing in other projects and also in newly created projects, but once its disabled, it's gone.
Deleting and reinstalling vcpkg (+ all the libraries that were installed) did not help. The only way is to make a new project and copy over all the source files to the new project (.cpp .h...) 
Is that a bug? Isn't there a way to re-enable it after it was disabled? And also even though it's "disabled", the options in side of it are all active that is the project still links normally to where libraries are installed and removing the vcpkg folder would cause errors.
image

image

Thanks for taking the time

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/14729

Fix issue: 
When 'Use vcpkg' set to No, the vcpkg tab disappear, user can't enable it again. That due the PropertyGroup condition, which is wrong for the VcpkgPageSchema.

Test with latest vcpkg and vs16.8.2

### DIFF
--- a/scripts/buildsystems/msbuild/vcpkg.targets
+++ b/scripts/buildsystems/msbuild/vcpkg.targets
@@ -4,14 +4,16 @@
   <Import Condition="'$(VcpkgPropsImported)' != 'true'" Project="vcpkg.props" />
 
   <!-- Define properties derived from those defined in vcpkg.props, in the project file or specified on the command line. -->
+   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


ZhilkinSerg (approved):
Works for me.

PR #14816, opened 2020-11-27
[lapack-reference] add to CMAKE_FIND_LIBRARY_SUFFIXES additional suffixes

xandox:
add to `CMAKE_FIND_LIBRARY_SUFFIXES` additional suffixes with versions for make find_library to able to find libgfortran

Under ubuntu (do not check for other distributives) libgfortran always has some version suffix and cmake find_library not able to find them without additional suffixes in CMAKE_FIND_LIBRARY_SUFFIXES as result FindLAPACK.cmake always fails and as result installing port lapack under ubuntu always fails. 

find_library(gfortran)

yes


xandox:
@Neumann-A How you think is it ok?

Neumann-A:
@xandox Normally there should be a symlink called `libgfortran.so`.

JackBoosY:
@xandox Did you test the usage?

JackBoosY (approved):

Neumann-A:
It would be better to restore the original CMAKE_FIND_LIBRARY_SUFFIXES after the find_library call for gfortran. Alternatively, adding additional search paths to the find_library call could also work?

strega-nil:
LGTM! Thanks @xandox!

PR #14848, opened 2020-11-30
[vcpkg_apply_patches] Work around /etc/gitconfig: Permission denied

ras0219:
The current CI system on linux and osx is experiencing issues due to some problems with `/etc/gitconfig`. This PR works around the issue entirely by instructing git to not read the system config file at all when applying patches.

The environment variable `GIT_CONFIG_NOSYSTEM` is documented [here](https://git-scm.com/docs/git).

An example PR that hit this issue is https://github.com/microsoft/vcpkg/pull/14825.

@strega-nil @JackBoosY


ras0219:
@strega-nil Turns out I was wrong about this being the problem with #14825. I think it's likely a good change regardless, but I've suggested the actual fix on that PR as a review.

PR #14859, opened 2020-11-30
[atk] Switch to GitHub mirror

ras0219:
#13126 hit an issue where ftp.gnome.org timed out. This PR switches `atk` to use the GitHub mirror instead.


JackBoosY:
Hmm...should I take over this PR? Or it's invalid now?

PR #14861, opened 2020-11-30
[gdal baseline] add gdal:x64-osx=fail

strega-nil:
See title.


JackBoosY:
Any reason to do that?

PhoebeHui:
Does it fail with https://github.com/microsoft/vcpkg/pull/14866#issue-530080498, I summited the PR to fix it in baseline.

strega-nil:
Fixed by #14866 in a correct way

Issue #14834, opened 2020-11-29
./vcpkg install curl[tool] fails


martti-cloudkey:
-- Installing: /home/martti/temp/vcpkg/packages/curl_x64-linux/share/curl/curl-config
-- Fixing pkgconfig file: /home/martti/temp/vcpkg/packages/curl_x64-linux/lib/pkgconfig/libcurl.pc
-- CHECK_LIB_lber_RELEASE:CHECK_LIB_lber_RELEASE-NOTFOUND
CMake Error at scripts/cmake/vcpkg_fixup_pkgconfig.cmake:233 (message):
  Library "lber" was not found! If it is a system library use the
  SYSTEM_LIBRARIES parameter for the vcpkg_fixup_pkgconfig call! Otherwise,
  correct the *.pc file
Call Stack (most recent call first):
  scripts/cmake/vcpkg_fixup_pkgconfig.cmake:299 (vcpkg_fixup_pkgconfig_check_files)
  ports/curl/portfile.cmake:185 (vcpkg_fixup_pkgconfig)
  scripts/ports.cmake:136 (include)

Error: Building package curl:x64-linux failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `./vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: curl:x64-linux
  Vcpkg version: 2020.11.12-unknownhash

Additionally, attach any relevant sections from the log files above.

JackBoosY:
Will be fixed in #14807.

klalumiere:
For some reasons I'm not able to reproduce the issue anymore, either on my machine or in a Docker container (I used `rikorose/gcc-cmake`). Maybe the issue will also be solved for you too if you pull `master`. :thinking:

PR #14350, opened 2020-11-02
[curl] Fixes pkgconfig configuration file

Matioupi:
- What does your PR fix?
This PR fixes the libcurl.pc pkgconfig configuration file for windows debug build. It also enables tool on non windows platforms (except uwp)

- Which triplets are supported/not supported? Have you updated the CI baseline?

test triplets : x64-windows / x64-windows-static / x86-windows / x86-windows-static / x64-linux / arm64-linux
(tested with/without tool feature enabled) and x64-uwp (core feature only)

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Yes, to the best of my knowledge...

PR #14865, opened 2020-12-01
[curl] Mitigate issue #14834: port broken on Linux

klalumiere:
**Describe the pull request**

- What does your PR fix?

Mitigate issue #14834: port broken on Linux
    
The more complete fix will come from the pull request #14807. However, since curl on Linux is broken right now, I think it might be worth it to merge this branch quickly so that curl will work again on Linux.

- Which triplets are supported/not supported? Have you updated the CI baseline? Nothing changed here


ports/curl/portfile.cmake
```diff
@@ -182,7 +182,7 @@ endif()
 if(VCPKG_TARGET_IS_WINDOWS)	
     vcpkg_fixup_pkgconfig()
 elseif(VCPKG_TARGET_IS_LINUX)
-    vcpkg_fixup_pkgconfig(SYSTEM_LIBRARIES pthread dl c)
+    # vcpkg_fixup_pkgconfig(SYSTEM_LIBRARIES pthread dl c)
```
klalumiere, line 185: This was introduced ~10 days ago in #14350. The fix to make this work is developed in pull request #14807.

JackBoosY:
No, this will break the port using curl pkgconig.

klalumiere:
> No, this will break the port using curl pkgconig.

Well, yes and no 🙂 . As explained above in my comment, this line was added ~10 days ago in #14350. It only affect the behaviour on Linux, and right now, because of this line, the port doesn't work at all on Linux 
(neither does anything that depends, directly or transitively, on curl). Hence, I'd argue that this is a good first step , and then your pull request #14807 will provide a more complete fix that allows to use pkgconfig on Linux. 🙂

klalumiere:
> No, this will break the port using curl pkgconig.

Well, yes and no 🙂 . As explained above in my comment, this line was added ~10 days ago in #14350. It only affect the behaviour on Linux, and right now, because of this line, the port doesn't work at all on Linux 
(neither does anything that depends, directly or transitively, on curl). Hence, I'd argue that this is a good first step , and then your pull request #14807 will provide a more complete fix that allows to use pkgconfig on Linux. 🙂

klalumiere:
I was doing some tests for pull request #14807, and in the midst of these tests, I've realized that for I reason that I don't understand, I'm not able to reproduce the problem anymore. I guess that's a good news, although I'm curious and I'd like to understand what happen :laughing: . I'm closing and I'm going to leave a comment on issue #14834 to see if the issue is solved for other people to. I'm going to continue to follow #14807 (I'll add a comment there in ~1h if everything goes well).

klalumiere:
I was doing some tests for pull request #14807, and in the midst of these tests, I've realized that for I reason that I don't understand, I'm not able to reproduce the problem anymore. I guess that's a good news, although I'm curious and I'd like to understand what happen :laughing: . I'm closing and I'm going to leave a comment on issue #14834 to see if the issue is solved for other people to. I'm going to continue to follow #14807 (I'll add a comment there in ~1h if everything goes well).

PR #14888, opened 2020-12-02
[libigl] Always install as header-only library

FabienPean:
In #14376, triplet *-windows-dynamic version was updated to rely on the library as header-only while leaving the *-static triplet intact. However, the static library archive still suffers from the same reported issues: huge size and not all combination of explicitly instantiated templates exist. This PR forces installs of libigl as header-only library for both static and dynamic triplets. (Particularly useful for vcpkg on linux where the default triplet has static linkage by default)

Let me know if this PR is an agreeable change on your side.


FabienPean:
The PR shows failure for another port.

ports/libigl/install-extra-headers.patch
```diff
@@ -2,19 +2,17 @@ diff --git a/cmake/libigl.cmake b/cmake/libigl.cmake
...
 +++ b/cmake/libigl.cmake
-@@ -560,6 +560,85 @@ export(
+@@ -560,6 +560,75 @@ export(
  # Install headers for core library
  install_dir_files(core)
  install_dir_files(copyleft)
 + 
 +if (LIBIGL_WITH_EMBREE)
 +    install_dir_files(embree)
-+#    if(NOT LIBIGL_USE_STATIC_LIBRARY)
```
JackBoosY, line 12: This issue seems like a upsteam issue, can you report this to the upstream?
And does the upstream accept this changes?

JackBoosY (changes requested):

JackBoosY:
Maybe we should add an new feature `header-only` to this port instead of force it to be a header only port.

FabienPean:
Added header-only as a feature

ports/libigl/portfile.cmake
```diff
@@ -14,7 +14,7 @@ vcpkg_from_github(
 )
 
 set(LIBIGL_BUILD_STATIC OFF)
-if (VCPKG_LIBRARY_LINKAGE STREQUAL static)
+if (NOT "header-only" IN_LIST FEATURES AND VCPKG_LIBRARY_LINKAGE STREQUAL static)
```
JackBoosY, line 17: Should this judgment be `("header-only" IN_LIST FEATURES) OR (VCPKG_LIBRARY_LINKAGE STREQUAL static)` ?

JackBoosY (changes requested):

ports/libigl/portfile.cmake
FabienPean: The full block is
> set(LIBIGL_BUILD_STATIC OFF)
if (NOT "header-only" IN_LIST FEATURES AND VCPKG_LIBRARY_LINKAGE STREQUAL static)
    set(LIBIGL_BUILD_STATIC ON)
endif()

So it sounds correct to me like that. LIBIGL is built in static mode only when there is static linkage and header-only is *not* demanded by the user.

JackBoosY: Fine.

JackBoosY (approved):

BillyONeal (changes requested):
The "header-only" feature doesn't comply with our guidelines for features, namely, that they do not provide "alternative" selections. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-control-alternatives-in-published-interfaces

Crucially, any downstream consumer who is passing -ligl (or similar to link the static lib) would be broken by turning on this "feature".

It might be acceptable to make such a chance unconditional if the current users of the port think that makes sense.

/cc @patrikhuber @deGravity @xarthurx

ras0219-msft (approved):
Since the comments in the upstream project indicate that it is a "header only library", I think the right way forward is to always just do that.

With this change, LGTM!

PR #14927, opened 2020-12-03
[vcpkg] Error on '#' in version strings to avoid confusion

ras0219:
With versioning, we will now be displaying versions to users either as `text` or `text#port-version`. If a user sees this output and attempts to input it directly as a version constraint, it will not match and cause confusion.

This PR makes `#`s illegal in all version strings, reserving it for use for port-version suffixes. It then suggests how to fix the issue for users.

As a drive by, this PR also refactors out the several copies of deserializing `version-xyz`'s.

Example:
```json
{
  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg/master/scripts/vcpkg.schema.json",
  "name": "versiontest",
  "version-string": "2020-03-03",
  "description": [],
  "homepage": "",
  "dependencies": [
    { "name": "fmt", "version>=": "1.0.0#2" }
  ]
}
```
```
> vcpkg install --feature-flags=manifests,versions --dry-run
Errors occurred while parsing C:\src\vcpkg\testing\vcpkg.json
    $.dependencies[0].version>= (a version): invalid character '#' in version text. Did you mean "port-version": 2?
See https://github.com/Microsoft/vcpkg/tree/master/docs/users/manifests.md for more information.
```

+@strega-nil


JackBoosY:
That's awesome!

strega-nil (approved):

PR #14950, opened 2020-12-05
[seal] Allow disabling SEAL_THROW_ON_TRANSPARENT_CIPHERTEXT

kiromaru:
Allow disabling the option SEAL_THROW_ON_TRANSPARENT_CIPHERTEXT. Some
third party projects depend on this option being disabled.

- What does your PR fix? Allows disabling a configuration option for SEAL

Tested x64-windows-static, x64-linux, x64-osx.
SEAL only supports static linkage in Windows, updated the CI baseline to reflect this.

Yes


scripts/ci.baseline.txt
```diff
@@ -1380,7 +1380,10 @@ sdl2-net:arm-uwp=fail
 sdl2-net:x64-uwp=fail
 # https://github.com/microsoft/vcpkg/issues/10918
 seal:arm-uwp=fail
+seal:arm64-windows=fail
 seal:x64-uwp=fail
+seal:x64-windows=fail
+seal:x86-windows=fail
```
JackBoosY, line 1386: Why add these 3 lines?

JackBoosY (changes requested):

scripts/ci.baseline.txt
kiromaru: The SEAL library works only with static linkage in Windows. The CMake file makes sure of this, and will generate an error if it detects that dependencies are not available as static. This check was not working correctly before, the library would compile but would fail when trying to use it in a project. We pushed a hot-fix to GitHub to fix this, and the pull request reflects this new commit. As a result builds fail correctly now when non-static linkage is detected, so the failures in those three triplets are expected.

JackBoosY: If this failure is by design, please add `Support: !static` to `CONTROL` file instead of adding this.

JackBoosY: And please bump the port version. See [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#update-the-version-field-in-the-control-file-of-any-modified-ports).

kiromaru: I've bumped the port version and added the "Supports" line to `CONTROL`. I reverted the changes to ci.baseline.txt as well.

kiromaru:
Thank you for the review, we will actually update the version of SEAL to 3.6.1 which has just been released and include the change there.

PR #14962, opened 2020-12-05
[geogram] Fix windows dynamic build usage

FabienPean:
Fix the use of geogram library for triplet x64-windows. The bug has been reported upstream, but no news from maintainers so far. The PR fixes the package for current vcpkg version, and may need to be reverted when a new version of the library is released.


NancyLi1013:
Hi @FabienPean 

Thanks for your PR.

Could you please describe what the problem is in dynamic usage?

FabienPean:
Hi, sure !
The preprocessor definition "GEO_DYNAMIC_LIBS" is never defined during the CMake config step https://github.com/alicevision/geogram/issues/17,
which is necessary to have the proper declimport/declexport settings on windows in this file [geogram/src/lib/geogram/api/defs.h#L69](https://github.com/alicevision/geogram/blob/8b2ae6148c7ab1564fa2700673b4275296ce80d3/src/lib/geogram/api/defs.h#L69)
That's at least my understanding of the situation.

ports/geogram/fix-cmake-config-and-install.patch
```diff
@@ -73,8 +73,11 @@ diff --git a/src/lib/geogram/CMakeLists.txt b/src/lib/geogram/CMakeLists.txt
 index 49cb2ba..82c2f39 100755
 --- a/src/lib/geogram/CMakeLists.txt
 +++ b/src/lib/geogram/CMakeLists.txt
-@@ -62,7 +62,14 @@ if(WIN32)
+@@ -62,7 +62,17 @@ if(WIN32)
  endif()
++if(WIN32 AND VORPALINE_BUILD_DYNAMIC)
++    target_compile_definitions(geogram PUBLIC GEO_DYNAMIC_LIBS)
++endif()
```
NancyLi1013, line 80: Could you please make a new patch for this changes?

ras0219-msft (approved):

NancyLi1013 (approved):

NancyLi1013:
The regressions caused by `gsoap `will be fixed by PR #16270.

PR #14971, opened 2020-12-06
[ms-gltf] Add new port

luncliff:
### Changes

New port for https://github.com/microsoft/glTF-SDK in Windows / MacOS platforms.

* https://github.com/microsoft/glTF-SDK/pull/66 updated the library's dependent RapidJSON's version. So the code can be built with VcPkg now.  
* The `version-string` will be updated **when the new version is released**. (== Next of [`r1.9.5.0`](https://github.com/microsoft/glTF-SDK/releases))

#### What does your PR fix?

There was no port request for this.

#### Which triplets are supported/not supported? Have you updated the CI baseline?

Linux is not supported.  
`ci.baseline.txt` will be updated soon. 

#### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Yes!


PhoebeHui:
@luncliff, could you format the vcpkg.json file? See github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#manifest for solution.

luncliff:
@PhoebeHui, I was waiting for the next release of the library since last December but seems like it can take more than months... :(  
**The port's version is `r1.9.5.0` but the ref commit is not**. Can this be allowed? I'd like to hear opinions about it.  

In https://github.com/microsoft/glTF-SDK/pull/66, I update some build configurations and source code. Fortunately the project's TC passed, so I'd like to use the changes if possible.

scripts/ci.baseline.txt
```diff
@@ -997,6 +997,7 @@ mpir:x64-windows-static-md=skip
 mpir:arm64-windows=skip
 mpir:x64-osx=skip
 mpir:x64-linux=skip
+ms-gltf:x64-linux=fail
```
PhoebeHui, line 1000: Could you remove this in ci.baseline.txt, and add "supports":  "!linux" to vcpkg.json file?

PhoebeHui (changes requested):
@luncliff, I think it's fine, we already have the comments about the ref commit,  we can update that with new release in future.

ports/ms-gltf/vcpkg.json
```diff
@@ -0,0 +1,23 @@
...
+  "dependencies": [
+    "rapidjson"
+  ],
+  "default-features": [
+    "test"
+  ],
+  "features": {
+    "samples": {
+      "description": "Build with GLTF samples"
+    },
```
PhoebeHui, line 15: Have you test this feature locally?

luncliff: I tested for `x64-windows` and `x86-windows`, but not for `x64-osx`. I will check the part again.

scripts/ci.baseline.txt
luncliff: got it!

PhoebeHui:

PhoebeHui (approved):

luncliff:
@PhoebeHui Appreciate the changes and testing. (Sadly, things didn't go well last week)

I added a small patch because of `<filesystem>` support in Mac OS (`x64-osx` triplets).
It basically adds 2 more changes.

* source: `_LIBCPP_NO_EXPERIMENTAL_DEPRECATION_WARNING_FILESYSTEM` when `__APPLE__`
  Like warning for `std::experimental::filesystem` in MS STL, headers in MacOS SDK reports the deprecation and it is considered error. The library is used only in samples so I placed a macro definition.
* cmake: `CMAKE_OSX_DEPLOYMENT_TARGET` to `10.15`.   
  This version allows use of `std::filesystem::path`.
  Normally it is fine with CMake configuration, but I found that ninja places the minimum osx version to `10.11`.

PhoebeHui (approved):

dan-shaw (approved):

PR #14977, opened 2020-12-07
[vcpkg] x_vcpkg_install_local_dependencies fixes

sandercox:
This patch addresses two issues.  
First is a bug in the argument parsing introduced when switching to the PARSE_ARGV style parsing. As this was simply missing an argument where to start the parsing from. As this is in an experimental feature it probably wasn't found earlier.

Second issue is that install command also works on `INTERFACE` targets, but that there is no way to get a `$<TARGET_FILE_NAME:target>` on that as these targets simply don't have it. So skipping the install code when target is of type `INTEFACE_LIBRARY`


BillyONeal (approved):

ras0219-msft:
Thanks for the fixes!

PR #14333, opened 2020-10-31
[Qt6|6.1.0] Add new ports. 

Neumann-A:
closes #15029 

Current Status:
- added all 6.1.0 ports which I know of (does it work? Who knows?)

Current questions:
- Port naming? qt6-qtbase? Only qtbase? (since qt wants to get rid of the version number anyway)
- Features? Too many? Will never be all tested. (at least I will only be going for the default set.)

Depends:
- [x] #14308
- [x] #14395 (ZSTD)
- [x] #14530 (harfbuzz)
- [x] another PR copying pdbs for `vcpkg_copy_tools`
- [x] https://bugreports.qt.io/browse/QTBUG-88457 (qtdeclarative x64-windows-static; not in beta5; probably in beta6) 
- [x] open upstream issues
- [x] https://bugreports.qt.io/browse/QTBUG-89204 (solved by own patch)
- [x] https://bugreports.qt.io/browse/QTBUG-89203 (solved by own patch)
- [x] https://bugreports.qt.io/browse/QTBUG-89280 (solved by own patch)
- [x] https://bugreports.qt.io/browse/QTBUG-89374 (solved by own patch)
- [x] fix import prefix correction in vcpkg_fixup_cmake_targets as mentioned in 
 https://github.com/microsoft/vcpkg/pull/12215#discussion_r454579077. (Corrected it in another way for this PR)
- [x] #15049
- [x] #15144

PR #15049, opened 2020-12-10
[vcpkg_from_git] new options TAG and X_OUT_REF

Neumann-A:
- TAG github tag to checkout
- OUT_REF github commit id related to tag or ref
(useable for automatic updates of ports if used with a version tag)

I use this in my two Qt PRs to directly checkout from `code.qt.io` and enable automatic updates of the refs on tag change


ras0219-msft:
We explicitly ask for REFs and not tags in the shared helpers in order to defend against tag republishing. If this functionality is just for qt6 updates, it seems more appropriate to keep private to the qt ports.

Neumann-A:
@ras0219-msft: I hope qt is stable enough to not do that. This PR does not change the behavior it still requires REF. It just checks out by TAG if available and compare the result against the given REF. So tag republishing will still be detected. If a port is doing tag republishing regularly the port simply needs to be updated to remove the TAG option without any other changes.

ras0219:
Thanks for the explanation; I didn't read the code closely enough. I can potentially see use in the TAG field as an additional check on top of the REF field.

Could you link to exactly how you've used OUT_REF?

Neumann-A:
@ras0219 see https://github.com/microsoft/vcpkg/pull/14333/files#r542232703

scripts/cmake/vcpkg_from_git.cmake
```diff
@@ -25,11 +25,18 @@ The url of the git repository.
...
+The git tag to download.
+
 ### PATCHES
 A list of patches to be applied to the extracted sources.
 
 Relative paths are based on the port directory.
 
+### OUT_REF
+Maintainers only. Used for automatic REF updates for ports using a versioned TAG.
+Will return early!
```
ras0219-msft, line 38:
```suggestion
### X_OUT_REF (internal only)
This parameter is used for automatic REF updates for certain ports in the central vcpkg catalog. It should not be used by any ports outside the central catalog and within the central catalog it should not be used on any user path. This parameter may change behavior incompatibly or be removed at any time.
```

scripts/cmake/vcpkg_from_git.cmake
```diff
@@ -25,11 +25,18 @@ The url of the git repository.
 ### REF
 The git sha of the commit to download.
 
+### TAG
+The git tag to download.
+
```
ras0219-msft, line 30:
```suggestion
### TAG
An optional git tag to be verified against the `REF`. If the remote repository's tag does not match the specified `REF`, the build will fail.

```

Additionally, `TAG` needs to be added as an optional parameter to the `Usage:` schema block:
```
    [TAG <v1.0.2>]
```

scripts/cmake/vcpkg_from_git.cmake
```diff
@@ -41,7 +48,7 @@ Relative paths are based on the port directory.
 include(vcpkg_execute_in_download_mode)
 
 function(vcpkg_from_git)
-  set(oneValueArgs OUT_SOURCE_PATH URL REF)
+  set(oneValueArgs OUT_SOURCE_PATH URL REF TAG OUT_REF)
```
ras0219-msft, line 51:
```suggestion
  set(oneValueArgs OUT_SOURCE_PATH URL REF TAG X_OUT_REF)
```

ras0219-msft (approved):
Thanks for the explanation and pointer. LGTM given the changes I've suggested.

JackBoosY:
Should be good now.

scripts/cmake/vcpkg_from_git.cmake
```diff
@@ -41,7 +48,7 @@ Relative paths are based on the port directory.
 include(vcpkg_execute_in_download_mode)
 
 function(vcpkg_from_git)
-  set(oneValueArgs OUT_SOURCE_PATH URL REF)
+  set(oneValueArgs OUT_SOURCE_PATH URL REF TAG X_OUT_REF)
```
vicroms, line 51: @BillyONeal please remind me. 
As this is internal and documented should it be prefixed with `X_` or `Z_`?

PR #15068, opened 2020-12-11
[oneDNN] New port

jacobkahn:
Adds a new port for the [oneDNN](https://github.com/oneapi-src/oneDNN) deep learning library.

cc @vpirogov, @dzarukin, @echeresh, @mgouicem, @akharito, @densamoilov

cc @syhw, @padentomasello, @tlikhomanenko, @vineelpratap, @xuqiantong, @avidov

`x64 & !uwp`. Tested myself on `x64-linux`.

Yes


mgouicem:
great to see that oneDNN is making its way here! And thanks for the port @jacobkahn :-)
One quick question, is there any reason to link to Intel MKL? oneDNN without MKL should perform on par or better than without (and the build with MKL is mostly here for testing purpose).

 Do you have any particular workloads where you see oneDNN with MKL faster than oneDNN without MKL?

jacobkahn:
@mgouicem — happy to help 👍  - we're about to use the port in [flashlight](https://github.com/facebookresearch/flashlight) for our new CPU/OpenCL backends

I have zero current benchmarks or supporting evidence that linking with MKL is faster besides that it used to be a bit faster with very old versions of MKL-DNN that I last benchmarked on. If it's just as good without, I'll happily gut the dependency (it makes life easier!). Thanks for the pointer there.

ports/onednn/portfile.cmake
```diff
@@ -0,0 +1,32 @@
...
+    OPTIONS ${DNNL_OPTIONS}
+)
+vcpkg_install_cmake()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake)
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+  file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/bin")
```
JackBoosY, line 27: Is the folder bin empty during static build?

JackBoosY (changes requested):

ports/onednn/portfile.cmake
jacobkahn: @JackBoosY I actually think it is but that I forgot to remove it. Just removed it/we'll see what CI says about the Windows build.

JackBoosY (approved):

PR #15102, opened 2020-12-14
[opencv4] set LD_LIBRARY_PATH to fix dynamic build

### LINKED ISSUE #15101: [opencv4:x64-linux-dynamic] build failure
Seems that building opencv4 with default options (opencv4[core,dnn,jpeg,opengl,png,tiff,webp]) and dynamic linkage on Linux is somewhat broken:

**Host Environment**
- OS: Ubuntu 20.04 / Mint 20
- Compiler: g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
- Custom triplet:

```cmake
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)

```

**To Reproduce**
`./vcpkg install --recurse --overlay-triplets=/home/dweckmann/custom-triplets opencv4`

**Failure logs** 
```bash
...
-- Building x64-linux-dbg
    Command failed: /home/dweckmann/Work/vcpkg/downloads/tools/cmake-3.18.4-linux/cmake-3.18.4-Linux-x86_64/bin/cmake --build . --config Debug --target install -- -v -j9
    Working Directory: /home/dweckmann/Work/vcpkg/buildtrees/opencv4/x64-linux-dbg
      /home/dweckmann/Work/vcpkg/buildtrees/opencv4/install-x64-linux-dbg-out.log

  scripts/cmake/vcpkg_build_cmake.cmake:96 (vcpkg_execute_build_process)
  scripts/cmake/vcpkg_install_cmake.cmake:27 (vcpkg_build_cmake)
  ports/opencv4/portfile.cmake:362 (vcpkg_install_cmake)
```

[install-x64-linux-dbg-out.log](https://github.com/microsoft/vcpkg/files/5687920/install-x64-linux-dbg-out.log)

**Additional context**
The error seems to be:

```bash
/home/dweckmann/Work/vcpkg/installed/x64-linux/tools/protobuf/protoc: error while loading shared libraries: libprotoc.so.3.14.0.0: cannot open shared object file: No such file or directory
``` 

A possible fix is to set LD_LIBRARY_PATH to `vcpkg/installed/x64-linux/lib:vcpkg/installed/x64-linux/debug/lib ` so `protoc` tools can find libprotoc.so.3.14.0.0.

dweckmann:
**Describe the pull request**

- What does your PR fix? 

Fixes #15101

The idea is to add a special option (`ADD_LIB_TO_LD_LIBRARY_PATH`) to `vcpkg_build_cmake()`, so it can handle, as it does for `PATH`, dynamic .so used by tools. The name of the option has been arbitrary chosen, feel free to suggest anything better.

Only Unix builds should be impacted.

### DIFF
--- a/docs/maintainers/vcpkg_build_cmake.md
+++ b/docs/maintainers/vcpkg_build_cmake.md
@@ -18,6 +18,9 @@ be passed.
 ### ADD_BIN_TO_PATH
 Adds the appropriate Release and Debug `bin\` directories to the path during the build such that executables can run against the in-tree DLLs.
 
+### ADD_LIB_TO_LD_LIBRARY_PATH
+Adds the appropriate Release and Debug `lib\` directories to the LD_LIBRARY_PATH on Unix during the build such that executables can run against the in-tree .so.
+
 ## Notes:
 This command should be preceeded by a call to [`vcpkg_configure_cmake()`](vcpkg_configure_cmake.md).
 You can use the alias [`vcpkg_install_cmake()`](vcpkg_configure_cmake.md) function if your CMake script supports the
--- a/ports/opencv4/portfile.cmake
+++ b/ports/opencv4/portfile.cmake
@@ -359,7 +359,7 @@ vcpkg_configure_cmake(
         ${ADDITIONAL_BUILD_FLAGS}
 )
 
-vcpkg_install_cmake()
+vcpkg_install_cmake(ADD_LIB_TO_LD_LIBRARY_PATH)
 vcpkg_fixup_cmake_targets(CONFIG_PATH "share/opencv" TARGET_PATH "share/opencv")
 vcpkg_copy_pdbs()
 
--- a/scripts/cmake/vcpkg_build_cmake.cmake
+++ b/scripts/cmake/vcpkg_build_cmake.cmake
@@ -19,6 +19,9 @@ be passed.
 ### ADD_BIN_TO_PATH
 Adds the appropriate Release and Debug `bin\` directories to the path during the build such that executables can run against the in-tree DLLs.
 
+### ADD_LIB_TO_LD_LIBRARY_PATH
+Adds the appropriate Release and Debug `lib\` directories to the LD_LIBRARY_PATH on Unix during the build such that executables can run against the in-tree .so.
+
 ## Notes:
 This command should be preceeded by a call to [`vcpkg_configure_cmake()`](vcpkg_configure_cmake.md).
 You can use the alias [`vcpkg_install_cmake()`](vcpkg_configure_cmake.md) function if your CMake script supports the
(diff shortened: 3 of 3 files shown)


JackBoosY:
I think it's not correct... @strega-nil what do you think?

JackBoosY:
The reason for the issue in #15101 is that `vcpkg_copy_tool_dependencies` does not do the right thing for *inx.

dweckmann:
What is the right thing to do in `vcpkg_copy_tool_dependencies` ? Copying .so won't work until using the correct `-rpath` at link time (ie `-Wl,-rpath,"\$ORIGIN"`), which certainly requires a patch for protobuf at least. It is cleaner and safer than setting LD_LIBRARY_PATH, but at least LD_LIBRARY_PATH is easy to set and should not break anything. PLease, let me know what to do.

JackBoosY:
```
cl : Command line warning D9025 : overriding '/Z7' with '/Zi'
jmemansi.c(234): fatal error C1051: program database file, 'D:\buildtrees\gdal\x64-windows-dbg\gdal301_d.pdb', has an obsolete format, delete it and recompile
```

JackBoosY:
@dweckmann I'm not sure whether modifying `LD_LIBRARY_PATH` will affect the priority of finding dependent libraries at build time, but it is very possible.
Can copy `libprotoc.so.3.14.0.0` to tools/protobuf solve this issue?

JackBoosY:
If that does not solve the issue, we should modify the cmake file of opencv4 to **temporarily** add `LD_LIBRARY_PATH` before executing the protobuf command, and **restore** it after running the command.

dweckmann:
> If that does not solve the issue, we should modify the cmake file of opencv4 to **temporarily** add `LD_LIBRARY_PATH` before executing the protobuf command, and **restore** it after running the command.

I will try, however, this doesn't seems to be correct as the path itself depend of the build type. You can only set it correctly inside `vcpkg_build_cmake()`. BTW, normally LD_LIBRARY_PATH is only read at runtime. LIBRARY_PATH is for link time. (https://stackoverflow.com/questions/4250624/ld-library-path-vs-library-path#:~:text=LIBRARY_PATH%20is%20used%20by%20gcc,been%20successfully%20compiled%20and%20linked.)

JackBoosY:
@dweckmann Fine, I'm not familiar with `LD_LIBRARY_PATH`, I think that's okay.

JackBoosY (approved):

dweckmann:
> @dweckmann I'm not sure whether modifying `LD_LIBRARY_PATH` will affect the priority of finding dependent libraries at link time, but it is very possible.
> Can copy `libprotoc.so.3.14.0.0` to _tools/protobuf_ solve this issue?

Unfortunately, as I though, it doesn't work. However, maybe we can try to "patch" the binary with `chrpath` to tell `protoc` to look for `../lib`. This could be a smarter solution. But, if this is working, I guess this may be done for all tools, not only protoc

dweckmann:
ok, I think I found the real problem, the runpath for protobuf is `@ORIGIN/../lib` and it should be `@ORIGIN/../../lib`. It is computed so certainly because the original binart is in `/bin` not `/bin/protobuf`. The real problem is not in opencv nor in vcpkg_build_cmake(), but in protobuf. Feel free to close this PR and associated issue.

dweckmann:
BTW, the problem is even worse. By not setting LD_LIBRARY_PATH nor `rpath` or `runpath`, some tools use the system libraries. I now understand why dynamic linkage is not usable on linux...

JackBoosY:
@dweckmann  Yes, that's what billy said.

PR #15114, opened 2020-12-14
[vcpkg registries] support versions

strega-nil:
This PR merges the Registries changes and the versioning changes, so that one can use both at the same time.

There is one major difference between this PR and the RFC (#13590), which is that instead of version files looking like:

```json
[
  ...
]
```

version files look like:

```
{
  "versions": [
    ...
  ]
}
```

this is to support interop between this PR and existing demos and the like;
fixing this, along with perhaps renaming `port_versions` to `port-versions` should be done after this is merged,
should be a trivial change.

This is the first PR resulting from splitting #15054 into multiple PRs.

The e2e tests for filesystem registries will come in the git registries PR, since there's already test code written for that.


toolsrc/src/vcpkg/base/files.cpp
```diff
@@ -1192,7 +1192,7 @@ namespace vcpkg::Files
             {
             }
 
-#if defined(WIN32)
+#if defined(_WIN32)
```
strega-nil, line 1195: drive-by fix.

toolsrc/src/vcpkg/base/files.cpp
```diff
@@ -1228,7 +1228,7 @@ namespace vcpkg::Files
                 ec.clear();
                 if (fd == -1)
                 {
-                    fd = ::open(native.c_str(), 0);
+                    fd = ::open(native.c_str(), O_RDWR | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
```
strega-nil, line 1231: This allows `::open` to create a file that doesn't already exist.

toolsrc/src/vcpkg/versiondeserializers.cpp
```diff
@@ -21,6 +21,7 @@ namespace
 
         Optional<std::string> visit_string(Json::Reader& r, StringView sv) override
         {
+            (void)GIT_TREE;
```
strega-nil, line 24:
```suggestion
```

toolsrc/src/vcpkg/vcpkgpaths.cpp
```diff
@@ -222,7 +233,12 @@ namespace vcpkg
             fs::path m_manifest_path;
             Configuration m_config;
 
+            fs::path m_registries_path;
+
             FeatureFlagSettings m_ff_settings;
+
+            fs::path registries_work_tree_dir() const { return m_registries_path / fs::u8path("git"); }
+            fs::path registries_dot_git_dir() const { return registries_work_tree_dir() / fs::u8path(".git"); }
```
strega-nil, line 241: These lines should be removed.

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,350 +18,715 @@ namespace
...
+        return std::move(res);
+    }
 
     View<StringView> RegistryDeserializer::valid_fields() const
     {
         static const StringView t[] = {
             RegistryImplDeserializer::KIND,
+            RegistryImplDeserializer::BASELINE,
             RegistryImplDeserializer::PATH,
+            RegistryImplDeserializer::REPO,
```
strega-nil, line 721: Should be deleted.

strega-nil (commented):
These are the things I noticed and am fixing now, along with some notes.

toolsrc/src/vcpkg/portfileprovider.cpp
```diff
@@ -95,21 +86,25 @@ namespace vcpkg::PortFileProvider
...
                 auto maybe_scf = Paragraphs::try_load_port(fs, ports_dir);
-                if (auto scf = maybe_scf.get())
+                if (auto scfp = maybe_scf.get())
                 {
-                    if (scf->get()->core_paragraph->name == spec)
+                    auto& scf = *scfp;
+                    if (scf->core_paragraph->name == spec)
                     {
-                        return SourceControlFileLocation{std::move(*scf), ports_dir};
+                        auto res = std::make_unique<OverlayRegistryEntry>();
```
BillyONeal, line 104: This somewhat suggests a missing ctor of `OverlayRegistryEntry` since it's doing the usual "create + assign over" antipattern.

BillyONeal: Ditto below.

toolsrc/include/vcpkg/registries.h
```diff
@@ -17,40 +18,46 @@ namespace vcpkg
...
         // requires: static_cast<bool>(implementation)
-        Registry(std::vector<std::string>&& packages, std::unique_ptr<RegistryImpl>&& implementation);
+        Registry(std::vector<std::string>&& packages, std::unique_ptr<RegistryImplementation>&& implementation);
 
         Registry(std::vector<std::string>&&, std::nullptr_t) = delete;
 
         // always ordered lexicographically
         View<std::string> packages() const { return packages_; }
-        const RegistryImpl& implementation() const { return *implementation_; }
+        const RegistryImplementation& implementation() const { return *implementation_; }
```
BillyONeal, line 52: I think it would be a better design for `Registry` to replicate any function call that someone would call `implementation()` over for law-of-demeter reasons.

(That said this is preexisting in this PR so no change necessarily requested)

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
-            for (std::size_t idx = 0; idx < arr.size(); ++idx)
+            for (auto file : fs::directory_iterator(super_directory))
             {
-                r.visit_at_index(
-                    arr[idx], static_cast<int64_t>(idx), buffer, FilesystemVersionEntryDeserializer::instance);
-
-                auto it = res.versions.lower_bound(buffer.first);
-                if (it == res.versions.end() || it->first != buffer.first)
+                auto filename = fs::u8string(file.path().filename());
+                if (Strings::ends_with(filename, ".json"))
```
BillyONeal, line 178: Consider `if (!condition) { continue; }` to reduce extreme indenting (like 169)

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
+                std::make_unique<BuiltinRegistryEntry>(std::make_unique<GitRegistryEntry>(port_name.to_string()));
+            auto gre = res->git_entry.get();
+            for (auto&& version_entry : version_entries)
+            {
+                gre->port_versions.push_back(version_entry.version);
+                gre->git_trees.push_back(version_entry.git_tree);
+            }
             return res;
         }
+        else
```
BillyONeal, line 216: Delete else due to preceeding return?

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
+            // Fall back to current available version
+            auto port_directory = paths.builtin_ports_directory() / fs::u8path(port_name);
+            if (paths.get_filesystem().exists(port_directory))
+            {
+                auto found_scf = Paragraphs::try_load_port(paths.get_filesystem(), port_directory);
+                if (auto scfp = found_scf.get())
+                {
+                    auto& scf = *scfp;
+                    auto maybe_error = scf->check_against_feature_flags(port_directory, paths.get_feature_flags());
+                    if (maybe_error)
```
BillyONeal, line 227: Add {}s

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
+                {
+                    Checks::exit_with_message(VCPKG_LINE_INFO, res_baseline.error());
+                }
+                opt_baseline = res_baseline.get();
+                if (auto p = opt_baseline->get())
                 {
-                    return std::make_unique<FilesystemEntry>(std::move(*pentry));
+                    return std::move(*p);
                 }
                 else
```
BillyONeal, line 290: Delete else due to preceeding return?

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
+            {
+                Checks::exit_with_message(
+                    VCPKG_LINE_INFO,
+                    "Couldn't find explicitly specified baseline `\"default\"` in the baseline file.",
+                    m_baseline_identifier);
+            }
+
+            // attempt to check out the baseline:
+            auto maybe_path = get_git_baseline_json_path(paths, m_baseline_identifier);
+            if (auto path_to_git_baseline = maybe_path.get())
```
BillyONeal, line 278: Maybe flip this and delete the else for less indents?

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
 
-    struct FilesystemRegistry final : RegistryImpl
+    Optional<VersionT> BuiltinRegistry::get_baseline_version(const VcpkgPaths& paths, StringView port_name) const
     {
-        std::unique_ptr<RegistryEntry> get_port_entry(const VcpkgPaths& paths, StringView port_name) const override
-        {
-            const auto& fs = paths.get_filesystem();
-            auto entry_path = this->path_to_port_entry(paths, port_name);
-            if (!fs.exists(entry_path))
+        const auto& baseline = m_baseline.get([this, &paths]() -> Baseline {
```
BillyONeal, line 249: I'm really not a fan of lambdas that are so long they don't fit on a screen because it makes figuring out what `return` means therein somewhat obscure. Can this be made a named member `const Baseline& BuiltinRegistry::populate_baseline(const VcpkgPaths&)` or similar?

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -16,341 +18,693 @@ namespace
...
+                Checks::exit_with_message(VCPKG_LINE_INFO, res_baseline.error());
+            }
+        });
+
+        auto it = baseline.find(port_name);
+        if (it != baseline.end())
+        {
+            return it->second;
+        }
+        else
```
BillyONeal, line 382: Delete else due to included return?

toolsrc/src/vcpkg/registries.cpp
```diff
@@ -375,9 +729,163 @@ namespace vcpkg
...
+
+        auto versions_file_path = port_versions / relative_path_to_versions(port_name);
+
+        if (!fs.exists(versions_file_path))
+        {
+            return Strings::format("Couldn't find the versions database file: %s", fs::u8string(versions_file_path));
+        }
+
+        auto maybe_contents = fs.read_contents(versions_file_path);
+        if (auto contents = maybe_contents.get())
```
BillyONeal, line 760: Flip and delete else for -1 indent?

BillyONeal (commented):
I feel bad that I don't have substantial code review comments, because most of this is parsing code etc. and I don't think humans in general are very good at reviewing that without debugging through examples or having complete test coverage.

I do think this needs some tests before it is merged. (I don't if versions had test coverage but had I been around to review that I'd probably have made the same comment :/ )

toolsrc/include/vcpkg/base/files.h
```diff
@@ -236,6 +236,7 @@ namespace vcpkg::Files
         void current_path(const fs::path& path, LineInfo li);
 
         // waits forever for the file lock
+        // creates a non-existent file
```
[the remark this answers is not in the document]
strega-nil: It is attempting to say "if you pass a path that doesn't exist, this function creates that file and locks it."

toolsrc/include/vcpkg/base/jsonreader.h
```diff
@@ -75,13 +75,13 @@ namespace vcpkg::Json
         };
         std::vector<Path> m_path;
 
+    public:
```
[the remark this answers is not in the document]
strega-nil: Probably, but this is a small change to just allow a single additional public member.

toolsrc/src/vcpkg/install.cpp
```diff
@@ -839,17 +841,12 @@ namespace vcpkg::Install
...
-                    }
-                    else
-                    {
-                        return std::make_unique<PortFileProvider::BaselineProvider>(paths);
-                    }
-                }();
+                auto baseprovider = std::make_unique<PortFileProvider::BaselineProvider>(paths);
+                if (auto p_baseline = manifest_scf.core_paragraph->extra_info.get("$x-default-baseline"))
+                {
+                    paths.get_configuration().registry_set.experimental_set_builtin_registry_baseline(
```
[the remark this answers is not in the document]
strega-nil: Yes. This function will be removed, and it's important to say so, I think.

toolsrc/include/vcpkg/base/files.h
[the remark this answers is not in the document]
BillyONeal: See I interpreted this as "it is an error if the file exists" because "non-existent" file looks like a precondition.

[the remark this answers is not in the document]
strega-nil: mmh, I'll reword the comment.

ras0219-msft (commented):
Review still in progress (portfileprovider.cpp:299)

toolsrc/include/vcpkg/configuration.h
```diff
@@ -13,4 +17,7 @@ namespace vcpkg
         // taken care of in RegistrySet.
         RegistrySet registry_set;
     };
+
+    std::unique_ptr<Json::IDeserializer<Configuration>> make_configuration_deserializer(
+        const VcpkgCmdArguments& args, const fs::path& config_directory);
```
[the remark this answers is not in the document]
strega-nil: This is extremely hard to actually implement.

toolsrc/include/vcpkg/registries.h
```diff
@@ -17,40 +19,46 @@ namespace vcpkg
 {
     struct RegistryEntry
     {
-        // returns fs::path() if version doesn't exist
-        virtual fs::path get_port_directory(const VcpkgPaths& paths, const VersionT& version) const = 0;
+        virtual View<VersionT> get_port_versions() const = 0;
+
+        // precondition: version \in get_port_versions(). Otherwise, this function asserts.
```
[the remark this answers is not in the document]
strega-nil: This is actually how it works in the implementation, fwiw.

BillyONeal (approved):
I still would really like to see unit level tests for any parsing code that aren't present here.

toolsrc/src/vcpkg-test/registries.cpp
```diff
@@ -0,0 +1,69 @@
...
+    int get_tri_num(const RegistryImplementation& r)
+    {
+        if (auto tri = dynamic_cast<const TestRegistryImplementation*>(&r))
+        {
+            return tri->number;
+        }
+        else
+        {
+            return -1;
+        }
```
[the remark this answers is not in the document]
strega-nil: Given that this is basically a pattern match, not a lot of code, not going to use this suggestion.

BillyONeal (approved):

vicroms (approved):

toolsrc/include/vcpkg/vcpkgcmdarguments.h
```diff
@@ -134,6 +134,8 @@ namespace vcpkg
         std::unique_ptr<std::string> scripts_root_dir;
         constexpr static StringLiteral BUILTIN_PORTS_ROOT_DIR_ARG = "x-builtin-ports-root";
         std::unique_ptr<std::string> builtin_ports_root_dir;
+        constexpr static StringLiteral BUILTIN_PORT_VERSIONS_DIR_ARG = "x-builtin-port-versions-dir";
+        std::unique_ptr<std::string> builtin_port_versions_dir;
```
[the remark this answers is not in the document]
strega-nil: I think we should do a full bikeshed on these in the new year. I think they're still fine to bikeshed.

ras0219 (commented):
@strega-nil Please consider opening a new PR that addresses these issues

PR #15159, opened 2020-12-16
[comms] fixes debug build configuration

mathisloge:
**Describe the pull request**

- What does your PR fix? 
i've discoverd that the debug build configuration is broken as soon as one is including the project.
this PR moves the cmake files from lib / debug/lib to share. the root cause was that the target config cmake file had INTERFACE_INCLUDE_DIRECTORIES  set to debug/include/{port}

- Which triplets are supported/not supported? Have you updated the CI baseline? uwp /yes

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?yes


vicroms:
Please update the port version of all updated ports.

ports/comms-ublox/vcpkg.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "comms-ublox",
   "version-string": "0.20.0",
+  "port-version": 2,
```
vicroms, line 4:
```suggestion
  "port-version": 1,
```

Port versions start at 0 (omitting `"port-versions"` defaults to 0) so this is technically port version 1.
Same for the other ports.

ports/comms/vcpkg.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "comms",
   "version-string": "3.1.3",
+  "port-version": 2,
```
vicroms, line 4:
```suggestion
  "port-version": 1,
```

ports/commsdsl/vcpkg.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "commsdsl",
   "version-string": "3.6.0",
+  "port-version": 2,
```
vicroms, line 4:
```suggestion
  "port-version": 1,
```

vicroms (changes requested):

ports/comms-ublox/vcpkg.json
mathisloge: oh. didn't notice. thanks for the hint.

ports/comms/portfile.cmake
```diff
@@ -16,9 +16,10 @@ vcpkg_configure_cmake(
         -DCC_NO_UNIT_TESTS=ON
 )
 vcpkg_install_cmake()
-
-file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
-file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/LibComms/cmake TARGET_PATH share/LibComms)
```
vicroms, line 19: I noticed that the vcpkg heuristics detect some extra targets that will cause link errors.

```
The package comms:x86-windows provides CMake targets:

    find_package(LibComms CONFIG REQUIRED)
    # Note: 1 target(s) were omitted.
    target_link_libraries(main PRIVATE comms cc::comms "cc_cxxtest" cc::comms_champion)
```

For me only target `cc::comms` actually works. 
So I would recommend adding a `usage` file to output the correct message.

mathisloge: Okay. Will add that tomorrow. The cc::comms_champion is currently disabled since the whole part might be separated from the core library. It just enables plugin bindings for the gui to test the protocols directly. But that is a feature which needs some work after the separation. If the separation will not take place, I'll need to make some patches to get qt correctly linked.

mathisloge: added usage file

ports/comms/usage
```diff
@@ -0,0 +1,4 @@
+The package comms provides CMake targets:
+
+    find_package(LibComms CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE cc::comms)
```
JackBoosY, line 4: Can you please add an empty line below?

JackBoosY (changes requested):

ports/comms/usage
mathisloge: done.

JackBoosY (approved):

JackBoosY:
LGTM.

vicroms (approved):

PR #15170, opened 2020-12-16
[qt5-base] mysql plugin added

jepessen:
The pull request creates the qsqlmysql database plugin for qt5-base.

It adds a dependency from libmysql in order to build it before the plugin since it's requested.


jepessen:
Hi.
It's possible in some way to see the error logs written during the tests, like[ this one](https://dev.azure.com/vcpkg/public/_build/results?buildId=46650&view=logs&j=e43d08a3-46a6-5c05-c22b-b8b3a7ac1517&t=341b2070-72fd-5058-0746-f056592aaa91&l=3312)?

Because I can try to fix for all configuration that I have, I can't test every of them (I don't have linux machines or osx for example) so I cannot perform local tests and I cannot check what's wrong with them.

JackBoosY:
@jepessen 
1. Click `Details` in pipeline test results below.
2. Click `View more details on Azure Pipelines`
3. Click `[NUM] artifacts produced`
4. Click the three vertical dots on the far right of the error log entry
5. Click `Download artifacts`

jepessen:
@JackBoosY ok thanks a lot

JackBoosY:

JackBoosY:
@Neumann-A said this should be added as a qt5-base feature.

jepessen:
Ok I'll add it as feature ASAP.

jepessen:
I've set the mysql plugin as qt5-base feature.

ports/qt5-base/portfile.cmake
```diff
@@ -112,6 +123,12 @@ if(QT_OPENSSL_LINK)
     list(APPEND CORE_OPTIONS -openssl-linked)
 endif()
 
+if(VCPKG_TARGET_IS_WINDOWS)
+if(QT_MYSQL_PLUGIN)
+    list(APPEND CORE_OPTIONS -sql-mysql)
+endif()
+endif()
```
JackBoosY, line 130:
```suggestion
if(VCPKG_TARGET_IS_WINDOWS)
    if(QT_MYSQL_PLUGIN)
        list(APPEND CORE_OPTIONS -sql-mysql)
    endif()
endif()
```

JackBoosY (changes requested):

JackBoosY (approved):

JackBoosY:
LGTM.

ports/qt5/CONTROL
```diff
@@ -1,8 +1,9 @@
 Source: qt5
 Version: 5.15.1
+Port-Version: 1
 Homepage: https://www.qt.io/
 Description: Qt5 Application Framework
-Build-Depends: qt5-base[core]
+Build-Depends: qt5-base[core,mysqlplugin]
```
vicroms, line 6: Why add this as a default feature?

JackBoosY: Yeah, I think it should be removed.

jepessen: I think it should be there, because there's already sqlite and postgresql. Since they are all plugin, or we put them all together at once or none of them in my opinion. There's no reason to include postgresql plugin by default and not mysql, or vice versa. Since postgresql plugin is already installed by default with qt5 I've followed the same logic.

jepessen:
I think it should be there, because there's already sqlite and postgresql. Since they are all plugin, or we put them all together at once or none of them in my opinion. There's no reason to include postgresql plugin by default and not mysql, or vice versa. Since postgresql plugin is already installed by default with qt5 I've followed the same logic (it was the reason why I did't added it with a separate feature at start).

ports/qt5/CONTROL
JackBoosY: One last question: does feature `mysqlplugin` conflict with other features?

jepessen: I don't think so, all check passes and it adds only the qmysql dll plugin. It does not change anything already existent.

JackBoosY: All the checks passed because only `mysqlplugin` is the default feature.

jepessen: I'll try to build qt5-base and qt5 without the feature enabled and I'll let you know if there are errors.

jepessen: I've tried to build qt5 with almost all modules (except webengine and wayland) and it works. This is the command:
```
.\vcpkg.exe install qt5[3d,activeqt,charts,connectivity,datavis3d,declarative,doc,essentials,extras,gamepad,graphicaleffects,imageformats,latest,location,mqtt,multimedia,networkauth,purchasing,quickcontrols,quickcontrols2,remoteobjects,script,scxml,sensors,serialbus,serialport,speech,svg,tools,translations,virtualkeyboard,webchannel,webglplugin,websockets,webview,xmlpatterns]:x64-windows
```
So the mysql plugin module should not interfere with others.

JackBoosY: okay.

jepessen: There's something else that I need to check?

JackBoosY: @jepessen Nope, waiting for merge.

jepessen: Ok thanks a lot!

JackBoosY:
No...no... the `mysql` doesn't support x86 now, so if you try to install `qt5` directly, it will shows us:
```
CMake Error at scripts/cmake/vcpkg_fail_port_install.cmake:97 (message):
  Architecture 'x86' not supported by libmysql!
```

I will open a PR to remove the default feature.

PR #15172, opened 2020-12-17
[vcpkg] Improve versioning files generators

vicroms:
This PR extracts the improvements to the generators from #14999 

Port versions are now parallelized to drastically improve generation time and the baseline is now generated from the output of `port_versions/` which is also way faster than calling `x-package-info`.


scripts/generateBaseline.py
```diff
@@ -1,50 +1,71 @@
...
+        if not os.path.exists(port_file_path):
+            print(
+                f'Error: No version file for {port_name}.\n', file=sys.stderr)
+            continue
+        sys.stderr.write(
+            f'\rProcessed {i}/{total_count} ({i/total_count:.2%})')
+        with open(port_file_path, 'r') as db_file:
+            try:
+                versions_object = json.load(db_file)
+                if len(versions_object['versions']) > 0:
```
Hoikas, line 37:
```suggestion
                if versions_object['versions']:
```

Empty dicts evaluate to `False`.

vicroms: Thanks!

toolsrc/src/vcpkg/commands.porthistory.cpp
```diff
@@ -241,7 +215,24 @@ namespace vcpkg::Commands::PortHistory
...
 
             auto json_string = Json::stringify(root, vcpkg::Json::JsonStyle::with_spaces(2));
-            System::printf("%s\n", json_string);
+
+            if (maybe_output_file.has_value())
+            {
+                auto output_file_path = fs::u8path(maybe_output_file.value_or_exit(VCPKG_LINE_INFO));
+                auto& fs = paths.get_filesystem();
+                std::error_code ec;
+                fs.write_contents(output_file_path, json_string, ec);
```
ras0219, line 224:
```suggestion
                fs.write_contents(output_file_path, json_string, VCPKG_LINE_INFO);
```
`write_contents` should do the right thing if passed VCPKG_LINE_INFO.

ras0219 (approved):

PR #15180, opened 2020-12-17
[tesseract]: find_package(lz4) in TesseractConfig.cmake

mheyman:
This fixes an issue I was having with building opencv4[contrib]. OpenCV[contrib] uses tesseract and tesseract uses lz4 but doesn't make the lz4:lz4 target available. It does make targets for other dependencies available so I figured this small change was okay..


NancyLi1013 (commented):
Could you please also bump the `Port-Version` in CONTROL file?

NancyLi1013:
It seems like `lz4 `is not required for `tesseract` directly. `tesseract` depends on `libarchive `and `leptonica`. 

`Lz4 `is required for `libarchive`. I think we should fix this on `libarchive `instead of `tesseract`. Other targets are the same case.

Could you please help confirm and try?

mheyman:
I updated libarchive to have a simple libarchiveConfig.cmake that tesseract could depend on. Now opencv can use lz4::lz4 because it does a find_package(tesseract) which does a find_package(libarchive) which does a find_package(lz4).

portfile version numbers also bumped.

cenit:
#15089 ?

cenit:
if this one gets merged, I will fix mine to still enhance CI coverage for the future... we cannot have these problems slip in again :)

cenit:
also, because you are using a find_package(libarchive), you can remove many find_packages in the tesseractConfig.cmake which were just transitively added from libarchive.
Also, please take care of spelling, with caps and lowercases. Your spelling of LibLZMA is just asking for troubles on case sensitive filesystems

ports/libarchive/portfile.cmake
```diff
@@ -64,6 +64,22 @@ vcpkg_configure_cmake(
...
+endif()
+if(FEATURES MATCHES "lzma")
+    set(LIBARCHIVE_CONFIG "${LIBARCHIVE_CONFIG}find_package(liblzma REQUIRED)\n")
+endif()
+if(FEATURES MATCHES "zstd")
+    set(LIBARCHIVE_CONFIG "${LIBARCHIVE_CONFIG}find_package(zstd REQUIRED)\n")
+endif()
+if(LIBARCHIVE_CONFIG)
+    file(WRITE ${CURRENT_PACKAGES_DIR}/share/libarchive/libarchiveConfig.cmake "${LIBARCHIVE_CONFIG}")
+endif()
```
NancyLi1013, line 81: Can we make a patch to add these dependencies that are required by the features?

ports/tesseract/portfile.cmake
```diff
@@ -47,7 +47,8 @@ find_package(ZLIB REQUIRED)
 find_package(BZip2 REQUIRED)
 find_package(LibXml2 REQUIRED)
 find_package(zstd REQUIRED)
```
NancyLi1013, line 49:
```suggestion
```
These packages have been included in `libarchive`. So there is no need to add them.

ports/tesseract/portfile.cmake
```diff
@@ -47,7 +47,8 @@ find_package(ZLIB REQUIRED)
...
-find_package(LibLZMA REQUIRED)" TESSERACT_CONFIG "${TESSERACT_CONFIG}")
+find_package(LibLZMA REQUIRED)
```
NancyLi1013, line 50:
```suggestion
```
The same as above.

NancyLi1013:
@cenit 

I noticed that you also update these changes for `libarchive `and `tesseract` in PR #15089. 

So let's revert these changes in this PR and we use the former way to fix the dependency `lz4`, what do you think?

cenit:
I updated these changes also in #15089 because you're right, it was not tesseract directly depending on lz4 but its dependency libarchive is.
Since I'd had to merge these changes in that PR sooner or later, I already did it while I had some time.
That PR already contains latest requested you just asked here (and also I used find_dependency in order to properly manage REQUIRED keyword upstream). We are just duplicating work I think...

NancyLi1013:
@cenit 

Thanks for instant response and also for your hard work. 
I think we can merge this PR first, then you update the changes in your PR.

@mheyman 

Since this PR includes the same change as PR #15089.

Could you please revert this changes in this PR and only add `find_package(lz4 REQUIRED)" TESSERACT_CONFIG "${TESSERACT_CONFIG}")`?

I'm so sorry for my poor consideration, which leads you to do repeated work.

NancyLi1013:
Since PR #15089 has fixed this issue, which has been merged to master.

So this PR is not needed now. Still thanks for your contribution @mheyman.

PR #15187, opened 2020-12-18
Write manifest "vcpkg install" output to stdout

PazerOP:
First pull request to this project, so there might be things I'm overlooking.

In https://github.com/microsoft/vcpkg/commit/c9027488970352953a59505234775b367d4cfd2a, the output from the `vcpkg install` command run by manifest mode was written to a file. This had the side effect of no longer writing that output to stdout/err, which can be a very unfortunate change if your dependencies managed by vcpkg take a long time to install: you will sit there with the only output being "Running vcpkg install" at the very beginning, and "Running vcpkg install - done" at the very end. Since there is otherwise no progress indicator, I believe it beneficial to have that `vcpkg install` output visible in the console.

CMake's `execute_process` functionality has support for `ECHO_OUTPUT_VARIABLE` and `ECHO_ERROR_VARIABLE`, but unfortunately these only apply to the variable output methods, not the file output methods. This pull request stores the output of the `vcpkg install` command into a new variable `_VCPKG_MANIFEST_INSTALL_LOGTEXT`, and then after the `vcpkg install` command is finished, the output is written into the `vcpkg-manifest-install.log` file.

No other changes were made besides vcpkg.cmake. I believe this question is intended for users who are making changes to ports.

I believe so, but again, my understanding is that this question is intended for people that are making changes to ports.


scripts/buildsystems/vcpkg.cmake
```diff
@@ -402,11 +401,16 @@ if(VCPKG_MANIFEST_MODE AND VCPKG_MANIFEST_INSTALL AND NOT _CMAKE_IN_TRY_COMPILE
                 "--x-install-root=${_VCPKG_INSTALLED_DIR}"
                 ${_VCPKG_ADDITIONAL_MANIFEST_PARAMS}
                 ${VCPKG_INSTALL_OPTIONS}
-            OUTPUT_FILE "${_VCPKG_MANIFEST_INSTALL_LOG}"
-            ERROR_FILE "${_VCPKG_MANIFEST_INSTALL_LOG}"
+            OUTPUT_VARIABLE _VCPKG_MANIFEST_INSTALL_LOGTEXT
+            ERROR_VARIABLE _VCPKG_MANIFEST_INSTALL_LOGTEXT
+            ECHO_OUTPUT_VARIABLE
+            ECHO_ERROR_VARIABLE
```
strega-nil, line 407: This needs to guard the version of CMake, since we support back to 3.5 and this was added in 3.18. The best way to do this, I think, is:

```cmake
if(CMAKE_VERSION VERSION_GREATER "3.18")
    set(echo_outputs ECHO_OUTPUT_VARIABLE ECHO_ERROR_VARIABLE)
else()
    set(echo_outputs)
endif()
...
execute_process(
  ...
  ${echo_outputs})
```

PazerOP: It is possible to simply reuse and append `ECHO_OUTPUT_VARIABLE` and `ECHO_ERROR_VARIABLE` to one of the existing variables that is used in the `execute_process` command. Is this considered bad practice?

```CMake
if (CMAKE_VERSION VERSION_GREATER "3.18")
    set(_VCPKG_ADDITIONAL_MANIFEST_PARAMS ${_VCPKG_ADDITIONAL_MANIFEST_PARAMS} ECHO_OUTPUT_VARIABLE ECHO_ERROR_VARIABLE)
endif()
```

PazerOP: I made the suggested change, but used `VERSION_GREATER_EQUAL` instead of `VERSION_GREATER`. Also, I named the variable `_VCPKG_MANIFEST_INSTALL_ECHO_PARAMS` for consistency.

strega-nil (approved):

JackBoosY (approved):

PR #15213, opened 2020-12-19
[libass] Update to 0.15.0

wangqr:
**Describe the pull request**

- What does your PR fix?
Update libass to 0.15.0

All triplets are supported. Just like previous version.

Yes


ports/libass/libass.def
```diff
@@ -1,24 +1,34 @@
 EXPORTS
+ass_library_version
```
JackBoosY, line 2: Why change this file? Is the order of function symbols related to construction?

JackBoosY (changes requested):

ports/libass/libass.def
wangqr: A new function, `ass_track_set_feature` is added to libass in 0.15.0. So this file is regenerated from the new `ass.h`. The order of functions in `libass.def` now matches `ass.h` from upstream.

If we want to minimize change, we can append `ass_track_set_feature` to `libass.def` instead of regenerating whole file.

JackBoosY: @wangqr Fine, that's reasonable.

JackBoosY (approved):

ports/libass/portfile.cmake
```diff
@@ -1,10 +1,9 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO libass/libass
-    REF 98727c3b78f44cb3bbc955fcf5d977ebd911d5ca
-    SHA512 d466108180cea598b817f89aa21a1021ed2a763580d9aad51b054aa120186af48ab4264907e49ddcb38479a28d87d5431751a28afee9cb83ad7623f002d99c57
+    REF 0.15.0
+    SHA512 a832da5246577cf815481bb9e4bebabc74bc1d0f5c50faa098f4150a379d801d7e6d1bd7f9f578143a9412e258c5296d08c4fc2d04cc33f1751e613c2583214c
```
JackBoosY, line 5: Seems libass installed the pkgconfig, can you please add `vcpkg_fixup_pkgconfig()` to line 24?

JackBoosY (changes requested):

wangqr:
The CIs are failing due to harfbuzz.pc (the PkgConfig file) not found. Currently harfbuzz.pc is not installed if harfbuzz is built with CMake (harfbuzz/harfbuzz#896) and upstream has no interest in fixing this. Building harfbuzz using meson should be able to install harfbuzz.pc properly, but this is still waiting to be merged into vcpkg (#12860).

wangqr:
Force-pushed to rebase this PR after #15263

JackBoosY:
Check failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=46997&view=artifacts&pathAsName=false&type=publishedArtifacts).

wangqr:
The dependency PR #12860 has been merged, so pkgconfig is no longer failing.

JackBoosY (approved):

JackBoosY:
Seems good, don't know why the pipeline didn't archived the installed file paths.
Already tested on `x86-windows` and `x64-windows-static` locally.

PR #15252, opened 2020-12-22
Fixed port of quirc - patch did no longer apply / modified to fix build

Markus87:
* Quirc did no longer build for any triplet because the patch was for a previous version of quirc
* Modified the patch to get the windows triplets working again with the new version of quirc


Markus87:
I dont know how this did not break any tests.
Are not all ports tested by CI? If so how to get a port into the automatic tests?

JackBoosY:
@Markus87 Maybe the original error has been fixed by other means, and the patch is no longer needed.
So does this PR still exist?

Markus87:
@JackBoosY Indeed the original patch would no longer be needed, but a new patch is required to make it build with msvc.
I just checked the latest code in the quirc repository and it seems fixed there, so updating quirc to the latest version should make the patch obsolete.

I will try to update quirc to the latest version. If that works can I use this pull request for updating the port or should I start a new one?

JackBoosY:
@Markus87 Both patch or update are fine.

JackBoosY (approved):

JackBoosY:
```
CMake Error at ports/quirc/portfile.cmake:24 (file):
  file COPY cannot find
  "/mnt/vcpkg-ci/buildtrees/quirc/src/b114181c40-f1e71e0d5b.clean/license":
  No such file or directory.
```

Will fix later.

JackBoosY:
LGTM.

vicroms:
Thanks for the PR!

PR #15254, opened 2020-12-22
[boost-modular-builder] let fallback to default tool path

xandox:
Let fallback to default boost build path.

This fix works for me now but I am not sure it is correct way to fix a problem.
I use vcpkg in manifest mode and try cross compile my project for android. I redefine `VCPKG_TARGET_ARCHITECTURE` and `VCPKG_CMAKE_SYSTEM_NAME` for `boost-build` port in overplayed triplet file, but any checks in this script assumes that you use vcpkg not in manifest mode. Maybe it's better to check here if we in manifest mode but there are not any cmake variable which can helps to detect this mode (VCPKG_MANIFEST_MODE not defined, VCPKG_MANIFEST_INSTALL is OFF).


PhoebeHui:
@xandox, could you open an issue for this?

xandox:
@PhoebeHui sure but for what concrete?

PhoebeHui:
@xandox, I'm also not sure if it's a correct fix, since it affects a community triplet, this is not usual scenario we're working on, but we accept the PRs for fixing the issue, I suggest to open an issue with details so others may replicate this problem.

BTW, what does the issue have to do with manifest? this is not be a problem if not in manifest mode?

xandox:
Tried to explain everything in #15291

PhoebeHui:
@xandox, could you please resolve the conflicts?

@BillyONeal, could you help further review?

vicroms (approved):
This looks OK to me.

PR #15281, opened 2020-12-24
[librsync] fix librsync linkage

### LINKED ISSUE #15132: librsync faiture linking
**Describe the bug**
I found and installed the librsync library. It was installed as x64-windows-static since it can only compile into a static library. Then I wrote find_library and find_path in my cmakelists. CMake has successfully found both the library and the path to include. But when compiling, the linker cannot find symbols from Librsync

**Environment**
- OS: Windows 10
- Compiler: MSVC 19.28.29335.0

**To Reproduce**
1. ./vcpkg install librsync
2. CmakeList.txt
`find_library(RSYNC_LIB "rsync.lib")
find_path(RSUNC_INCLUDE "librsync.h")
add_library(patcher ${SOURCES} ${HEADERS})
target_link_libraries(patcher PRIVATE ${RSYNC_LIB} )
target_include_directories(patcher PUBLIC ${RSUNC_INCLUDE})`
3. Use any function from librsync
4. See linking error

**Expected behavior**
Success linking with librsync

Amf1k:
**Describe the pull request**

- What does your PR fix? Fixes #15132

- Which triplets are supported/not supported? Have you updated the CI baseline? No

### DIFF
--- a/ports/librsync/portfile.cmake
+++ b/ports/librsync/portfile.cmake
@@ -33,6 +33,13 @@ if(EXISTS ${CURRENT_PACKAGES_DIR}/debug/lib/rsync.dll)
     file(RENAME ${CURRENT_PACKAGES_DIR}/debug/lib/rsync.dll ${CURRENT_PACKAGES_DIR}/debug/bin/rsync.dll)
 endif()
 
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+    vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/include/librsync_export.h
+        "#  ifdef LIBRSYNC_STATIC_DEFINE"
+        "#  if 1 /* LIBRSYNC_STATIC_DEFINE */"
+    )
+endif()
+
 file(INSTALL
     ${SOURCE_PATH}/COPYING
     DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright
--- a/ports/librsync/CONTROL
+++ b/ports/librsync/CONTROL
@@ -1,5 +1,6 @@
 Source: librsync
 Version: 2020-09-16
+Port-Version: 1
 Description: librsync is a library for calculating and applying network deltas, with an interface designed to ease integration into diverse network applications.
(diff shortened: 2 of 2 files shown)


ports/librsync/portfile.cmake
```diff
@@ -17,6 +17,7 @@ vcpkg_configure_cmake(
         -DBUILD_RDIFF:BOOL=OFF 
         -DENABLE_COMPRESSION:BOOL=OFF
         -DENABLE_TRACE:BOOL=OFF
+        -DLIBRSYNC_STATIC_DEFINE:BOOL=ON
```
NancyLi1013, line 20: This seems invalided. Have you tested it on your machine?

NancyLi1013 (changes requested):

ports/librsync/portfile.cmake
Amf1k: I tested, but apparently I made a mistake and I had this key added to the project via add_compile_definitions. I think now we need to look for another solution, apparently we will have to somehow wrap everything in a cmake package

NancyLi1013: Yes, since there are no CMake config and targets exported on upstream. We might consider to use another way to solve this problem, for example:

```
    vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/include/librsync_export.h
        "#ifdef LIBRSYNC_STATIC_DEFINE"
        "#if 1"
    )
endif()
```

NancyLi1013 (commented):
Please also add Port-Version: 1 to CONTROL file.

ports/librsync/portfile.cmake
Amf1k: Yes, this is a good and simple option. I corrected it a bit and add a comment

Amf1k:
> Please also add Port-Version: 1 to CONTROL file.

done.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your PR @Amf1k.

vicroms:
Thanks for the PR!

PR #15284, opened 2020-12-24
[imgui] Rewrite build process related to bindings

JackBoosY:
Inherited from #15063.

- Use INTERFACE library instead of adding binding-related code files to imgui core.

Related: https://github.com/ocornut/imgui/pull/1713


JackBoosY:
Waiting for the upstream approval.

PR #14008, opened 2020-10-13
[freetype] Update to 2.10.4

kevinlul:
pkgconfig patch was upstreamed

- What does your PR fix? Regular port update https://sourceforge.net/projects/freetype/files/freetype2/2.10.4/ with major security vulnerability fix

- Which triplets are supported/not supported? Have you updated the CI baseline? Should be no changes

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? Yes

PR #15334, opened 2020-12-28
[freetype] Update to 2.10.4

kevinlul:
**Describe the pull request**

- What does your PR fix?

Regular port update https://sourceforge.net/projects/freetype/files/freetype2/2.10.4/ with major security vulnerability fix, plus the upstream vtk patch for compatibility with this revision from https://discourse.vtk.org/t/building-opencascade-7-5-0-with-vtk-9-0-1

Closes #14008. I made a new PR due to the changes on `master` and branched off the last commit that affected freetype and vtk since I could not build anything on HEAD #15332 but I could on 191e01f 

It should work everywhere but since vtk takes really long to build, I've only tested against x86-windows.

### DIFF
--- a/ports/vtk/fix-freetype.patch
+++ b/ports/vtk/fix-freetype.patch
@@ -1,0 +1,26 @@
+diff --git a/Rendering/FreeType/vtkFreeTypeTools.cxx b/Rendering/FreeType/vtkFreeTypeTools.cxx
+index c54289dc..e6d9b14f 100644
+--- a/Rendering/FreeType/vtkFreeTypeTools.cxx
++++ b/Rendering/FreeType/vtkFreeTypeTools.cxx
+@@ -378,7 +378,7 @@ FTC_CMapCache* vtkFreeTypeTools::GetCMapCache()
+ }
+ 
+ //----------------------------------------------------------------------------
+-FT_CALLBACK_DEF(FT_Error)
++FT_Error
+ vtkFreeTypeToolsFaceRequester(
+   FTC_FaceID face_id, FT_Library lib, FT_Pointer request_data, FT_Face* face)
+ {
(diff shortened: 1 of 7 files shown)


ports/vtk/fix-freetype.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/Rendering/FreeTypeFontConfig/vtkFontConfigFreeTypeTools.cxx b/Rendering/FreeTypeFontConfig/vtkFontConfigFreeTypeTools.cxx
+index 5c1908cf..41ed97e7 100644
+--- a/Rendering/FreeTypeFontConfig/vtkFontConfigFreeTypeTools.cxx
++++ b/Rendering/FreeTypeFontConfig/vtkFontConfigFreeTypeTools.cxx
+@@ -29,7 +29,7 @@ vtkStandardNewMacro(vtkFontConfigFreeTypeTools);
+ namespace
+ {
+ // The FreeType face requester callback:
+-FT_CALLBACK_DEF(FT_Error)
++static FT_Error
```
JackBoosY, line 23: What is wrong here? Does the upstream accept this changes and know this bug?

JackBoosY (changes requested):

ports/vtk/fix-freetype.patch
kevinlul: See the linked thread for the details. The short version that I understand is that `FT_CALLBACK_DEF` is actually an internal freetype macro that got removed in 2.10.4, and the patches I incorporate and reference replace the usage of this macro with what it actually expands to.

JackBoosY (approved):
Did you test the usage?

kevinlul:
Behaviour of freetype is fine for me since that was working in the old PR but I cannot really say anything meaningful about vtk since I didn't expect this problem with a seemingly unrelated port.

vicroms:
Thanks for the PR!

PR #15352, opened 2020-12-28
[boost-python] Adds `python-debugging=on` to `B2_OPTIONS_DBG` (fixes #15350)

### LINKED ISSUE #15350: [boost-python] `python39_d.dll` not found when `BOOST_DEBUG_PYTHON` is defined in a custom triplet
As a workaround to #5097, I decided to define `BOOST_DEBUG_PYTHON` in our custom triplet, which is [the official way to use Python Debug libraries with Boost](https://www.boost.org/doc/libs/1_75_0/libs/python/doc/html/building/python_debugging_builds.html) (this could be seen as a fix to #5097, but I'm going to let the author of the issue decide if it is). However, linking failed saying that `python39_d.dll` could not be found. Investigations lead me to the conclusion that the `B2` option `python-debugging` should be set to `on`  when `BOOST_DEBUG_PYTHON` is defined.

**Proposed solution**
Building on #15274, we can add `python-debugging=on` to `B2_OPTIONS_DBG` when `BOOST_DEBUG_PYTHON` is inside the `VCPKG_CXX_FLAGS_DEBUG`. A pull request is coming.

I see no alternative to fix the current issue. However, the initial goal was to find a workaround to #5097.

As a workaround to #5097, one alternative is to _not_ use `BOOST_DEBUG_PYTHON`, i.e. to always link to the release library `python39.dll`. However, as [frivard-coveo commented](https://github.com/microsoft/vcpkg/issues/5097#issuecomment-746910534), when trying to do this in a build where the `CMAKE_BUILD_TYPE` is `Debug`, the script `applocal.ps1` did not copy `python39.dll` inside the binaries directory because it searches for DLLs inside `vcpkg_installed/x64-windows/debug`. Hence, one alternative is to modify the port `python3` so that the release library `python39.dll` is copied inside `vcpkg_installed/x64-windows/debug`. This is, to say the least, unconventional. Another alternative is to modify `applocal.ps1` so that it can search inside both `vcpkg_installed/x64-windows/debug` and `vcpkg_installed/x64-windows` when the `CMAKE_BUILD_TYPE` is `Debug`. This alternative seems to me like a significant change, which we may want to think about before doing.

klalumiere:
**Describe the pull request**
Adds `python-debugging=on` to `B2_OPTIONS_DBG` when `BOOST_DEBUG_PYTHON` is inside `VCPKG_CXX_FLAGS_DEBUG`.

- What does your PR fix? Fixes #15350


PhoebeHui (approved):

PR #15371, opened 2020-12-29
build libiconv and realsense2 for android

xandox:
1. make vcpkg_configure_make works with detected C and CXX compilers not only for WIN32 host system
2. make libiconv not empty package for android
3. add libusb as dependency of realsense2 under android

- Which triplets are supported/not supported? arm64-android

yes


JackBoosY:
Get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=47113&view=artifacts&pathAsName=false&type=publishedArtifacts).

xandox:
@JackBoosY thank you very match. 
Can you restart x86_windows tests, because they failed with 
```console
-- Downloading https://download.gnome.org/sources/gmime/3.2/gmime-3.2.6.tar.xz...
-- Downloading https://download.gnome.org/sources/gmime/3.2/gmime-3.2.6.tar.xz... Failed. Status: 28;"Timeout was reached"
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:184 (message):
      
      Failed to download file.
      If you use a proxy, please set the HTTPS_PROXY and HTTP_PROXY environment
      variables to "https://user:password@your-proxy-ip-address:port/".
      Otherwise, please submit an issue at https://github.com/Microsoft/vcpkg/issues

  ports/gmime/portfile.cmake:6 (vcpkg_download_distfile)
```

Also is it possible to get config.log files for osx? They usually were placed to `buildtrees`/${PORT}/${VCPKG_TARGET_TRIPLET}` directories.

xandox:
@JackBoosY do not bother about config.log, I found one mac computer.

xandox:
Problem with x64-osx is in next:
1. CMake detect C and CXX compilers in (how I understand) in XCode build tools (`/Library/Developer/CommandLineTools/usr/bin/cc` and `/Library/Developer/CommandLineTools/usr/bin/c++`) and they work for cmake somehow.
2. `configure` script finds other compilers (probably `/usr/bin/gcc` or `/usr/bin/clang`) if not set CC and CXX variables explicitly.
3. For some reasons (I can't find out what reasons it is) `/Library/Developer/CommandLineTools/usr/bin/cc` can't build simple program like:
```c
#include <stdio.h>
int main() { return 0;}
```
with error: `test.c:1:10: fatal error: 'stdio.h' file not found`

so now I do not set CC and CXX int `vcpkg_conigure_make` if `CMAKE_HOST_APPLE` 

If someone know reasons for strange behavior for `/Library/Developer/CommandLineTools/usr/bin/cc` I will be very thankful for explanation.

xandox:
It still does not work as expected

JackBoosY:
Since I'm lack of android knowledge, I can't help you to fix that.

PR #15372, opened 2020-12-29
[onednn] Fix share path infix to match find_package() name

jacobkahn:
I forgot to check in this line when sending https://github.com/microsoft/vcpkg/pull/15068. My mistake.

Same triplets as before. No change.

My initial tests on the PR adding the port were using this change. I've confirmed that this works on `x64-linux`.

Yes.

### DIFF
--- a/ports/onednn/portfile.cmake
+++ b/ports/onednn/portfile.cmake
@@ -17,7 +17,9 @@ vcpkg_configure_cmake(
 )
 vcpkg_install_cmake()
 
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake)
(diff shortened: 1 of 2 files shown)


NancyLi1013 (commented):
Could you please also add `Port-Version: 1` to vcpkg.json file?

NancyLi1013 (commented):
Seems this port doesn't support x86, arm and uwp. Could you please add the unsupported message to the top of portfile.cmake?

`vcpkg_fail_port_install(ON_ARCH "x86" "arm" ON_TARGET "uwp")`

NancyLi1013 (approved):

vicroms:
Thanks for the PR!

PR #15383, opened 2020-12-30
[directxmath] port updated for cmake improvements

walbourn:
In PR 15210 I added directxmath as a new port using the CMake integration upstream. This works fine on Windows and installs for x64-linux, but trying to use it from Linux with ``find_package`` fails.

This is due to case-sensitivity in the Linux filesystem combined with a CamelCase ``PROJECT_NAME``, and the specific logic of ``find_package`` behavior (it needs ``<PackageName>Config.cmake`` or ``<lowercase-packagename>-config.cmake``). This is fixed upstream in [this PR](https://github.com/microsoft/DirectXMath/commit/b8e0ea25f776eb73935bf4ccba3129080ed4d995) and this port updates to points to this version.

Also moved the installed locations of DirectXMath headers into ``include/DirectXMath`` to avoid any accidental include order issues since those headers are in the Windows SDK as well.


PhoebeHui (changes requested):
@walbourn, thanks for the PR!

Could you also update the version in CONTROL file?

walbourn:
> Could you also update the version in CONTROL file?

I've not updated the version of the library. It's just fixed the CMake.

PhoebeHui (changes requested):
@walbourn, I suggest to use the version that we referenced in portfile.cmake.
https://github.com/microsoft/DirectXMath/releases

PhoebeHui (approved):
Thanks for your updates!

vicroms:
Thanks for the PR!

PR #15394, opened 2020-12-30
[IXWebsocket ] Upgraded to 11.0.4 version

ludekvodicka:
**Describe the pull request**


ports/ixwebsocket/portfile.cmake
```diff
@@ -27,6 +27,8 @@ vcpkg_configure_cmake(
 vcpkg_install_cmake()
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/lib/cmake)
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/lib/cmake)
```
vicroms, line 31: I suspect that IXWebSocket is now generating CMake targets. If that's the case I would recommend doing this instead:

```suggestion
vcpkg_install_cmake()
vcpkg_fixup_cmake_targets()

file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
```

vicroms (changes requested):

ports/ixwebsocket/portfile.cmake
ludekvodicka: Thanks for the update. I'm still new to vcpkg and CMake too, so I didn't know that. 
I just tested your proposal and it's working correctly.

Maybe one more question to you. I'm not able to figure out why PDB file isn't generated for IXWebsocket when built via vcpkg.  When the library is compiled directly via cmake, I have .PDB file correctly generated.

vicroms: Vcpkg uses [`/Z7` which embeds the debug symbols into the library files](https://docs.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-160#z7), so no `.pdb` files are generated for `lib` files. 

https://github.com/microsoft/vcpkg/blob/4a542c97fb1c5bf47d672da5950ef4c517285a5c/scripts/toolchains/windows.cmake#L28-L31

I gave the CMakeLists for IXWebsockets a quick look and noticed that it can only produce static libraries. So it needs a 

```cmake
```

at the top of the portfile.

Also, I would change the fixup line to `vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ixwebsocket TARGET_PATH share/${port})` to fix an issue where integration files end up installed in `share/ixwebsocket/ixwebsocket/*.cmake`.

ludekvodicka: Thanks for a perfect reply. I didn't know that and also didn't find it anywhere in the doc. This makes sense.

Regarding other changes, I'm not the original author of the port, I only wanted to help with the update. But I can implement the mentioned change. I will send a new PR.

ludekvodicka:
@vicroms Just added two mentioned changes. I hope it's ok.

vicroms (approved):

PhoebeHui (approved):
ixwebsocket[core,mbedtls,openssl,ssl]:x64-windows test passed.

vicroms:
Thanks for the PR!

ludekvodicka:
@vicroms Thanks for the guidance and assistance.

PR #15298, opened 2020-12-25
[openssl] Update to 1.1.1i

lebdron:
- Update openssl to 1.1.1i.
- Add support for arm64-osx triplet.

PR #15404, opened 2020-12-31
[many ports] Disambiguate Python usage.

Hoikas:
In #15298 there are CI failures that appear to result from ports using the wrong Python interpreter, potentially the one built by #14891. This fixes all cmake based ports to correctly specify the path to the python executable instead of relying on magical fairy dust for finding the acquired python in the path.

- What does your PR fix?
See above

No change

Yes

### DIFF
--- a/ports/azure-kinect-sensor-sdk/portfile.cmake
+++ b/ports/azure-kinect-sensor-sdk/portfile.cmake
@@ -13,8 +13,6 @@ vcpkg_from_github(
 )
 
 vcpkg_find_acquire_program(PYTHON3)
-get_filename_component(PYTHON3_DIR "${PYTHON3}" DIRECTORY)
-vcpkg_add_to_path("${PYTHON3_DIR}")
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     docs K4A_BUILD_DOCS
@@ -31,6 +29,7 @@ vcpkg_configure_cmake(
     -DBUILD_EXAMPLES=OFF
     -DWITH_TEST=OFF
     -DIMGUI_EXTERNAL_PATH=${CURRENT_INSTALLED_DIR}/include/bindings
+    -DPython3_EXECUTABLE=${PYTHON3}
 )
 
 vcpkg_install_cmake()
--- a/ports/azure-kinect-sensor-sdk/CONTROL
+++ b/ports/azure-kinect-sensor-sdk/CONTROL
@@ -1,6 +1,6 @@
 Source: azure-kinect-sensor-sdk
 Version: 1.4.0-alpha.0
-Port-Version: 6
+Port-Version: 7
 Homepage: https://github.com/microsoft/Azure-Kinect-Sensor-SDK
 Description: Azure Kinect SDK is a cross platform (Linux and Windows) user mode SDK to read data from your Azure Kinect device.
 Build-Depends: azure-c-shared-utility, glfw3, gtest, imgui, libusb, spdlog, cjson, ebml, libjpeg-turbo, matroska, libsoundio, libyuv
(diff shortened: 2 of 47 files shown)


PhoebeHui (approved):
LGTM, thanks for the PR!

JackBoosY:
I think we need to add `vcpkg-cmake-wrapper` to python to let all python-dependent ports use the default way to find python in vcpkg, instead of having to manually add python paths to these ports.

Hoikas:
As far as I can tell, Python is being found in the default way in all of these ports. I don't understand why CI in #15298 is using the `python.exe` installed by the python3 port. CMake by default looks in the registry, so I'm not sure how it's finding our python executable. The error does exhibit behavior that *might* occur with the python3 `vcpkg-cmake-wrapper` that will (hopefully) be introduced by #15221. While this solution is somewhat more verbose in terms of the cmake invocation, it eliminates a potential source for errors.

NancyLi1013:
@Hoikas

`graphqlparser `has the same problem with these ports on `python`.

```
-- Found PythonInterp: D:/installed/x64-windows/tools/python3/python.exe (found suitable version "3.9", minimum required is "2") 
CMake Error at CMakeLists.txt:29 (MESSAGE):
  Python 2 is required.
```

Hoikas:
Thanks! It looks like upstream used to have special handling for building the vcpkg port, but it was removed.

Hoikas:
Re-reading @JackBoosY's comment, I wonder if this is a problem with the PR CI environment? The last CI run in #15353 didn't seem to have these problems. @NancyLi1013 Did you see this failure with `graphqlparser` on CI or your local machine?

NancyLi1013:
@Hoikas 

I can reproduce the failures with `graphqlparser `on my local if I installed `python3 `port.

vicroms (approved):
LGTM

ports/azure-kinect-sensor-sdk/portfile.cmake
```diff
@@ -31,6 +29,7 @@ vcpkg_configure_cmake(
     -DBUILD_EXAMPLES=OFF
     -DWITH_TEST=OFF
     -DIMGUI_EXTERNAL_PATH=${CURRENT_INSTALLED_DIR}/include/bindings
+    -DPython3_EXECUTABLE=${PYTHON3}
```
PhoebeHui, line 32: Should this be Python_EXECUTABLE? and other 4 places.

Hoikas: It depends on the upstream CMakeLists. This port [uses](https://github.com/microsoft/Azure-Kinect-Sensor-SDK/blob/6dea37e560d6904180e403dfcd56346fb10863ab/CMakeLists.txt#L219) `find_package(Python3)`, so we have to use `Python3_EXECUTABLE`. If it used `find_package(Python)`, then we would use `Python_EXECUTABLE`. In the case of `find_package(PythonInterp)`, then we want `PYTHON_EXECUTABLE`.

Generally, ports tend to use `Python3` the most, followed by `PythonInterp`, and `Python` least frequently.

Hoikas:
It looks like the usage of the python3 port's Python is the result of #14317, specifically [this portion](https://github.com/microsoft/vcpkg/commit/b7056e9f1f34f18b6648b1f1d9c4e9d31f04e91c#diff-b93dff802371176cfb748f08cd609b9185a5003aba8f1240a78d4f8d8146ee4fR309). It seems to me like this part of the toolchain is in error because the tools in that directory are built for the target and therefore may not be usable on the host. That is why x64-windows-static checks are failing. The rest of the failures are, as mentioned in #15457 due to Python 3's python.exe hiding Python 2's python.exe.

Hoikas:
I am closing this in favor of fixing the issue with the python3 port's vcpkg-cmake-wrapper, as suggested by @JackBoosY, in #15220.

PR #15416, opened 2021-01-02
[libheif] update to 1.10.0

0xC0000054:
Update libheif to the latest version 1.10.0

This fixes an access violation when encoding an image with the dynamically linked version of x265.


JonLiu1993 (approved):

vicroms:
Thanks for the PR!

EwoutH:
I tried updating libheif to [v1.11.0](https://github.com/strukturag/libheif/tree/v1.11.0), but I couldn't figure out from which file the SHA512 hash is generated. Could you give a few pointers?

vicroms:
@EwoutH 

Change the REF to point to v1.11.0 in the `vcpkg_from_github` helper, then try to install the port (remove the old version if you have it installed).  The installation will fail due to SHA mismatch but it will tell you what the actual SHA should be.

PR #15419, opened 2021-01-02
[docs] fix "CMake script" -> "makefile" for vcpkg_(n)build_make

tueda:
This patch replaces "your CMake script" with "your makefile" in the documentation of `vcpkg_build_(n)make.md`, which is probably copied from `vcpkg_build_cmake.md`.

I am not sure if such a small document fix would be accepted as a PR; if not, please feel free to squash it with others.


PhoebeHui (approved):
LGTM, thanks for the PR!

PR #7202, opened 2019-07-09
[sqlite3-tool]Fix build error on arm/uwp platform.

JackBoosY:
Related #7192.

PR #15427, opened 2021-01-03
[msdfgen] New port 

Haeri:
**Describe the pull request**

New port for [msdfgen](https://github.com/Chlumsky/msdfgen)

Only static build

Yes


ports/msdfgen/portfile.cmake
```diff
@@ -0,0 +1,32 @@
...
+)
+
+vcpkg_configure_cmake(
+    SOURCE_PATH ${SOURCE_PATH}
+    PREFER_NINJA
+)
+
+vcpkg_install_cmake()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/msdfgen TARGET_PATH share/msdfgen)
```
NancyLi1013, line 20:
```suggestion
vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/msdfgen)
```
`TARGET_PATH` is not needed here. Since the path is default to _share/${PORT}_. Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_fixup_cmake_targets.md

ports/msdfgen/portfile.cmake
```diff
@@ -0,0 +1,32 @@
...
+
+# move exe to tools
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+    file(INSTALL "${CURRENT_PACKAGES_DIR}/bin" DESTINATION "${CURRENT_PACKAGES_DIR}/tools/" RENAME msdfgen)
```
NancyLi1013, line 24: We might consider to use `vcpkg_cpoy_tools()` to handle these executable files.

ports/msdfgen/portfile.cmake
```diff
@@ -0,0 +1,32 @@
...
+    file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/bin ${CURRENT_PACKAGES_DIR}/debug/bin)
+endif()
+
+# cleanup
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
+
+# license
+file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/msdfgen RENAME copyright)
```
NancyLi1013, line 32:
```suggestion
file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
```

NancyLi1013 (changes requested):

NancyLi1013:
The failures on x64-uwp and arm-uwp:

```
CheckForUnsupportedScenarios:
         The specified XPath query did not capture any nodes.
     6>D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.dir\package.appxManifest : error APPX0703: Manifest references file 'msdfgen-standalone.exe' which is not part of the payload. [D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj]
     6>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj" (default targets) -- FAILED.
     3>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\ALL_BUILD.vcxproj" (default targets) -- FAILED.
     1>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\install.vcxproj" (default targets) -- FAILED.
```
Could you please help look into this? @Haeri

If this port doesn't support uwp, please add `supports: !uwp` to vcpkg.json and add `vcpkg_fail_port_install(ON_TARGET "uwp")` to portfile.cmake.

ports/msdfgen/portfile.cmake
Haeri: @NancyLi1013  0ddd70718b588d79b4542269cac5b662266c668d resolves your suggestion. I manually wrote msdfgen instead of ${PORT} here.. not sure what the policy is but my reasoning is that the tool could be called something else like generator etc.. Let me know if I should change it.

Haeri:
> The failures on x64-uwp and arm-uwp:
> 
> ```
> CheckForUnsupportedScenarios:
>          The specified XPath query did not capture any nodes.
>      6>D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.dir\package.appxManifest : error APPX0703: Manifest references file 'msdfgen-standalone.exe' which is not part of the payload. [D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj]
>      6>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj" (default targets) -- FAILED.
>      3>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\ALL_BUILD.vcxproj" (default targets) -- FAILED.
>      1>Done Building Project "D:\buildtrees\msdfgen\x64-uwp-dbg\install.vcxproj" (default targets) -- FAILED.
> ```
> 
> Could you please help look into this? @Haeri
> 
> If this port doesn't support uwp, please add `supports: !uwp` to vcpkg.json and add `vcpkg_fail_port_install(ON_TARGET "uwp")` to portfile.cmake.

I have no idea what UWP is doing... Never dealt with it before. But from the error message it could be that it is expecting a `msdfgen-standalone.exe` since the solution has that name but the the binary actually being renamed to msdfgen.exe in the CMakeLists.txt:
```cmake
add_executable(msdfgen-standalone main.cpp)
set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive OUTPUT_NAME msdfgen)
```
Not sure if this is maybe even a cmake bug?

NancyLi1013:
@Haeri Thanks for your further investigation.

I think there is no need to rename `msdfgen-standalone` as `msdfgen`. We can update as the follows:

Update `set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive OUTPUT_NAME msdfgen)
`

 as

`set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive)`

ports/msdfgen/portfile.cmake
```diff
@@ -0,0 +1,31 @@
...
+)
+
+vcpkg_install_cmake()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/msdfgen)
+
+# move exe to tools
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+    vcpkg_copy_tools(TOOL_NAMES msdfgen AUTO_CLEAN)
+endif()
```
NancyLi1013, line 25:
```suggestion
vcpkg_copy_tools(TOOL_NAMES msdfgen-standalone AUTO_CLEAN)

if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/bin" "${CURRENT_PACKAGES_DIR}/debug/bin")
endif()
```

NancyLi1013 (changes requested):

Haeri:
> @Haeri Thanks for your further investigation.
> 
> I think there is no need to rename `msdfgen-standalone` as `msdfgen`. We can update as the follows:
> 
> Update `set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive OUTPUT_NAME msdfgen) `
> 
> as
> 
> `set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive)`

@NancyLi1013 I was hesitant to remove the renaming as the `msdfgen.exe` is referenced multiple time in the repository, namely in the `bin/example.bat` and `README.md`.

ports/msdfgen/portfile.cmake
Haeri: `AUTO_CLEAN` seems to be cleaning up after itself so calling `file(REMOVE_RECURSE` is not needed as far as I understand. At least when I call it, there is no bin folder remaining.

NancyLi1013:
@Haeri 
Thanks for your clarification about this point.

Sorry for my fault. We should not change the property in this case. So just keep it as before.

For the failures on uwp, which is caused by the executable name, I also have no idea about this issue.


@strega-nil Could you please help take a look about this issue?

We noticed that there is still `msdfgen-standalone.exe` instead of `msdfgen.exe` in package.appxManifest on uwp platform.

```
F:\15427\vcpkg\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.dir\package.appxManifest : error APPX0703: Manifest references file 'msdfgen-standalone.exe' which is not part of the payload. [F:\15427\vcpkg\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj]
     6>Done Building Project "F:\15427\vcpkg\buildtrees\msdfgen\x64-uwp-dbg\msdfgen-standalone.vcxproj" (default targets) -- FAILED.
```

ports/msdfgen/portfile.cmake
NancyLi1013: Since this is static library, there is no dlls generated in bin folder. So you're right.

Haeri:
#7202 Seems to have had the same issue, so maybe @JackBoosY knows more?

NancyLi1013:
Thanks again for your feedback. @Haeri 

We might consider to drop the property of renaming this executable. Namely remove `set_target_properties(msdfgen-standalone PROPERTIES ARCHIVE_OUTPUT_DIRECTORY archive OUTPUT_NAME msdfgen)`.

Besides this, I cannot figure out other solution to avoid this problem.

Haeri:
@NancyLi1013 I got around 90 code hits on github.com for `msdfgen.exe` so I think maybe it would be better to drop UWP instead. Or maybe there is a way to create a symlink from `msdfgen-standalone.exe` to `msdfgen.exe` so UWP doesn't complain about `msdfgen-standalone.exe` not existing and people relying on the existence of `msdfgen.exe` also being happy?

Haeri:
Interesting... this worked.. I was going to solve this with a symlink but seeing that they are very finicky especially on windows and the binary being only 200kb I think this could be a viable solution?

JackBoosY:
@Haeri I think that's okay, because the uwp tools can't be used directly. So please add it as a feature, make a patch to fix that.

Haeri:
@JackBoosY Sorry I did a lot of changes recently... not sure which one you are referring to... Also I don't know what you mean with `add it as a feature`?

JackBoosY:
@Haeri You can add a cmake option `MSDFGEN_BUILD_TOOLS` (`(option(MSDFGEN_BUILD_TOOLS ) "Build tools" OFF)`) to its CMakeLists.txt, and add judgment `if (MSDFGEN_BUILD_TOOLS)` to `add_executable(..)`.

Haeri:
@JackBoosY There is already an option natively given in the CMake. Its called `MSDFGEN_BUILD_MSDFGEN_STANDALONE`:
https://github.com/Chlumsky/msdfgen/blob/9af250c7d6780a41dcaf536c05e3e1987a1bdcd7/CMakeLists.txt#L102

JackBoosY:
@Haeri You shoud:
1. Add feature `tools` in _CONTROL_ file.
2. In `portfile.cmake`:
```cmake
vcpkg_from_github(...)

set(BUILD_TOOLS OFF)
if ("tools" IN_LIST FEATURES)
    if (VCPKG_TARGET_IS_UWP)
        message("Tools couldn't be built on UWP, disable it automatically.")
    else()
        set(BUILD_TOOLS ON)
    endif()
endif()
...
vcpkg_configure_cmake(
...
    options
        -DMSDFGEN_BUILD_MSDFGEN_STANDALONE=${BUILD_TOOLS}
)
...
```

ports/msdfgen/portfile.cmake
```diff
@@ -0,0 +1,43 @@
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
JackBoosY, line 1: Any reason to do that?
I think you should leave a comment here.

JackBoosY (changes requested):

ports/msdfgen/portfile.cmake
Haeri: The author didn't do any dllexport so the dynamic target fails.

JackBoosY, line 1:
```suggestion
# No symbols are exported in msdfgen source
vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```

JackBoosY (changes requested):

JackBoosY (approved):

JackBoosY:
@Haeri Ready for review now?

NancyLi1013 (approved):

PR #15445, opened 2021-01-04
[opencl] Fix dynamic build on linux

dweckmann:
**Describe the pull request**

- What does your PR fix?

the libraries installation paths are wrong for dynamic path. This pr just fix that

Only linux dynamic build is impacted by this PR


PhoebeHui (changes requested):
@dweckmann, thanks for the PR!

Could you update the 'Port-Version' to 6 in CONTROL file?

dweckmann:
> @dweckmann, thanks for the PR!
> 
> Could you update the 'Port-Version' to 6 in CONTROL file?

oups... I always forgot to do this :) It should be good now.

Thanks for review !

PhoebeHui (approved):

PR #15457, opened 2021-01-05
[python2] Add feature tools

JackBoosY:
Add feature `tools` to build and install python executables.
To fix the issue:
```
Traceback (most recent call last):
  File "D:\buildtrees\cppcms\src\20a4ad93d4-26c9d29282.clean\bin\cppcms_tmpl_cc", line 14, in <module>
    import StringIO
ModuleNotFoundError: No module named 'StringIO'
```

### DIFF
--- a/ports/cppcms/CONTROL
+++ b/ports/cppcms/CONTROL
@@ -1,6 +1,7 @@
 Source: cppcms
-Version: 1.2.1-1
+Version: 1.2.1
+Port-Version: 2
 Homepage: https://github.com/artyom-beilis/cppcms
 Description: CppCMS is a Free High Performance Web Development Framework (not a CMS) aimed at Rapid Web Application Development
-Build-Depends: icu, pcre, openssl, zlib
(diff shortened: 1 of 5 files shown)


JackBoosY:
We should consider to remove `PYTHON2` and `PYTHON3` in `vcpkg_find_acquire_program`.

JackBoosY:
We should also refer to `manual-link` to create a separate tool folder to avoid adding their path to `CMAKE_PREFIX_PATH`.

Hoikas:
The python2 port doesn't support macOS or linux AFAIK. Further, the Python executables built by static triplets have undesirable edge cases on Windows. It might be better to wait until after #15424 for this.

Hoikas:
I've pushed an update to the python3 vcpkg-cmake-wrapper in #15221 that should fix this problem.

JackBoosY:
Since #15221 has partially solved the python issue, I will close this PR now.

PR #15460, opened 2021-01-05
update trantor to 1.1.1

kkzi:
**update trantor to 1.1.1**

none

x64-windows

yes

### DIFF
--- a/ports/trantor/CONTROL
+++ b/ports/trantor/CONTROL
@@ -1,5 +1,5 @@
 Source: trantor
-Version: 1.1.0
+Version: 1.1.1
 Homepage: https://github.com/an-tao/trantor
 Description: A non-blocking I/O cross-platform TCP network library, using C++14.
 Build-Depends: openssl, c-ares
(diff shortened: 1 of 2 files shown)


NancyLi1013 (commented):
Why do you rewrite these two files?

kkzi:
> Why do you rewrite these two files?

update trantor version from 1.1.0 to 1.1.1

NancyLi1013:
You just need to update the changes instead of rewriting all contents.

For example, you can only update Version field 1.1.0 as 1.1.1  in CONTROL file. Since there are no any changes about other fields.

kkzi:
> You just need to update the changes instead of rewriting all contents.
> 
> For example, you can only update Version field 1.1.0 as 1.1.1 in CONTROL file. Since there are no any changes about other fields.

done

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your PR @kkzi.

PR #15484, opened 2021-01-06
[paraview] fix building with `mpi` feature enabled

albertziegenhagel:
This PR fixes building `paraview` with the `mpi` feature enabled. To my knowledge that feature has never been tested and thus was actually never working before.

The following changes where made:
  - Apply a fix to `vtk` so that some headers get installed that are required by ParaView.
  - Enable the `FiltersParallelFlowPaths` module in `vtk` when both `paraview` and `mpi` features are enabled.
  - Download the source for `IceT` before building ParaView.

~~Additionally I have added the `mpi` feature to the `vcpkg-ci-paraview` test port. This should make sure that the combination will be tested in the future. I am not sure what the policy about such testing ports is, so let me know if you would prefer that I revert that change.~~

edit: I had to disable the `mpi` feature of the `vcpkg-ci-paraview` test port again, since it enables the `parallel` feature of `hdf5` which is in conflict with the `cpp` feature that is required by some other port.


ports/vtk/fix-install-rendering-parallel-header.patch
```diff
@@ -0,0 +1,11 @@
+--- a/Rendering/Parallel/CMakeLists.txt	Wed Jun 24 18:25:19 2020
++++ b/Rendering/Parallel/CMakeLists.txt	Tue Aug 18 15:29:13 2020
+@@ -34,7 +34,7 @@
+ vtk_module_add_module(VTK::RenderingParallel
+   CLASSES ${classes}
+   SOURCES ${sources}
+-  PRIVATE_HEADERS ${private_headers})
++  HEADERS ${private_headers})
```
JackBoosY, line 8: Is this a upstream bug?

JackBoosY (changes requested):

JackBoosY:
I will test it locally.

ports/vtk/fix-install-rendering-parallel-header.patch
Neumann-A: The question is more if all headers are required or only a few. Upstream paraview is tightly entangled with VTK although they try to minimize those deps they still happen. There are also some private headers which paraviews needs of vtk which get manually copied in the vtk portfile.

albertziegenhagel: Oh it is just a single header `vtkCompositeZPassFS.h` which is used by ParaView here:
https://gitlab.kitware.com/paraview/paraview/-/blob/510ee33bbc32de0882dd753fd8ceb61eb2a309ed/Remoting/Views/vtkIceTCompositePass.cxx#L50

I did totally overlook the part in the VTK portfile that explicitly copied some headers for ParaView...

I was thinking whether to open a MR in upstream VTK about that issue, but was not feeling to good about it, since building ParaView against an external VTK installation is not officially supported and I wasn't sure whether they will like those changes.

I will change this PR to copy the header similar to the other ones and remove applying the patch. Thanks for the hint!

Neumann-A: Open up a issue in ParaView about the private VTK header. I know that at least Ben Boeckel wants to have the two things cleanly separated. See https://gitlab.kitware.com/paraview/paraview/-/issues/19515

albertziegenhagel: I have force-pushed the mentioned changes here and have opened an issue over at ParaView: https://gitlab.kitware.com/paraview/paraview/-/issues/20425

JackBoosY: @albertziegenhagel The official replied to you.

albertziegenhagel: @JackBoosY They did reply, but did not not yet answer the question where they would like the issues to be solved (either VTK can install the required headers or ParaView should stop using them). While making VTK install the headers is probably easy to be implemented, making ParaView not using them will most likely need a deeper insight into ParaView itself than what I currently have.

As for this MR I do think the current approach to copy the additional header is the best we can do at the moment.

JackBoosY:
Testing...

JackBoosY:
When building `vtk[openvr,qt,python,atlmfc,paraview,opengl]`:
```
CMake Error at ThirdParty/hdf5/CMakeLists.txt:26 (message):
  An external MPI-aware HDF5 requires that VTK be built with MPI support as
  well.
```

Does that related to your changes?

albertziegenhagel:
@JackBoosY: Sorry for the late reply. The error is not related to this PR. This happens when `hdf5[parallel]` is already installed and you are trying to install `vtk` without the `mpi` feature.

See [this comment in another PR](https://github.com/microsoft/vcpkg/pull/12971#issuecomment-679002803) and the following discussion.

JackBoosY:
Can you please solve the file conflicts by revert vtk.json, merge this branch from master and regenerate the version record?

Thanks.

albertziegenhagel:
> Can you please solve the file conflicts by revert vtk.json, merge this branch from master and regenerate the version record?

Done!

JackBoosY:
```
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
```
Seems your changes is incorrect. An extra slash.

JackBoosY:
Pinging @albertziegenhagel for response. Is work still being done for this PR?

PhoebeHui:
Closing this PR since it seems that no progress is being made, Please reopen if work is still being done.

PR #15200, opened 2020-12-18
[tool-meson] Rev vcpkg's meson build

RichLogan:
This is a rev of vcpkg's meson version to fix an upstream regression introduced in https://github.com/microsoft/vcpkg/pull/12927 that causes issues with the building of dav1d: https://github.com/microsoft/vcpkg/pull/15009.

I'm not a fan of tracking an unreleased version but wanted to fix the existing issue now.

PR #15491, opened 2021-01-07
[vcpkg baseline][poppler] Fix issue with using libcurl.

Hoikas:
This fixes an issue where poppler will attempt to build with libcurl. Libcurl's cmake config file does not properly set its CMake policies so certain `if()` calls fail, causing a cascade failure in the poppler. Should fix issue seen in #15200.

- What does your PR fix?
See above

No change

Yes


JackBoosY:
Can we completely solve this issue instead of simply disabling curl?

Hoikas:
Sure. This just seemed like the quickest fix because poppler does not list curl as a dependency.

JackBoosY (approved):

JackBoosY:

ras0219-msft:
Thanks for the fix!

PR #15500, opened 2021-01-07
[openigtlink] Fix cmake config location

dweckmann:
**Describe the pull request**

- What does your PR fix? 

Currently the openigtlink config files are generated in a igtl-3.0 subdirectory making the standard cmake `find_package(OpenIGTLink CONFIG REQUIRED)` fail. Even worse, the `OpenIGTLinkConfig.cmake` contained wrong hard coded path, set during build time. The package was just unusable.

This PR address this issues.

### DIFF
--- a/ports/openigtlink/portfile.cmake
+++ b/ports/openigtlink/portfile.cmake
@@ -11,14 +11,12 @@ vcpkg_configure_cmake(
     PREFER_NINJA
     OPTIONS
          -DBUILD_TESTING=OFF
+         -DOpenIGTLink_INSTALL_PACKAGE_DIR=share/${PORT}
 )
 
 vcpkg_install_cmake()
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
-
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/igtl/cmake)
-
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
 
 vcpkg_copy_pdbs()
--- a/ports/openigtlink/CONTROL
+++ b/ports/openigtlink/CONTROL
@@ -1,3 +1,5 @@
 Source: openigtlink
 Version: 3.0
+Port-Version: 1
+Homepage: https://github.com/openigtlink/OpenIGTLink
 Description: OpenIGTLink is an open-source network communication interface specifically designed for image-guided interventions.
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -4242,7 +4242,7 @@
     },
     "openigtlink": {
       "baseline": "3.0",
-      "port-version": 0
+      "port-version": 1
     },
     "openimageio": {
       "baseline": "2.2.10.0",
--- a/versions/o-/openigtlink.json
+++ b/versions/o-/openigtlink.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "6f814a06b22cc250df4eb31553c6df7a803567b7",
+      "version-string": "3.0",
+      "port-version": 1
+    },
+    {
       "git-tree": "df766287f1c9c4d379143fab2456a64ae7ae8452",
       "version-string": "3.0",
       "port-version": 0


NancyLi1013 (changes requested):

ports/openigtlink/portfile.cmake
```diff
@@ -11,14 +11,12 @@ vcpkg_configure_cmake(
...
     OPTIONS
          -DBUILD_TESTING=OFF
+         -DOpenIGTLink_INSTALL_PACKAGE_DIR=share/${PORT}
 )
 
 vcpkg_install_cmake()
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
-
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/igtl/cmake)
```
NancyLi1013, line 20: Why do you remove this?

NancyLi1013 (changes requested):

ports/openigtlink/portfile.cmake
dweckmann: > Why do you remove this?

Because it seems unneeded now as the cmake configs are generated at the right place, with the right content. Should I still call it ?

NancyLi1013: If there is absolute path in cmake configs or targets, we should still call it. If not, it's unnecessary.
Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_fixup_cmake_targets.md

NancyLi1013:
Have you tested the usage? Does it work fine? It seems that there is something wrong with include path. 

I tried to use it, the header` igtl/igtlSocket.h` cannot be found.

Could you please help check?

dweckmann:
> Have you tested the usage? Does it work fine? It seems that there is something wrong with include path.
> 
> I tried to use it, the header` igtl/igtlSocket.h` cannot be found.
> 
> Could you please help check?

We use #include <igtlSocket.h>. Upstreams install it on /usr/include/openigtl/. So I'm not sure what is the right prefix. Anyway, before my patch doesn't change the location of the header (same as before). Not sure what to do...

ports/openigtlink/portfile.cmake
dweckmann: There is paths which were definitively wrong before my patch. Now, it should point to relative directory.

dweckmann:
> Have you tested the usage? Does it work fine? It seems that there is something wrong with include path.
> 
> I tried to use it, the header` igtl/igtlSocket.h` cannot be found.
> 
> Could you please help check?

I've just double checked. In fact, openigtlink cmake config reports `D:/Work/vcpkg/installed/x64-windows/include/igtl` as `OpenIGTLink_INCLUDE_DIRS` variable. Therefore, I cannot see how `#include <igtl/igtlSocket.h>` could work. As I didn't touch upstream cmake, I assume using the provided cmake config from upstream would force you to write `#include <igtlSocket.h>`, if you don't add by yourself  `..vcpkg/installed/x64-windows/include` to the target include directories. We could do that by patching upstream openigtl cmake files, but I'm not sure it is the right thing to do.

NancyLi1013:
Generally, we use the header like this if it is installed in _D:/Work/vcpkg/installed/x64-windows/include/igtl_.

`#include <igtl/igtlSocket.h>`

If there is no `igtl `subdirectory, we can use `#include <igtlSocket.h>`.

dweckmann:
> Generally, we use the header like this if it is installed in _D:/Work/vcpkg/installed/x64-windows/include/igtl_.
> 
> `#include <igtl/igtlSocket.h>`
> 
> If there is no `igtl `subdirectory, we can use `#include <igtlSocket.h>`.

Well, then we must patch upstream, as I didn't touch upstream cmake. On a linux system, the headers are in _/usr/include/igtl_, so it works as is if you use the provided cmake config or not.

I'm a bit confused about how this could have worked before ? The config file were clearly wrong, and still `OpenIGTLink_INCLUDE_DIRS` pointed to  something like _.../include/igtl_ .  It seems wrong to make `OpenIGTLink_INCLUDE_DIRS` point to _D:/Work/vcpkg/installed/x64-windows/include_

dweckmann:
According to official samples and tutorials directly from http://openigtlink.org/tutorials/ and specifically from https://www.na-mic.org/wiki/OpenIGTLink/Library/Tutorial, all use includes without "igtl" prefix :

Source code from http://svn.na-mic.org/NAMICSandBox/trunk/OpenIGTLink/Examples:

```c++
#include "igtlOSUtil.h"
#include "igtlMessageHeader.h"
#include "igtlTransformMessage.h"
#include "igtlImageMessage.h"
#include "igtlServerSocket.h"
#include "igtlStatusMessage.h"
#include "igtlPositionMessage.h"
...
```

So, I really guess `<igtl/igtlSocket.h>` works by side effect on unix system, but the samples says `<igtlSocket.h>`. 

@NancyLi1013, what do you think ?

NancyLi1013 (approved):

NancyLi1013:
@dweckmann 
Thanks for your further investigation about this issue.

I also checked that `igtl `has been added to `OpenIGTLink_INCLUDE_DIRS`, so we can use `#include <igtlSocket.h>` directly.

LGTM now, thanks for your efforts.

strega-nil:
Same thing here @dweckmann

NancyLi1013:
@dweckmann 

Could you please merge master to this PR?

PR #15506, opened 2021-01-07
[libraw] Fix non-determinant openmp behavior.

Hoikas:
This fixes an issue by which libraw may be built with or without support for `openmp` depending on the currently installed packages. This has been observed to cause a non-determinant build error in the `openimageio` port. Since `openmp` is not listed as a dependency of libraw, I have explicitly disabled it in the portfile to prevent these kinds of errors.

- What does your PR fix?
See above.

No change.

Yes

### DIFF
--- a/ports/libraw/portfile.cmake
+++ b/ports/libraw/portfile.cmake
@@ -26,6 +26,7 @@ vcpkg_configure_cmake(
     OPTIONS
         -DINSTALL_CMAKE_MODULE_PATH=${CURRENT_PACKAGES_DIR}/share/libraw
         -DCMAKE_DEBUG_POSTFIX=d
+        -DENABLE_OPENMP=OFF
 )
 
 vcpkg_install_cmake()
--- a/ports/libraw/CONTROL
+++ b/ports/libraw/CONTROL
@@ -1,5 +1,6 @@
 Source: libraw
 Version: 201903-3
+Port-Version: 1
 Build-Depends: lcms, jasper
 Homepage: https://www.libraw.org
 Description: raw image decoder library


ports/libraw/portfile.cmake
```diff
@@ -26,6 +26,7 @@ vcpkg_configure_cmake(
     OPTIONS
         -DINSTALL_CMAKE_MODULE_PATH=${CURRENT_PACKAGES_DIR}/share/libraw
         -DCMAKE_DEBUG_POSTFIX=d
+        -DENABLE_OPENMP=OFF
```
JackBoosY, line 29: Can we solve this completely instead of disabling it?

JackBoosY (changes requested):

ports/libraw/portfile.cmake
Hoikas: I'm not sure how to fix this in a clean way... LibRaw doesn't have a built in find module or export a cmake config, so I'm not sure how to add the openmp interface link libraries.

JackBoosY: I'll take a look later.

JackBoosY:
@Hoikas Can you provide me the repro step or the failure logs?

Hoikas:
IIRC when testing #15509 on linux, `./vcpkg install openimageio[libraw]` would trigger it sometimes.

JackBoosY:
@Hoikas I tried to use #15509 changes to repro this issue, but failed.
Is this a usage issue or build issue?

Hoikas:
It happens when building ports that consume libraw because it does not expose any interface libraries. I think to repro, you will need to have openmp installed first, then install libraw, then build a port that uses libraw (eg openimageio[libraw]) for the failure to happen.

JackBoosY:
@Hoikas Same result after using the step you provided: success.

JackBoosY:
@Hoikas  Ping for response.

Hoikas:
I just tried again and was unable to reproduce this against master.

JackBoosY:
@Hoikas  Why close this PR?

PR #15530, opened 2021-01-08
[curl] Use cmake_minimum_required() to ensure if(ON) works

ras0219:
This fixes a baseline issue with poppler discovered in #15424


Hoikas:
Duplicates #15491

PR #15538, opened 2021-01-09
[vcpkg] Show the cache file name at download time to provide more ancillary information in non-ARIA2 mode.

RonxBulld:
Show the cache file name at download time to provide more ancillary information in non-ARIA2 mode.

- What does your PR fix?
We often encounter download errors, and it is difficult to set the correct cache file name when copying files to local using other download tools.

This PR just adds the message content.

Yes.


PhoebeHui (approved):
LGTM, thanks for the PR!

PR #15571, opened 2021-01-11
[blas/lapack] force BLA_VENDOR in wrapper

Neumann-A:
to make sure we are actually picking up the selected blas/lapack from vcpkg.


JackBoosY:

PR #15603, opened 2021-01-12
[outcome] Replace Outcome single header based port with full fat cmake install port

### LINKED ISSUE #15563: Whether vcpkg will now accept LLFIO or not?
Historically, LLFIO did not operate a formal release process other than "whatever master branch currently is", and so was unsuitable for distribution by vcpkg.

Seeing as a user asked (https://github.com/ned14/llfio/issues/69), and that as of Nov 2020 LLFIO does now issue formal versioned releases with prebuilt binaries (https://github.com/ned14/llfio/releases), would this situation have changed?

I suspect the remaining problem will be LLFIO's dependencies Outcome and QuickCppLib. Outcome does have a formal versioned release process which tracks the Boost release cycle, and LLFIO is pinned to the last release of Outcome. However, QuickCppLib remains on a pure "whatever master branch currently is" release process.

ned14:
- What does your PR fix?

The last time vcpkg support for Outcome was attempted was in 2017, where a disagreement about how it ought to be implemented led to no support. This led to users submitting a single file include based edition of Outcome as a stand in. Given that four years have passed, and both vcpkg and Outcome have considerably matured and become far more popular with users, this PR tries again to make vcpkg's Outcome install based i.e. "full fat".

Every conceivable triplet should be supported. Outcome is very widely used in the C++ ecosystem across some very unusual platforms.

Almost. We do bundle an embedded library quickcpplib. This was already being silently bundled by the single file include edition, because the single file include contains quickcpplib, so in this sense nothing has changed. Quickcpplib is not intended for public usage outside this author's own libraries, and me not wanting it to be publicly usable was one of the primary reasons the 2017 submission of Outcome to vcpkg was not possible.

Related issues:

- https://github.com/microsoft/vcpkg/issues/15563 (Whether vcpkg will now accept LLFIO or not?)
- https://github.com/ned14/llfio/issues/69 (Full fat install based Outcome is a hard prerequisite for LLFIO to enter vcpkg)


ned14:
Now that all the CI tests have passed, @PhoebeHui @Pospelove both of you worked on bringing Outcome to vcpkg (thanks for that!). Can I ask that you review this PR, as obviously it'll be a breaking change for you, to make sure it's good?

Pospelove:
LGTM. Good job.

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
...
-file(INSTALL ${OUTCOME_HEADERS} DESTINATION ${CURRENT_PACKAGES_DIR}/include)
+vcpkg_install_cmake()
+
+# Looks like vcpkg_fixup_cmake_targets() can't be run twice, so do this by hand
+file(RENAME "${CURRENT_PACKAGES_DIR}/lib/cmake/quickcpplib" "${CURRENT_PACKAGES_DIR}/share/quickcpplib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/cmakelib" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/cmakelib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/scripts" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/scripts")
+
+# Must come AFTER the above, as it appears to hose the lib/cmake directory
+vcpkg_fixup_cmake_targets(CONFIG_PATH "lib/cmake/${PORT}" TARGET_PATH "share/${PORT}")
```
JackBoosY, line 37:
```suggestion
file(RENAME "${CURRENT_PACKAGES_DIR}/share/cmakelib" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/cmakelib")
file(RENAME "${CURRENT_PACKAGES_DIR}/share/scripts" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/scripts")

vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/quickcpplib TARGET_PATH share/quickcpplib)
vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/outcome)
```
In fact it can be, but only for `config.cmake` or `targets.cmake`.

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
...
+  URLS "https://github.com/ned14/outcome/releases/download/all_tests_passed_${OUTCOME_GITCOMMIT}/outcome-v2-all-sources-${OUTCOME_GITCOMMIT}.tar.xz"
+  FILENAME "outcome-v2-all-sources-${OUTCOME_GITCOMMIT}.tar.xz"
+  SHA512 ${OUTCOME_HASH}
+)
+vcpkg_extract_source_archive_ex(
+  SKIP_PATCH_CHECK
+  OUT_SOURCE_PATH SOURCE_PATH
+  ARCHIVE "${ALL_SOURCES_TARBALL}"
+  REF "outcome-${OUTCOME_GITCOMMIT}"
+)
```
JackBoosY, line 17:
```suggestion
vcpkg_from_github(
    OUT_SOURCE_PATH SOURCE_PATH
    REPO ned14/outcome
    REF all_tests_passed_${OUTCOME_GITCOMMIT}
    SHA512 ${OUTCOME_HASH}
    HEAD_REF develop
)
```

Can we use `vcpkg_from_github` instead?

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
-# header-only library
-vcpkg_from_github(
-    OUT_SOURCE_PATH SOURCE_PATH
-    REPO ned14/outcome
-    REF 34f3bd55e2bcaf246cb79efe64a5600e89b91b66 #v2.1.3
-    SHA512 83eba50e2095e7c768dacb3af5f82db117c3451f1d5bc2f73d716608d56f7b73006ec33d0f3842fdefd076f0e82b72ece5777868712f75e83eac93aa8adf351c
-    HEAD_REF develop
+set(OUTCOME_VERSION v2.2.0)
```
JackBoosY, line 1:
```suggestion
```

Only defined here, not used.

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
...
+# Looks like vcpkg_fixup_cmake_targets() can't be run twice, so do this by hand
+file(RENAME "${CURRENT_PACKAGES_DIR}/lib/cmake/quickcpplib" "${CURRENT_PACKAGES_DIR}/share/quickcpplib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/cmakelib" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/cmakelib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/scripts" "${CURRENT_PACKAGES_DIR}/share/quickcpplib/scripts")
+
+# Must come AFTER the above, as it appears to hose the lib/cmake directory
+vcpkg_fixup_cmake_targets(CONFIG_PATH "lib/cmake/${PORT}" TARGET_PATH "share/${PORT}")
+
+file(REMOVE_RECURSE
+    "${CURRENT_PACKAGES_DIR}/debug"
```
JackBoosY, line 40: Is outcome still a header only library now? What files are there in _debug_ folder?

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
...
+)
+
+# Fixup the exports files with our embedded quickcpplib unusualness
+file(READ "${CURRENT_PACKAGES_DIR}/share/quickcpplib/quickcpplibExports.cmake" quickcpplibExports)
+string(REPLACE "get_filename_component(_IMPORT_PREFIX \"\${_IMPORT_PREFIX}\" PATH)\nget_filename_component(_IMPORT_PREFIX \"\${_IMPORT_PREFIX}\" PATH)\nget_filename_component(_IMPORT_PREFIX \"\${_IMPORT_PREFIX}\" PATH)" "get_filename_component(_IMPORT_PREFIX \"\${_IMPORT_PREFIX}\" PATH)\nget_filename_component(_IMPORT_PREFIX \"\${_IMPORT_PREFIX}\" PATH)" quickcpplibExports "${quickcpplibExports}")
+file(WRITE "${CURRENT_PACKAGES_DIR}/share/quickcpplib/quickcpplibExports.cmake" "${quickcpplibExports}")
+
+file(READ "${CURRENT_PACKAGES_DIR}/share/outcome/outcomeExports.cmake" outcomeExports)
+string(REPLACE "get_filename_component(_DIR \"\${CMAKE_CURRENT_LIST_FILE}\" PATH)" "get_filename_component(_DIR \"\${CMAKE_CURRENT_LIST_FILE}\" PATH)\ninclude(\"\${_DIR}/../quickcpplib/quickcpplibExports.cmake\")" outcomeExports "${outcomeExports}")
+file(WRITE "${CURRENT_PACKAGES_DIR}/share/outcome/outcomeExports.cmake" "${outcomeExports}")
```
JackBoosY, line 51: Why change this? That's fixed by `vcpkg_fixup_cmake_targets` and it should be correct.

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,55 @@
...
+
 
-file(INSTALL ${SOURCE_PATH}/Licence.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+file(INSTALL "${SOURCE_PATH}/Licence.txt" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME "copyright")
+file(INSTALL "${CMAKE_CURRENT_LIST_DIR}/usage" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}")
```
JackBoosY, line 55: Since outcome provides `outcomeConfig.cmake` / `quickcpplibConfig.cmake`, we don't need this usage because vcpkg will detect the target and display the usage by default.

JackBoosY (changes requested):

ports/outcome/portfile.cmake
ned14: I tried using `vcpkg_fixup_cmake_targets()` twice exactly as you have it, as that seems more logical. But as I mentioned in the comment, that function hoses the lib/cmake directory, which causes the second call to fail.

ports/outcome/portfile.cmake
ned14: We cannot. The all sources tarball contains all the dependencies, and more importantly, the precise versions of those for which Outcome's CI pass succeeded.

ports/outcome/portfile.cmake
ned14: np

ports/outcome/portfile.cmake
ned14: Outcome is indeed a header only library.

They contain various stuff generated by either vcpkg or cmake install. It looked to me trash left over from copying i.e. stuff gets dumped in there, stuff gets copied out elsewhere, but because the directories weren't entirely empty they don't get deleted. I'd say you're safe deleting it, and besides vcpkg complains if you don't.

ports/outcome/portfile.cmake
ned14: As mentioned above, it appears that you can run `vcpkg_fixup_cmake_targets` exactly once. Hence me repeating the transformation, effectively speaking.

ports/outcome/portfile.cmake
ned14: Oh that's nice. Though I'd like to see if it detects the right one. I'll remove this for now, thanks.

JackBoosY:
I will help you to improve these changes later.

ports/outcome/portfile.cmake
JackBoosY: Which quickcpplib version (or submission ID) should it be?

BurningEnlightenment:
I would love to see the ABI permutation disabled (either by default or by user request via [library feature](https://github.com/microsoft/vcpkg/blob/2020.11-1/docs/specifications/feature-packages.md#a-proposed-user-experience)). Installing multiple library versions is prohibited under the vcpkg dependency model and so I think there is little (no?) gain from ABI permutation at the cost of more unreadable symbols. What do you think?

JackBoosY:
> I would love to see the ABI permutation disabled (either by default or by user request via [library feature](https://github.com/microsoft/vcpkg/blob/2020.11-1/docs/specifications/feature-packages.md#a-proposed-user-experience)). Installing multiple library versions is prohibited under the vcpkg dependency model and so I think there is little (no?) gain from ABI permutation at the cost of more unreadable symbols. What do you think?

@BillyONeal what do you think about?

JackBoosY:
@ned14  Can you test if my changes is correct?

ned14:
> > I would love to see the ABI permutation disabled (either by default or by user request via [library feature](https://github.com/microsoft/vcpkg/blob/2020.11-1/docs/specifications/feature-packages.md#a-proposed-user-experience)). Installing multiple library versions is prohibited under the vcpkg dependency model and so I think there is little (no?) gain from ABI permutation at the cost of more unreadable symbols. What do you think?
> 
> @BillyONeal what do you think about?

As current Outcome does not have a stable ABI until 2022, per-SHA ABI permutation is used to prevent collision between multiple copies of Outcome being brought into the same process. Some libraries may use Outcome from vcpkg, others may use Outcome standalone, still others may use a customised Outcome (there are quite a few forks going around because I refused certain changes, so various people stepped up with forks, and they haven't changed the namespace name).

In 2022 it is expected that we shall declare the ABI stable, and from thenceforth it shall be CI tested per commit to ensure it never, ever, changes again. At that point the ABI permutation will be dropped in the stable branch. If Outcome enters vcpkg, I should imagine that as part of the ordinary ports upgrade, vcpkg then gets the ABI stable Outcome.

ned14:
> @ned14 Can you test if my changes is correct?

I've refactored your changes a bit further, see what you think. Some notes:

- I think consolidating all the SHAs into a single file with instructions of how to upgrade them as a set is better UX for port maintainers.

- `usage` appears to be ignored if the port is targets based. Right now it prints:

    The package outcome:x86-windows provides CMake targets:

    find_package(outcome CONFIG REQUIRED)
    target_link_libraries(main PRIVATE outcome::hl)

    find_package(quickcpplib CONFIG REQUIRED)
    target_link_libraries(main PRIVATE quickcpplib::hl)

 This isn't useful, the latter find package is not necessary and is confusing, but `usage` does not appear to override.

- I added vcpkg features so external users can choose for the internally chosen `byte-lite` and `gsl-lite` to be the vcpkg versions. I have no idea how to discover what to copy, and my attempts to get cmake to go search the vcpkg directories quickly became complex. Indeed, if we could even figure out the vcpkg port SHA, that could work too, and I didn't want cmake to go yanking port files from vcpkg's github and scanning the contents with a regex match. I don't think it wise to just use the vcpkg edition of those by default, in the past upstream changes have broken Outcome and it took some months to get upstream to change how they did things in a way that Outcome could return to trunk. Sticking with the versions published by Outcome's CI as "known good" is safer, in my opinion.

- In other package managers they have the ability to run the test suite for a package in order to test if the chosen dependency versions from the package manager are compatible and non breaking. I don't see a facility for that here, but if there is one, I wouldn't object to Outcome using vcpkg's versions of the dependencies, if and only if the Outcome test suite passes with them.

JackBoosY:
@ned14
> I think consolidating all the SHAs into a single file with instructions of how to upgrade them as a set is better UX for port maintainers.

The reason I separated these two dependencies into two files is to add them as separate ports in the future. So I think it is appropriate to put REF and HASH in the corresponding cmake file.

> usage appears to be ignored if the port is targets based

Yes, I will take the usage file back.

> I added vcpkg features so external users can choose for the internally chosen byte-lite and gsl-lite to be the vcpkg versions.

#11758 is adding the feature of the specified version of the dependency of the port. Once it is merged, we can specify the version of byte-lite and gsl-lite. I don't know whether this question can be solved completely.

> In other package managers they have the ability to run the test suite for a package in order to test if the chosen dependency versions from the package manager are compatible and non breaking.

Yes, vcpkg currently lacks steps to test port validity and compatibility, but in the near future, I hope we can support them.
Each PR of the upgraded port version will test whether the port that depends on these ports is built correctly, if it fails, we will not merge into the master.

ned14:
> The reason I separated these two dependencies into two files is to add them as separate ports in the future. So I think it is appropriate to put REF and HASH in the corresponding cmake file.

I'm not keen on exposing QuickCppLib as a public port. I don't want people to see it and have expectations that they can use it directly. This was one of the major reasons I would not support a vcpkg port of my libraries back in 2017, because it adds support costs to me. If that's vcpkg's intent, think the single header include based port, which you already have, is plenty sufficient as is.

I would support QuickCppLib as a vcpkg port if and only if:

- Its port name is somehow adorned with "this is a port for internal use only". This could be as a simple as a prefix e.g. `vcpkginternal_QuickCppLib`.
- If internal use only ports do not appear in the default list of ports (i.e. a flag must be passed to display all including internal ports).

Obviously all this is open source, so you guys can do what you want. However maintaining good relations between packagers and devs is important long term. I have good relations with most packagers of my libraries, I do my best to support them, it's a quid pro quo.

> > I added vcpkg features so external users can choose for the internally chosen byte-lite and gsl-lite to be the vcpkg versions.
> 
> #11758 is adding the feature of the specified version of the dependency of the port. Once it is merged, we can specify the version of byte-lite and gsl-lite. I don't know whether this question can be solved completely.

Versioning makes package management considerably more complex in general. People think they want versioning, and locally I suppose they do, but to be honest I don't think people generally appreciate the wider problems versioning brings into package management. Python's packaging story has become what it has due to versioning, and they're desperately trying to dig themselves out of that hole now.

For me personally, the whole point of bothering with package management is that for some given release X of all the packages, that every package within that release has been confirmed to work well when combined with every other package within that release. Ensuring that is a lot of tedious work, but it's most of the point of Debian or Fedora or any binary packages based Linux distro. I mean, anyone can go make their own Linux distro, but it's all that interoperability _validation and testing_ is where the real value add occurs.

For me personally, same applies to a C++ package manager, and if a C++ package manager doesn't supply that, I don't see the point of buying into it for non-toy production software.

cpp-pm are currently transitioning to a matrix test configuration, so as each package get updated, they are tested by CI with all the other packages. If an upstream upgrades breaks something downstream, you now have to go fix downstream. This may kill off cpp-pm as they lack full time employees to go do all that donkey work (e.g. tons of stuff depends on OpenSSL, so an OpenSSL upgrade may break lots of other packages in tedious ways), but it's the right engineering choice, in my opinion.

> > In other package managers they have the ability to run the test suite for a package in order to test if the chosen dependency versions from the package manager are compatible and non breaking.
> 
> Yes, vcpkg currently lacks steps to test port validity and compatibility, but in the near future, I hope we can support them.
> Each PR of the upgraded port version will test whether the port that depends on these ports is built correctly, if it fails, we will not merge into the master.

This isn't useful for header only libraries, obviously. Also, I find it frequent that my library source code compiles just fine after an upstream upgrade, but it's the test suite which fails because upstream broke some assumption in my code. I had a nasty recent experience with `span<T>` magically changing upstream, indeed I upcalled it to WG21 and it now looks they'll be undoing that change to span. Library code compiled fine, it was the test suite breaking that discovered the upstream change.

In other words, I really do think you need to support test suites, and I think you need to throw a ton of Azure pipelines at testing downstream dependencies every time something they depend upon gets PRed.

JackBoosY:
@ned14 So, I assume that the current changes are good, right?

ned14:
I've fixed the outcome package features so they actually work. If the current set of changes suit you, feel free to merge.

JackBoosY (approved):

JackBoosY:
Already test all features succesfully on `x86-windows` and `x64-windows-static`.

ras0219-msft:
Hey @ned14, thanks for the PR!

To adhere to our maintainer policy, we do need to always use the vcpkg copies of the dependencies. I'm also sympathetic to your points about upstream breaking you however :).

To resolve this tension, I've patched in the ability to build a single object that ensures outcome builds given the upstream versions. I'm using https://github.com/ned14/outcome/blob/develop/doc/src/snippets/intro_example.cpp; please let me know if there's a better cpp file to ensure compiler coverage. I use `add_library(... OBJECT ...)` to ensure this is as cheap as possible -- just a single compiler execution.

As you also mentioned, this doesn't cover runtime breakages. To handle this case, I'd like to add running the tests in CI, restricted to a single (or small set of) triplet. However, I can't see any way to enable building the tests without also enabling a _lot_ of other stuff (python generator scripts, etc). Until we find a way to do that, I've added a big nastygram explaining to all users about potential incompatibility and explaining how to use our new versioning support to pin the versions of those dependencies back to the "known good" ones.

The way running the tests in CI would work is we will add an off-by-default feature `run-tests` which we will enable in the CI plan. This feature would then issue `vcpkg_build_cmake(TARGET test)`, which should morally run `make test`.

We're definitely aligned on how much of a package manager's value is in ensuring interoperability between packages! That's exactly why we have the policy to disallow embedded dependencies; ensuring you can use package A _and_ package B that both depend on package C.

>I would support QuickCppLib as a vcpkg port if and only if:
>
>Its port name is somehow adorned with "this is a port for internal use only". This could be as a simple as a prefix e.g. vcpkginternal_QuickCppLib.

This is absolutely reasonable; I'd propose marking this in two ways:
1. `ned11-internal-quickcpplib` as the name
2. The description can be short and simple, noting its internal nature: "NOT FOR EXTERNAL CONSUMPTION. a set of internal scripts used by ned11's libraries"

Using vcpkg in the name would be misleading to future maintainers who might think that it's a vcpkg internal bit and thus fair game to depend upon/modify within vcpkg :). We've already taken similar approaches for vcpkg-specific helpers, such as `boost-vcpkg-helpers` and `boost-modular-build-helper` and we've had zero issues with users attempting to depend on those for outside projects.

ned14:
> To resolve this tension, I've patched in the ability to build a single object that ensures outcome builds given the upstream versions. I'm using https://github.com/ned14/outcome/blob/develop/doc/src/snippets/intro_example.cpp; please let me know if there's a better cpp file (or files) to ensure proper coverage. I use add_library(... OBJECT ...) to ensure this is as cheap as possible -- just a single compiler execution.

I think this sufficient, though as you suggest, we can do better.

> As you also mentioned, this doesn't cover runtime breakages. To handle this case, I'd like to add running the tests in CI, restricted to a single (or small set of) triplet. However, I can't see any way to enable building the tests without also enabling a lot of other stuff (python generator scripts, etc). Until we find a way to do that, I've added a big nastygram explaining to all users about potential incompatibility and explaining how to use our new versioning support to pin the versions of those dependencies back to the "known good" ones.

There is a special unit test in Outcome, core-result.cpp, which is intentionally designed to be standalone for this exact use case. If we can successfully compile core-result.cpp, I think it likely that vcpkg's versions of the dependencies will be okay. If the compiled binary executes without failure, I think it very highly likely they will be okay. Note however that Outcome only barely uses either library, whereas LLFIO which uses Outcome uses both heavily.

Can you show me an example of how to do this `run-tests` feature correctly, and I'll attempt an implementation based on core-result.cpp?

> ned14-internal-quickcpplib as the name
> The description can be short and simple, noting its internal nature: "NOT FOR EXTERNAL CONSUMPTION. a set of internal scripts used by ned11's libraries"

So you would prefer splitting out the quickcpplib package, but named as the above.

I do have a question here about proper form. Strictly speaking, it is quickcpplib which has the dependencies on byte-lite and span-lite. However, it has no unit tests for those, and even if I added one, it would be nothing like as good a test as a unit test from LLFIO would be. Equally, as an internal package, I suppose there are zero expectations that it should behave well on its own.

Based on all of the above, can I summarise a work plan before I start work on it:

1. Break QuickCppLib into its own vcpkg package, named internally, reusing your changes here.
2. Have Outcome test it works during install by trying to compile core-result.cpp.
3. Have Outcome declare to vcpkg a CI test based on executing core-result.cpp so we can spot quickly if upstream dependency version changes have broken Outcome.

Optional:

4. Given the tight integration and dependency chaining between all these packages going on here, it seems to me it might make sense to also add into this PR the package for LLFIO, because I have a horrible feeling that if I add LLFIO later, I'll need to modify these packages again anyway. So might as well get it all done at once?

ras0219:
To implement `run-tests`, we need to do something like:
```
# CONTROL
Source: outcome
Build-Depends: byte-lite

Feature: run-tests
Description: Build and run the dependency validation test
```
```cmake
# portfile.cmake
vcpkg_check_features(
    OUT_FEATURE_OPTIONS FEATURE_OPTIONS
    # This translates the "run-tests" feature into a `-DOUTCOME_BUILD_SINGLE_DEP_TEST=ON/OFF`
    run-tests OUTCOME_BUILD_SINGLE_DEP_TEST
)

vcpkg_configure_cmake(
  OPTIONS ${FEATURE_OPTIONS}
)

if("run-tests" IN_LIST FEATURES)
    # This runs "make test"
    vcpkg_build_cmake(TARGET test)
endif()

vcpkg_install_cmake()
```
I've omitted a bunch of existing lines for brevity/clarity; if it gives you any trouble I'd be happy to be more verbose.

Then, to get `run-tests` to be enabled up in CI, we need to do a bit of a hack. Essentially, we need to add a do-nothing port that depends on `outcome[run-tests]` in the triplets that we want "lit up". The do-nothing port will live in `scripts/test_ports/`; you can see a few other "only-available-in-CI" ports we have there for similar testing reasons.
```
# /scripts/test_ports/outcome-run-tests/CONTROL
Source: outcome-run-tests
Version: 0
Description: outcome-run-tests
Build-Depends: outcome[run-tests] (x64&linux)
```
```cmake
# /scripts/test_ports/outcome-run-tests/portfile.cmake
set(VCPKG_POLICY_EMPTY_PACKAGE enabled)
```
Those changes should be everything needed to get the test built and run during x64-linux.

>1. Break QuickCppLib into its own vcpkg package, named internally, reusing your changes here.
> ...
> ...it seems to me it might make sense to also add into this PR the package for LLFIO...

Breaking up quickcpplib could be done now or could be done when you make a PR for `LLFIO`; plus/minus the run-tests topic above, I think the current state of this PR is quite mergeable if you'd like to make the new `outcome` immediately available. I don't know how complicated the LLFIO port will be, but if you think it will have anything "interesting" in it I'd recommend sequencing it as a separate PR.

ned14:
@ras0219 That's very useful thanks. I'll try to find some time next week to try implementing the next revision for this PR.

ned14:
Ok, this is my work in progress intended for you guys to review and inspect and make sure that I have the form correct. Note that the LLFIO smoke test does not link currently on x86 windows, so it isn't ready for merge yet. I'll tackle that on Thursday. But I wanted to see if the CI passes for all the other targets, and to check if I followed instructions correctly in terms of adding in CI execution of dependency smoke tests etc etc.

JackBoosY:
@ned14  Please ping me if this PR is ready for review.

ned14:
@JackBoosY @ras0219 All known problems with this changeset are fixed, and this is ready for review. I note that status-code isn't building on some of the platforms for some reason. And indeed there is a wide set of CI failures in general (I am out of time for this morning).

Note that the cmake config files installed are missing the link dependencies on their vcpkg sourced dependencies. So, Outcome does not indicate it needs `quickcpplib::hl`, and LLFIO does not indicate it needs `outcome::hl` and `quickcpplib::hl`. This isn't ideal, but I reckon it's probably also not the end of the world. Let me know what you think on this.

ned14:
FYI that the CI passing except for the UWP stuff (I very much doubt LLFIO can compile on that, ever) and ARM64 Windows.

I tried disabling UWP support in the CONTROL file, but it appears to have no effect.

ARM64 Windows should work just fine, but LLFIO cmake fails to find a working `<filesystem>`, and it bombs out.

JackBoosY:
For status-code:
```
     7>D:\buildtrees\status-code\src\4d12ab129b-50ac9e8fd8.clean\include\com_code.hpp(64,1): error C2664: '_com_error::_com_error(HRESULT,const wchar_t *)': cannot convert argument 2 from 'IErrorInfo *' to 'const wchar_t *' [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-not-posix.vcxproj]
       D:\buildtrees\status-code\src\4d12ab129b-50ac9e8fd8.clean\include\com_code.hpp(64,22): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-not-posix.vcxproj]
       C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\include\comdef.h(73,5): message : see declaration of '_com_error::_com_error' [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-not-posix.vcxproj]
     7>D:\buildtrees\status-code\src\4d12ab129b-50ac9e8fd8.clean\include\com_code.hpp(80,1): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-not-posix.vcxproj]
     6>D:\buildtrees\status-code\src\4d12ab129b-50ac9e8fd8.clean\include\com_code.hpp(64,1): error C2664: '_com_error::_com_error(HRESULT,const wchar_t *)': cannot convert argument 2 from 'IErrorInfo *' to 'const wchar_t *' [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-noexcept.vcxproj]
       D:\buildtrees\status-code\src\4d12ab129b-50ac9e8fd8.clean\include\com_code.hpp(64,22): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-noexcept.vcxproj]
       C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\include\comdef.h(73,5): message : see declaration of '_com_error::_com_error' [D:\buildtrees\status-code\x64-uwp-rel\test-status-code-noexcept.vcxproj]
```

For llfo:
```
CMake Error at quickcpplib/repo/cmakelib/QuickCppLibUtils.cmake:84 (message):
  FATAL: <filesystem> is not available, and neither libstdc++ nor libc++ STLs
  will link a program
  CMakeLists.txt:229 (indented_message)
```

ned14:
> For status-code:

Thanks for this added info, I've worked around the problem in status-code.

> For llfo:
> 
> ```
> CMake Error at quickcpplib/repo/cmakelib/QuickCppLibUtils.cmake:84 (message):
>   FATAL: <filesystem> is not available, and neither libstdc++ nor libc++ STLs
>   will link a program
>   CMakeLists.txt:229 (indented_message)
> ```

LLFIO's cmake tries many ways of finding a `<filesystem>` or a `<experimental/filesystem>` before it gives up with the message above. Note its test is to run a test program, so any link failures, or failures to load a DLL, which fail the test. It's possible the logic is broken, but more likely the platform in question has a foobared `<filesystem>`.

Note that if you force C++ 17, which is supposed to include `<filesystem>`, that may make it work.

ned14:
Out of interest, is there some way of retrieving the cmake install or build failure logs from Azure pipelines? I've searched around the web UI, and I'm probably missing something obvious, but I couldn't find them. I think Outcome ought to work fine on uwp, so it should not fail there. I assume that because LLFIO has explicitly disabled itself on uwp, once Outcome passes on uwp, all CI tests ought to pass.

JackBoosY:
@ned14 llfo is indeed disabled on uwp, the current regression of llfo is arm64-windows.
For other regressions, I will provide you the failure logs tomorrow.

JackBoosY:
For llfo: 
```
CMake Error at quickcpplib/repo/cmakelib/QuickCppLibUtils.cmake:84 (message):
  FATAL: <filesystem> is not available, and neither libstdc++ nor libc++ STLs
  will link a program
  CMakeLists.txt:229 (indented_message)
```

For status-code:
```
     8>G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\bin\Release\test-status-code-noexcept.exe : fatal error LNK1120: 4 unresolved externals [G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code-noexcept.vcxproj]
       _GenerateAppxPackageFile:
         C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\MakeAppx.exe pack /l /h sha256 /f generate-tables.dir\Release\package.map.txt /o /p G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\AppPackages\generate-tables\generate-tables_1.0.0.0_ARM_Test\generate-tables_1.0.0.0_ARM.msix  
     8>Done Building Project "G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code-noexcept.vcxproj" (default targets) -- FAILED.
     7>Link:
         C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\bin\HostX64\arm\link.exe /ERRORREPORT:QUEUE /OUT:"G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\bin\Release\test-status-code.exe" /INCREMENTAL:NO /NOLOGO WindowsApp.lib /MANIFEST /ManifestFile:"test-status-code.dir\Release\test-status-code.exe.intermediate.manifest" /MANIFESTUAC:NO /DEBUG /PDB:"G:/15603/vcpkg/buildtrees/status-code/arm-uwp-rel/bin/Release/test-status-code.pdb" /SUBSYSTEM:CONSOLE /OPT:REF /OPT:ICF /TLBID:1 /APPCONTAINER /WINMD:NO /WINMDFILE:"G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\bin\Release\test-status-code.winmd" /DYNAMICBASE /NXCOMPAT /IMPLIB:"G:/15603/vcpkg/buildtrees/status-code/arm-uwp-rel/Release/test-status-code.lib" /MACHINE:ARM  /machine:ARM /nologo "test-status-code.dir\Release\main.obj"
     4>_GenerateAppxPackageFile:
         generate-tables -> G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\AppPackages\generate-tables\generate-tables_1.0.0.0_ARM_Test\generate-tables_1.0.0.0_ARM.msix
     7>main.obj : error LNK2019: unresolved external symbol "unsigned long __cdecl system_error2::win32::GetLastError(void)" (?GetLastError@win32@system_error2@@YAKXZ) referenced in function "private: static class system_error2::status_code_domain::string_ref __cdecl system_error2::_com_code_domain::_make_string_ref(long,wchar_t *)" (?_make_string_ref@_com_code_domain@system_error2@@CA?AVstring_ref@status_code_domain@2@JPA_W@Z) [G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code.vcxproj]
     7>main.obj : error LNK2019: unresolved external symbol "unsigned long __cdecl system_error2::win32::FormatMessageW(unsigned long,void const *,unsigned long,unsigned long,wchar_t *,unsigned long,void *)" (?FormatMessageW@win32@system_error2@@YAKKPBXKKPA_WKPAX@Z) referenced in function "private: static class system_error2::status_code_domain::string_ref __cdecl system_error2::_nt_code_domain::_make_string_ref(long)" (?_make_string_ref@_nt_code_domain@system_error2@@CA?AVstring_ref@status_code_domain@2@J@Z) [G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code.vcxproj]
     7>main.obj : error LNK2019: unresolved external symbol "int __cdecl system_error2::win32::WideCharToMultiByte(unsigned int,unsigned long,wchar_t const *,int,char *,int,char const *,int *)" (?WideCharToMultiByte@win32@system_error2@@YAHIKPB_WHPADHPBDPAH@Z) referenced in function "private: static class system_error2::status_code_domain::string_ref __cdecl system_error2::_com_code_domain::_make_string_ref(long,wchar_t *)" (?_make_string_ref@_com_code_domain@system_error2@@CA?AVstring_ref@status_code_domain@2@JPA_W@Z) [G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code.vcxproj]
     7>main.obj : error LNK2019: unresolved external symbol "void * __cdecl system_error2::win32::GetModuleHandleW(wchar_t const *)" (?GetModuleHandleW@win32@system_error2@@YAPAXPB_W@Z) referenced in function "private: static class system_error2::status_code_domain::string_ref __cdecl system_error2::_nt_code_domain::_make_string_ref(long)" (?_make_string_ref@_nt_code_domain@system_error2@@CA?AVstring_ref@status_code_domain@2@J@Z) [G:\15603\vcpkg\buildtrees\status-code\arm-uwp-rel\test-status-code.vcxproj]
```

For outcome: success. Don't know why its regression happened.

JackBoosY:
I know why.
Building `outcome[core,run-tests]:x64-uwp`:
```
Microsoft (R) Build Engine version 15.9.21+g9802d43bc3 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1009: Project file does not exist.
Switch: test.vcxproj

```

ned14:
@JackBoosY That's super useful, thank you. I had forgotten that on ARM, Windows uses the ARM mangling convention. I'll try to grab some time to do more work on this later this week.

ned14:
MSVC targeting ARM will compile with `<filesystem>` just fine, but when you run the resulting executable you get:

```
This version of %1 is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher
Return value: FAILED_TO_RUN
Source file was:

#include <filesystem>
int main() {
  try { return std::filesystem::path("hi").empty(); } catch(std::filesystem::filesystem_error) { return 1; }
}
```
@BillyONeal Is it known that MSVC's C++ 17 `<filesystem>` does not work on ARM Windows? Or is this an artefact of whatever ARM CI that Azure is using, that maybe it doesn't emulate enough of a real ARM Windows or something?

(Just to be clear here, we are not talking about ARM UWP where `<filesystem>` almost certainly cannot work, the failing CI target about is supposedly plain ARM Windows where I would have thought you would get a full fat Win32 API)

ned14:
@JackBoosY @ras0219 I suspect the reason for the LLFIO ARM Windows failure is because as part of detection of a working `<filesystem>` we compile a test program and try running it. If the host is x64, and we are cross compiling for ARM, that obviously fails, and no `<filesystem>` is found.

Normally LLFIO's cmake detects cross compile using `CMAKE_SYSTEM_PROCESSOR STREQUAL ${CMAKE_HOST_SYSTEM_PROCESSOR}`, so it is possible that that is not working as cmake documents when vcpkg does a build for ARM Windows?

A useful datapoint for you might be that if I build targeting ARM on Windows here, I get:

CMAKE_SYSTEM_PROCESSOR = AMD64
CMAKE_HOST_SYSTEM_PROCESSOR = AMD64

This suggests a bug in cmake which ought to be fixed. In the meantime, a suggested workaround would be useful.

scripts/test_ports/outcome-run-tests/CONTROL
```diff
@@ -0,0 +1,4 @@
+Source: outcome-run-tests
+Version: 0
+Description: Ensures that outcome built with vcpkg version of dependencies produces working executables.
+Build-Depends: outcome[run-tests]
```
ras0219, line 4:
```suggestion
Build-Depends: outcome[run-tests]
Supports: x64 & linux
```

This will restrict running tests to only a "native" build

scripts/test_ports/llfio-run-tests/CONTROL
```diff
@@ -0,0 +1,4 @@
+Source: llfio-run-tests
+Version: 0
+Description: Ensures that LLFIO built with vcpkg version of dependencies produces working executables.
+Build-Depends: llfio[run-tests], llfio[status-code,run-tests]
```
ras0219, line 4:
```suggestion
Build-Depends: llfio[status-code,run-tests]
Supports: x64 & linux
```

This will restrict running tests to only a "native" build.

ports/ned14-internal-quickcpplib/portfile.cmake
```diff
@@ -0,0 +1,74 @@
...
+
+vcpkg_install_cmake()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/quickcpplib)
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/cmakelib" "${CURRENT_PACKAGES_DIR}/share/ned14-internal-quickcpplib/cmakelib")
+file(RENAME "${CURRENT_PACKAGES_DIR}/share/scripts" "${CURRENT_PACKAGES_DIR}/share/ned14-internal-quickcpplib/scripts")
+
+file(INSTALL "${CURRENT_PORT_DIR}/usage" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}")
```
ras0219, line 73: I think it would be cleaner to just not provide any usage instructions here, since it is not intended to be used.

ports/llfio/usage
```diff
@@ -0,0 +1,15 @@
+The package llfio provides three CMake targets:
+
+    find_package(llfio CONFIG REQUIRED)
+
+- If you want the header-only form of LLFIO:
+    target_link_libraries(main PUBLIC llfio::hl)
+
+- If you want the static library form of LLFIO:
```
ras0219, line 8: Because vcpkg only builds one version of the library, we should avoid confusing the user with conflicting instructions; instead, it would be best to have a `usage.static` and `usage.dynamic` in the port directory and install only the applicable one:

```cmake
file(INSTALL "${CURRENT_PORT_DIR}/usage.${VCPKG_LIBRARY_LINKAGE}" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME usage)
```
The usage can then list only the exported form from this particular built package.

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,82 @@
...
+)
+
+# Outcome expects status-code to live inside its include directory
+file(COPY "${CURRENT_INSTALLED_DIR}/include/status-code/"
+    DESTINATION "${SOURCE_PATH}/include/outcome/experimental/status-code/include/"
 )
+file(COPY "${CURRENT_INSTALLED_DIR}/include/status-code/detail/"
+    DESTINATION "${SOURCE_PATH}/include/outcome/experimental/status-code/include/detail/"
+)
+
```
ras0219, line 54:
```suggestion
```

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,82 @@
...
+set(VCPKG_BUILD_TYPE release)
+
+# Use Outcome's own build process, skipping examples and tests.
+vcpkg_configure_cmake(
+    SOURCE_PATH "${SOURCE_PATH}"
+    PREFER_NINJA
+    OPTIONS
+        -DPROJECT_IS_DEPENDENCY=On
+        -Dquickcpplib_FOUND=1
+        "-DCMAKE_CXX_FLAGS=-I${CURRENT_INSTALLED_DIR}/include"
```
ras0219, line 66:
```suggestion
```
This interferes with the flags we want to set inside `vcpkg_configure_cmake()`; if we must pass additional flags, we can append to `VCPKG_CXX_FLAGS`:
```cmake
string(APPEND VCPKG_CXX_FLAGS " \"-I${CURRENT_INSTALLED_DIR}/include\"")
string(APPEND VCPKG_C_FLAGS " \"-I${CURRENT_INSTALLED_DIR}/include\"")
vcpkg_configure_cmake( ... )
```
but it would be much more robust to patch in
```cmake
find_path(INCLUDE_DIR the_required_header.h)
include_directories(${INCLUDE_DIR})
```
to the top of the main CMakeLists.txt

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,82 @@
...
+        -DOUTCOME_ENABLE_DEPENDENCY_SMOKE_TEST=ON  # Leave this always on to test everything compiles
+        -DCMAKE_DISABLE_FIND_PACKAGE_Git=ON
+)
+
+if("-DOUTCOME_ENABLE_DEPENDENCY_SMOKE_TEST=ON" IN_LIST OUTCOME_FEATURE_OPTIONS)
```
ras0219, line 71:
```suggestion
if("run-tests" IN_LIST FEATURES)
```

ports/outcome/portfile.cmake
```diff
@@ -1,13 +1,82 @@
...
     HEAD_REF develop
+    PATCHES
+      outcome-prune-sources.patch
+)
+
+vcpkg_check_features(
+    OUT_FEATURE_OPTIONS OUTCOME_FEATURE_OPTIONS
+    FEATURES
+      run-tests OUTCOME_ENABLE_DEPENDENCY_SMOKE_TEST
+)
```
ras0219, line 37:
```suggestion
```

If `OUTCOME_FEATURE_OPTIONS` isn't going to be passed into the build, this isn't needed (we can check features directly against `FEATURES`).

ports/status-code/CONTROL
```diff
@@ -0,0 +1,4 @@
+Source: status-code
+Version: 1.0.0-ab3cd821
+Homepage: https://github.com/ned14/status-code
+Description: Proposed SG14 status_code for the C++ standard.
```
ras0219, line 4: It would be neat to include paper number(s) in the description if they exist (for searchability)

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+      SHA512 589d3bc7bca98ca8d05ce9f5cf009dd98b8884bdf3739582f2f6cbf5a324ce95007ea041450ed935baa4a401b4a0242c181fb6d2dcf7ad91587d75f05491f50e
+      HEAD_REF master
+  )
+endif()
+
+vcpkg_check_features(
+    OUT_FEATURE_OPTIONS LLFIO_FEATURE_OPTIONS
+    FEATURES
+      status-code LLFIO_USE_EXPERIMENTAL_SG14_STATUS_CODE
+      run-tests LLFIO_ENABLE_DEPENDENCY_SMOKE_TEST
```
ras0219, line 28:
```suggestion
```

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+        -DCMAKE_DISABLE_FIND_PACKAGE_Git=ON
+)
+
+# Build the shared library and static library editions
+vcpkg_build_cmake(TARGET _sl)
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+    vcpkg_build_cmake(TARGET _dl)
+endif()
+
+if("-DLLFIO_ENABLE_DEPENDENCY_SMOKE_TEST=ON" IN_LIST OUTCOME_FEATURE_OPTIONS)
```
ras0219, line 64:
```suggestion
if("run-tests" IN_LIST FEATURES)
```

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO ned14/llfio
+    REF 9294f05d48220a6a5d6d307da5fe8f236dd020cf
+    SHA512 00bd49fe568ba2a840917de8d5b8d398d7480b2431f99f8c3b282fc60785d8664869fd83e8ac56ffc9a1eaa0b991c5d8a97ab5e53e20e5567f3d75577ad4feb4
+    HEAD_REF develop
+)
+
+if(WIN32)
```
ras0219, line 14:
```suggestion
if(VCPKG_TARGET_IS_WINDOWS)
```
WIN32 here refers to the host machine, not the target

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+file(COPY "${CURRENT_INSTALLED_DIR}/include/quickcpplib"
+    DESTINATION "${SOURCE_PATH}/quickcpplib/repo/include/"
+)
+file(COPY "${CURRENT_INSTALLED_DIR}/share/ned14-internal-quickcpplib/"
+    DESTINATION "${SOURCE_PATH}/quickcpplib/repo/"
+)
+
+# LLFIO expects ntkernel-error-category to live inside its include directory
+file(REMOVE_RECURSE "${SOURCE_PATH}/include/llfio/ntkernel-error-category")
+if(WIN32)
```
ras0219, line 41:
```suggestion
if(VCPKG_TARGET_IS_WINDOWS)
```

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+endif()
+
+vcpkg_configure_cmake(
+    SOURCE_PATH "${SOURCE_PATH}"
+    PREFER_NINJA
+    OPTIONS
+        -DPROJECT_IS_DEPENDENCY=On
+        -Dquickcpplib_FOUND=1
+        -Doutcome_FOUND=1
+        "-DCMAKE_CXX_FLAGS=-I${CURRENT_INSTALLED_DIR}/include"
```
ras0219, line 52:
```suggestion
```

See discussion below about `VCPKG_CXX_FLAGS`

ports/llfio/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+        ${LLFIO_FEATURE_OPTIONS}
+        -DLLFIO_ENABLE_DEPENDENCY_SMOKE_TEST=ON  # Leave this always on to test everything compiles
+        -DCMAKE_DISABLE_FIND_PACKAGE_Git=ON
+)
+
+# Build the shared library and static library editions
+vcpkg_build_cmake(TARGET _sl)
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+    vcpkg_build_cmake(TARGET _dl)
+endif()
```
ras0219, line 62:
```suggestion
if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
    vcpkg_build_cmake(TARGET _dl)
else()
    vcpkg_build_cmake(TARGET _sl)
endif()
```
I'm not entirely sure if this is the correct fix, however vcpkg ports must produce only one variant (static/shared) of libraries. The variant requested by the user is defined by `VCPKG_LIBRARY_LINKAGE`.

`vcpkg_configure_cmake()` will also automatically define the standard CMake variable for requesting only one flavor: `BUILD_SHARED_LIBS` (https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)

ras0219:
Yes, I think it's a known bug when doing cross-compilation on windows that we do not set `CMAKE_SYSTEM_PROCESSOR` correctly. For now, it's fine to mark `arm64-windows` et al as unsupported in our CI (adding `llfio:arm64-windows=fail` to the end of `/scripts/ci.baseline.txt`).

JackBoosY:
Please use `VCPKG_HOST_IS_*` when cross-compiling. See [documentation](https://github.com/microsoft/vcpkg/blob/6b19f55def8ae1b37ff93df4cf3b95ed31b21b60/scripts/cmake/vcpkg_common_definitions.cmake#L9).

scripts/test_ports/outcome-run-tests/CONTROL
ned14: I assume we do want the test to run on Windows as well, so used "Supports: x64"

scripts/test_ports/llfio-run-tests/CONTROL
ned14: I assume we do want the test to run on Windows as well, so used "Supports: x64"

ports/ned14-internal-quickcpplib/portfile.cmake
ned14: The usage instructions now say "don't use this port". If you don't supply any usage, I believe a default usage instructions are auto generated right?

ports/llfio/usage
ned14: Fixed.

ports/outcome/portfile.cmake
ned14: I disagree it is more robust to patch. All my libraries assume a certain layout of header includes, as you've probably noticed, so I reckon twiddling the CXX flags is more robust.

ports/outcome/portfile.cmake
ned14: Fixed.

ports/outcome/portfile.cmake
ned14: Fixed.

ports/status-code/CONTROL
ned14: Fixed.

ports/llfio/portfile.cmake
ned14: Fixed.

ports/llfio/portfile.cmake
ned14: As you've probably noticed, none of my libraries observe `BUILD_SHARED_LIBS`, they always generate targets for header only, static, and shared libraries (and indeed entire target trees for each of the sanitisers and static analysers for each of those again). LLFIO very strongly encourages the static library build over all others, and everything assumes it is always built, because it always is. All target trees generate uniquely named files, nothing ever collides, so apart from it taking slightly longer it is benign.

ned14:
> Yes, I think it's a known bug when doing cross-compilation on windows that we do not set `CMAKE_SYSTEM_PROCESSOR` correctly. For now, it's fine to mark `arm64-windows` et al as unsupported in our CI (adding `llfio:arm64-windows=fail` to the end of `/scripts/ci.baseline.txt`).

I've attempted a workaround, everything works on ARM Windows just fine, so permitting that would be nice. Especially as ARM Windows market share will only increase going forth.

I think I've addressed most of the other feedback, barring mistake or oversight. The UWP tests seem to still fail, advice on that would be welcome.

ned14:
> I'll see if I can get some time to try installing the UWP toolchain and see what the cause is.

Alas UWP appears to be currently broken in VS2019 16.9.0 Preview 4.0, which I am forced to use because LLFIO ICEs the entire 16.8.x series. I guess we await Preview 5.0 or RC1.

ned14:
I couldn't get UWP working, so I disabled it. All tests now pass.

ned14:
Any news on this? It's been a month now since last update.

PR #15624, opened 2021-01-13
[glslang] Provide glslang-default-resource-limits library.

wolfpld:
**Describe the pull request**

- What does your PR fix?

This PR makes the glslang-default-resource-limits library publicly available for linking. This is required by third party tools using glslang, in order to get the required resource limits data structure. This struct is internally used by the glslang library and its ABI does change between releases, so it's not possible to maintain this data consistently on the third party side.

No change here, in comparison to what was previously available.

Yes, with the exception of bumping the `version` field, instead of increasing the `port-version`, as this was the convention used previously in this port.


ports/glslang/CONTROL
```diff
@@ -1,4 +1,4 @@
 Source: glslang
-Version: 2019-03-05-1
+Version: 2019-03-05-2
```
NancyLi1013, line 2:
```suggestion
Version: 2019-03-05
Port-Version: 2
```
Now we recommend to use `Port-Version` field here. Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#port-version

ports/glslang/CMakeLists-targets.patch
```diff
@@ -106,3 +106,16 @@ index f918d7a..329800f 100644
...
+diff --git a/StandAlone/CMakeLists.txt b/StandAlone/CMakeLists.txt
+index 5cea53d9..43e45534 100644
+--- a/StandAlone/CMakeLists.txt
++++ b/StandAlone/CMakeLists.txt
+@@ -46,7 +46,7 @@ if(ENABLE_GLSLANG_INSTALL)
+     install(TARGETS spirv-remap
+             RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+             
+-    if(BUILD_SHARED_LIBS)
++    if(BUILD_SHARED_LIBS OR TRUE)
```
NancyLi1013, line 118: Frankly speaking, I don't know why you add `TRUE `here.

NancyLi1013 (changes requested):

ports/glslang/CMakeLists-targets.patch
wolfpld: That's understandable and I share your concern about this. However, adding `OR TRUE` is what this port does, for whatever reason, as you can see in the already existing portion of the `CMakeLists-targets.patch`. Therefore, the above change is consistent with previous work done here. Changing this to a proper solution is out of the scope of this PR, which is about adding a missing library.

NancyLi1013: Thanks for your clarification.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your PR @wolfpld.

hiradyazdan:
I am not sure how relevant this is, but since this PR relates to `glslang-default-resource-limits` I was wondering it should have resovled likning the target correctly, but I'm getting this error:

```shell
error LNK2001: unresolved external symbol "struct TBuiltInResource const glslang::DefaultTBuiltInResource" (?DefaultTBuiltInResource@glslang@@3UTBuiltInResource@@B)
SDFRay4D.exe : fatal error LNK1120: 1 unresolved externals
```

If I pull the actual repository I can add this library and link the target correctly, but I am not sure if this package cmake config links the target as in `glslang::default-resource-limits` which I assume is causing this error.

hiradyazdan:
I am not sure how relevant this is, but since this PR relates to `glslang-default-resource-limits` I was wondering it should have resovled likning the target correctly, but I'm getting this error:

```shell
error LNK2001: unresolved external symbol "struct TBuiltInResource const glslang::DefaultTBuiltInResource" (?DefaultTBuiltInResource@glslang@@3UTBuiltInResource@@B)
....exe : fatal error LNK1120: 1 unresolved externals
```

If I pull the actual repository I can add this library and link the target correctly, but I am not sure if this package cmake config links the target as in `glslang::default-resource-limits` which I assume is causing this error.

PR #15634, opened 2021-01-14
[openal-soft] Fix issue with cmake config.

Hoikas:
On Windows, openal-soft config cmake (OpenALConfig.cmake) by default generates a file that contains an absolute path to an interface link library on the build system:
```cmake
set_target_properties(OpenAL::OpenAL PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/AL"
  INTERFACE_LINK_LIBRARIES "C:/Program Files (x86)/Windows Kits/10/Lib/10.0.18362.0/um/x86/WinMM.Lib"
)
```

This can break if the package is restored from a binary cache on a system where:
* the OS is not installed on drive C.
* the OS is not 64-bit -- or the package was built on 32-bit Windows but restored on 64-bit Windows.
* the Windows SDK version does not match exactly (Visual Studio offers this as an option in the installer).

To fix this, we disable the WINMM and DSOUND backends in openal-soft. This does not result in any lost functionality because as of Windows Vista, WINMM and DSOUND are implemented by WASAPI in Windows itself. The consumer has no knowledge or control over the backend except at build-time. Further, it matches the behavior of the port on Linux and macOS where only a single audio backend is explicitly requested.

- What does your PR fix?
See above.

No change.

Yes


JackBoosY:
Why do these files contain absolute paths?
It should be, example:
`INTERFACE_LINK_LIBRARIES "WinMM.Lib"`

Hoikas:
Upstream is directly feeding the system library paths into the OpenAL `target_link_libraries()`.
- https://github.com/kcat/openal-soft/blob/openal-soft-1.20.1/CMakeLists.txt#L879
- https://github.com/kcat/openal-soft/blob/openal-soft-1.20.1/CMakeLists.txt#L1198

It looks like some major work upstream to fix everything. The easiest fix is, IMO, to disable the libraries that effectively do nothing since Windows Vista.

JackBoosY (approved):

PR #15669, opened 2021-01-15
[google-cloud-cpp] update to v1.23.0

coryan:
Update `google-cloud-cpp` to the latest release (v1.23.0)

N/A

No change.

Yes.

### DIFF
--- a/ports/google-cloud-cpp/vcpkg.json
+++ b/ports/google-cloud-cpp/vcpkg.json
@@ -1,6 +1,6 @@
 {
   "name": "google-cloud-cpp",
-  "version-string": "1.22.0",
+  "version-string": "1.23.0",
   "description": "C++ Client Libraries for Google Cloud Platform APIs.",
   "homepage": "https://github.com/googleapis/google-cloud-cpp",
   "supports": "!uwp",
--- a/port_versions/baseline.json
+++ b/port_versions/baseline.json
@@ -2201,7 +2201,7 @@
       "port-version": 0
     },
     "google-cloud-cpp": {
-      "baseline": "1.22.0",
+      "baseline": "1.23.0",
       "port-version": 0
     },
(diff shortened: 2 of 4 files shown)


coryan:
The OS/X failure seems unrelated, but happy to be proven wrong.

ras0219 (approved):

JonLiu1993 (approved):

ports/google-cloud-cpp/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "google-cloud-cpp",
-  "version-string": "1.22.0",
+  "version-string": "1.23.0",
```
vicroms, line 3:
```suggestion
  "version-semver": "1.23.0",
```

From what I've seen in their GitHub, this library uses SemVer consistently.

vicroms (changes requested):

ports/google-cloud-cpp/vcpkg.json
coryan: We do not follow semver, this is documented in each our our libraries:

"Please note that the Google Cloud C++ client libraries do **not** follow Semantic Versioning."

https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/pubsub/README.md#google-cloud-pubsub-c-client-library

Furthermore, we strive for API (and **not** ABI) compatibility, and we **do** break the API without increasing the major version number, e.g.:

https://github.com/googleapis/google-cloud-cpp/blob/master/CHANGELOG.md#v1210---2020-12

vicroms: I see, thanks for the clarification! In this case then `version-string` is correct 👍

vicroms (approved):

PR #15683, opened 2021-01-16
[simage] Fix optional dependencies

ras0219-msft:


JackBoosY (approved):

ports/simage/CONTROL
```diff
@@ -1,5 +1,6 @@
 Source: simage
 Version: 1.8.0
+Port-Version: 1
 Description: Image file format library abstraction layer
-Build-Depends: giflib (linux|uwp), libjpeg-turbo (linux|uwp), libpng (linux|uwp), libsndfile (linux|uwp), tiff (linux|uwp), libvorbis (linux|uwp)
+Build-Depends: giflib (!windows&!osx), libjpeg-turbo (!windows&!osx), libpng (!windows&!osx), libsndfile, tiff (!windows&!osx), libvorbis, zlib (!windows&!osx)
```
dg0yt, line 5: `!windows` also excludes `uwp`. This is different than before, and it doesn't align with the `OSX_OR_WINDOWS` variable in `portfile.cmake`. What is the desired state for UWP? 
Now this caused to let the tiff update to 4.3.0 pass through CI without testing simage, but hits ports which happen to install tiff before simage for UWP: #17519, #18393

PR #15693, opened 2021-01-16
[vcpkg] More versioning docs

vicroms:
This PR adds more documentation for versioning

- [X] Getting started ([Rendered](https://github.com/vicroms/vcpkg/blob/versioning/docs/docs/users/versioning.getting-started.md))
- [X] Reference: Version schemes
- [X] Reference: Version files
- [X] Reference: Constraints


soumyamahunt:
Is there any plan to support `msbuild` too? Also is there plan on `dependabot` integration?

vicroms:
@soumyamahunt 
Yes, this will work with MSBuild projects. 
I added a link to an example of how to use manifests in MSBuild.

soumyamahunt:
> @soumyamahunt
> Yes, this will work with MSBuild projects.
> I added a link to an example of how to use manifests in MSBuild.

What about enabling versions feature flag? Is this feature flag temporary, when versioning available for general release it won't be needed anymore?

vicroms:
We plan to enable the flag by default once we have evidence from users of this feature that the feature is stable enough.

PR #15706, opened 2021-01-17
[icu] Add support for macos rpath prefix macro

adam-bloom:
**Describe the pull request**

- What does your PR fix?
This PR implements a temporary solution for https://unicode-org.atlassian.net/browse/ICU-21458. It alters the ID fields of the built dylibs to add support for the @rpath macro on macOS. The new code is commented that it should be removed and replaced with calls to new configure options, if ICU-21458 is resolved.

Only macOS is affected.

Yes

### DIFF
--- a/ports/icu/portfile.cmake
+++ b/ports/icu/portfile.cmake
@@ -169,6 +169,127 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
         WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}"
         LOGNAME "make-build-${RELEASE_TRIPLET}")
 
+    # remove this block if https://unicode-org.atlassian.net/browse/ICU-21458
+    # is resolved and use the configure script instead
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+        if(DEFINED CMAKE_INSTALL_NAME_DIR)
+            set(ID_PREFIX "${CMAKE_INSTALL_NAME_DIR}")
+        else()
+            set(ID_PREFIX "@rpath")
+        endif()
+
+        # add ID_PREFIX to libicudata
+        vcpkg_execute_build_process(
+            COMMAND /usr/bin/install_name_tool -id "${ID_PREFIX}/libicudata.${ICU_VERSION_MAJOR}.dylib"
+            "libicudata.${VERSION}.dylib"
+            WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}/lib"
+            LOGNAME "make-build-fix-rpath-${RELEASE_TRIPLET}"
+        )
+
+        # add ID_PREFIX to libicui18n
+        vcpkg_execute_build_process(
+            COMMAND /usr/bin/install_name_tool -id "${ID_PREFIX}/libicui18n.${ICU_VERSION_MAJOR}.dylib"
+            "libicui18n.${VERSION}.dylib"
+            WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}/lib"
+            LOGNAME "make-build-fix-rpath-${RELEASE_TRIPLET}"
+        )
+
+        # add ID_PREFIX to libicui18n dependencies
+        vcpkg_execute_build_process(
+            COMMAND /usr/bin/install_name_tool -change "libicuuc.${ICU_VERSION_MAJOR}.dylib"
+                                                   "${ID_PREFIX}/libicuuc.${ICU_VERSION_MAJOR}.dylib"
+                                                   "libicui18n.${VERSION}.dylib"
+            WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}/lib"
+            LOGNAME "make-build-fix-rpath-${RELEASE_TRIPLET}"
+        )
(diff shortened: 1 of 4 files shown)


adam-bloom:
The macOS end-to-end tests are failing due to:
```
Failed to take the filesystem lock on /Volumes/data/work/1/s/.vcpkg-root:
    Resource busy
```
I don't believe this has anything to do with the changes in this PR, since this PR does not touch core vcpkg itself. Any ideas?

ports/icu/portfile.cmake
```diff
@@ -169,6 +169,127 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
...
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+        if(DEFINED CMAKE_INSTALL_NAME_DIR)
+            set(ID_PREFIX "${CMAKE_INSTALL_NAME_DIR}")
+        else()
+            set(ID_PREFIX "@rpath")
+        endif()
+
+        # add ID_PREFIX to libicudata
+        vcpkg_execute_build_process(
+            COMMAND /usr/bin/install_name_tool -id "${ID_PREFIX}/libicudata.${ICU_VERSION_MAJOR}.dylib"
```
JackBoosY, line 183: Should use `find_program(install_name_tool install_name_tool)` then change this to:
`COMMAND ${install_name_tool} -id "${ID_PREFIX}/libicudata.${ICU_VERSION_MAJOR}.dylib"`

ports/icu/portfile.cmake
```diff
@@ -169,6 +169,127 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
...
+            "libicudata.${VERSION}.dylib"
+            WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}/lib"
+            LOGNAME "make-build-fix-rpath-${RELEASE_TRIPLET}"
+        )
+
+        # add ID_PREFIX to libicui18n
+        vcpkg_execute_build_process(
+            COMMAND /usr/bin/install_name_tool -id "${ID_PREFIX}/libicui18n.${ICU_VERSION_MAJOR}.dylib"
```
JackBoosY, line 191: See comment above.

JackBoosY (changes requested):

JackBoosY (approved):

JackBoosY:
I'd like @strega-nil to review this PR also.

JackBoosY (approved):

strega-nil:
@adam-bloom This is unfortunately a change we cannot accept; changing the output depending on whether a tool is found is unacceptable for us. How common is `install_name_tool`? When does it exist on the system? We can discuss better solutions depending on that.

adam-bloom:
> @adam-bloom This is unfortunately a change we cannot accept; changing the output depending on whether a tool is found is unacceptable for us. How common is `install_name_tool`? When does it exist on the system? We can discuss better solutions depending on that.

`install_name_tool` is in /usr/bin on macOS, which is a part of the write restricted system partition. I don't have access to older macOS installations, so I'm not sure when it was first included. The `man` page has a date of 2009 there - so likely 10.6 was the first inclusion. Additionally, Apple Developer Tools is listed as a vcpkg requirement in the readme, and another copy of  `install_name_tool` is included with Xcode command line tools: `/Library/Developer/CommandLineTools/usr/bin/install_name_tool`.

In other words, I think every macOS system running vcpkg should be compatible with this. If, however, someone attempted to build an osx triplet from another OS (I'm not sure if that would even work without the macOS SDK...), they might be missing this tool and the step would skip.

I'll note that I initially did not have the if present check, since I assumed all osx vcpkg triplet runs would be from macOS systems that have this tool. I added that at the request of @JackBoosY

strega-nil:
I agree with using `find_program`, but it should be `REQUIRED`, given what you're saying :)

ports/icu/portfile.cmake
```diff
@@ -169,6 +169,136 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
...
+    # is resolved and use the configure script instead
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+        if(DEFINED CMAKE_INSTALL_NAME_DIR)
+            set(ID_PREFIX "${CMAKE_INSTALL_NAME_DIR}")
+        else()
+            set(ID_PREFIX "@rpath")
+        endif()
+
+        # install_name_tool may be missing if cross-compiling
+        find_program(INSTALL_NAME_TOOL install_name_tool)
```
strega-nil, line 182:
```suggestion
        find_program(INSTALL_NAME_TOOL install_name_tool REQUIRED)
```

then remove the `if` :)

strega-nil (changes requested):

adam-bloom:
Is there a use case for vcpkg where a osx triplet is built on a non-macOS system? Or is this definitely safe to mark as required?

JackBoosY:
@strega-nil The judgment is `if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")`, I think it's fine.
Do you have any other questions?

@adam-bloom  Nope.

strega-nil:
yeah, the issue I have is making it non-`REQUIRED`; since we'll always be building on macOS, we should `REQUIRED` the program.

JackBoosY:
So I think it's good now.

JackBoosY:
Already merged from master.

strega-nil (approved):

PR #15719, opened 2021-01-17
[glslang] Update to 11.1.0

fknfilewalker:
**Describe the pull request**
Update of glslang to version 11.1.0. Previous one was over a year old and had no support for KHR Ray Tracing. I also had to remove Shaderc from the ports since it is broken for some time now. Even with the update of glslang it does not build and only prevents the tests for glslang from succeeding. Additional I added linux and macOS builds, which were not configured in the previous port.

Works with `glslang:x64-windows` `glslang:x64-linux` `glslang:x64-osx`

I can also fix shaderc after this one.


JonLiu1993 (approved):

JonLiu1993 (approved):

ports/glslang/CMakeLists-targets.patch
```diff
@@ -1,121 +1,122 @@
...
  
  if(ENABLE_GLSLANG_INSTALL)
--    install(TARGETS OGLCompiler
--            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+-    install(TARGETS OGLCompiler EXPORT OGLCompilerTargets
 +    install(TARGETS OGLCompiler EXPORT glslangConfig
-+            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-+            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+             ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+-	install(EXPORT OGLCompilerTargets DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
```
ras0219-msft, line 21: This appears to be removing a blessed upstream name for the package: `OGLCompiler`.

ports/glslang/CMakeLists-targets.patch
```diff
@@ -1,121 +1,122 @@
...
- endif(WIN32)
- 
+@@ -111,28 +111,28 @@ endif(WIN32)
  if(ENABLE_GLSLANG_INSTALL)
--    if(BUILD_SHARED_LIBS)
-+    if(BUILD_SHARED_LIBS OR TRUE)
+     if(BUILD_SHARED_LIBS)
          if (ENABLE_SPVREMAPPER)
-             install(TARGETS SPVRemapper
+-            install(TARGETS SPVRemapper EXPORT SPVRemapperTargets
```
ras0219-msft, line 32: This appears to be removing a blessed upstream name for the package: `SPVRemapper`

ports/glslang/CMakeLists-targets.patch
```diff
@@ -1,121 +1,122 @@
...
-@@ -46,7 +46,7 @@ if(ENABLE_GLSLANG_INSTALL)
-     install(TARGETS spirv-remap
-             RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
-             
--    if(BUILD_SHARED_LIBS)
-+    if(BUILD_SHARED_LIBS OR TRUE)
-         install(TARGETS glslang-default-resource-limits
-                 LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
-     endif()
+-        install(TARGETS HLSL EXPORT HLSLTargets
```
ras0219-msft, line 78: This appears to be removing a blessed upstream package name: `HLSL`

ras0219-msft (changes requested):
Since it upstream has introduced official target/package names, we must begin providing those; after the `vcpkg_fixup_cmake_targets()` call, the `OGLCompiler*` files and `SPVRemapper*` files can be moved to `share/oglcompiler/` and `share/spvremapper/` respectively.

For compatibility, we can ensure the previous names still work by appending text to the end of the `share/glslang/glslang-config.cmake` of the form:
```
include(CMakeDependencyMacro)
find_dependency(OGLCompiler CONFIG)
find_dependency(SPVRemapper CONFIG)

add_library(glslang::glslang INTERFACE IMPORTED)
set_target_properties(glslang::glslang PROPERTIES INTERFACE_LINK_LIBRARIES "glslang")
# repeat for all previous names (SPVRemapper, etc)
```
This ensures that the previously supported pattern of:
```
find_package(glslang CONFIG REQUIRED)
target_link_libraries(main PRIVATE glslang::glslang)
```
continues to work, but the new upstream targets are also supported so existing projects (like shaderc) will hopefully be much easier to make work.

ports/glslang/CMakeLists-targets.patch
Luke-Filewalker: I think the commit diff is a bit difficult to read in this case. The TARGETS OGLCompiler is preserved in my commit. As it is with HLSL and SPVRemapper 

https://github.com/Luke-Filewalker/vcpkg/blob/952a2a3cca7570b1da7fd07d42b487fb9124fe47/ports/glslang/CMakeLists-targets.patch#L10-L23
```patch
diff --git a/OGLCompilersDLL/CMakeLists.txt b/OGLCompilersDLL/CMakeLists.txt
index 0b007d45..5bba8245 100644
--- a/OGLCompilersDLL/CMakeLists.txt
+++ b/OGLCompilersDLL/CMakeLists.txt
@@ -42,7 +42,7 @@ if(WIN32)
 endif(WIN32)
 
 if(ENABLE_GLSLANG_INSTALL)
-    install(TARGETS OGLCompiler EXPORT OGLCompilerTargets
+    install(TARGETS OGLCompiler EXPORT glslangConfig
             ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
-	install(EXPORT OGLCompilerTargets DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
+	install(EXPORT glslangConfig DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
 endif(ENABLE_GLSLANG_INSTALL)
```

Also the previous build produced 
```
find_package(glslang CONFIG REQUIRED)
target_link_libraries(main PRIVATE glslang::HLSL glslang::SPIRV glslang::glslang glslang::OGLCompiler)
```
as an output, exactly like the new one.
Or do you mean 
```
find_package(glslang CONFIG REQUIRED)
target_link_libraries(main PRIVATE glslang::glslang)
```
should link all the targets at once?

Luke-Filewalker:
Pls someone respond: https://github.com/microsoft/vcpkg/pull/15719#discussion_r560776238

NancyLi1013:
@Luke-Filewalker 

The latest version has been updated to 11.2.0. So could you please update the changes to this PR?

https://github.com/KhronosGroup/glslang/releases/tag/11.2.0

Also since there is LICENSE.txt file in source codes. I think it's unnecessary to add copyright.
https://github.com/KhronosGroup/glslang/blob/11.2.0/LICENSE.txt

axsaucedo:
Is this something planned for the next release of vcpkg? We'll be using this branch for the time being as we need glslang to be in pair with our submodule-in-repo build

axsaucedo:
Just testing the branch against our project but it seems I'm getting an error when binding the libraries:
```
"C:\...\build\test\test_kompute.vcxproj" (default target) (1) ->
(Link target) ->
  LINK : fatal error LNK1104: cannot open file 'glslang.lib' [C:\...\build\test\test_kompute.vcxproj]
```

Luke-Filewalker:
You can always use the relatively old version that is currently included in the original vcpkg repo. I will update this one once I find the time. But after that, I don't know how long it takes to get this merged.

MrROBUST:
Hello. Can someone check why the 'vcpkg\packages\glslang_x64-windows\include\glslang' folder does not contain 'OSDependent' subfolder after port installation? Neither with this PR nor with the current port.

JonLiu1993:
@Luke-Filewalker ,Can you deal with the conflicting files?

JonLiu1993:
@Luke-Filewalker ,Can you please resolve the conflicts against master?

Luke-Filewalker:
Updated it to 11.4.0 and passes all checks except windows x86. During test "Validate version file" the following happens:

```
Error: While reading versions for port glslang from file: C:\a\1\s\versions\g-\glslang.json
       File declares version `11.4.0` with SHA: fa1249fa3b58d1e38362f3b27f020d6a5295fcd2
       But local port with the same verion has a different SHA: 1991049d53f75d60a03bf5f8d58f26a8f4c85cbf
```
No idea why the SHA is wrong just for x86

Luke-Filewalker:
passes all checks now

ports/glslang/portfile.cmake
```diff
@@ -1,45 +1,49 @@
...
+if(NOT BUILD_BINARIES)
+  file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/bin)
+else()
+  file(RENAME ${CURRENT_PACKAGES_DIR}/bin ${CURRENT_PACKAGES_DIR}/tools)
+endif()
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include
+                    ${CURRENT_PACKAGES_DIR}/debug/bin)
+
+# Handle copyright
+file(COPY ${CMAKE_CURRENT_LIST_DIR}/copyright DESTINATION ${CURRENT_PACKAGES_DIR}/share/glslang)
```
JonLiu1993, line 49: @Luke-Filewalker ,We should replace "share/glslang" to "share/${PORT}"

ports/glslang/portfile.cmake
```diff
@@ -1,45 +1,49 @@
...
-  file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/bin)
-else()
-  file(RENAME ${CURRENT_PACKAGES_DIR}/bin ${CURRENT_PACKAGES_DIR}/tools)
-endif()
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include
-                    ${CURRENT_PACKAGES_DIR}/debug/bin)
-
-# Handle copyright
-file(COPY ${CMAKE_CURRENT_LIST_DIR}/copyright DESTINATION ${CURRENT_PACKAGES_DIR}/share/glslang)
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
JonLiu1993, line 1: Please revert the portfile.cmake format

JonLiu1993 (changes requested):

JonLiu1993 (approved):

JonLiu1993:
@Luke-Filewalker ,Please revert the portfile.cmake format,Keep it consistent with the previous file

Luke-Filewalker:
> 
> 
> @Luke-Filewalker ,Please revert the portfile.cmake format,Keep it consistent with the previous file
> 
> ports/glslang/portfile.cmake

Not exactly sure what you mean

JonLiu1993:
> > @Luke-Filewalker ,Please revert the portfile.cmake format,Keep it consistent with the previous file
> > ports/glslang/portfile.cmake
> 
> Not exactly sure what you mean

Please keep the original format of EOL.

JonLiu1993:
@ras0219-msft @ras0219 ,Could you help review this pr？Thanks

ports/glslang/CMakeLists-targets.patch
ras0219-msft: I was concerned about a possible `find_package(OGLCompiler)`, but I realize now that this is a `Targets.cmake`, not a `-config.cmake`.

Thanks for the explanation!

Luke-Filewalker:
Update to newest 1.5.0

versions/g-/glslang.json
```diff
@@ -1,9 +1,14 @@
...
-      "version-string": "2019-03-05",
-      "port-version": 3
+      "git-tree": "83d03e1409d04d16851f766f7b6f0f9f6154004b",
+      "version-string": "11.5.0",
+      "port-version": 0
+    },
+    {
+      "git-tree": "8e277afea288c18b4f80cbce0a6884b99d47b944",
+      "version-string": "11.4.0",
+      "port-version": 0
```
JonLiu1993, line 11: We should delete this

ports/glslang/vcpkg.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "glslang",
-  "version-string": "2019-03-05",
-  "port-version": 3,
+  "version-string": "11.5.0",
```
JonLiu1993, line 3: We should use "version" instaead of "version-string"

JonLiu1993 (changes requested):

ras0219-msft (approved):
We can't simply delete the entire history of a package; that would break all current users.

Instead, I looked into updating shaderc to fix the conflicts. While updating it, I noticed that our patches are actually modifying upstream's install to add a `glslang::` namespace. This violates our policy of adhering to upstream as much as possible.

While unravelling that difference, I also found a much simpler way to fix the `find_package(glslang)` without needing to maintain dozens of lines of patchfiles moving forward. I've pushed this improvement to your branch.

Assuming CI passes, LGTM. Thanks for the PR and sorry for the very long wait on this!

PR #15729, opened 2021-01-18
[vcpkg baseline] [mesa] Fix conflicts with angle

NancyLi1013:
**Describe the pull request**

The following flies are conflicts with `angle`. 
```
Installed by angle:x64-windows
    bin/libGLESv2.dll
    bin/libGLESv2.pdb
    debug/bin/libGLESv2.dll
    debug/bin/libGLESv2.pdb
    debug/lib/libGLESv2.lib
    lib/libGLESv2.lib
```
So use `-D gles-lib-suffix=_mesa` to fix this.


NancyLi1013:
@Neumann-A 

Could you please help confirm if we can remove these libs directly in `mesa` port?

Neumann-A:
don't remove them prefix them and maybe fix the pkgconfig file. 
angle libraries should probably also be prefixed.

NancyLi1013:
Thanks for your reply.

Do you mean add the prefix such as `angle_libGLESv2` and `mesa_libGLESv2` for the generated libs and dlls?

Neumann-A:
furthermore mesa conflicts with the opengl port on windows but in general the mesa opengl is a better alternativ then the windows provided opengl.

NancyLi1013:
> Yes but somebody but somebody needs to take the real name of libGLESv2. mesa has an explicit configure option to add the prefix.

So we need to make the prefix optional, right?



> furthermore mesa conflicts with the opengl port on windows but in general the mesa opengl is a better alternativ then the windows provided opengl.

For `opengl`, do we need to handle conflicts now?

Neumann-A:
> So we need to make the prefix optional, right?

make it optional via an added feature in mesa and make it the default for now until vcpkg has a better story to deal with those issues

 > For opengl, do we need to handle conflicts now?

probably. IMHO: building mesa and testing in CI is actual more important than copying opengl from the system.

NancyLi1013:
@Neumann-A 

I noticed that there is the option `gles-lib-suffix` in mesa. It seems you tried to use `-D gles-lib-suffix=_mesa` before, but you commented this option out. 

I think this option can solve the conflicts with angel, why do you not use it?

Neumann-A:
> I think this option can solve the conflicts with angel, why do you not use it?

I simply forgot about the conflict. But this should really be a default-feature and not hardcoded in the portfile.

ports/mesa/portfile.cmake
```diff
@@ -168,7 +168,7 @@ endif()
 vcpkg_configure_meson(
     SOURCE_PATH "${SOURCE_PATH}"
     OPTIONS 
-        #-D gles-lib-suffix=_mesa
+        -D gles-lib-suffix=_mesa
```
NancyLi1013, line 171:
```suggestion
        -Dgles-lib-suffix=_mesa

```

JackBoosY (approved):

PR #15752, opened 2021-01-19
[libsass/sassc] add new port/tool

Neumann-A:
Required by GTK 4.0 
extracted from #13100 for easier review and merge

### DIFF
--- a/ports/libsass/CONTROL
+++ b/ports/libsass/CONTROL
@@ -1,0 +1,7 @@
+Source: libsass
+Version: 3.6.4
+Port-Version: 0
+Description: LibSass - Sass compiler written in C++
+Homepage: https://github.com/sass/libsass
+Supports: !uwp
+
--- a/ports/libsass/portfile.cmake
+++ b/ports/libsass/portfile.cmake
@@ -1,0 +1,22 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO sass/libsass
+    REF 8d312a1c91bb7dd22883ebdfc829003f75a82396 # 3.6.4
+    SHA512  41e532c081804c889c3e7f9169bd90d7fc944d13817d5e3a4f8c19608ebb630db2b2b0e9c61a59a2446076f4093d5ec4cb4f767aa06fa0bf0f0687094e1a2913
+    HEAD_REF master
+    PATCHES remove_compiler_flags.patch
+)
+
+vcpkg_configure_make(
+    SOURCE_PATH ${SOURCE_PATH}
+    AUTOCONFIG
+    OPTIONS
+
+)
+vcpkg_install_make(MAKEFILE GNUmakefile)
+vcpkg_fixup_pkgconfig()
+vcpkg_copy_pdbs()
+
+
+# Handle copyright
+file(INSTALL ${SOURCE_PATH}/COPYING DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
--- a/ports/sassc/portfile.cmake
+++ b/ports/sassc/portfile.cmake
@@ -1,0 +1,34 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO sass/sassc
+    REF 46748216ba0b60545e814c07846ca10c9fefc5b6 # 3.6.1
+    SHA512 98c5943ec485251fd5e3f41bcfe80dbbc6e2f334d4b6947895d3821b30009c40fb7cb944403304cede70360a5dd0ac103262644ef37a56e0fa76163657fbcc32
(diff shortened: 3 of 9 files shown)


ports/sassc/portfile.cmake
```diff
@@ -0,0 +1,34 @@
...
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO sass/sassc
+    REF 46748216ba0b60545e814c07846ca10c9fefc5b6 # 3.6.1
+    SHA512 98c5943ec485251fd5e3f41bcfe80dbbc6e2f334d4b6947895d3821b30009c40fb7cb944403304cede70360a5dd0ac103262644ef37a56e0fa76163657fbcc32
+    HEAD_REF master
+    PATCHES remove_compiler_flags.patch
+)
+
+find_library(LIBSASS_DEBUG sass PATHS "${CURRENT_INSTALLED_DIR}/debug/lib/" NO_DEFAULT_PATH)
+find_library(LIBSASS_RELEASE sass PATHS "${CURRENT_INSTALLED_DIR}/lib/" NO_DEFAULT_PATH)
```
ras0219-msft, line 11: portfiles can't use `find_library()` because they'll use the library extensions of the host, not the target. The best option available is probably to either 
1. Patch the build to run pkg-config, since it seems that libsass produces a .pc file
2. Implement a big switch on the various targets (much worse)

ras0219-msft (changes requested):

ports/sassc/portfile.cmake
Neumann-A: > portfiles can't use find_library() because they'll use the library extensions of the host, not the target. The best option available is probably to either.

No they don't. `find_library` normally does not work in CMake script mode so it is manually setup in vcpkg to lookup the library for the target.
https://github.com/microsoft/vcpkg/blob/97c11b2eeae2ecaa56e7e01d51cc0f1b83de6f10/scripts/cmake/vcpkg_common_definitions.cmake#L125-L130

ports/libsass/CONTROL
```diff
@@ -0,0 +1,7 @@
+Source: libsass
+Version: 3.6.4
+Port-Version: 0
+Description: LibSass - Sass compiler written in C++
+Homepage: https://github.com/sass/libsass
+Supports: !uwp
+
```
NancyLi1013, line 7:
```suggestion
```

ports/libsass/portfile.cmake
```diff
@@ -0,0 +1,22 @@
...
+    REF 8d312a1c91bb7dd22883ebdfc829003f75a82396 # 3.6.4
+    SHA512  41e532c081804c889c3e7f9169bd90d7fc944d13817d5e3a4f8c19608ebb630db2b2b0e9c61a59a2446076f4093d5ec4cb4f767aa06fa0bf0f0687094e1a2913
+    HEAD_REF master
+    PATCHES remove_compiler_flags.patch
+)
+
+vcpkg_configure_make(
+    SOURCE_PATH ${SOURCE_PATH}
+    AUTOCONFIG
+    OPTIONS
```
NancyLi1013, line 13:
```suggestion
```
Seems there are no any `OPTIONS` arguments here. So it's unnecessary to add it.

ports/libsass/portfile.cmake
```diff
@@ -0,0 +1,22 @@
...
+
```
NancyLi1013, line 14:
```suggestion
```

ports/libsass/portfile.cmake
```diff
@@ -0,0 +1,22 @@
...
+)
+vcpkg_install_make(MAKEFILE GNUmakefile)
+vcpkg_fixup_pkgconfig()
+vcpkg_copy_pdbs()
+
+
```
NancyLi1013, line 20:
```suggestion
```

NancyLi1013 (changes requested):

ports/sassc/portfile.cmake
ras0219-msft: That's.... a bit horrifying. But I guess it makes this work for now.

ras0219-msft:
Thanks for the explanation and the new ports!

ports/sassc/portfile.cmake
Neumann-A: I don't know why you find it horrifying. Why would you ever want to have something different then the TARGET libraries if you can only build for the TARGET?

The one important thing for crosscompiling:
CMake does not know how to set it up. As such its setup will always be for the HOST since that is the default for everything in CMake. 

As such the user has to provide all the information in the toolchain. CMAKE_FIND_LIBRARY_ is in general not mentioned there since you typically don't crosscompile from linux to windows or the other way around the defaults are ok. Meaning x64-linux to arm-linux will have the same library suffix/prefix but is just located in a different search path

PR #15764, opened 2021-01-19
[opencolorio] Install OpenColorIOConfig.cmake

### LINKED ISSUE #15765: [opencolorio] installs incorrect CMake file
****
Installing opencolorio on Windows-64 installed these cmake files:
./installed/x64-windows/share/opencolorio/OpenColorIO-debug.cmake
./installed/x64-windows/share/opencolorio/OpenColorIO-release.cmake
./installed/x64-windows/share/opencolorio/OpenColorIO.cmake

But trying to use the package with find_package(OpenColorIO) looks for [ OpenColorIOConfig.cmake, opencolorio-config.cmake, or FindOpenColorIO.cmake ] according to CMake's output.  As a result, CMake fails to find the package.

Additionally, the instructions after a successful install say to use "opencolorio" with find_package(), but the cmake files on disk are named "OpenColorIO" with upper case letters.

**Proposed solution**

The .cmake files installed by installing the package should follow the persnickety naming rules of CMake.

ras0219:
Enable `find_package(OpenColorIO CONFIG REQUIRED)`.

Fixes #15765 

As a drive by, this also changes the version scheme to semver based on observation of upstream's release naming.

### DIFF
--- a/port_versions/baseline.json
+++ b/port_versions/baseline.json
@@ -4190,7 +4190,7 @@
     },
     "opencolorio": {
       "baseline": "1.1.1",
-      "port-version": 4
+      "port-version": 5
     },
     "opencolorio-tools": {
       "baseline": "1.1.1",
--- a/port_versions/o-/opencolorio.json
+++ b/port_versions/o-/opencolorio.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "c011ef89b247b0bc54a08016fe23793defd398a5",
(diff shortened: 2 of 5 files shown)


wrosecrans:
I quickly tested this on my machine and confirmed cmake found the package.  I believe this solves #15765

NancyLi1013 (approved):

ras0219-msft:
Thanks @wrosecrans for testing the fix!

PR #15767, opened 2021-01-20
[vcpkg] fix edit command on windows

strega-nil:
cc @ras0219


ras0219-msft:
Thanks for the quick fix Nicole :)

PR #15808, opened 2021-01-22
[vcpkg_configure_cmake] Introduce REQUIRE_ALL_PACKAGES

### LINKED ISSUE #15526: [gdal:x64-windows-static] missing geos dependencies
**Host Environment**
- OS: Windows 10 build 1607
- Compiler: Visual C++ 2017 00370-20001-21499-AA108 / VisualStudio.15.Release/15.9.29+28307.1300

**To Reproduce**
- `.\vcpkg install gdal --triplet=x64-windows-static`
- Write a basic program using gdal and compile/link with cmake.

Linking fails with missing symbols errors from the geos library.
Manually adding the geos package with cmake solves the linking errors (well... almost as it may generate conflicts with windows's `io.h` and `./include/geos/io.h` but that's another story).

Looking at `./ports/gdal/dependency_win.cmake` we see these declarations for geos
```
  # Setup geos libraries + include path
  set(GEOS_INCLUDE_DIR "${CURRENT_INSTALLED_DIR}/include" )
  set(GEOS_LIBRARY_REL "${CURRENT_INSTALLED_DIR}/lib/geos_c.lib" )
  set(GEOS_LIBRARY_DBG "${CURRENT_INSTALLED_DIR}/debug/lib/geos_cd.lib" )
```

But when geos is installed, it comes with two static libs, not one:
2021-01-08 16_10_05-lib

So I guess the geos declarations should be modified this way to be complete:
```
  # Setup geos libraries + include path
  set(GEOS_INCLUDE_DIR "${CURRENT_INSTALLED_DIR}/include" )
  set(GEOS_LIBRARY_REL "${CURRENT_INSTALLED_DIR}/lib/geos_c.lib ${CURRENT_INSTALLED_DIR}/lib/geos.lib" )
  set(GEOS_LIBRARY_DBG "${CURRENT_INSTALLED_DIR}/debug/lib/geos_cd.lib ${CURRENT_INSTALLED_DIR}/debug/lib/geosd.lib" )
```

ras0219:
A recurring problem with many builds are uses of autodetected dependencies; depending on the specific install order, different CI runs results in different results for detection which results in flaky, unrobust behavior.

This PR introduces new flags to `vcpkg_configure_cmake()` to control this: `REQUIRE_ALL_PACKAGES`, `DISABLE_PACKAGES`, and `OPTIONAL_PACKAGES`.

A deficiency in the current PR implementation is that disables are also applied to transitive `find_package()`/`find_dependency()` calls; this is problematic in that it (in the most general case) requires each port to calculate the transitive closure of packages used under the current configuration and compute a disable list using all that information. Ideally, we would only apply `DISABLE_PACKAGES` to the first level of `find_package()`, which allows these settings to only concern themselves with local information.


cenit:
Very nice and very useful. Might help also reducing patchsets in some ports (and improve their maintainability too)

cenit:
Is it necessary to remove the QUEIT keyword from ${ARGS}? Or putting REQUIRED at the end always wins?

scripts/buildsystems/vcpkg.cmake
```diff
@@ -605,20 +623,20 @@ macro(${VCPKG_OVERRIDE_FIND_PACKAGE_NAME} name)
         else()
             set(Boost_COMPILER "-vc140")
         endif()
-        _find_package(${ARGV})
+        _find_package(${ARGV} ${_REQUIRED})
     elseif("${name}" STREQUAL "ICU" AND EXISTS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include/unicode/utf.h")
```
Neumann-A, line 627: everything below should probably be installed within a port via a vcpkg-cmake-wrapper.cmake instead of being in the toolchain.

cenit: i agree. They belong to a time in which vcpkg-cmake-wrapper.cmake was not existing...

ras0219: Yep, but I didn't want to make that change in this PR.

cenit:
Just few years ago @Neumann-A (or me) would have triggered a full rebuild just after this PR was merged, enabling the REQUIRE_ALL_PACKAGES by default. And I would have really enjoyed the result of that hard work!
It would be a very reasonable default, to have more reproducible builds. But I understand that it would be very very difficult, with many broken ports at first. The outcome would be a better vcpkg too

ras0219:
We unfortunately can no longer do such "break the world" operations with the new versioning and registries features -- we've lost "total" atomicity in exchange for stability.

However, we do intend to make this a requirement for all new ports and to eventually apply this policy to every current port.

ports/gdal/vcpkg-cmake-wrapper.cmake
```diff
@@ -1,12 +1 @@
...
-find_path(GDAL_INCLUDE_DIR NAMES gdal.h HINTS ${CURRENT_INSTALLED_DIR})
-
-find_library(GDAL_LIBRARY_DEBUG NAMES gdal_d gdal_i_d gdal NAMES_PER_DIR PATH_SUFFIXES lib PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug" NO_DEFAULT_PATH REQUIRED)
-find_library(GDAL_LIBRARY_RELEASE NAMES gdal_i gdal NAMES_PER_DIR PATH_SUFFIXES lib PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}" NO_DEFAULT_PATH REQUIRED)
-
-select_library_configurations(GDAL)
-
-set(GDAL_INCLUDE_DIRS ${GDAL_INCLUDE_DIR})
-set(GDAL_LIBRARIES ${GDAL_LIBRARY})
+_find_package(${ARGS} CONFIG)
```
Neumann-A, line 1: This is indirectly providing a `FindGDAL.cmake` via a vcpkg created `-config.cmake`. Maybe finally decide to support a CMake module directory?

ports/gdal/portfile.cmake
```diff
@@ -183,32 +185,140 @@ else()
...
+        set(cppstdlib stdc++)
+    endif()
+
     vcpkg_configure_make(
         SOURCE_PATH ${SOURCE_PATH}
         AUTOCONFIG
         COPY_SOURCE
-        OPTIONS ${CONF_OPTS}
+        OPTIONS_RELEASE
+            "LIBS=${release_libs} -lm -lpthread -l${cppstdlib}"
```
Neumann-A, line 217: Does this generate a correct `*.pc` file with the dependencies in `requries` or will it just add all libraries to the link flags?

ras0219: This is just modifying the LIBS variable to get things linking.

Neumann-A: Yeah I know. But sometimes those get added to the generated pc file in `Libs:`.

ports/gdal/vcpkg-cmake-wrapper.cmake
ras0219: The reason why we don't use the module directory approach is twofold:

1. _Many_ projects do something like:
```
set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
```
which would break the "obvious" way to provide a module path.
2. Because vcpkg packages are the "producers", they should provide `-config.cmake` files, which are essentially strictly superior to Find modules (notably they support `find_dependency()`).

I took this approach for GDAL to replace the official target with a config file, since I felt that we could provide the full official interface easily enough.

Neumann-A: Concerning 1:
The reason projects does this is because the cmake docs tell them to set it. In my opinion, this is a case of outdated/Wrong docs since `CMAKE_MODULE_PATH` can be provided either by a toolchain or by a superproject/superbuild so only setting the value is conceptually wrong and needs correction. (i.e. `list(PREPEND)` or similar is the correct way to go. See e.g. VTK). Modifying&backuping the `CMAKE_MODULE_PATH` is for example also done by `VTK-Config.cmake`

Concerning 2:
If the information is provided by vcpkg it does not matter here if it is `<package>-config.cmake` or `Find<package>.cmake` because you can encode the same information in both. The config way would only be superior if you also supply a `<configfile>Version.cmake`. Although the `find_dependency` part is correct you just can wrap it into a minimal function in a `Find<package>.cmake`  to workaround that. 

My main nitpick here ist that with a `CMAKE_MODULE_PATH` you wouldn't need to use the indirection with the `vcpkg-cmake-wrapper.cmake` here. Furthermore redirection of **all** calls to `FindGDAL.cmake` is possibly breaking downstream users with a custom `FindGDAL.cmake` which does not define the same targets and/or variables as the vcpkg one without a way to bail out. So chose your poison very carefully.

ras0219-msft: Agreed that the current practice of many projects is "wrong"/"bad", but patching the world is never a good solution.

It is true that we are possibly overriding a package's custom FindGDAL.cmake, however these custom find scripts are (in general) fundamentally broken in the face of static libraries, because there's no way for them to know the precise dependencies that were consumed. Additionally, as long as we adhere to "only use vcpkg-cmake-wrapper to wrap built-in CMake modules", the probability of custom FindXYZ.cmake scripts is much lower (but non-zero -- we've both seen it happen).

Neumann-A: > because there's no way for them to know the precise dependencies that were used during the build

Unless they are using the `FindPkgConfig` module + `pkg_check_modules` in the `Find<Package>.cmake`.  If they also use the `IMPORTED_TARGET` option and link against the created target it should be always correct (unless there is something wrong with the pc files generated by vcpkg). Unfortunately, I have also seen that the results of `pkg_check_modules` is sometimes only used as hinting for `find_library` calls etc.  (Source: Qt6)

Since I saw a mention of changes to FindGDAL in the CMake 3.20 release notes you might want to have look at in the CMake RC to make sure everything is ok with the config here.

ports/gdal/portfile.cmake
```diff
@@ -183,32 +185,141 @@ else()
...
+        OUTPUT_VARIABLE release_libs
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    set(ENV{PKG_CONFIG_PATH} "${CURRENT_INSTALLED_DIR}/debug/lib/pkgconfig${VCPKG_HOST_PATH_SEPARATOR}${backup_PKG_CONFIG_PATH}")
+    execute_process(
+        COMMAND "${PKGCONFIG}" --libs libtiff-4 libxml-2.0 libpng json-c netcdf libopenjp2 geos expat
+        OUTPUT_VARIABLE debug_libs
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    set(ENV{PKG_CONFIG_PATH} "${backup_PKG_CONFIG_PATH}")
```
JackBoosY, line 204: Should be separarte this to a new vcpkg function?

JackBoosY (changes requested):

ports/gdal/portfile.cmake
ras0219: That may be reasonable, but I want to scope this PR to just the primary feature (`REQUIRE_ALL_PACKAGES`). We can consider extracting this procedure out in the future.

wrobelda:
Don't want to sound need or anything, but it is a blocker for a number of PRs right now, which is a bummer. Has this stalled?

ras0219-msft:
Note https://github.com/microsoft/vcpkg/pull/18393 and https://github.com/microsoft/vcpkg/pull/18473 for examples of transitive dependency issues.

dg0yt:
> Note #18393 and #18473 for examples of transitive dependency issues.

Or now simage on Windows: https://github.com/microsoft/vcpkg/pull/17519, https://github.com/microsoft/vcpkg/pull/17519

JackBoosY:

cenit:
😞 😢 😭

JackBoosY:
@cenit Don't worry, I believe this PR will continue.

PR #15822, opened 2021-01-22
[vcpkg] Remote-install (Github downloading and installing)

Mr-Leshiy:
It is a new feature ```remote-install``` command.
The purpose to add such new functionality, is to make process if distribution of some new libraries with the vcpkg in the more easy way.  I am introducing a new way how it is possible to install library using vcpkg.

It is the first version of such feature, it is only works with the github repo, but of course it is also possible to extend it and make it more flexible.

**So how it works ?**
Lets analyze the following example of the usage of the new command ```remote-install```.
Example:
```
./vcpkg remote-install nnoops-long-arith-lib --author-name=Mr-Leshiy
```
1. So here it is we prvide a package name ```nnoops-long-arith-lib``` and the author of such package ```--author-name=Mr-Leshiy```.
Package name it is should be a github repository name and author name - an owner of this repository on the github. So it is directly corresponds to the github url of the repository ```https://github.com/Mr-Leshiy/nnoops-long-arith-lib```.
2. Vcpkg will go the such repository add try to download an archive with the concrete naming which should corresponds to the repository name, in our example it will find the ```nnoops-long-arith-lib-vcpkg.zip``` archive. In such archive should be description how to install your library with the vcpkg (examples of such instructions you can find in the ```ports``` directory of the vcpkg repository)
3. Next step is to create directory with the name of the package in our example ```nnoops-long-arith-lib``` in the ports folder. 
```vcpkg/ports/nnoops-long-arith-lib```
4. After that download an archive ```nnoops-long-arith-lib-vcpkg.zip``` from the github repository to the ```vcpkg/ports/nnoops-long-arith-lib``` and unarchive it in this directory.
5. The final step is just execute an install process for such package. Executes  ```Install::perform_and_exit(install_args, paths, default_triplet); ```. 

I was inspired of the Rust and Golang programming languages, how it is possible to manage dependencies and directly point the destination of the dependency for the Github.
I think it will help for the wide use of the vcpkg and for the faster integration process of the new libraries which are not part of the vcpkg at this time.

### DIFF
--- a/toolsrc/include/vcpkg/base/downloads.h
+++ b/toolsrc/include/vcpkg/base/downloads.h
@@ -24,11 +24,15 @@ namespace vcpkg::Downloads
                                      const fs::path& path,
                                      const std::string& sha512);
 
+    std::string download_file(Files::Filesystem& fs, View<std::string> urls, const fs::path& download_path);
+
     // Returns url that was successfully downloaded from
     std::string download_file(Files::Filesystem& fs,
                               View<std::string> urls,
                               const fs::path& download_path,
                               const std::string& sha512);
+
+    void download_file(Files::Filesystem& fs, const std::string& url, const fs::path& download_path);
 
     void download_file(Files::Filesystem& fs,
                        const std::string& url,
--- a/toolsrc/include/vcpkg/input.h
+++ b/toolsrc/include/vcpkg/input.h
@@ -1,6 +1,10 @@
 #pragma once
 
+#include <vcpkg/base/cstringview.h>
+
 #include <vcpkg/packagespec.h>
+#include <vcpkg/triplet.h>
+#include <vcpkg/vcpkgpaths.h>
 
 namespace vcpkg::Input
 {
--- a/toolsrc/include/vcpkg/remote_install.h
+++ b/toolsrc/include/vcpkg/remote_install.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <vcpkg/commands.interface.h>
+#include <vcpkg/vcpkgpaths.h>
+
+#include <string>
+
+namespace vcpkg::RemoteInstall
+{
+    extern const CommandStructure COMMAND_STRUCTURE;
+
+    void perform_and_exit(const VcpkgCmdArguments& args, const VcpkgPaths& paths, Triplet default_triplet);
+
+    struct RemoteInstallCommand : Commands::TripletCommand
+    {
+        virtual void perform_and_exit(const VcpkgCmdArguments& args,
+                                      const VcpkgPaths& paths,
+                                      Triplet default_triplet) const override;
+    };
+}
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by ahoarau** [comment]
Super interesting. I'm seeing the next use to be like:
```
./vcpkg remote-install <repo> --author-name=<owner> --branch=feature/fa
./vcpkg remote-install <repo> --author-name=<owner> --tag=v1.2.3
./vcpkg remote-install <repo> --author-name=<owner> --commit=dhdh8w892hd
```

and even better:
```
./vcpkg remote-install my.private.repo.com --commit=dhdh8w892hd
```

Mr-Leshiy:
@ahoarau , Thank you for your feedback. Yes it is good point with the configurable tag or branch or commit of the repo. I think it make sense. Generally, I am going to make such functionality more flexible for the user. But now it is impossible, because I need to extend the arguments parse possibilities from the command line in the vcpkg, and such changes leads to the current tests failure. So if my suggestion will be suitable for the vcpkg, I will actively improve it !

ras0219:
Hi @Mr-Leshiy, thanks for the PR!

There are several important pieces that need to be addressed before we would move in this direction:
1. How does this work with manifest files?
2. What does this offer over registries, and why does that justify increasing the interface complexity of vcpkg? (https://github.com/microsoft/vcpkg/blob/master/docs/specifications/registries.md)
3. How does this work with versioning? A fundamental tenant of vcpkg is that remote state changes should not change the results of your build. (https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md)

I suspect the best way forward for this would be to make it easier to produce very simple git registries. I'd strongly encourage ironing out the design in issues before moving to implementation, because it could change dramatically.

Finally, the tool sources are soon to be extracted to https://github.com/Microsoft/vcpkg-tool so any eventual implementation will need to be made there :)

Thanks again!

Mr-Leshiy:
@ras0219 , Thank you for your answer.
This feature offers a simpler way how to add a new library to the vcpkg without your approving, and whole maintaining of the build for the vcpkg will be on the library side (meaning that all protfile.cmake and CONTROL files will be stored in the library directly). For me it seems more easier way how to add new library to the vcpkg.
But learning your reasons why you choose approach with registries (Legal requirements, Technical requirements etc.) I realise that my feature does not stands with these requirements.
About the versioning. It is working generally like in the original vcpkg, because my solution just download the build instructions from the remote repository, puts it into the ports directory, so next steps from the original vcpkg installing.

Can you detailed explain what do you mean with the 'simple git registries'.

@ras0219 , should I close this PR ? And is it nessary to move this feature to the vcpkg-tool rep ? What do you think ?

I am a C++ enthusiast who do make the usage of c++ more easier, and I found vcpkg very powerful tool how I can manage my dependecies in the projects. So I will be happy to somehow improve the vcpkg or to bring some new functionality :)

ras0219:
By "simple git registries", I mean a git registry with only one package in it.

Even though the documentation is sparse, I'd recommend trying to create your own git-based registry using the reference documentation in https://github.com/microsoft/vcpkg/blob/master/docs/specifications/registries.md. It's still in alpha and we know the documentation is insufficient, but we'd love to hear your feedback so far about any issues you encounter and how it does (or doesn't) meet your needs.

Mr-Leshiy:
@ras0219 , got it ! I will look into the git registries. Any feedback I will post in the 'issues'.
This PR I will close.

P.S. @ras0219 , are the vcpkg team going to add a 'Discussions' forum, https://docs.github.com/en/discussions/quickstart ? I think it is a good place for the discussion a new features, or to ask some questions etc.

PR #15831, opened 2021-01-23
[spirv-cross,spirv-headers,spirv-tools] 2021-01-15 update

barcharcraz:
**Describe the pull request**

- update spirv-tools to version 2020.6
- update spirv-headers to version 1.5.4.raytracing.fixed (a required update to get spirv-tools to build)
- update spirv-cross to version 2021-01-15

Additionally removed several no-longer-needed patches from spirv-tools.

I think that spirv-tools and spirv-cross actually will build shared libs on windows now, but I have not gone and gotten that working.


JackBoosY:
Most of them are good, but I have some questions:
1. `spirv-cross` generated additional cmake configuration files _share/spirv_cross_c/cmake/spirv_cross_cConfig.cmake_, they should be deleted.
2. `spirv-tools` generated two pkgconfig _lib/pkgconfig/SPIRV-Tools-shared.pc_ and _lib/pkgconfig/SPIRV-Tools.pc_ at the same time. Is this expected?
3. `spirv-tools` generated shared library _lib/libSPIRV-Tools-shared.so_ on `x64-linux`, that should be incorrect.
4. Note `glslang` generates spirv binary and header files(_include/SPIRV/GLSL.ext.AMD.h_, _lib/SPIRV.lib_ and _tools/spirv-remap.exe_), which may conflict with `spirv` hidden files or violate our policy (use the internal port of vcpkg instead of generating a separate copy).

barcharcraz:
why should we delete the config files from spirv-cross?

I'll check the package config files later today

why is the library incorrect on linux, I didn't touch the library names at all and I don't want to force the package to generate libraries with a linuxey name if it doesn't do that on it's own (I don't want to change the library names from upstream).

barcharcraz:
also I didn't touch glslang, I will check if spirv has an embedded dependency on glslang, but if not I'm not sure how to proceed there. I don't want to modify either package if they both generate those headers from the glsl specification xml.

JackBoosY:
> why should we delete the config files from spirv-cross?

Because all cmake configuration files should be installed in _share/NAME_, so I guess `vcpkg_fixup_cmake_targets(CONFIG_PATH share/spirv_cross_c/cmake TARGET_PATH share/spirv_cross_c)` is missing in `spirv-cross`'s `portfile.cmake`.

> why is the library incorrect on linux

Becuase our triplet `x64-linux` is a static triplet, all libraries should be built as static.

> also I didn't touch glslang

Yes, glslang's issue doesn't related to your changes, I just notify it.

barcharcraz:
@JackBoosY I have addressed the review comments, and also moved to a newer version of spirv-tools which required a newer (non released!) version of spirv-headers, I wrote the version as the date I accessed the repo.

JackBoosY:
Waiting for merge #16904.

ports/spirv-tools/portfile.cmake
```diff
@@ -33,17 +32,23 @@ vcpkg_configure_cmake(
         -DSPIRV_WERROR=OFF
         -DSPIRV_SKIP_EXECUTABLES=${SKIP_EXECUTABLES} # option SPIRV_SKIP_TESTS follows this value
         -DENABLE_SPIRV_TOOLS_INSTALL=${TOOLS_INSTALL}
+        -DSPIRV_TOOLS_BUILD_STATIC=ON
+        -DBUILD_SHARED_LIBS=OFF
```
JackBoosY, line 36:
```suggestion
```
We don't need to add this since `vcpkg_configure_make` already did that.

ports/spirv-tools/portfile.cmake
```diff
@@ -33,17 +32,23 @@ vcpkg_configure_cmake(
...
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/SPIRV-Tools-opt TARGET_PATH share/SPIRV-Tools-opt)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/SPIRV-Tools-reduce TARGET_PATH share/SPIRV-Tools-reduce)
+vcpkg_fixup_pkgconfig()
 
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/bin") # only static linkage, i.e. no need to preserve .dll/.so files
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+file(REMOVE "${CURRENT_PACKAGES_DIR}/bin/SPIRV-Tools-shared.dll")
+file(REMOVE "${CURRENT_PACKAGES_DIR}/lib/libSPIRV-Tools-shared.so")
+file(REMOVE "${CURRENT_PACKAGES_DIR}/debug/lib/libSPIRV-Tools-shared.so")
```
JackBoosY, line 51: Since you've already set `-DSPIRV_TOOLS_BUILD_STATIC=ON`, are these files exist?

ports/spirv-headers/portfile.cmake
```diff
@@ -2,8 +2,8 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO KhronosGroup/SPIRV-Headers
-    REF 1.5.1.corrected
-    SHA512 92447b1b1eca6f0253265f36d67b00c0c79e93f6a707e27bd8239bd2f693c468a92b7f7c3bb3fde6bb091383baaff42d3b0bbfeb9ff5f952d8a0b9626b03848e
+    REF bcf55210f13a4fa3c3d0963b509ff1070e434c79
+    SHA512 d0553b95f28b77209862059cd0a8c15ca3340f33e13d9bb75340ced07a5aa07b8b9eaa1bdc42daa0dbf78679c3b1ef3d344c73b17518061249cdc67000568c37
```
JackBoosY, line 6: Can we use a stable version instead of this?

JackBoosY (changes requested):

ports/spirv-tools/portfile.cmake
barcharcraz: it's built in all cases but not linked to be the tools.

ports/spirv-headers/portfile.cmake
barcharcraz: nope, the tools depend on unreleased code :(

JackBoosY (approved):

PR #15832, opened 2021-01-23
[vcpkg] clarify the actions of vcpkg_check_linkage

barcharcraz:
**Describe the pull request**

I got confused by the behavior of `vcpkg_check_linkage(ONLY_STATIC_LIBRARY)` in triplets that request dynamic libraries, so I documented it's behavior in this case.

### DIFF
--- a/docs/maintainers/vcpkg_check_linkage.md
+++ b/docs/maintainers/vcpkg_check_linkage.md
@@ -14,6 +14,8 @@ vcpkg_check_linkage(
 ### ONLY_STATIC_LIBRARY
 Indicates that this port can only be built with static library linkage.
 
+Note: If the user requested a dynamic build ONLY_STATIC_LIBRARY will result in a note being printed, not a fatal error.
+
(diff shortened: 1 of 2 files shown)


JackBoosY:
We have discussed this, here are the reasons:

Building a DLL that static links the CRT will have a separate copy of the CRT for every DLL library built this way – this causes serious size, correctness, and general surprise problems.
 
Building a static library against the DLL crt is totally fine however – you can static link in several different static libs and the final binary will just depend on the single (DLL) CRT.
 
The overarching rule is that it’s Ok to switch from static to dynamic (going down the dependency tree) but it’s Not Ok to switch from DLL to static.

barcharcraz:
Yes, although this PR does not have anything to do with CRT linkage, it's just documenting that vcpkg_check_linkage allows including static only libraries in dynamic builds, no matter what the CRT setting is set to.

JackBoosY:
If just add a comment here, I think it's acceptable.

JackBoosY (approved):

PR #15850, opened 2021-01-24
[drogon] Upgrade to v1.3.0

kotori2:
**Describe the pull request**

- What does your PR fix? 
Upgrades `drogon` version.
Fix cross compiling on ARM64 by adding a feature switch to disable building `drogon_ctl` tools because they are required to run at build time.
Also it fixes a linking error on macOS since upstream library `c-ares` requires it. https://github.com/an-tao/drogon/issues/688

I tested the following triplets:
```
drogon[ctl]:x64-windows
drogon:x64-windows
drogon[ctl]:x86-windows
drogon:x86-windows
drogon:arm64-windows (Cross compile with host x64)
drogon[ctl]:x64-osx
```

yes

### DIFF
--- a/ports/drogon/vcpkg.patch
+++ b/ports/drogon/vcpkg.patch
@@ -57,14 +57,3 @@ index 6df5dac..b79036d 100755
  
    # Find mysql, only mariadb client liberary is supported
    find_package(MySQL)
-@@ -251,7 +249,10 @@ if(BUILD_EXAMPLES)
- endif(BUILD_EXAMPLES)
- 
- if(BUILD_CTL)
-+  message("BUILD_CTL")
-   add_subdirectory(drogon_ctl)
-+else()
-+  message("no BUILD_CTL")
- endif(BUILD_CTL)
- 
- if(COZ_PROFILING)
--- a/ports/drogon/vcpkg.patch
+++ b/ports/drogon/vcpkg.patch
@@ -1,8 +1,25 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index 96d6c6a..9d3dd24 100755
+index 6df5dac..b79036d 100755
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -72,9 +72,9 @@ if(WIN32)
+@@ -5,13 +5,11 @@ project(drogon)
+ message(STATUS "compiler: " ${CMAKE_CXX_COMPILER_ID})
+ include(CheckCXXSourceRuns)
+ check_cxx_source_runs(
+-        ${PROJECT_SOURCE_DIR}/cmake/tests/binary_compatibility_test.cc
+-        cross_compiling)
++        "int main(){return 0;}"
++        not_cross_compiling)
+ 
+-if(cross_compiling)
++if(!not_cross_compiling)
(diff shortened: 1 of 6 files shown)


kotori2:
~~Imagine the CI is dead~~
Could someone re-initiate CI check please
```
    zlib[core]:x64-linux -> 1.2.11#9
Failed to take the filesystem lock on /agent/_work/1/s/.vcpkg-root:
    Device or resource busy
Exception: /agent/_work/1/s/scripts/azure-pipelines/end-to-end-tests-prelude.ps1:56
Line |
  56 |          throw "'$Script:CurrentTest' had a step with a nonzero exit c …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | 'default baseline 2' had a step with a nonzero exit code

##[error]PowerShell exited with code '1'.
```

JonLiu1993 (approved):

ports/drogon/vcpkg.patch
```diff
@@ -31,3 +48,23 @@ index 96d6c6a..9d3dd24 100755
...
+-  endif(pg_FOUND)
++  endif(PostgreSQL_FOUND)
+ 
+   # Find mysql, only mariadb client liberary is supported
+   find_package(MySQL)
+@@ -251,7 +249,10 @@ if(BUILD_EXAMPLES)
+ endif(BUILD_EXAMPLES)
+ 
+ if(BUILD_CTL)
++  message("BUILD_CTL")
```
dan-shaw, line 64: Can we remove comments?

ports/drogon/vcpkg.patch
```diff
@@ -1,8 +1,25 @@
...
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -72,9 +72,9 @@ if(WIN32)
+@@ -5,13 +5,11 @@ project(drogon)
+ message(STATUS "compiler: " ${CMAKE_CXX_COMPILER_ID})
+ include(CheckCXXSourceRuns)
+ check_cxx_source_runs(
+-        ${PROJECT_SOURCE_DIR}/cmake/tests/binary_compatibility_test.cc
+-        cross_compiling)
++        "int main(){return 0;}"
```
dan-shaw, line 11: If tests do not work, can we pass an option to skip or just comment them?

dan-shaw (approved):

ports/drogon/vcpkg.patch
kotori2: First of all `check_cxx_source_runs` takes raw code as argument, instead of file name. It is just a fix to make it work. 
This program requires the working `drogon_ctl` binary if they were compiled. This test mainly achive that, if someone wants to cross compile ARM binary on x86, this will automatically stop them from compiling `drogon_ctl` tool, which will fail the compile. 
Also for the option, I already added a `ctl` `feature` for that. It's just a fool-proof check.

ports/drogon/vcpkg.patch
kotori2: Sorry that was a debug string. Already removed.

ras0219-msft:
LGTM, thanks for the PR!

PR #15859, opened 2021-01-25
[OpenCV] bump to 4.5.1 and to 3.4.13

### LINKED ISSUE #11875: [opencv3] patch to vcpkg build of 3.4.7-2
Library name: opencv3[contrib]

Thanks for working on vcpkg and the opencv port! (I hope to soon be moving to opencv4, but until then ...) 

The opencv3[contrib] feature will use hdf5 when it is available. But the opencv3 contrib /modules/hdf/CMakeLists.txt is broken on windows. It has some hand-crafted logic for finding hdf5, and that seems to work for the release build. But it does not work for the vcpkg debug build. Besides it is not necessary because the vpkg build of hdf5 generates a correct config script on windows. So the find package logic for hdf5 on windows should be the same as non-windows, which is simply ``find_package(HDF5)``.

The attached zip file has 0010-win32-hdf5.patch and with an updated portfile.cmake that uses the patch.
[opencv3-patch.zip](https://github.com/microsoft/vcpkg/files/4758957/opencv3-patch.zip)

cenit:
Fixes: #11875
~Depends on: #15788~
(in order to continue tests and debug, I have integrated also quirc fixes from PR above)


cenit:
I cannot reproduce the OpenMVG regression on linux locally.
Any help is appreciated.

NancyLi1013 (approved):

NancyLi1013:
```
Found the following errors:
Error: While reading versions for port opencv from file: C:\a\1\s\versions\o-\opencv.json
       Local port version `4.5.1` exists in version file but it's not the first entry in the "versions" array.
Error: While reading versions for port opencv3 from file: C:\a\1\s\versions\o-\opencv3.json
       Local port version `3.4.13` exists in version file but it's not the first entry in the "versions" array.
Error: While reading versions for port opencv4 from file: C:\a\1\s\versions\o-\opencv4.json
       Local port version `4.5.1` exists in version file but it's not the first entry in the "versions" array.

    vcpkg x-add-version --all

```
Seems need to update version again. @cenit

cenit:
I am sure I did it many times. But merging with master which contained an older version on my branch which already added a newer version made this problem. 
Maybe the problem was merging with master, I should have rebased? I will try my best to fix the problems

NancyLi1013:
Currently, the versions will check every time once there are changes, which is annoying. But we have no better way to solve this.

All checks have passed now. Is necessary to test these features in this PR?

cenit:
there is a vcpkg-ci-opencv port which should be already testing all reasonable features in all triplets.
Please check that CI really tested opencv in all configs, then we should be ready

NancyLi1013:
Seems CI didn't test all ports. I will rerun to retrigger CI.

cenit:
@NancyLi1013 why is CI rebuilding everything?
Also, regression on uwp is clearly not due to this PR... must be something already on master!

NancyLi1013:
@cenit 

I'm not sure why CI rebuild everything. I just renew the latest commit to retrigger CI since ports seem to be not built before.

For the regressions, I only know `armadillo `failed with downloading source codes since there is something wrong with the sever.

Other failures need to be checked whether these are problems for the ports or CI problems.

cenit:
@NancyLi1013 can you retrigger again? The error on x64-osx was a filesystem one, should disappear if retried.

NancyLi1013:
All checks have passed now.

Thanks again for your contribution @cenit.

PR #15756, opened 2021-01-19
[vcpkg/meson] fix some details

Neumann-A:
- Add ADD_BIN_TO_PATH to vcpkg_install_meson
- Make x86 on x64_x86 a native build (unless uwp)
- Make native compiler unuseable in cross builds. 
- Bonus: Fix single configuration builds. 

changes extracted from #13100

PR #15867, opened 2021-01-25
[vcpkg] vcpkg_configure_meson: Support macOS cross-compile

past-due:
**Describe the pull request**

NOTE: Depends on #15659 ~~(the first 3 commits in this PR are from #15659, for running CI)~~ - now merged!

Based on discussion on Discord with @ras0219, this takes the approach of a minimal fix to the meson cross-compile config.

While this fixes cross-compiling *many* meson-based ports, there is one exception:

Meson-based ports that require "native" architecture tools to be built as part of their build process may still have issues. But as @ras0219 noted (on Discord), that is best fixed by port-level fixes after PR https://github.com/microsoft/vcpkg/pull/15424 is merged that take advantage of that new "host triplet dependency" functionality. Regardless, this fix is needed for the rest.

- What does your PR fix?

Fixes `arm64-osx` cross-compile of meson-based ports.

(Building meson-based ports for `arm64-osx` would previously build them for the native architecture - i.e. building `harfbuzz:arm64-osx` on an Intel 64-bit Mac, would seemingly succeed **but** actually build an `x86_64` architecture library.)

`arm64-osx`

### DIFF
--- a/scripts/cmake/vcpkg_configure_meson.cmake
+++ b/scripts/cmake/vcpkg_configure_meson.cmake
@@ -280,6 +280,15 @@ function(vcpkg_internal_meson_generate_cross_file _additional_binaries) #https:/
     endif()
 endfunction()
 
+function(vcpkg_internal_meson_convert_compiler_flags_to_list _out_var _compiler_flags)
+    string(REGEX REPLACE "( |^)(-|/)" ";\\2" ${_out_var} "${_compiler_flags}")
+    if(VCPKG_TARGET_IS_OSX OR VCPKG_TARGET_IS_IOS)
+        # special-case handling for "-arch <arm64/etc>" on macOS/iOS - this must be split into separate list entries
+        string(REGEX REPLACE "-arch ([^;]+)" "-arch;\\1" ${_out_var} "${${_out_var}}")
+    endif()
+    set(${_out_var} "${${_out_var}}" PARENT_SCOPE)
+endfunction()
+
 function(vcpkg_internal_meson_generate_cross_file_config _config) #https://mesonbuild.com/Native-environments.html
     if(VCPKG_TARGET_IS_WINDOWS)
         set(L_FLAG /LIBPATH:)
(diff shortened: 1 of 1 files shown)


JackBoosY:

past-due:
Rebased on top of the latest `master` branch.

past-due:
@ras0219-msft @JackBoosY Just for clarity, I would like to encourage the merging of this PR as-is. (Additional port fixes will depend on #15424, but I feel that those are out-of-scope for this PR and best left for a subsequent PR.) 

This small change fixes a number of ports (`harfbuzz:arm64-osx`, and everything that depends on it, being a _very_ important one).

past-due:
@JackBoosY `x86_windows` failure appears to be unrelated to this PR:

```
Found the following errors:
Error: While reading versions for port log4cpp from file: C:\a\1\s\versions\l-\log4cpp.json
       Version `2.9.1#4` was not found in versions file.
       Run:

           vcpkg x-add-version log4cpp

       to add the new port version.

    vcpkg x-add-version --all
```

scripts/cmake/vcpkg_configure_meson.cmake
```diff
@@ -280,6 +280,15 @@ function(vcpkg_internal_meson_generate_cross_file _additional_binaries) #https:/
     endif()
 endfunction()
 
+function(vcpkg_internal_meson_convert_compiler_flags_to_list _out_var _compiler_flags)
+    string(REGEX REPLACE "( |^)(-|/)" ";\\2" ${_out_var} "${_compiler_flags}")
```
Neumann-A, line 284:
```suggestion
    string(REGEX REPLACE "( +|^)(--?|/)" ";\\2" ${_out_var} "${_compiler_flags}")
```

past-due: @Neumann-A I presume this is an additional fix ported from other work? (The PR currently preserves the existing regex.) Should it be ported to other instances of the `"( |^)(-|/)" ";\\2"` regex in this file? (Or is this best left for a separate PR, because it would seem to have a larger separate impact?)

Neumann-A: it is from #15756

If it is added here I can just throw part of the code in #15756 away when I solve the merge conflict. So I don't have to  introduce an additional change into this code position. 

And why is your `-arch` flag not separated by the above regex?

Neumann-A: and yeah why is this function not also applied to the nativ file generator?

past-due: The native file generator should not be getting arch flags to begin with, and as mentioned above:
> Meson-based ports that require "native" architecture tools to be built as part of their build process may still have issues. But as @ras0219 noted (on Discord), that is best fixed by port-level fixes after PR #15424 is merged that take advantage of that new "host triplet dependency" functionality. Regardless, this fix is needed for the rest.

(The PR is now https://github.com/microsoft/vcpkg-tool/pull/5)

Which will have to wait until a later PR.

past-due:
All checks out with the new regex, and I'm pretty happy with that as the solution. (It fixes many other issues as well with splitting `CFLAGS` / `CXXFLAGS`.)

@Neumann-A, @ras0219-msft?

ras0219-msft:
LGTM, this is a clear improvement over the status quo. I'm sure there will be some additional edge cases found in the future, but I don't think that is worth holding up the improvement now.

PR #15882, opened 2021-01-26
[ngspice] build codemodel extension libraries

marekr:
Something I left out previously in this port.

ngspice has "codemodels" which are DLLs with a custom extension of *.cm adding more simulation capability. This update adds them to the port output.

Note, I'm not sure if this is correct but ngspice on all platforms including the official windows binaries wants the code models to reside in /lib/ngspice. I am not sure if this is correct for vcpkg.


JackBoosY (changes requested):
Are these extensions required? So we can change it to a feature.

marekr:
> Are these extensions required? So we can change it to a feature.

The library itself can run without them. But the use case of ngspice may encounter the need for these codemodel extensions by default since these extensions have been around for SPICE since 1991.

JackBoosY:
@marekr So I suggest to add it as a feature, and set to the default feature.

marekr:
@JackBoosY done

JackBoosY (approved):

ras0219-msft:
LGTM, thanks for the fix!

PR #15860, opened 2021-01-25
[CGAL] switch dep to gmp instead of mpir

Neumann-A:
enable builds in CI again

closes #15881

PR #15886, opened 2021-01-26
[baseline][colmap] Fix find dependency freeimage

JackBoosY:
The default find method only find the release version of freeimage.

Related: #15860.

### DIFF
--- a/ports/colmap/fix-dependency-freeimage.patch
+++ b/ports/colmap/fix-dependency-freeimage.patch
@@ -1,0 +1,14 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index b05097c..cdd8ca4 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -96,7 +96,8 @@ find_package(Boost REQUIRED COMPONENTS
+ 
+ find_package(Eigen3 REQUIRED)
+ 
+-find_package(FreeImage REQUIRED)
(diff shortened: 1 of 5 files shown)


NancyLi1013 (approved):

Neumann-A:
you might also want to pass `GMP_LIBRARY=<somepath>/gmpd.lib` via `OPTIONS_DEBUG` so that GMP in debug gets linked (Could be a windows only detail).
Small tip to analyse build logs: search|grep for `installed/<triplet>/debug/lib` in release builds and `installed/<triplet>/lib` in debug builds to discover wrong lib linkage.

JackBoosY:
@Neumann-A  I think the dependency cgal cannot be built successfully. 
Because cgal depends on mpfr and mpir, while mpfr depends on gmp, and there is a conflict between gmp and mpir. 
So we can ignore this point for now.

Neumann-A:
@JackBoosY If you look into CGAL you actually see a `FindGMP`. mpir is just mentioned there in a find_library call as a gmp replacement. That is why #15860 switches the dependency mpir to gmp. The problem is that that find_library call is not configuration aware there and always links the release lib of gmp since the debug version has a debug suffix (gmpd).

JackBoosY:
@Neumann-A  Yep, but I can't test it in the master branch, so I suggest fix this after your PR merge.

JackBoosY:
Ping @dan-shaw for merge this PR quickly. Some PRs require this PR.

PR #15887, opened 2021-01-26
[boost-concept-check] Fix warning C4834 when install pagmo2

Cheney-W:
**Describe the pull request**
Since the change: https://github.com/microsoft/STL/pull/1474 , [nodiscard] was added before some operator() function, this change causes the warning C4834: discarding return value of function with 'nodiscard' attribute occurs when install port pagmo2. 

The change of STL is by design.
The warning was reported in the concept_check.hpp file, which comes from the dependency boost-concept-check of pagmo2，so I add a patch to modify the source code of boost-concept-check.

### DIFF
--- a/ports/boost-concept-check/fix-warning-c4834.patch
+++ b/ports/boost-concept-check/fix-warning-c4834.patch
@@ -7,7 +7,7 @@ index abbadb7..cab58e5 100644
        void test(boost::false_type)
        {
 -          f(first,second);
-+          //f(first,second);
++          std::ignore = f(first,second);
            Return r = f(first, second); // require operator()
            (void)r;
        }
--- a/ports/boost-concept-check/fix-warning-c4834.patch
+++ b/ports/boost-concept-check/fix-warning-c4834.patch
@@ -7,7 +7,7 @@ index abbadb7..cab58e5 100644
        void test(boost::false_type)
        {
 -          f(first,second);
-+          std::ignore = f(first,second);
++          (void) f(first,second);
            Return r = f(first, second); // require operator()
            (void)r;
        }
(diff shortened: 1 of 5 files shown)


NancyLi1013 (approved):

StarGate-One:
@Cheney-W  /cc @NancyLi1013 @longnguyen2004 @dmillard @ras0219-msft @BillyONeal 
If the STL change is by design from microsoft and the offending port is pagmo2 either 
- An issue should be raised with maintainers boost-concept-check at boost.org (https://github.com/boostorg/concept_check)
- An issue should be raised with the maintainers of pagmo2 port should work around the issue
- but either way, this is an upstream issue not a vcpkg issue
- how is this going to effect downstream consumption of this boost-concept-check port with others expecting it to function as boost released it

Cheney-W:
@StarGate-One 
Thanks for your comment. I have submitted an issue: https://github.com/boostorg/concept_check/issues/31 to https://github.com/boostorg/concept_check, until they provide a solution, this PR will remain in its current state.

StarGate-One:
> @StarGate-One
> Thanks for your comment. I have submitted an issue: [boostorg/concept_check#31](https://github.com/boostorg/concept_check/issues/31) to https://github.com/boostorg/concept_check, until they provide a solution, this PR will remain in its current state.

@Cheney-W
- You have no idea what kind of issues this can cause downstream for all the users of vcpkg using the 21 boost ports this will effect?
- So should the boost maintainers send all the downstream issues this creates to you to remedy?
/cc: @ras0219 @vicroms

StarGate-One:
@Cheney-W Thank you a-lot. 👍😊
Due diligence was done by giving a chance to respond on how the other libraries would be effected by this library change.✔
Have a wonderful day and an even better week!😎

NancyLi1013 (approved):

NancyLi1013:
@ras0219-msft Could you please help review and merge this PR?

Thanks.

ras0219-msft:
LGTM, thanks for getting it checked against upstream!

PR #15892, opened 2021-01-26
[libvpx] Fix arm-uwp builds for v142 toolset

jwtowner:
This adds improved support for targeting UWP as well as ARM/ARM64 for regular Windows Desktop with the v142 toolset. Without this change, libvpx for the arm-uwp triplet currently fails to build with v142. This change does not modify the existing behavior for v141 and earlier.

Supported triplets when VCPKG_PLATFORM_TOOLSET is set to v142:

arm-uwp*
arm64-uwp*
x86-uwp*
x64-uwp*
arm-windows*
arm64-windows*
x86-windows*
x64-windows*

### DIFF
--- a/ports/libvpx/portfile.cmake
+++ b/ports/libvpx/portfile.cmake
@@ -11,6 +11,7 @@ vcpkg_from_github(
     PATCHES
         0001-vcxproj-nasm.patch
         0002-Fix-nasm-debug-format-flag.patch
+        0003-add-uwp-and-v142-support.patch
 )
 
 vcpkg_find_acquire_program(PERL)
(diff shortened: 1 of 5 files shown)


jwtowner:
I followed the directions to update the port version number in the _CONTROL_ file and ran `vcpkg x-add-version libvpx`, but I'm stilling a failure on the _Validate Version Files_ check for x86_windows. I'm not sure what I did wrong. Any ideas? Thanks.

ports/libvpx/portfile.cmake
```diff
@@ -11,6 +11,7 @@ vcpkg_from_github(
     PATCHES
         0001-vcxproj-nasm.patch
         0002-Fix-nasm-debug-format-flag.patch
+        0003-add-uwp-and-v142-support.patch
```
NancyLi1013, line 14: Could you please submit these changes to upstream?

NancyLi1013:
For the failures on x86-windows, you can run .`\vcpkg.exe x-add-version libvpx --overwrite-version `. Then submit the changes.

ports/libvpx/portfile.cmake
jwtowner: Yep, I'd be glad to get these changes upstreamed. In the meantime, I may have another change to this package for ARMv7 support for non-Windows platforms in the near future.

jwtowner:
> For the failures on x86-windows, you can run .`\vcpkg.exe x-add-version libvpx --overwrite-version `. Then submit the changes.
Excellent, thank you!

ras0219-msft:
Looks good to me, thanks for the PR!

PR #15898, opened 2021-01-27
[vcpkg] Teach vcpkg.targets to emit a .tlog, enabling up-to-date checks

ras0219:
See title.

Even a no-op check on Windows can take several seconds due to running `vcvars.bat`. This change fixes this in two ways:

1. Add `Inputs` and `Outputs` to `VcpkgInstallManifestDependencies`, causing it to not execute `vcpkg.exe` if the manifest has not been changed.
2. Add steps to emit a `.tlog` file, plugging in to the VCTools/MSBuild "fast up to date" system. Without this step, modifying only `vcpkg.json` will not cause `Ctrl-B` to actually build anything; with this step, it will correctly invoke vcpkg.exe and rebuild anything that has changed (due to header changes etc).

Due to limitations in the fast up-to-date system, vcpkg.json will be compared against any `.ilk` files. We therefore need to always trigger a re-link via `Link_MinimalRebuildFromTracking=false` if vcpkg.json is modified.

@strega-nil @yuehuang010

### DIFF
--- a/scripts/buildsystems/msbuild/vcpkg.targets
+++ b/scripts/buildsystems/msbuild/vcpkg.targets
@@ -30,6 +30,7 @@
 
     <VcpkgConfigSubdir Condition="'$(VcpkgNormalizedConfiguration)' == 'Debug'">debug\</VcpkgConfigSubdir>
     <VcpkgApplocalDeps Condition="'$(VcpkgApplocalDeps)' == ''">true</VcpkgApplocalDeps>
+    <ProjectStateLine>VcpkgTriplet=$(VcpkgTriplet):$(ProjectStateLine)</ProjectStateLine>
   </PropertyGroup>
 
   <!-- Import property page 'Vcpkg' -->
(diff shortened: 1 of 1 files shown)


scripts/buildsystems/msbuild/vcpkg.targets
```diff
@@ -68,15 +68,30 @@
...
-    <Exec Command="%22$([System.IO.Path]::Combine($(VcpkgRoot), 'vcpkg.exe'))%22 install --x-wait-for-lock --triplet %22$(VcpkgTriplet)%22 --vcpkg-root %22$(VcpkgRoot)\%22 %22--x-manifest-root=$(VcpkgManifestRoot)\%22 %22--x-install-root=$(VcpkgInstalledDir)\%22"
-          StandardOutputImportance="High" />
+          Condition="'$(VcpkgEnabled)' == 'true' and '$(VcpkgEnableManifest)' == 'true' and '$(VcpkgManifestInstall)' == 'true'"
+          Inputs="@(_VcpkgInstallManifestDependenciesInputs)"
+          Outputs="$(TLogLocation)VcpkgInstallManifest.read.1u.tlog">
+    <Message Text="Installing vcpkg dependencies to $(VcpkgInstalledDir)" Importance="High" />
+    <MakeDir Directories="$(TLogLocation)" />
+    <WriteLinesToFile File="$(TLogLocation)VcpkgInstallManifest.read.1u.tlog"
+                      Lines="@(_VcpkgInstallManifestDependenciesInputs -> '^%(Identity)');"
+                      Encoding="Unicode"
```
strega-nil, line 84: Won't this result in a utf-16 file? Is this correct?

strega-nil (approved):
I don't fully understand MSBuild, but this LGTM.

scripts/buildsystems/msbuild/vcpkg.targets
ras0219: Yes and yes :) The `u` in `.1u.tlog` means unicode, or so I've been told.

scripts/buildsystems/msbuild/vcpkg.targets
```diff
@@ -68,15 +68,30 @@
...
-          Condition="'$(VcpkgEnabled)' == 'true' and '$(VcpkgEnableManifest)' == 'true' and '$(VcpkgManifestInstall)' == 'true'">
-    <Message Text="Installing vcpkg dependencies" Importance="High" />
-    <Exec Command="%22$([System.IO.Path]::Combine($(VcpkgRoot), 'vcpkg.exe'))%22 install --x-wait-for-lock --triplet %22$(VcpkgTriplet)%22 --vcpkg-root %22$(VcpkgRoot)\%22 %22--x-manifest-root=$(VcpkgManifestRoot)\%22 %22--x-install-root=$(VcpkgInstalledDir)\%22"
-          StandardOutputImportance="High" />
+          Condition="'$(VcpkgEnabled)' == 'true' and '$(VcpkgEnableManifest)' == 'true' and '$(VcpkgManifestInstall)' == 'true'"
+          Inputs="@(_VcpkgInstallManifestDependenciesInputs)"
+          Outputs="$(TLogLocation)VcpkgInstallManifest.read.1u.tlog">
+    <Message Text="Installing vcpkg dependencies to $(VcpkgInstalledDir)" Importance="High" />
+    <MakeDir Directories="$(TLogLocation)" />
+    <WriteLinesToFile File="$(TLogLocation)VcpkgInstallManifest.read.1u.tlog"
```
yuehuang010, line 82: I would create the tlog after the Exec below, in the case that the Exec fails, the tlog wouldn't be created and thus this target could run again.

PR #15913, opened 2021-01-27
[sciplot] Add new port

CaeruleusAqua:
Add new port for sciplot


scripts/ci.baseline.txt
```diff
@@ -1381,6 +1381,8 @@ ryu:x64-windows-static=fail
 ryu:x64-windows-static-md=fail
 ryu:x86-windows=fail
 ryu::arm64-windows=fail
+sciplot:arm-uwp=fail
+sciplot:x64-uwp=fail
```
JackBoosY, line 1385: If sciplot doesn't support uwp, please add `Supports: !uwp` to `CONTROL` instead of adding this.

JackBoosY (changes requested):

JackBoosY:
Please revert changes about the version record, then re-generate the version record.

JackBoosY (approved):

ports/sciplot/CONTROL
```diff
@@ -0,0 +1,5 @@
+Source: sciplot
+Version: 0.1
```
ras0219-msft, line 2: Consider using `vcpkg.json` and a version schema other than `version-string`: https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md#version-schemes

You can convert the CONTROL to vcpkg.json using `vcpkg format-manifest /path/to/CONTROL`.

ports/sciplot/portfile.cmake
```diff
@@ -0,0 +1,22 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO sciplot/sciplot
+    REF v0.1.0
+    SHA512 846657530639ec66af052dd26b09170fc2b87590cfc017ac633795af301ac60d46682d55b59f7a88e50669fa3644df625c3001486001a648b0fc913b0c602d89
+    HEAD_REF vcpkg
```
ras0219-msft, line 6:
```suggestion
    HEAD_REF master
```

ras0219-msft (changes requested):
Thanks for the PR! Overall, this looks great to me.

It looks like the name `sciplot` is already in use by another package in the native code ecosystem: https://packages.debian.org/buster/libsciplot-dev. While this other package is not yet in vcpkg and is quite old (1996!), it is still available on modern Linux systems and we'd like to avoid confusion.

Given the https://github.com/sciplot/sciplot/blob/master/LICENSE file, I'd suggest `allanleal-sciplot`. As a secondary option, `sciplot-sciplot`, though I think this is much less clear.

CaeruleusAqua:
I have to discuss this with the library maintainer. Give me some time!

allanleal:
Hi @ras0219-msft . I understand your concert. 

I would say, however, that the chances for confusion among users between `sciplot` and `libsciplot-dev` are very minor because the vcpkg package `sciplot` will clearly indicate what project it actually represents and users of `sciplot` will learn at [sciplot.github.io](https://sciplot.github.io) that it is available at vcpkg under the name `sciplot`.

It would be truly great if we could have `sciplot` here, to be consistent across other package managers.

ras0219-msft:
>It would be truly great if we could have sciplot here, to be consistent across other package managers.

I agree, consistency across package managers is important. What other package managers are you considering?

https://repology.org/projects/?search=sciplot
https://libraries.io/pypi/sciplot
https://anaconda.org/conda-forge/r-sciplot

I don't mean to be difficult -- I really am concerned about user confusion when moving between ecosystems. As an example of this, the Guidelines Standard Library actually could not be named `gsl`, because there exists prior art of the GNU Scientific Library. Instead, the guidelines standard library was named `ms-gsl`: https://github.com/microsoft/vcpkg/blob/master/ports/ms-gsl/CONTROL.

CaeruleusAqua:
@allanleal Hey, I don't mind if the package is called "allenleal-sciplot". Should we just continue with "allenleal-sciplot" to close the deal?

allanleal:
I'm in the process of creating a conda-forge package `sciplot`. There is no name conflict there. It would be absolutely great if we could also have `sciplot` in `vcpkg` (there is no package here with this name, and the other ones have either a similar name or seem not to be maintained anymore). 

You can check status here: https://github.com/conda-forge/staged-recipes/pull/14071

allanleal:
@ras0219-msft conda-forge has just accepted the package under the name `sciplot`. 

https://github.com/conda-forge/staged-recipes/pull/14071

It will be soon packaged and available for download via `conda install conda-forge::sciplot`. 

I would really appreciate if we could also go with the name `sciplot` here in vcpkg.

ras0219-msft:
As of this writing, given that the original SciPlot project no longer has an accessible website (
http://www.ae.utexas.edu/\~rwmcm/SciPlot.html) and no longer has an accessible vcs system outside of Debian's mirror (https://salsa.debian.org/debian/sciplot), we are willing to merge this library under the originally proposed "sciplot". If the older sciplot project is submitted, it will need to be submitted under a different name (`rwmcm-sciplot`, etc).

ras0219-msft (approved):

allanleal:
Hi @ras0219-msft , many thanks for this. @CaeruleusAqua , many thanks also for creating this vcpkg port.

JackBoosY (approved):

PR #15926, opened 2021-01-28
[vcpkg] Emit parse errors to match '<file>:<line>:<col>: error: <text>' for better IDE compatibility

ras0219:
See title.

The built-in error log parsers in Visual Studio (MSBuild) expect the filename to come first on the error line. With this change, vcpkg parse errors are picked up by the error log and you can jump directly to the offending file location via double-click.

https://github.com/dotnet/msbuild/blob/ec6ed832e6a4adf94a2e6ff7e2f34d8e2da1cc05/src/Shared/CanonicalError.cs#L78

### DIFF
--- a/toolsrc/src/vcpkg-test/json.cpp
+++ b/toolsrc/src/vcpkg-test/json.cpp
@@ -234,7 +234,7 @@ TEST_CASE ("JSON track newlines", "[json]")
     auto res = Json::parse("{\n,", fs::u8path("filename"));
     REQUIRE(!res);
     REQUIRE(res.error()->format() ==
-            R"(Error: filename:2:1: Unexpected character; expected property name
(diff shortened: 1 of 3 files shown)


NancyLi1013:
The failures on x64-linux and x64-osx: 

```
../src/vcpkg-test/json.cpp:236: FAILED:
  REQUIRE( res.error()->format() == R"(Error: filename:2:1: Unexpected character; expected property name
   on expression: ,
                  ^
)" )
with expansion:
  "filename:2:1: error: Unexpected character; expected property name
     on expression: ,
                    ^
  "
  ==
  "Error: filename:2:1: Unexpected character; expected property name
     on expression: ,
                    ^
  "
```

NancyLi1013 (approved):

BillyONeal (approved):

PR #15929, opened 2021-01-28
[physx] Install PDBs for static library.

Hoikas:
The physx pdbs are copied into the `bin` directory, which is deleted under static configurations. This prevents debug information from being available to users in this configuration. This PR ensures pdbs are copied into `lib` when the library is built statically.

- What does your PR fix?
See above

No change

Yes

### DIFF
--- a/ports/physx/CONTROL
+++ b/ports/physx/CONTROL
@@ -1,4 +1,4 @@
 Source: physx
 Version: 4.1.1
-Port-Version: 6
+Port-Version: 7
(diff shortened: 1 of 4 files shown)


JackBoosY:
If building static physx generates pdb, it must be incorrect.

Hoikas:
Why mark this as invalid? If you build and consume the port, you will find that it produces debug information in pdbs instead of embedding it into the library. If there is a way to make CMake include the debug information in the .lib file, then I will update the pull request to do that.

JackBoosY:
@Hoikas  See #15934.

Hoikas:
Closing in favor of #15934.

PR #15934, opened 2021-01-29
[physx] Use /Z7 instead of /Zi to integrate the debug info when building static

JackBoosY:
When building windows-static, we should always use Z7 to embed debugging information in the static library.

Related: #15929.

### DIFF
--- a/ports/physx/fix-compiler-flag.patch
+++ b/ports/physx/fix-compiler-flag.patch
@@ -1,0 +1,38 @@
+diff --git a/physx/source/compiler/cmake/uwp/CMakeLists.txt b/physx/source/compiler/cmake/uwp/CMakeLists.txt
+index 2abab31..e3cd075 100644
+--- a/physx/source/compiler/cmake/uwp/CMakeLists.txt
++++ b/physx/source/compiler/cmake/uwp/CMakeLists.txt
+@@ -40,10 +40,10 @@ ENDIF()
+ 
+ # Cache the CXX flags so the other CMakeLists.txt can use them if needed
+ SET(PHYSX_CXX_FLAGS "/Wall /d2Zi+ /MP /WX /W4 /GF /GS- /GR- /Gd ${PHYSX_FP_MODE} ${PHYSX_WARNING_DISABLES}" CACHE INTERNAL "PhysX CXX")
+-SET(PHYSX_CXX_FLAGS_DEBUG "/Od ${WINCRT_DEBUG} /Zi" CACHE INTERNAL "PhysX Debug CXX Flags")
+-SET(PHYSX_CXX_FLAGS_CHECKED "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Checked CXX Flags")
+-SET(PHYSX_CXX_FLAGS_PROFILE "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Profile CXX Flags")
+-SET(PHYSX_CXX_FLAGS_RELEASE "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Release CXX Flags")
++SET(PHYSX_CXX_FLAGS_DEBUG "/Od ${WINCRT_DEBUG} /Z7" CACHE INTERNAL "PhysX Debug CXX Flags")
++SET(PHYSX_CXX_FLAGS_CHECKED "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Checked CXX Flags")
++SET(PHYSX_CXX_FLAGS_PROFILE "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Profile CXX Flags")
++SET(PHYSX_CXX_FLAGS_RELEASE "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Release CXX Flags")
+ 
+ # These flags are local to the directory the CMakeLists.txt is in, so don't get carried over to OTHER CMakeLists.txt (thus the CACHE variables above)
+ SET(CMAKE_CXX_FLAGS ${PHYSX_CXX_FLAGS})
+diff --git a/physx/source/compiler/cmake/windows/CMakeLists.txt b/physx/source/compiler/cmake/windows/CMakeLists.txt
+index c772333..a0b30d8 100644
+--- a/physx/source/compiler/cmake/windows/CMakeLists.txt
++++ b/physx/source/compiler/cmake/windows/CMakeLists.txt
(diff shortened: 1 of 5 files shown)


Hoikas:
This won't work because `PHYSX_CXX_FLAGS` are set on line 83 of the `portfile.cmake`.

Hoikas:
Previous comment in error, please ignore :smile:

Hoikas:
Will this prevent PDBs from being generated for dynamic builds?

JackBoosY:
@Hoikas  Nope, only for static build.

NancyLi1013 (approved):

ports/physx/fix-compiler-flag.patch
```diff
@@ -0,0 +1,38 @@
...
+ # Cache the CXX flags so the other CMakeLists.txt can use them if needed
+ SET(PHYSX_CXX_FLAGS "/Wall /d2Zi+ /MP /WX /W4 /GF /GS- /GR- /Gd ${PHYSX_FP_MODE} ${PHYSX_WARNING_DISABLES}" CACHE INTERNAL "PhysX CXX")
+-SET(PHYSX_CXX_FLAGS_DEBUG "/Od ${WINCRT_DEBUG} /Zi" CACHE INTERNAL "PhysX Debug CXX Flags")
+-SET(PHYSX_CXX_FLAGS_CHECKED "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Checked CXX Flags")
+-SET(PHYSX_CXX_FLAGS_PROFILE "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Profile CXX Flags")
+-SET(PHYSX_CXX_FLAGS_RELEASE "/Ox ${WINCRT_NDEBUG} /Zi" CACHE INTERNAL "PhysX Release CXX Flags")
++SET(PHYSX_CXX_FLAGS_DEBUG "/Od ${WINCRT_DEBUG} /Z7" CACHE INTERNAL "PhysX Debug CXX Flags")
++SET(PHYSX_CXX_FLAGS_CHECKED "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Checked CXX Flags")
++SET(PHYSX_CXX_FLAGS_PROFILE "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Profile CXX Flags")
++SET(PHYSX_CXX_FLAGS_RELEASE "/Ox ${WINCRT_NDEBUG} /Z7" CACHE INTERNAL "PhysX Release CXX Flags")
```
Neumann-A, line 16: Since this is CMake shouldn't this simply be CMAKE_CXX_FLAGS ...?

Hoikas: The PhysX cmake unconditionally overwrites `CMAKE_CXX_FLAGS` with the contents of `PHYSX_CXX_FLAGS` and the appropriate variable as seen here.

Neumann-A: So just do `SET(PHYSX_CXX_FLAGS "${CMAKE_CXX_FLAGS}" CACHE INTERNAL "PhysX Debug CXX Flags")` to stop the nonsense. PhysX will write the same values back into CMAKE_CXX_FLAGS.

Hoikas: `PHYSX_CXX_FLAGS` and these values are concatenated, so the result of that operation would be to effectively have both /Z7 and /Zi in the resulting `CMAKE_CXX_FLAGS`. The approach in this patch is correct.

Neumann-A: No this PR is not correct since it is not taking custom toolchains into account. 

depending on how `PHYSX_CXX_FLAGS(_CONFIG)` are put into `CMAKE_CXX_FLAGS(_CONFIG)` the right solution is either to set `PHYSX_CXX_FLAGS(_CONFIG)` to be empty or set them to `CMAKE_CXX_FLAGS(_CONFIG)`

Hoikas: Have you read the PhysX cmake? It has hard assumptions that will prevent custom toolchains from working even without this PR.

cenit:
Please take @Neumann-A review into consideration. Ports overwriting flags are really troublesome. They should use vcpkg flags and stop there

Hoikas:
@cenit @Neumann-A Please read the upstream PhysX cmake. It already manually sets per-compiler flags.

ports/physx/fix-compiler-flag.patch
cenit: in any case vcpkg should not put any patch in place to support even further those positions, that's the point. We already have proper flags in place, we should use them

cenit:
@Hoikas again: 

> in any case vcpkg should not put any patch in place to support even further those positions, that's the point. We already have proper flags in place, we should use them

Hoikas:
@cenit I personally welcome any pull request that will address this issue and fix the issue by which debugging information is not properly exported by this port. Do you intend to submit it instead of downvoting PR comments?

ports/physx/fix-compiler-flag.patch
Neumann-A: >Have you read the PhysX cmake? It has hard assumptions that will prevent custom toolchains from working even without this PR.

No I have not. If it is a hard requirement to built certain or all targets within a directory they should have used `add_compile_options` or for single targets `target_compile_options` instead of messing around with global user supplied values. 

From my point of view only `/arch:SSE2 /d2Zi+` are required for x86 while the arch flag can be dropped for x64. For me the setting of the compiler flags looks more like a toolchain while in itself.

cenit:
It's just a review of this PR. Which means, @JackBoosY maybe can just try avoid setting those flags and just inherit those passed through CMAKE_CXX_FLAGS by vcpkg itself.
The patch in this case might just resolve in removing /Zi without adding /Z7 back

ports/physx/fix-compiler-flag.patch
Neumann-A: Also the `/arch` flag on x86 is only required if they are actually using SSE2 intrinsics somewhere. (It could as well be an default optimization for SSE2 here without ever using those intrinsics.)

Hoikas: I agree that it would be nice if NVIDIA had done a better job with their cmake. Again, I personally welcome a better pull request or a review suggestion that supplies a better fix. As it stands, this is an annoying issue with an *official* triplet (x64-windows-static) that needs to be fixed.

JackBoosY:
The upstream does not consider the use of preset flags at all, I prefer to only fix the settings in windows here.

cenit:
77873f4 is very similar to what I meant! :)
Thank you @JackBoosY

ras0219-msft:
This LGTM, thanks everyone for chiming in on finding the right solution!

PR #15937, opened 2021-01-29
[meson] fix single-config builds

cenit:
Otherwise, single-config builds are broken

refs: https://github.com/microsoft/vcpkg/pull/12860#issuecomment-769601841 and beyond


ras0219-msft (approved):

JackBoosY (approved):

Neumann-A:
fun fact: A similar fix is also included in #15756 and I totally forgot about it.

ras0219-msft:
Thanks for the fix!

PR #15960, opened 2021-01-30
[vcpkg] fix X_VCPKG_APPLOCAL_DEPS_INSTALL

dhrdlicka:
**Describe the pull request**
This pull request fixes applocal.ps1 getting invoked with double `CMAKE_INSTALL_PREFIX` in install targets when `X_VCPKG_APPLOCAL_DEPS_INSTALL` is enabled:
```
C:/Users/David/Documents/hello/build/cmake_install.cmake(49):  message(-- Installing app dependencies for hello... )
-- Installing app dependencies for hello...
C:/Users/David/Documents/hello/build/cmake_install.cmake(50):  execute_process(COMMAND C:/Program Files/WindowsApps/Microsoft.PowerShell_7.1.1.0_x64__8wekyb3d8bbwe/pwsh.exe -noprofile -executionpolicy Bypass -file C:/Users/David/AppData/Local/Programs/vcpkg/scripts/buildsystems/msbuild/applocal.ps1 -targetBinary ${CMAKE_INSTALL_PREFIX}/C:/Program Files (x86)/Hello/bin/hello.exe -installedDir C:/Users/David/Documents/hello/build/vcpkg_installed/x64-windows/debug/bin -OutVariable out )
```

### DIFF
--- a/scripts/buildsystems/vcpkg.cmake
+++ b/scripts/buildsystems/vcpkg.cmake
@@ -526,6 +526,9 @@ function(x_vcpkg_install_local_dependencies)
     if(_VCPKG_TARGET_TRIPLET_PLAT MATCHES "windows|uwp")
         cmake_parse_arguments(PARSE_ARGV 0 __VCPKG_APPINSTALL "" "DESTINATION" "TARGETS")
         _vcpkg_set_powershell_path()
+        if(NOT IS_ABSOLUTE ${__VCPKG_APPINSTALL_DESTINATION})
+            set(__VCPKG_APPINSTALL_DESTINATION "\${CMAKE_INSTALL_PREFIX}/${__VCPKG_APPINSTALL_DESTINATION}")
+        endif()
         foreach(TARGET IN LISTS __VCPKG_APPINSTALL_TARGETS)
             get_target_property(TARGETTYPE ${TARGET} TYPE)
             if(NOT TARGETTYPE STREQUAL "INTERFACE_LIBRARY")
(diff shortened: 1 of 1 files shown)


scripts/buildsystems/vcpkg.cmake
```diff
@@ -553,7 +553,7 @@ if(X_VCPKG_APPLOCAL_DEPS_INSTALL)
             set(PARSED_TARGETS "")
 
             # Destination - [RUNTIME] DESTINATION argument overrides this
-            set(DESTINATION "${CMAKE_INSTALL_PREFIX}/bin")
+            set(DESTINATION "bin")
```
JackBoosY, line 556: I think we should use other methods to deal with this situation. We always need `CMAKE_INSTALL_PREFIX` here.

dhrdlicka: Using it there breaks CPack and other scenarios that might override the prefix at install time such as `cmake --install build --prefix <custom prefix>`.

dhrdlicka:
It seems like I overlooked `x_vcpkg_install_local_dependencies` always assuming a relative destination path, which was the real cause behind the double prefix bug I tried to fix. Added a check for that in the new commit.

JackBoosY:
cc @sandercox for review this PR.

scripts/buildsystems/vcpkg.cmake
ras0219-msft: Using the local path is correct here.

ras0219-msft:
LGTM, thanks for the fix!

PR #15965, opened 2021-01-31
[cairo] Make freetype and fontconfig optional.

Hoikas:
This moves freetype and fontconfig into a default feature. Cairo supports three font backends, with freetype+fontconfig being only one of them. Cairo supports the Windows font API, and Fontconfig is very slow to build on Windows and has poor support for community triplets x86-windows-static-md and x64-windows-static-md.

This also fixes an unrelated issue in the cairo port whereby the dependency library `glib` would not be properly found by the target `unofficial::cairo::cairo`.

- What does your PR fix?
See above

No change

Yes


Neumann-A:
#13100 will change cairo to its native build system which is autoconf

ras0219-msft:
This LGTM while #13100 is still in progress. The added feature makes sense for pre- and post-meson.

PR #15983, opened 2021-02-01
[x64-windows-release] add a single config community triplet based on x64-windows and make it work on some ports

cenit:
It might be very useful to catch regressions for single-config triplets. What do you think?
Also, debug config might be even in a worse state. Shall we test that in case you agree? Shall we test both, separately of course?

I will add all failed ports to ci.baseline.txt at the end, and hope for merge later.


JackBoosY:
cc @BillyONeal for review this PR.

cenit:
OK PR is ready to consider if it's interesting for merge.
I made sure that vcpkg-ci-opencv builds fine in single-config release mode, and that's enough for me :)

JackBoosY:
We will have a discuss with this.

BillyONeal:
This seems correct to me but I think we would want to add at least 3 machines to the pool to handle the additional configuration and we're at the cap of the number of cores we can request from Azure right now. (We want the pool to be able to go to at least 2 concurrent full rebuilds plus one more meaning post this change we would need at least 21 machines and we are currently set to 18)

dg0yt:
I would very much appreciate seeing this implemented soon. For vcpkg, this would increase the number of parallel runs, but it would reduce the number CI runs which result from additional PRs required to repair broken updates. But the net effect would be even more positive: The users of vcpkg create another pile of wasted CI runs when they stumble over broken updates.

Last not least, limiting CI to one configuration (usually release) is probably a popular measure to limit build time with the whole fan-in of ports like Qt or GDAL.

cenit:
> (usually release)

debug variants are broken in principle, as #16110 testifies

JackBoosY:
To be honest, the behavior of directly operating without judging the build type is very common in `portfile.cmake`.

dg0yt:
> To be honest, the behavior of directly operating without judging the build type is very common in `portfile.cmake`.

Which importance is given to the build type in the PR reviews?

Unfortunately, it is hard to test for port maintainers: You must modify the triplet files. You cannot use environment variables or command line options. And there is no reminder to handle such configurations in the documentation.

Wouldn't it be possible to handle an environment variable in `scripts/ports.cmake`? I would propose to do this after loading the target triplet file so that the environment variable can override defaults added by the user. And of course it needs to go into the API information.
With this change, it would be possible to ask maintainers to test port maintainers it at least one configuration single-type configuration.

dg0yt:
Side note: 
I would also suggest to define the exact spelling of the build types. I see them being tested with `STREQUAL "debug"` and with `MATCHES "[dD][eE][bB][uU][gG]"`. Normalization would also help with API identification.

JackBoosY:
@dg0yt In the past, our review of the build type was not strict, but now if we need to face it strictly, I think we need to provide the `vcpkg_` function to completely replace it.

dg0yt:
> debug variants are broken in principle, as #16110 testifies

I see. TLDNR: 

 > ... the include folder at the moment cannot come from the debug at all.

e.g. (fictive)
~~~
$ VCPKG_BUILD_TYPE=debug ./vcpkg install zlib
...
The folder /include is empty or not present. This indicates the library was not correctly installed.
    /home/kpa/Projekte/vcpkg/vcpkg/ports/zlib/portfile.cmake---
~~~

however:
~~~
$ VCPKG_BUILD_TYPE=release ./vcpkg install zlib
....
-- Configuring x64-linux-rel
-- Building x64-linux-rel
...
~~~

BillyONeal:
As I mentioned in https://github.com/microsoft/vcpkg/pull/15983#issuecomment-772183796, we don't have the hardware available to add more triplets to normal testing right now. Maybe we can reevaluate this next time we get budget requests etc. with our bosses (August? Ish? Maybe? Not positive)

We might look at deprecating arm-uwp in the medium term (moving it into community and out of regularly tested status) which might free up most of the hardware needed to test this.

dg0yt:
If a full CI run is too much, how about an early warning system which builds
- ports that have changed,
- reference ports for scripts that have changed?

This would at least cover the fact most developers probably don't test all supported triplets.

JackBoosY:
As you can see, regressions on many ports have appeared.
There are many ports that have not been specially processed according to `VCPKG_BUILD_TYPE`. Especially for the `install` command and the `copy` command.
This can also be caused if _debug/include_ is not processed in debug only mode.

JackBoosY:
I'd like @ras0219-msft to join this discussion.

cenit:
> As you can see, regressions on many ports have appeared.
> There are many ports that have not been specially processed according to `VCPKG_BUILD_TYPE`. Especially for the `install` command and the `copy` command.
> This can also be caused if _debug/include_ is not processed in debug only mode.

that’s right. and unfortunately it’s not just new ports, it’s also older ports that do not work anymore in release-only triplets. it’s “frustrating” to not have this check in ci

JackBoosY:
> that’s right. and unfortunately it’s not just new ports, it’s also older ports that do not work anymore in release-only triplets. it’s “frustrating” to not have this check in ci

May be we should do some hack in the install and copy command like execute_process. And I think we lack the policy for this part. @ras0219 @ras0219-msft what do you think about this?

cenit:
opencv is not tested anymore in single config, due to a cascade failure. I will try to find the culprit, should be in the ports added in 90044be because I remember opencv was tested last time

Please @BillyONeal reconsider this PR. For people using vcpkg for real work, this triplet is very important and keeping it under control is mandatory

JackBoosY:
The point is that we have not added review documents for a single configuration before, and there is a large amount of unprocessed code for a single configuration in the master branch.
The question is how to solve this issue once and for all in a brief introduction? Such as overloading `file(RENAME)`?

cenit:
ok finding the problem was easy. libwebp is considered broken in singleConfig by CI, but the problem is that it is not, and in fact i cannot reproduce it locally.

this is the error in https://dev.azure.com/vcpkg/public/_build/results?buildId=59426&view=logs&jobId=7ef837f1-cdda-595f-049e-ab56fcd4d1d2 : 

```
[1/1] cmd /c "cd .. && "D:/downloads/tools/cmake-3.21.1-windows/cmake-3.21.1-windows-i386/bin/cmake.exe" "D:/buildtrees/libwebp/src/1690184832-82bbc619db.clean" "-DWEBP_BUILD_ANIM_UTILS=OFF" "-DWEBP_BUILD_GIF2WEBP=OFF" "-DWEBP_BUILD_IMG2WEBP=OFF" "-DWEBP_BUILD_VWEBP=OFF" "-DWEBP_HAVE_SDL=OFF" "-DWEBP_BUILD_WEBPINFO=OFF" "-DWEBP_BUILD_WEBPMUX=OFF" "-DWEBP_BUILD_EXTRAS=OFF" "-DWEBP_NEAR_LOSSLESS=ON" "-DWEBP_ENABLE_SIMD=ON" "-DWEBP_BUILD_CWEBP=OFF" "-DWEBP_BUILD_DWEBP=OFF" "-DWEBP_ENABLE_SWAP_16BIT_CSP=OFF" "-DWEBP_UNICODE=ON" "-DCMAKE_MAKE_PROGRAM=D:/downloads/tools/ninja/1.10.2-windows/ninja.exe" "-DBUILD_SHARED_LIBS=ON" "-DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=C:/a/1/s/scripts/toolchains/windows.cmake" "-DVCPKG_TARGET_TRIPLET=x64-windows-release" "-DVCPKG_SET_CHARSET_FLAG=ON" "-DVCPKG_PLATFORM_TOOLSET=v142" "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY=ON" "-DCMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP=TRUE" "-DCMAKE_VERBOSE_MAKEFILE=ON" "-DVCPKG_APPLOCAL_DEPS=OFF" "-DCMAKE_TOOLCHAIN_FILE=C:/a/1/s/scripts/buildsystems/vcpkg.cmake" "-DCMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION=ON" "-DVCPKG_CXX_FLAGS=" "-DVCPKG_CXX_FLAGS_RELEASE=" "-DVCPKG_CXX_FLAGS_DEBUG=" "-DVCPKG_C_FLAGS=" "-DVCPKG_C_FLAGS_RELEASE=" "-DVCPKG_C_FLAGS_DEBUG=" "-DVCPKG_CRT_LINKAGE=dynamic" "-DVCPKG_LINKER_FLAGS=" "-DVCPKG_LINKER_FLAGS_RELEASE=" "-DVCPKG_LINKER_FLAGS_DEBUG=" "-DVCPKG_TARGET_ARCHITECTURE=x64" "-DCMAKE_INSTALL_LIBDIR:STRING=lib" "-DCMAKE_INSTALL_BINDIR:STRING=bin" "-D_VCPKG_ROOT_DIR=C:/a/1/s" "-D_VCPKG_INSTALLED_DIR=D:/installed" "-DVCPKG_MANIFEST_INSTALL=OFF" "-G" "Ninja" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=D:/packages/libwebp_x64-windows-release""
FAILED: ../CMakeCache.txt 
cmd /c "cd .. && "D:/downloads/tools/cmake-3.21.1-windows/cmake-3.21.1-windows-i386/bin/cmake.exe" "D:/buildtrees/libwebp/src/1690184832-82bbc619db.clean" "-DWEBP_BUILD_ANIM_UTILS=OFF" "-DWEBP_BUILD_GIF2WEBP=OFF" "-DWEBP_BUILD_IMG2WEBP=OFF" "-DWEBP_BUILD_VWEBP=OFF" "-DWEBP_HAVE_SDL=OFF" "-DWEBP_BUILD_WEBPINFO=OFF" "-DWEBP_BUILD_WEBPMUX=OFF" "-DWEBP_BUILD_EXTRAS=OFF" "-DWEBP_NEAR_LOSSLESS=ON" "-DWEBP_ENABLE_SIMD=ON" "-DWEBP_BUILD_CWEBP=OFF" "-DWEBP_BUILD_DWEBP=OFF" "-DWEBP_ENABLE_SWAP_16BIT_CSP=OFF" "-DWEBP_UNICODE=ON" "-DCMAKE_MAKE_PROGRAM=D:/downloads/tools/ninja/1.10.2-windows/ninja.exe" "-DBUILD_SHARED_LIBS=ON" "-DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=C:/a/1/s/scripts/toolchains/windows.cmake" "-DVCPKG_TARGET_TRIPLET=x64-windows-release" "-DVCPKG_SET_CHARSET_FLAG=ON" "-DVCPKG_PLATFORM_TOOLSET=v142" "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY=ON" "-DCMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP=TRUE" "-DCMAKE_VERBOSE_MAKEFILE=ON" "-DVCPKG_APPLOCAL_DEPS=OFF" "-DCMAKE_TOOLCHAIN_FILE=C:/a/1/s/scripts/buildsystems/vcpkg.cmake" "-DCMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION=ON" "-DVCPKG_CXX_FLAGS=" "-DVCPKG_CXX_FLAGS_RELEASE=" "-DVCPKG_CXX_FLAGS_DEBUG=" "-DVCPKG_C_FLAGS=" "-DVCPKG_C_FLAGS_RELEASE=" "-DVCPKG_C_FLAGS_DEBUG=" "-DVCPKG_CRT_LINKAGE=dynamic" "-DVCPKG_LINKER_FLAGS=" "-DVCPKG_LINKER_FLAGS_RELEASE=" "-DVCPKG_LINKER_FLAGS_DEBUG=" "-DVCPKG_TARGET_ARCHITECTURE=x64" "-DCMAKE_INSTALL_LIBDIR:STRING=lib" "-DCMAKE_INSTALL_BINDIR:STRING=bin" "-D_VCPKG_ROOT_DIR=C:/a/1/s" "-D_VCPKG_INSTALLED_DIR=D:/installed" "-DVCPKG_MANIFEST_INSTALL=OFF" "-G" "Ninja" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=D:/packages/libwebp_x64-windows-release""
-- The C compiler identification is MSVC 19.29.30133.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe - skipped
-- Performing Test HAVE_BUILTIN_BSWAP16
-- Performing Test HAVE_BUILTIN_BSWAP16 - Failed
-- Performing Test HAVE_BUILTIN_BSWAP32
-- Performing Test HAVE_BUILTIN_BSWAP32 - Failed
-- Performing Test HAVE_BUILTIN_BSWAP64
-- Performing Test HAVE_BUILTIN_BSWAP64 - Failed
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE  
-- Performing Test HAVE_PTHREAD_PRIO_INHERIT
-- Performing Test HAVE_PTHREAD_PRIO_INHERIT - Failed
-- Performing Test PTHREAD_CREATE_UNDETACHED
-- Performing Test PTHREAD_CREATE_UNDETACHED - Failed
-- Found OpenGL: opengl32   
-- Performing Test HAVE_MATH_LIBRARY
-- Performing Test HAVE_MATH_LIBRARY - Success
Broken installation of vcpkg port zlib
CMake Warning (dev) at D:/installed/x64-windows-release/share/zlib/vcpkg-cmake-wrapper.cmake:13 (_find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  CMake variable ZLIB_ROOT is set to:

    D:/installed/x64-windows-release

  For compatibility, CMake is ignoring the variable.
  C:/a/1/s/scripts/buildsystems/vcpkg.cmake:738 (include)
  D:/downloads/tools/cmake-3.21.1-windows/cmake-3.21.1-windows-i386/share/cmake-3.21/Modules/FindPNG.cmake:54 (find_package)
  C:/a/1/s/scripts/buildsystems/vcpkg.cmake:784 (_find_package)
  cmake/deps.cmake:67 (find_package)
  CMakeLists.txt:51 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found ZLIB: D:/installed/x64-windows-release/lib/zlib.lib (found version "1.2.11") 
-- Found PNG: D:/installed/x64-windows-release/lib/libpng16.lib (found version "1.6.37") 
CMake Error at D:/installed/x64-windows-release/share/jpeg/vcpkg-cmake-wrapper.cmake:5 (message):
  Broken installation of vcpkg port libjpeg-turbo
  C:/a/1/s/scripts/buildsystems/vcpkg.cmake:738 (include)
  cmake/deps.cmake:67 (find_package)
  CMakeLists.txt:51 (include)


See also "D:/buildtrees/libwebp/x64-windows-release-rel/CMakeFiles/CMakeOutput.log".
See also "D:/buildtrees/libwebp/x64-windows-release-rel/CMakeFiles/CMakeError.log".
```

but locally i cannot find any  problem with libwebp, nor libjpeg-turbo...

cenit:
might it be that there are some problems in PR cleanup procedure? that would explain also the problems about qt-activex that @Neumann-A is not able to reproduce and that created that huge cone of destruction of port-testing in CI

cenit:
well...
something suspicious is happening: why the x64-windows-release pipeline is building also 

```
  double-conversion:x64-windows:       pass: 2e250ff6f5106947eccdc13972ecd0a755b664cc6fa267e33b32cf27da0a5165
                      libpng:x64-windows:       pass: 0ddf0d1fbec664e51356848e89f9669e95345e42eaacf28d3c19a7f7ae4f3d44
                    freetype:x64-windows:       pass: a66a632f68267161dd8307e98e6b5db97f98307e07ad5d5f8e7e3c67fc52446e
                    harfbuzz:x64-windows:       pass: 230284e3ab21614bbff974ad7729ab74ed236ab66ae88e4bcac317c0b2f490a7
               libjpeg-turbo:x64-windows:       pass: 847bcf6f234b16e00f900df45829cfd8d9866811285096614d293c77545699a5
                     openssl:x64-windows:       pass: 9d26d03e2fe0ca73ab5d1587e7055d35e60e66344783665903e24b3ca7cbd798
                       libpq:x64-windows:       pass: cae83637c96ff64f53b65a0f63a7971cd15b33278c5fe01413b6383dd354b222
                       pcre2:x64-windows:       pass: 6cf65d40d93b8484ad3e8d6f425f21365167e33858f316ec13a3dbe548ad05e2
                     sqlite3:x64-windows:       pass: c4d902cfd6ed303c0fb91aa221a6dbbf86363fdc37aa9a083f8edd9401713d40
                        zstd:x64-windows:       pass: 43bde27e5c9402888c655d5d38940b4214d19724a0303638953c087ff1dc0a79
                      qtbase:x64-windows:       pass: 0cb58a6d470733126f71a9a125bfbee41cedc8602fbbd17a39f614d2c9450e37

```
x64-windows ports? it should not, and maybe that's where the confusion is coming, some wrapper that are finding wrong libraries afterwards (that's a problem, but still cannot understand why it's happening)

dg0yt:
@cenit Mind the host dependencies. Maybe you also want to set the host triplet, for a consistent universe.

Regarding the JPEG wrapper, I tried to consider release-only triplets, but I didn't test it actively. It will need another fix soon anyway (CMake < 3.12).

cenit:
ok, looks like that's the only reason, even if it looks strange to have host dependencies for those libraries...

dg0yt:
> ok, looks like that's the only reason, even if it looks strange to have host dependencies for those libraries...

Well, it is the same for the (default!) x86-windows triplet on x64-host.

For example Qt6, for it is family of build tools (qmake, moc, lrelease etc.):

~~~
$ grep -B1 '"host"' ports/qt*/vcpkg.json
ports/qt5-base/vcpkg.json-      "name": "vcpkg-pkgconfig-get-modules",
ports/qt5-base/vcpkg.json:      "host": true
--
ports/qtactiveqt/vcpkg.json-      "name": "qtactiveqt",
ports/qtactiveqt/vcpkg.json:      "host": true,
--
ports/qtbase/vcpkg.json-      "name": "qtbase",
ports/qtbase/vcpkg.json:      "host": true,
--
ports/qtdeclarative/vcpkg.json-      "name": "qtdeclarative",
ports/qtdeclarative/vcpkg.json:      "host": true,
--
ports/qtquick3d/vcpkg.json-      "name": "qtquick3d",
ports/qtquick3d/vcpkg.json:      "host": true,
--
ports/qtscxml/vcpkg.json-      "name": "qtscxml",
ports/qtscxml/vcpkg.json:      "host": true,
--
ports/qtshadertools/vcpkg.json-      "name": "qtshadertools",
ports/qtshadertools/vcpkg.json:      "host": true,
--
ports/qttools/vcpkg.json-      "name": "qttools",
ports/qttools/vcpkg.json:      "host": true,
~~~

cenit:
[OT] qtbase (qt6) requires qt built for host to build itself for target?? wow, didn't know, strange, ok! [/OT]

Neumann-A:
@cenit: This also applies to qt5 due to `moc uic rcc` (there are more but these are the most prominent one everybody knows) which are all tools which need to run on the host to build qt.

cenit:
ah right!

BillyONeal:
> Please @BillyONeal reconsider this PR. For people using vcpkg for real work, this triplet is very important and keeping it under control is mandatory

The budget problems blocking this PR remain unchanged from before. It will be June 2022 before budget changes may happen again; if we wanted to add this now we would need to remove an equivalent amount of compute (for example by removing one of the other big triplets or 2 small ones)

cenit:
this might be considered a small triplet, because it is only one config and the release one too (heavier for compute, but much lower for storage than debug)

what about just triggering it in PR and not also in CI, in the meantime?

cenit:
@dg0yt i removed your check in https://github.com/microsoft/vcpkg/pull/15983/commits/ba82eea0c996d98cd98ccdf00214ccf0f4d55234 because I think it can be more dangerous than useful.
Custom triplet might define release-only at a port level, so that check can trigger a failure for a basic port too often...

dg0yt:
> @dg0yt i removed your check in [ba82eea](https://github.com/microsoft/vcpkg/commit/ba82eea0c996d98cd98ccdf00214ccf0f4d55234) because I think it can be more dangerous than useful.

YMMV. 

> Custom triplet might define release-only at a port level, so that check can trigger a failure for a basic port too often...

Good point.

BillyONeal:
> this might be considered a small triplet, because it is only one config and the release one too (heavier for compute, but much lighter for storage than debug)

Storage costs approximately nothing.

Screenshot 2021-09-14 105231

> what about just triggering it in PR and not also in CI, in the meantime?

That would be even worse because PRs consume more compute than CI does.

dg0yt:
> > what about just triggering it in PR and not also in CI, in the meantime?
> 
> That would be even worse because PRs consume more compute than CI does.

As a start, I would be happy if CI would just build *modified* ports (e.g. just zlib, not all affected ports). I think that most single-config problems are in building a port, not in consuming it.
As a further mitigation, heavy ports might be skipped via baseline.

cenit:
the host/target mechanism is broken for single-config configs, which are seen as non-host :(
the tool are built for x64-windows and that triplet contains dbg+rel, and expects all target triplets to do so

dg0yt:
> In CI it does not work because ffmpeg dependency is tested fully-featured

There is a test-port, vcpkg-ci-ffmpeg, which you perhaps want to skip via baseline. It uses feature [all-nonfree].

cenit:
> There is a test-port, vcpkg-ci-ffmpeg, which you perhaps want to skip via baseline. It uses feature [all-nonfree].

i know about that port :)
but i cannot skip it, because otherwise opencv would be a cascade automatically.
It's better to try to fix it, or at last resort reduce its feature set in x64-windows-release triplet from CONTROL file

cenit:
@BillyONeal do you know why it's not possible to define --host-triplet when running "./vcpkg ci ..." ?

dg0yt:
> > There is a test-port, vcpkg-ci-ffmpeg, which you perhaps want to skip via baseline.
>
> but i cannot skip it, because otherwise opencv would be a cascade automatically.

Neither "opencv" nor "vcpkg-ci-opencv" depend on "vcpkg-ci-ffmpeg"?

cenit:
> > > There is a test-port, vcpkg-ci-ffmpeg, which you perhaps want to skip via baseline.
> > 
> > 
> > but i cannot skip it, because otherwise opencv would be a cascade automatically.
> 
> Neither "opencv" nor "vcpkg-ci-opencv" depend on "vcpkg-ci-ffmpeg"?

opencv depends on vcpkg-ci-ffmpeg, but i am not sure that skipping it automatically resumes the base ffmpeg port... do you think so? In any case, fixing the host-triplet (which I cannot do in CI mode...), the only regression remaining on my pc is tensorflow as a dependency of ffmpeg, after that everything should be ok

dg0yt:
> opencv depends on vcpkg-ci-ffmpeg

This is not possible (unless changed in this PR). vcpkg-ci-ffmpeg is from test_ports which is only available via overlay during ci, but not to normal users.
opencv doesn't even depend on ffmpeg:

~~~
$ ./vcpkg depend-info --overlay-ports=scripts/test_ports opencv:x64-windows
vcpkg-cmake: 
vcpkg-cmake-config: 
libjpeg-turbo: 
liblzma: vcpkg-cmake, vcpkg-cmake-config
zlib: 
libpng: vcpkg-cmake, vcpkg-cmake-config, zlib
libwebp[unicode, simd, nearlossless]: 
protobuf: 
quirc: 
tiff[lzma, jpeg, zip]: libjpeg-turbo, liblzma, vcpkg-cmake, zlib
opencv4[webp, tiff, quirc, png, jpeg, dnn]: libjpeg-turbo, libpng, libwebp, protobuf, quirc, tiff, vcpkg-cmake, vcpkg-cmake-config, zlib
opencv[webp, tiff, quirc, png, jpeg, dnn]: opencv4
~~~

dg0yt:
And for a really long list, vcpkg-ci-opencv with some ffmpeg features:
~~~
./vcpkg depend-info --overlay-ports=scripts/test_ports vcpkg-ci-opencv:x64-windows
opengl: 
zlib: 
libjpeg-turbo: 
freeglut: 
vcpkg-cmake-config: 
vcpkg-cmake: 
szip: vcpkg-cmake, vcpkg-cmake-config
jasper: freeglut, libjpeg-turbo, opengl, vcpkg-cmake, vcpkg-cmake-config
lcms: 
libiconv: 
liblzma: vcpkg-cmake, vcpkg-cmake-config
libpng: vcpkg-cmake, vcpkg-cmake-config, zlib
openblas: 
bzip2: 
brotli: 
libraw: jasper, lcms
polyclipping: 
libogg: 
libwebp[unicode, simd, nearlossless]: 
libxml2: libiconv, liblzma, vcpkg-cmake, vcpkg-cmake-config, zlib
openexr: zlib
openjpeg: 
lz4: vcpkg-cmake, vcpkg-cmake-config
openssl: 
kubazip: 
jxrlib: 
minizip: zlib
gflags: 
zstd: vcpkg-cmake, vcpkg-cmake-config
blas: openblas
vcpkg-gfortran: 
curl[ssl, sspi, schannel, non-http]: vcpkg-cmake, vcpkg-cmake-config, zlib
utfcpp: 
tiff[lzma, jpeg, zip]: libjpeg-turbo, liblzma, vcpkg-cmake, zlib
stb: 
freetype[zlib, png, bzip2, brotli]: brotli, bzip2, libpng, vcpkg-cmake, vcpkg-cmake-config, zlib
sqlite3: vcpkg-cmake, vcpkg-cmake-config
irrlicht: bzip2, libjpeg-turbo, libpng, vcpkg-cmake, vcpkg-cmake-config, zlib
giflib: vcpkg-cmake
rapidjson: 
hdf5[szip, zlib]: szip, vcpkg-cmake, vcpkg-cmake-config, zlib
poly2tri: 
sdl2: 
pugixml: 
proj4[tiff, database]: sqlite3, sqlite3, tiff, vcpkg-cmake, vcpkg-cmake-config
pcre2: 
zziplib: zlib
vcpkg-pkgconfig-get-modules: 
netcdf-c[netcdf-4, dap]: curl, hdf5, vcpkg-cmake, vcpkg-cmake-config
pegtl-2: 
llvm[lld, enable-terminfo, target-x86, enable-bindings, enable-zlib, enable-rtti, disable-clang-static-analyzer, disable-abi-breaking-checks, enable-threads, disable-assertions, default-targets, default-options, tools, compiler-rt, clang]: zlib
expat: 
freeimage: jxrlib, libjpeg-turbo, libpng, libraw, libwebp, openexr, openjpeg, tiff, zlib
eigen3: 
glew: opengl, vcpkg-cmake, vcpkg-cmake-config
glog: gflags
egl-registry: 
double-conversion: 
cuda: 
harfbuzz: freetype
icu: 
imgui[freetype]: freetype
jsoncpp: 
lapack-reference[noblas, blas-select]: blas, vcpkg-cmake-config, vcpkg-gfortran
leptonica: giflib, libjpeg-turbo, libpng, libwebp, openjpeg, tiff, zlib
libarchive[openssl, lzma, zstd, lz4, libxml2, bzip2]: bzip2, liblzma, libxml2, lz4, openssl, zlib, zstd
libharu[notiffsymbols]: libpng, tiff, vcpkg-cmake, zlib
assimp: irrlicht, kubazip, minizip, poly2tri, polyclipping, rapidjson, stb, utfcpp, zlib
libtheora: libogg
vtk: double-conversion, eigen3, expat, freetype, glew, hdf5, jsoncpp, libharu, libjpeg-turbo, liblzma, libogg, libpng, libtheora, libxml2, lz4, netcdf-c, pegtl-2, proj4, pugixml, sqlite3, tiff, utfcpp, zlib
cudnn: cuda
ceres: eigen3, glog
ade: 
tesseract: leptonica, libarchive
tbb: 
ffmpeg[swscale, swresample, avformat, avfilter, avdevice, avcodec]: vcpkg-cmake, vcpkg-pkgconfig-get-modules
gdcm: expat, openjpeg, zlib
quirc: 
qt5-base: double-conversion, egl-registry, freetype, harfbuzz, icu, libjpeg-turbo, libpng, openssl, pcre2, sqlite3, vcpkg-pkgconfig-get-modules, zlib, zstd
halide[target-x86, jit]: llvm, vcpkg-cmake, vcpkg-cmake-config
protobuf: 
lapack: lapack-reference
ogre[zziplib, freeimage, overlay, assimp]: assimp, freeimage, freetype, imgui, pugixml, sdl2, zlib, zziplib
opencv4[webp, sfm, quirc, qt, png, opengl, openexr, tiff, nonfree, openmp, lapack, tbb, jpeg, vtk, jasper, ipp, halide, ffmpeg, ovis, eigen, dnn, cudnn, cuda, contrib, gdcm, ade]: ade, blas, ceres, cuda, cudnn, eigen3, ffmpeg, gdcm, gflags, glog, halide, hdf5, jasper, lapack, libjpeg-turbo, libpng, libwebp, ogre, openexr, opengl, protobuf, qt5-base, quirc, tbb, tesseract, tiff, vcpkg-cmake, vcpkg-cmake-config, vtk, zlib
opencv[webp, sfm, quirc, qt, png, opengl, openexr, tiff, nonfree, openmp, lapack, tbb, jpeg, vtk, jasper, ipp, halide, ffmpeg, ovis, eigen, dnn, cudnn, cuda, contrib, gdcm, ade]: opencv4
vcpkg-ci-opencv: opencv
~~~

cenit:
this is how i am testing locally, so that it might be clearer  ;)

./vcpkg install vcpkg-ci-opencv --overlay-ports=scripts/test_ports --triplet x64-windows-release --host-triplet x64-windows-release --recurse --keep-going

and it works, now, start to end.
I still have to fix vcpkg-ci-ffmpeg (problems with bazel but should be already documented in port), then i have to understand how to trigger --host-triplet in CI

cenit:
> You can try to set the VCPKG_DEFAULT_HOST_TRIPLET environment variable e.g by adding as a variable to the single-config AZP job.

good idea.
I tried with https://github.com/microsoft/vcpkg/pull/15983/files#diff-b810c6816d42fd5c52f27e1ca0da28fd595a2bd00cbb448b443eff8e1739fe57R66 but it does not work

cenit:
another note: is host triplet considered in binary cache hash? It doesn't seem so

cenit:
`vcpkg-ci-opencv` can be built again in single config triplet (release only, debug-only is broken by design in vcpkg)

cenit:
@JackBoosY @BillyONeal if this PR is considered unmergeable, do you think it might be interesting to extract commits to fix ports for single-config triplets and at least having *now* working, with the high probability that tomorrow they will be broken again?

ports/zlib/vcpkg-cmake-wrapper.cmake
```diff
@@ -2,9 +2,6 @@ set(ZLIB_ROOT "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}")
 find_path(ZLIB_INCLUDE_DIR NAMES zlib.h PATHS "${ZLIB_ROOT}/include" NO_DEFAULT_PATH)
 find_library(ZLIB_LIBRARY_RELEASE NAMES zlib  z PATHS "${ZLIB_ROOT}/lib" NO_DEFAULT_PATH)
 find_library(ZLIB_LIBRARY_DEBUG   NAMES zlibd z PATHS "${ZLIB_ROOT}/debug/lib" NO_DEFAULT_PATH)
-if(NOT ZLIB_INCLUDE_DIR OR NOT ZLIB_LIBRARY_RELEASE OR (NOT ZLIB_LIBRARY_DEBUG AND EXISTS "${ZLIB_ROOT}/debug/lib"))
-    message("Broken installation of vcpkg port zlib")
-endif()
```
JackBoosY, line 7: Another PR related: https://github.com/microsoft/vcpkg/pull/19999

JackBoosY:
I think it's fine, I would accept all the changes except the triplet and baseline.
Also, I have some question, see comment above.

ports/zlib/vcpkg-cmake-wrapper.cmake
cenit: yes, that PR was born from this one and a comment from @dg0yt

dg0yt:
> I think it's fine, I would accept all the changes except the triplet and baseline.

Hm. IMO the main added value of this PR is exactly the triplet, baseline, and CI job.

At the moment, vcpkg has little to offer for single-config users and contributors. Not even ready-to-use triplet files. You can easily ask contributors to test all features of a port, but you cannot ask to test single-config.

JackBoosY:
Ping @BillyONeal for reply.

BillyONeal:
> Ping @BillyONeal for reply.

I don't have anything to add beyond what I've already mentioned. It isn't that the vcpkg team doesn't want this, it's that we don't have the budget to do it.

cenit:
> > Ping @BillyONeal for reply.
> 
> I don't have anything to add beyond what I've already mentioned. It isn't that the vcpkg team doesn't want this, it's that we don't have the budget to do it.

so if I remove the CI/PR job, would you merge this PR? are all corrections ok? (not that I feel so confident with jam scripts, for example... but the fixes seem to work ok locally and in CI)

cenit:
because at least we could mark again a day in which vcpkg-ci-opencv can be built from master in single-config (release only) mode. It will be broken in record time, I am sure, but maybe I will be surprised!

cenit:
now PR is "clean", no added jobs in future CI runs, it just fixes some ports for single-config triplets

dg0yt:
This PR includes changes to ports which are competing with other PRs. Even changes which are *not* fixing something which is really broken. This should not be pushed through in a batch PR.

JackBoosY:
Can you please solve the file conflicts?

cenit:
> Even changes which are not fixing something which is really broken

?
you mean about the checks of existance of debug/lib folder? Those checks are broken right now in single-config triplets.
Some ports, which are definitely doing wrong things, create debug artifacts also in single-config triplets, so those checks are broken right now and should be removed (also, as I already said, they are not correct in the idea itself because ports might be single-config at a port-level)

JackBoosY:
Can you please modify this PR title to suitable the changes?

dg0yt:
> > Even changes which are not fixing something which is really broken
> 
> ?
> you mean about the checks of existance of debug/lib folder? Those checks are broken right now in single-config triplets.
> Some ports, which are definitely doing wrong things, create debug artifacts also in single-config triplets, so those checks are broken right now and should be removed (also, as I already said, they are not correct in the idea itself because ports might be single-config at a port-level)

For zlib, the condition is wrong, but it is not a `FATAL_ERROR`. #19999 was updated to require not more but a single binary. For the libjpeg-turbo port I want to submit a similar change, but need to wait for #19034.
Again, the zlib check already helped uncover a serious issue in another port. Whether these checks are desired or not is unrelated to single-config support.

JackBoosY:
So can you please remove the triplet part and the baseline part?

cenit:
i have removed the baseline and the CI. i don't understand why should i remove also the triplet since it is in the community folder. It might be used by more people than other triplets already there, imho

JackBoosY (approved):

JackBoosY:
Oh sorry, it's community triplet. LGTM.

BillyONeal (approved):
Bikeshed: I think we should call it no-debug or release-only, or only-release or something like that to avoid people thinking they somehow need to build the "release" triplet before shipping their software.

cenit:
@yurybura can you please review user-config.jam for boost-modular-build in this PR?

yurybura (changes requested):
Could you revert all changes in ports/boost-modular-build-helper/user-config.jam?
Actual master version should works.

ports/zlib/vcpkg-cmake-wrapper.cmake
dg0yt, line 7: Please remove the zlib changes. It does not break single config builds, and it is subject of #19999.

dg0yt (changes requested):

ports/zlib/vcpkg-cmake-wrapper.cmake
cenit: sorry, but i have to repeat again that yes, it breaks single config builds. There are ports that build both configs also when triplet is single config and those break this check. Also, triplet might be single-config per port and that too would be a deal breaker for this kind of check.
Instead of fixing tens of ports, easier to remove this check (whose added value is still not clear to me).
About #19999, the lower the precedence, at least while not accepted, right?

dg0yt: - It doesn't break. It issues a message, not a `FATAL_ERROR`.
- For usefulness cf. https://github.com/microsoft/vcpkg/issues/20057#issuecomment-917690013. The check catched a package badly messing with library lookup which may have caused hard problems at runtime.
- If review of #19999 results in a request to remove the check, I'm fine with it. But it is not a drive-by for #15983.

BillyONeal:
> Bikeshed: I think we should call it no-debug or release-only, or only-release or something like that to avoid people thinking they somehow need to build the "release" triplet before shipping their software.

The maintainers discussed this in person and think the name requested originally by @cenit is fine.

BillyONeal:
I merged with the tensorflow 2.6 update. @cenit and @yurybura please figure out the best resolution for how the boost bits are affected; marking requires:author-response until that's dealt with.

After that I think we will merge this.

yurybura (changes requested):
Please revert file `ports/boost-modular-build-helper/user-config.jam`.
B2 build tool will handle `<variant>release` and `<variant>debug` by himself.

cenit:
it was not and that's why i touched the jam script (icu debug errors in many tested ports).

i will try it again asap restoring the script

JackBoosY:
Please resolve the file conflicts.

Thanks.

JackBoosY:
The tensorflow / tensorflow-cc regression are baseline regressions, not related to this changes.

BillyONeal:
> The tensorflow / tensorflow-cc regression are baseline regressions, not related to this changes.

That is https://github.com/microsoft/vcpkg/issues/21758

JackBoosY:
@cenit Can you please resolve the file conflicts?
I almost forgot this PR 😥

BillyONeal:
Many of the changes that were in this PR have been separated into smaller PRs and already landed in `master`; I reverted the ports that were left with naught but whitespace changes and similar in https://github.com/microsoft/vcpkg/pull/15983/commits/5793a2aa1c0a3b8e0ec31c272134e2851c28e3ae and rebuilt the version database to get this to pass.

BillyONeal (approved):

BillyONeal:
@Cenit Thank you so much for continuing with this despite the long review and fixing ports process to get here!

cenit:
it took 4 years but the triplet is finally official

BillyONeal:
> it took 4 years but the triplet is finally official

All you needed was <s>God-of-Quota</s>Azure to want the thing :)

PR #15984, opened 2021-02-01
[IGNORE PLEASE] Azure SDK for C++: verify on VcPkg's CI/CD

antkmsft:
Please ignore this PR - I am testing if a release we are going to publish relatively soon would have any problems, so we catch them early.

### DIFF
--- a/ports/azure-core-cpp/CONTROL
+++ b/ports/azure-core-cpp/CONTROL
@@ -1,0 +1,21 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+Source: azure-core-cpp
+Version: 1.0.0-beta.5
+Description: Microsoft Azure Core SDK for C++
+  This library provides shared primitives, abstractions, and helpers for modern Azure SDK client libraries written in the C++.
+Homepage: https://github.com/Azure/azure-sdk-for-cpp/tree/master/sdk/core/azure-core
+Default-Features: http
+
+Feature: http
+Build-Depends: azure-core-cpp[core, curl], azure-core-cpp[core, curl, winhttp] (windows&!uwp)
+Description: Build all HTTP transport implementations, depending on the platform
+
+Feature: curl
+Build-Depends: azure-core-cpp[core], curl
+Description: Build an HTTP transport implementation with LibCURL
+
+Feature: winhttp
+Build-Depends: azure-core-cpp[core]
+Description: Build an HTTP transport implementation with WinHTTP
--- a/ports/azure-identity-cpp/CONTROL
+++ b/ports/azure-identity-cpp/CONTROL
@@ -1,0 +1,9 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+Source: azure-identity-cpp
+Version: 1.0.0-beta.3
+Build-Depends: azure-core-cpp
+Description: Microsoft Azure Identity SDK for C++
+  This library provides common authentication-related abstractions for Azure SDK.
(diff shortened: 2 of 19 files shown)


ports/azure-core-cpp/CONTROL
```diff
@@ -0,0 +1,21 @@
...
+Feature: http
+Build-Depends: azure-core-cpp[core, curl], azure-core-cpp[core, curl, winhttp] (windows&!uwp)
+Description: Build all HTTP transport implementations, depending on the platform
+
+Feature: curl
+Build-Depends: azure-core-cpp[core], curl
+Description: Build an HTTP transport implementation with LibCURL
+
+Feature: winhttp
+Build-Depends: azure-core-cpp[core]
```
ras0219, line 20:
```suggestion
```

You don't need to list "core" here -- this is a feature within this package, so it automatically requires the package to be installed.

ports/azure-identity-cpp/CONTROL
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+Source: azure-identity-cpp
+Version: 1.0.0-beta.3
+Build-Depends: azure-core-cpp
```
ras0219, line 6: Note that this implicitly depends upon default features; use `azure-core-cpp[core]` if you don't want to require http.

ports/azure-core-cpp/CONTROL
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+Source: azure-core-cpp
```
ras0219, line 4: I recommend you look at producing `vcpkg.json` files instead -- most importantly, these files allow you to use new version fields such as `version-semver` which will allow users to take advantage of our new versioning feature.

https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md

You can convert from CONTROL to vcpkg.json using `vcpkg format-manifest /path/to/CONTROL`

ahsonkhan:
@antkmsft Do we need to fix the SHA value?

The CI is failing with sha mismatch errors (surprisingly only on win x86). Any ideas why?
https://dev.azure.com/vcpkg/public/_build/results?buildId=48820&view=logs&j=0a61404f-5c45-5632-e83c-408b7fcca1d6&t=a3e8ae58-f7e4-5657-8106-10b6dc738157
```text
Found the following errors:
Error: While reading versions for port azure-core-cpp from file: C:\a\1\s\versions\a-\azure-core-cpp.json
       File declares version `1.0.0-beta.5` with SHA: 1bcedf8a5777f656642f0a4f9ae6ad7352fa2087
       But local port with the same verion has a different SHA: c7b7a6dcc2fae649ddfe068fe48388ac3d4167fc

           vcpkg x-add-version azure-core-cpp

```

ports/azure-core-cpp/CONTROL
antkmsft: Thank you for all this feedback! We will address it for the next release, and try to add the version-semver.

ras0219-msft:
Yes, as part of adding versioning support to the tool, we need to maintain a database of all previous versions of the port.

The procedure you probably want to use for automation is:
```
> generate files
> git commit -am "temp"
> vcpkg x-add-version azure-core-cpp
> vcpkg x-add-version azure-core-common-cpp
> ...
> git commit --amend -am "[azure-sdk-for-cpp] Automatic update to $version"
```

This double-commit process is needed because we reference the git tree object for the port in the version database; the first commit creates all the intermediate git objects so that `x-add-version` can read them.

ras0219:
Should this be closed?

ahsonkhan:
Yes, we can close this draft PR. Thank you.

PR #16010, opened 2021-02-03
[directxsdk] Update DirectX SDK port

walbourn:
The legacy DirectX SDK is no longer hosted on Microsoft Downloads due to the [retirement of SHA-1 signed](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/sha-1-windows-content-to-be-retired-august-3-2020/ba-p/1544373) content.

This port does a number of things:

* It uses the archive.org copy. That has been verified to be genuine, and the SHA-256 hash inside this port has not changed.
* There's a warning from the port noting this is deprecated
* Per the guidance in this [blog post](https://walbourn.github.io/the-zombie-directx-sdk/), I trimmed out the headers/libs known to conflict with the Windows SDK
* Added optional features to add legacy **xinput1-3** and **xaudio2-7** which conflicts with the Windows 8 SDK / Windows 10 SDK headers -and- requires legacy setup. This supports opt-in use of the legacy bits.
* Added optional feature **xp** to add all the content needed to build with the Windows 7.1A SDK

> Note that the download speeds from archive.org are noticeably slower than they were on Microsoft Downloads.

### DIFF
--- a/ports/directxsdk/portfile.cmake
+++ b/ports/directxsdk/portfile.cmake
@@ -1,6 +1,6 @@
 vcpkg_fail_port_install(ON_TARGET "LINUX" "OSX" "UWP" "ANDROID" ON_ARCH "arm")
 
-message(WARNING "Build ${PORT} is deprecated, and requires the use of the DirectSetup legacy REDIST solution. See https://aka.ms/dxsdk for more information.")
+message(WARNING "Build ${PORT} is deprecated, untested in CI, and requires the use of the DirectSetup legacy REDIST solution. See https://aka.ms/dxsdk for more information.")
 
 vcpkg_download_distfile(ARCHIVE
     URLS "https://archive.org/download/dxsdk_2010/DXSDK_Jun10.exe"
--- a/ports/directxsdk/portfile.cmake
+++ b/ports/directxsdk/portfile.cmake
@@ -13,7 +13,6 @@ vcpkg_extract_source_archive_ex(
     ARCHIVE ${ARCHIVE}
 )
 
-
 # See https://walbourn.github.io/the-zombie-directx-sdk/
 set(INC_DIR "${SOURCE_PATH}/Include")
 set(LIB_DIR "${SOURCE_PATH}/Lib/${VCPKG_TARGET_ARCHITECTURE}")
--- a/ports/directxsdk/portfile.cmake
+++ b/ports/directxsdk/portfile.cmake
@@ -1,7 +1,9 @@
 vcpkg_fail_port_install(ON_TARGET "LINUX" "OSX" "UWP" "ANDROID" ON_ARCH "arm")
 
+message(WARNING "Build ${PORT} is deprecated, and requires the use of the DirectSetup legacy REDIST solution. See https://aka.ms/dxsdk for more information.")
+
 vcpkg_download_distfile(ARCHIVE
-    URLS "https://download.microsoft.com/download/a/e/7/ae743f1f-632b-4809-87a9-aa1bb3458e31/DXSDK_Jun10.exe"
+    URLS "https://archive.org/download/dxsdk_2010/DXSDK_Jun10.exe"
     FILENAME "DXSDK_Jun10.exe"
     SHA512 4869ac947a35cd0d6949fbda17547256ea806fef36f48474dda63651f751583e9902641087250b6e8ccabaab85e51effccd9235dc6cdf64e21ec2b298227fe19
 )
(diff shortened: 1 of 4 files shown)


ports/directxsdk/portfile.cmake
```diff
@@ -12,49 +14,136 @@ vcpkg_extract_source_archive_ex(
 )
```
NancyLi1013, line 16:
```suggestion
```

NancyLi1013 (changes requested):

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your contribution @walbourn.

ports/directxsdk/vcpkg.json
```diff
@@ -1,7 +1,18 @@
 {
   "name": "directxsdk",
-  "version-string": "jun10",
-  "description": "DirectX SDK",
-  "homepage": "https://www.microsoft.com/en-us/download/details.aspx?id=6812",
-  "supports": "windows & !arm & !uwp"
+  "version-string": "jun10b",
```
Hoikas, line 3:
```suggestion
  "version-string": "jun10",
  "port-version": 1,
```

This isn't really a new version of the SDK, just a change for the port to actually work.

Hoikas (commented):
Great work!

ras0219-msft (changes requested):
This LGTM with @Hoikas' suggested change. Thanks a ton @walbourn for figuring out a way forward here :)

ports/directxsdk/vcpkg.json
walbourn: Actually, I would recommend the version update. This version of the port doesn't just blindly install everything in the DXSDK like the previous version did. It installs a specific subset of files and provides opt-in features for others.

Hoikas: The new version is `jun10#1` instead of `jun10` - this is done by setting the `port-version` as shown above. The `version-string` should only change if the upstream provider is responsible for the update. See the [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#port-version) for more details.

walbourn: OK...

NancyLi1013:
Just need to run `vcpkg x-add-version directxsdk` and submit the changes to solve the versions check on x86-windows.

```
Found the following errors:
Error: While reading versions for port directxsdk from file: C:\a\1\s\versions\d-\directxsdk.json
       Version `jun10#1` was not found in versions file.
       Run:

           vcpkg x-add-version directxsdk

       to add the new port version.
```

NancyLi1013:
Ping @ras0219-msft for merge this PR.

polelf:
should make the following modifications, otherwise there will be link errors
For example
```
vcpkg install directxsdk 
vcpkg install qt5
```

patch:
```
 ports/directxsdk/portfile.cmake | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/ports/directxsdk/portfile.cmake b/ports/directxsdk/portfile.cmake
index 86fa8a149..3b8a41722 100644
--- a/ports/directxsdk/portfile.cmake
+++ b/ports/directxsdk/portfile.cmake
@@ -52,8 +52,8 @@ endif()
 
 #install(DIRECTORY "${SOURCE_PATH}/Include" DESTINATION ${CURRENT_PACKAGES_DIR}/include)
 file(COPY "${SOURCE_PATH}/Include/" DESTINATION ${CURRENT_PACKAGES_DIR}/include/${PORT})
-file(COPY ${RELEASE_LIBS} ${OTHER_LIBS} DESTINATION ${CURRENT_PACKAGES_DIR}/lib)
-file(COPY ${DEBUG_LIBS} ${OTHER_LIBS} DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib)
+file(COPY ${RELEASE_LIBS} ${OTHER_LIBS} DESTINATION ${CURRENT_PACKAGES_DIR}/lib/${PORT})
+file(COPY ${DEBUG_LIBS} ${OTHER_LIBS} DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib/${PORT})
 
```

NancyLi1013:
@walbourn 

Could you please address the above suggestion from @polelf ?

polelf:
ports/directxsdk/portfile.cmake



polelf@tom.com
 
From: NancyLi1013
Date: 2021-02-05 17:54
To: microsoft/vcpkg
CC: qiujun; Mention
Subject: Re: [microsoft/vcpkg] [directxsdk] Update DirectX SDK port (#16010)
@walbourn
Could you please address the above suggestion from @polelf ?

ras0219:
Hmm, it's not clear what to do here. While having the libraries directly in `lib/` may cause issues with qt5, if they aren't in `lib/` then they won't be linked by MSBuild users. If we do need to move them, they should go into `lib/manual-link` and _not_ `lib/${PORT}`.

@walbourn You mentioned above that you've removed some things that conflict with the Win10 SDK; are you aware of any remaining conflicts?

walbourn:
@ras0219  I did not change the locations of any of the install locations in this port in this update. I only changed the content. OTW, whatever issue qt5 is having is not new to my PR here.

The problem is that qt5 shouldn't be including legacy DirectX SDK libs as that's not supported by the DirectX SDK EULA in the first place. Exactly which filenames are the problem here @polelf ?

walbourn:
@ras0219 It looks like this PR is still blocked "awaiting changes" but I don't have any pending to take... The change from @polelf would make the new port work differently than the old one in terms of file locations.

walbourn:
@NancyLi1013 , @ras0219 ,  @ras0219-msft, I've tried:

```
vcpkg install directxsdk:x86-windows
vcpkg install qt5:x86-windows
```

```
vcpkg install directxsdk:x64-windows
vcpkg install qt5:x64-windows
```

I've never gotten any link errors.  I looked at the libraries installed by the **qt5** port and *none* of them conflict with the filenames of the libraries published by the **directxsdk** port with any of the features. Unless @polelf has some additional details on the repro or link errors, I don't see any problem here to fix. There's no reason to move the lib location from the previous version of this port.

This PR is still blocked 'requested changes' by @ras0219-msft but there's nothing to change at this point.

walbourn:
@NancyLi1013 , @ras0219 ,  @ras0219-msft, I've tried:

```
vcpkg install directxsdk:x86-windows
vcpkg install qt5:x86-windows
```

```
vcpkg install directxsdk:x64-windows
vcpkg install qt5:x64-windows
```

I've never gotten any link errors.  I looked at the libraries installed by the **qt5** port and *none* of them conflict with the filenames of the libraries published by the **directxsdk** port with any of the features. Unless @polelf has some additional details on the repro or link errors, I don't see any problem here to fix. There's no reason to move the lib location from the previous version of this port.

This PR is still blocked 'requested changes' by @ras0219-msft but there's nothing to change at this point.

polelf:
@walbourn you need clean vcpkg cache

ras0219-msft:
I need to do some investigation about the interaction here; no current action required by @walbourn

ras0219-msft:
Ok, I think the right path forward is to disable directxsdk in our CI. It is a deprecated port that only exists for legacy reasons, using it is against the EULA so you can't actually ship anything based on it. It's therefore not worth spending hours on flaky conflicts between itself and the world.

It will still be available for customers and should _generally_ work (thanks again @walbourn) after printing a nastygram.

PR #16033, opened 2021-02-04
[sdl2] Fix vulkan feature dependencies

past-due:
SDL2 ships with its own copy of required Vulkan headers, so the `vulkan` port dependency here is unneeded (and unused).

### DIFF
--- a/ports/sdl2/vcpkg.json
+++ b/ports/sdl2/vcpkg.json
@@ -1,15 +1,12 @@
 {
   "name": "sdl2",
   "version-string": "2.0.14",
-  "port-version": 1,
+  "port-version": 2,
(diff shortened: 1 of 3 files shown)


JackBoosY:
@past-due  Did you test the feature vulkan manually without install vulkan?

past-due:
> @past-due Did you test the feature vulkan manually without install vulkan?

Yes. In fact, the include paths in SDL2's CMakeLists.txt specify its built-in Vulkan headers directly:
https://github.com/SDL-mirror/SDL/blob/cf50b90a204fbc80df2d1003d01864b6ceab5e0c/CMakeLists.txt#L280-L284
https://github.com/SDL-mirror/SDL/tree/master/src/video/khronos

(It doesn't bother trying to find Vulkan headers separately.)

JackBoosY (approved):

ras0219-msft:
This LGTM, since the vulkan API is "stable"

PR #16045, opened 2021-02-04
Nanodbc/fix/config cmake files path linux

samyy321:
**Describe the pull request**

On Linux the `*cmake.config` files are not found after VCPKG install (see https://github.com/nanodbc/nanodbc/issues/260).
This PR provides the right path for those files (see the path on nanodbc CMake install https://github.com/nanodbc/nanodbc/blob/master/CMakeLists.txt#L268)

No changes

I updated the `port version` to `2` instead of `1` because it seems that it was reset to `0` without changing the package version (https://github.com/microsoft/vcpkg/commit/75bd2efe0dce862dfa025209b8cecc61a00a1852#diff-a787227381783470350e4b19cf34e27eabfefd76122c6836cab686377e59e10d). The consequence is that the `baseline.json` already contains a `port version` to `1` with same package version (https://github.com/microsoft/vcpkg/blob/master/versions/n-/nanodbc.json#L11), so to avoid conflict I updated to `port version` `2`.


ports/nanodbc/CONTROL
```diff
@@ -1,5 +1,6 @@
 Source: nanodbc
 Version: 2.13.0
+Port-Version: 2
```
JonLiu1993, line 3: Port-Version should be 1

samyy321: Is it ok to overwrite the already existing line in `nanodbc.json` as mentioned in my comment above ? It's the behavior with `x-add-version --overwrite-version nanodbc`

JackBoosY: @samyy321 You should just increase the port version once in one PR.

samyy321: Ok so I don't touch the `versions/-n/nanodbc.json` file (revert the commit) and I just put `Port-Version` in `ports/nanodbc/CONTROL` to `1`?

By the way I noticed that `versions/-n/nanodbc.json` contains 2 objects with same value for `version-string` and `port-version` ` {
      "git-tree": "921e300d4f37f15072f30eaaf382d2d27f598ac6",
      "version-string": "2.13.0",
      "port-version": 0
    },
    {
      "git-tree": "8323483e55b9eb32dd5f7801468e3daed8dff134",
      "version-string": "2.13.0",
      "port-version": 0
    },`, will it be a source of conflict ?

JackBoosY: Since we mistakenly lowered the Port-Version from 1 to 0 during the last update, I think it’s good now.

JackBoosY (changes requested):
Did you test the usage locally?

samyy321:
> 
> 
> Did you test the usage locally?

Yes, tested with `vcpkg install` and manifest files.
Also tested with `"version>="` in manifest files it works with `"version>=": "2.13.0#2"`.

samyy321:
It doesn't work with other versions on same vcpkg git branch (for example `"version>=": "2.13.0#1"` or `"version>=": "2.12.4"` in manifest) but it seems to be the case on the head of vcpkg master branch so maybe I misuse/misunderstood the feature or it's not finished?

samyy321:
> 
> 
> Vcpkg install fails if manifest file contains a dependency version constraint different than the one used in the current vcpkg git commit (for example if I use `"version>=": "2.13.0#1"` or `"version>=": "2.12.4"` instead of `"version>=": "2.13.0#2"` in my local vcpkg branch) but it seems to be the case on the head of vcpkg master branch so maybe I misuse/misunderstood the feature or it's not finished?
> 
> Example output of `vcpkg-manifest-install.log` after install fail:
> 
> ```
> Error: no version entry for nanodbc at version 2.13.0#1
> We are currently using the version in the ports tree (2.13.0#2).
> ```

Ok I actually misused it 😅, I didn't have a `builtin-baseline` fields in my `vcpkg.json`, so added it it works now.

ras0219-msft (approved):

JonLiu1993 (approved):

JackBoosY (approved):

samyy321:
Why still not merged ?
Waiting for next VCPKG release or some more checks ?
I noticed that it's the same for older approved pull requests but find nothing about merging in documentation.

JonLiu1993:
@ras0219-msft ，Please merge this pr.

JackBoosY:

PR #16063, opened 2021-02-05
Add disable-exceptions feature to portfile for tbb

aggieNick02:
This feature passes `exceptions=0` to tbbbuild in the non-windows case (which eventually defines `TBB_USE_EXCEPTIONS` as `0` in cpp code), and sets `TBB_USE_EXCEPTIONS=0` in the `vcxproj` files in the windows case. I wasn't sure if this counts as exposing an "alternative", but expect a decent number of tbb clients would want this option available. Happy to discuss as this is my first portfile modification attempt.

The effect of the feature is that it removes the `try`/`catch(...)` wrappers around user code run by TBB. While these exception facilities can be nice in some cases, their removal allows for much easier debugging of a crash due to an unhandled exception in code that a TBB client provides to a TBB algorithm. With the `try`/`catch(...)` wrappers removed, the unhandled exception and crash dump are generated at the point of the thrown exception, versus significantly later in a different thread with the originally throwing thread no longer having the stack from when the exception was thrown.

All triplets should be supported with this feature.

### DIFF
--- a/ports/tbb/CMakeLists.txt
+++ b/ports/tbb/CMakeLists.txt
@@ -1,9 +1,13 @@
 project(tbb CXX)
 
+option(DISABLE_EXCEPTIONS "Set exceptions=0 for make to turn off exception support in TBB" OFF)
 file(GLOB SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/*)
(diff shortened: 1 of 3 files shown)


aggieNick02:
Forgot to update version in portfile. Closing this PR and I'll try again, starting in draft mode.

PR #16069, opened 2021-02-06
[docs/users/manifests.md] Prioritize examples. Fix example. Add MSBuild integration documentation.

ras0219-msft:
See title. This is in response to feedback from users that the manifests documentation gave the impression of high complexity, and more directly that seeing an example upfront really helped their understanding.

The "complex" manifest example had two syntax errors out of the box:
1. `"supports"` is not a valid feature field
2. Misplaced comma in the features list

Finally, we only had documentation for CMake and not MSBuild in the manifests documentation.

### DIFF
--- a/docs/users/manifests.md
+++ b/docs/users/manifests.md
@@ -319,4 +319,4 @@ To use manifests with MSBuild, first you need to use an [existing integration me
 
 As part of your project's build, vcpkg automatically be run and install any listed dependencies to `vcpkg_installed/` adjacent to the `vcpkg.json` file; these files will then automatically be included in and linked to your MSBuild projects.
 
-It's critical that all project files consuming the same `vcpkg.json` use the same triplet; if you need to use different triplets for different projects in your solution, they must consume from different `vcpkg.json` files.
+It's critical that all project files consuming the same `vcpkg.json` use the same triplet in a single build; if you need to use different triplets for different projects in your solution, they must consume from different `vcpkg.json` files.
--- a/docs/users/manifests.md
+++ b/docs/users/manifests.md
@@ -31,6 +31,25 @@ and a little more information on [CMake](#cmake-integration) integration.
 Check out the [manifest cmake example](../examples/manifest-mode-cmake.md) for an example project using CMake and
 manifest mode.
 
+## Simple Example Manifest
+
+```json
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg/master/scripts/vcpkg.schema.json",
+  "name": "my-application",
+  "version": "0.15.2",
+  "dependencies": [
+    "boost-system",
+    {
+      "name": "cpprestsdk",
+      "default-features": false
+    },
+    "libxml2",
+    "yajl"
+  ]
+}
+```
+
 ## Writing a Manifest
 
 A manifest is a JSON-formatted file named `vcpkg.json` which lies at the root of your project.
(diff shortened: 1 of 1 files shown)


docs/users/manifests.md
```diff
@@ -302,3 +312,11 @@ since the CMake integration won't break as long as you depending on the exact na
...
 
 with a `vcpkg.json` in the same directory as `CMakeLists.txt` should Just Work!
+
+## MSBuild Integration
+
+To use manifests with MSBuild, first you need to use an [existing integration method](integration.md#With%20MSBuild). Then, simply add a vcpkg.json above your project file (such as in the root of your source repository) and set the property `VcpkgEnableManifest` to `true`. You can set this property via the IDE in `Project Properties -> Vcpkg -> Use Vcpkg Manifest`.
+
+As part of your project's build, vcpkg automatically be run and install any listed dependencies to `vcpkg_installed/` adjacent to the `vcpkg.json` file; these files will then automatically be included in and linked to your MSBuild projects.
+
+It's critical that all project files consuming the same `vcpkg.json` use the same triplet; if you need to use different triplets for different projects in your solution, they must consume from different `vcpkg.json` files.
```
strega-nil, line 322: This paragraph is really confusing to me. I think it's trying to document our stance on #16083, but I can't tell; I think it should be changed.

docs/users/manifests.md
```diff
@@ -302,3 +312,11 @@ since the CMake integration won't break as long as you depending on the exact na
 ```
 
 with a `vcpkg.json` in the same directory as `CMakeLists.txt` should Just Work!
+
+## MSBuild Integration
+
+To use manifests with MSBuild, first you need to use an [existing integration method](integration.md#With%20MSBuild). Then, simply add a vcpkg.json above your project file (such as in the root of your source repository) and set the property `VcpkgEnableManifest` to `true`. You can set this property via the IDE in `Project Properties -> Vcpkg -> Use Vcpkg Manifest`.
```
strega-nil, line 318:
```suggestion
To use manifests with MSBuild, first you need to use an [existing integration method](integration.md#with-msbuild). Then, simply add a vcpkg.json above your project file (such as in the root of your source repository) and set the property `VcpkgEnableManifest` to `true`. You can set this property via the IDE in `Project Properties -> Vcpkg -> Use Vcpkg Manifest`.
```

strega-nil (changes requested):

docs/users/manifests.md
autoantwort: And it is not true. We now use one `vcpkg.json` for projects with different triplets. But they must have different installed dirs. You can for example add 
```xml
  <PropertyGroup Label="Vcpkg">
    <VcpkgInstalledDir>$(VcpkgRoot)$(VcpkgPlatformTarget)_installed\</VcpkgInstalledDir>
  </PropertyGroup>
```
to a PropertySheet that is consumed by every project in the solution. Then it works.

docs/users/manifests.md
```diff
@@ -302,3 +312,11 @@ since the CMake integration won't break as long as you depending on the exact na
...
+
+As part of your project's build, vcpkg automatically be run and install any listed dependencies to `vcpkg_installed/` adjacent to the `vcpkg.json` file; these files will then automatically be included in and linked to your MSBuild projects.
```
autoantwort, line 320: Is that the case? For me they are installed in `$(VcpkgRoot)\installed\` and I have not changed any default.

ras0219-msft: The second paragraph is dependent upon the first -- you have to have set `VcpkgEnableManifest` to `true` and added a `vcpkg.json`.

docs/users/manifests.md
ras0219-msft: Do you have a suggestion that makes this clearer? It is documenting our stance on #16083.

ras0219-msft: I have not suggested changing install directories because that is not a publicly documented capability of our MSBuild integration.

docs/users/manifests.md
autoantwort: Yeah I have done that. `VcpkgEnableManifest` is `true` and I use a `vcpkg.json`, but the installed dir is still `$(VcpkgRoot)\installed\`. Maybe this is a bug, but that is the current situation for me.

strega-nil: That's not good...

docs/users/manifests.md
strega-nil: I think our stance is not this; we do support having multiple triplets with a single manifest file. We just don't support it well, yet. We need to fix that.

strega-nil:
```suggestion
It's critical that all project files consuming the same `vcpkg.json` use the same triplet in a single build; if you need to use different triplets for different projects in your solution, they must consume from different `vcpkg.json` files.
```

ras0219-msft:
@autoantwort This sounds like a serious bug -- could you open an issue with a repro?

autoantwort:
> 
> 
> @autoantwort This sounds like a serious bug -- could you open an issue with a repro project?

Done. If you have questions simply ask.

PR #16082, opened 2021-02-07
[libarchive] Fix static linking dependencies

DDoSolitary:
Fix configuration of libarchive's dependencies zstd and zlib, which would otherwise cause undefined symbols error when linking statically.


ports/libarchive/portfile.cmake
```diff
@@ -42,6 +42,8 @@ else()
 endif()
 
 list(APPEND FEATURE_OPTIONS -DENABLE_ZLIB=ON)
+# Needed for configure_file
+set(ENABLE_ZLIB ON)
```
NancyLi1013, line 46: `ENABLE_ZLIB` has been set as ON by default. Why do you still set it again?
https://github.com/libarchive/libarchive/blob/v3.4.3/CMakeLists.txt#L205

NancyLi1013 (changes requested):
Please also bump `Port-Version` in CONTROL file.

Also use the command `vcpkg x-add-version libarchive` to update the files in `versions `folder.
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#update-the-version-files-in-versions-of-any-modified-ports

ports/libarchive/portfile.cmake
DDoSolitary: Yes it has been set for building libarchive itself. However, it is also used in `vcpkg-cmake-wrapper.cmake.in`, so we need to set it for the current file so that the [`configure_file` call below](https://github.com/microsoft/vcpkg/blob/9199e697a1c775bba06474e95b42347d54f3d62d/ports/libarchive/portfile.cmake#L72) works properly.

DDoSolitary:
Versions bumped as requested.

ports/libarchive/portfile.cmake
NancyLi1013: Since `zlib` is the required dependency, I think we should always require `find_dependency(ZLIB)` in vcpkg-cmake-wrapper.cmake.in.

It might be needed to update these codes
```
if(@ENABLE_ZLIB@)
    find_dependency(ZLIB)
endif()
```
as

` find_dependency(ZLIB)`.

But I also wonder why the option `ENABLE_ZLIB` is not set as ON in this file.

Perhaps it is related with `cmake_policy(SET CMP0012 NEW)`?
https://cmake.org/cmake/help/latest/policy/CMP0012.html

DDoSolitary: > Since zlib is the required dependency, I think we should always require find_dependency(ZLIB) in vcpkg-cmake-wrapper.cmake.in.

I'm not sure but probably because it's configurable for libarchive itself. The "required" property seems to be enforced by the port's author.

> But I also wonder why the option ENABLE_ZLIB is not set as ON in this file.

The `configure_file` function is called in `portfile,cmake` so neither `${FEATURE_OPTIONS}` nor libarchive's CMakeLists.txt will take effect. Other options work because although undocumented, `vcpkg_check_features` also [sets the options for its parent scope](https://github.com/microsoft/vcpkg/blob/5eea585548058540c3b73d087eb101c033b5fa85/scripts/cmake/vcpkg_check_features.cmake#L175), but it is not the case for `ENABLE_ZLIB`, which is manually appended to `FEATURE_OPTIONS`.

DDoSolitary:
I have to idea why the x64_linux CI tests failed.

JackBoosY:
```
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory
```

NancyLi1013:
@DDoSolitary 

The failures on x64-linux are not related with this PR. I have rerun to rebuild.

NancyLi1013 (approved):

DDoSolitary:
I'm unable to reproduce the x64-linux error locally. Running `./vcpkg install --overlay-ports=scripts/test_ports cmake:x64-linux` on my Arch Linux installation on WSL works without issues.

The errors in windows triplets seems irrelevant. They failed due to some network errors.

DDoSolitary:
@JackBoosY

> c++: error: /mnt/vcpkg-ci/installed/x64-linux/debug/lib//libglib-2.0.a: No such file or directory

It's quite surprising that it's trying to link against glib. The cmake port doesn't depend on the glib port at all (not even intermediately, glib is not installed when I built cmake in a clean vcpkg installation).

JackBoosY:
cc @ras0219 @ras0219-msft
Can you please take a look?

Thanks.

DDoSolitary:
Strangely, after removing libglib-2.0.so from /usr/lib, build fails with another error

```
CMake Error at /root/vcpkg/installed/x64-linux/share/cmake/Qt5Core/Qt5CoreConfig.cmake:95 (message):                              
  Library not found: glib-2.0
```

So I still have no idea why it's unable to find the glib library in the CI build.

NancyLi1013:
Depends on PR #15808.

NancyLi1013:
@DDoSolitary 

DDoSolitary:
@NancyLi1013 Rebased.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your patience @DDoSolitary.

NancyLi1013:
LGTM now, thanks for your patience @DDoSolitary.

ytbigvinny27 (approved):

PR #16085, opened 2021-02-07
[netcdf-c] Fix usage

JackBoosY:
- `netcdf-c` generates cmake configure files in directory `netcdf`, not `netcdf-c`. Fix this.
- Remove usage since vcpkg can automatic generate it.

Related: #15375.

PR #15375, opened 2020-12-29
[osgearth] Fix x64-windows-static-md

ankurvdev:
**Describe the pull request**
* Fix osgearth for x64-windows-static-md

- What does your PR fix? Fixes #15301

- Which triplets are supported/not supported? Have you updated the CI baseline?

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

PR #16088, opened 2021-02-07
[gdal] Fix usage and add dependencies on UNIX

### LINKED ISSUE #15526: [gdal:x64-windows-static] missing geos dependencies

JackBoosY:
Since gdal relies on many ports, and using static gdal library under Unix also needs to use these dependencies, fix the dependencies in Unix.
Add required dependencies giflib and zstd.

Depends on #16085.
Related: #15375.


ports/gdal/vcpkg-cmake-wrapper.cmake
```diff
@@ -8,5 +8,54 @@ find_library(GDAL_LIBRARY_RELEASE NAMES gdal_i gdal NAMES_PER_DIR PATH_SUFFIXES
 
 select_library_configurations(GDAL)
 
+if (UNIX)
```
ras0219, line 11: This should probably switch based on static/dynamic, not on `UNIX`. This likely means the file should be renamed to `vcpkg-cmake-wrapper.cmake.in`, this line should become `if("@VCPKG_LIBRARY_LINKAGE@" STREQUAL "static")`, and we should use `configure_file( @ONLY)`.

JackBoosY: @ras0219 I think this suitable for `WINDOWS & dynamic` and `LINUX & static`.
What do you think about?

JackBoosY:
Waiting for merge #16138.

ras0219-msft:
This may also be fixed by #15808

JackBoosY:
@ras0219-msft Since your PR may last a long time, do I need to continue the PR?

JackBoosY:
Replaced by #15808.

PR #16095, opened 2021-02-07
[vcpkg] Fix unsupported modifiers for X_VCPKG_APPLOCAL_DEPS_INSTALL.

crud89:
Issue #1653 and PR #14243 introduced support for automatic installation of dependencies when using `cmake install`. However, the current implementation raises an error, if the `INSTALL(TARGETS)` command is used with the `EXPORT` and/or `INCLUDES DESTINATION` modifier.

`INSTALL(TARGETS mytarget EXPORT mytargetConfig ...)` associates the specified targets with an config file that can be later be installed separately using `INSTALL(EXPORT mytargetConfig ...)`, whilst the `INCLUDES DESTINATION` specifies a list of include directories that will be added to the export configuration.

The [CMake documentation](https://cmake.org/cmake/help/latest/command/install.html#targets) also lists those modifers.


ras0219-msft:
Thanks for the PR!

PR #16107, opened 2021-02-08
[lua] support iOS triplets

luncliff:
### What does your PR fix?

* create a new patch with macro `__APPLE__` and `TARGET_OS_OSX`
   * iOS triplets' target platform prohibits `system()` function

* Separate `lua[tools]` for Lua interpreter/compiler
   * This may surprise port users who expect `luai`/`luac` as a default output
   * Given `lua[tools]`, when `VCPKG_TARGET_IS_IOS`, then `FATAL_ERROR`

* The `CONTROL` file is converted to `vcpkg.json`

* ~~Removed duplicated `vcpkg_configure_cmake` and `vcpkg_install_cmake` when `lua[cpp]` used~~
   * In my opinion such behavior is confusing for users. (They might try to link both `liblua.a` and `liblua-c++.a` together.)  
     Since `lua-c++` uses `std::exception` instead of `setjmp`, they should be separated unless the user knows the linkage result exactly.

### Which triplets are supported/not supported?

Support 4 ios triplets
* `arm64-ios`
* `arm-ios`
* `x64-ios`
* `x86-ios`

Yes


ports/lua/portfile.cmake
```diff
@@ -6,41 +6,36 @@ vcpkg_download_distfile(ARCHIVE
...
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+  FEATURES
+    cpp COMPILE_AS_CPP
+  INVERTED_FEATURES
+    tools SKIP_INSTALL_TOOLS
+)
+if(VCPKG_TARGET_IS_IOS)
+    if(FEATURE_OPTIONS MATCHES "-DSKIP_INSTALL_TOOLS=OFF")
```
ras0219-msft, line 23:
```suggestion
    if("tools" IN_LIST FEATURES)
```

ras0219-msft (changes requested):
Thanks for the PR!

The change to `lua[cpp]` is not permissible because it makes the feature non-additive -- if a user enables `[cpp]`, this change would make the package no longer provide the C symbols. This violates the maintainer guidelines: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-implement-alternatives

Currently, the reason why having `liblua.a` and `liblua-c++.a` both available and linkable is that the symbols do not conflict. Users can link both just fine; the "linkage result" is controlled by whether they've requested C symbols via `extern "C"` or not.

If the changes to `lua[cpp]` are reverted, this LGTM.

luncliff:
Thanks for the point, @ras0219-msft !
I will revert the part in next commit. :)

ports/lua/portfile.cmake
luncliff: Applied in c08daed. Nice trick!

ports/lua/fix-ios-system.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/src/loslib.c b/src/loslib.c
+--- a/src/loslib.c
++++ b/src/loslib.c
+@@ -143,7 +143,12 @@ static int os_execute (lua_State *L) {
+   const char *cmd = luaL_optstring(L, 1, NULL);
+   int stat;
+   errno = 0;
++#if defined(__APPLE__) && !TARGET_OS_OSX
```
JackBoosY, line 8: Does `defined(__APPLE__)` and `TARGET_OS_OSX` mean the same?

JackBoosY (changes requested):

ports/lua/fix-ios-system.patch
luncliff: No. AppleClang provides `__APPLE__` macro definition for target platform detection. Its purpose is to ensure the compilation is for Apple platform, which provides `TARGET_OS_IPHONE`, `TARGET_OS_SIMULATOR`, etc.  

`TARGET_OS_OSX` is `0` for the iPhone family and `1` for the general Mac desktop(x86_64). So the expression is saying "When the target platform is for Apple, but not Mac OS".

Reference: https://stackoverflow.com/a/49560690

JackBoosY (approved):

JackBoosY:
Waiting for merge #16138.

ports/lua/portfile.cmake
```diff
@@ -6,41 +6,50 @@ vcpkg_download_distfile(ARCHIVE
...
+        ${FEATURE_OPTIONS}
         -DCOMPILE_AS_CPP=OFF
     OPTIONS_DEBUG
         -DSKIP_INSTALL_HEADERS=ON
-        -DSKIP_INSTALL_TOOLS=ON
 )
-
 vcpkg_install_cmake()
 
-set(ENABLE_LUA_CPP 0)
```
ras0219-msft, line 26: The variable `ENABLE_LUA_CPP` is embedded into vcpkg-cmake-wrapper.cmake via `configure_file()` and must be defined as previous.

ras0219-msft (changes requested):
This PR changes building tools to be off-by-default, however that matches the general expected behavior of vcpkg packages.

ports/lua/portfile.cmake
luncliff: Revoked the change in 7d53833. Thanks for the point :D

luncliff:
> This PR changes building tools to be off-by-default, however that matches the general expected behavior of vcpkg packages.

I was working on multiple PR and forgot about that... 7d53833 added the `default-feature` for `lua[tools]`.

JackBoosY:
@ras0219-msft Should be good now.

luncliff:
Fixed collision after #16603. #15921 can collide with this PR, either.

luncliff:
Found #16943. Commenting here to remind me to rebase this. :)

ports/lua/fix-ios-system.patch
ras0219-msft: We need to also #include <TargetConditionals.h>, otherwise this is disabling system() on all apple platforms.

ports/lua/portfile.cmake
```diff
@@ -6,41 +6,52 @@ vcpkg_download_distfile(ARCHIVE
...
+  FEATURES
+    # cpp COMPILE_AS_CPP
+  INVERTED_FEATURES
+    tools SKIP_INSTALL_TOOLS
+)
+if(VCPKG_TARGET_IS_IOS)
+    if("tools" IN_LIST FEATURES)
+        message(FATAL_ERROR "lua[tools] is not supported for iOS platform build")
+    endif()
+endif()
```
ras0219-msft, line 26:
```suggestion
if(VCPKG_TARGET_IS_IOS AND "tools" IN_LIST FEATURES)
    message(FATAL_ERROR "lua[tools] is not supported for iOS platform build")
endif()
```

ports/lua/portfile.cmake
```diff
@@ -6,41 +6,52 @@ vcpkg_download_distfile(ARCHIVE
...
+    PATCHES
+        vs2015-impl-c99.patch
+        fix-ios-system.patch
 )
 
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+  FEATURES
+    # cpp COMPILE_AS_CPP
```
ras0219-msft, line 18:
```suggestion
```

ports/lua/portfile.cmake
```diff
@@ -6,41 +6,52 @@ vcpkg_download_distfile(ARCHIVE
...
+        message(FATAL_ERROR "lua[tools] is not supported for iOS platform build")
+    endif()
+endif()
+
+set(ENABLE_LUA_CPP 0)
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
     OPTIONS
+        ${FEATURE_OPTIONS}
```
ras0219-msft, line 33:
```suggestion
    OPTIONS_RELEASE
        ${FEATURE_OPTIONS}
    OPTIONS
```
And then re-include `-DSKIP_INSTALL_TOOLS=ON` so that we aren't building and installing tools in debug.

ports/lua/vcpkg.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "lua",
+  "version-string": "5.4.2",
+  "port-version": 1,
+  "description": "A powerful, fast, lightweight, embeddable scripting language",
+  "homepage": "https://www.lua.org",
+  "default-features": [
+    "tools"
+  ],
```
ras0219-msft, line 9:
```suggestion
```

We should not be building tools by default; now that it is a feature, users can easily opt-in via `lua[tools]`.

ras0219-msft (changes requested):

ports/lua/fix-ios-system.patch
luncliff: I just glanced at the header, and the way will be more explicit for `buildtrees` debugging. Good suggestions 👍

ports/lua/vcpkg.json
luncliff: Actually, it was what I intended when I open this PR. So the change is fine for me...
But 1 question for check... was there a change after https://github.com/microsoft/vcpkg/pull/16107#pullrequestreview-588317351?

JackBoosY:
Any progress here?

luncliff:
> Any progress here?

Sorry. I couldn't work on this recently. I will update in this weekend

JackBoosY (approved):

PR #16110, opened 2021-02-08
[CI] debug variant of #15983

cenit:
as per title, it's a parallel test of debug config, like release was for #15983


cenit:
@ras0219 a "debug"-only triplet is terribly broken. OTOH a "release"-only triplet is almost working (see the other PR, which has a green checkmark).
Is it completely unsupported and not even an interesting triplet? I might have some cases in which it might be useful, especially during development and when using a central building machine which would produce only debug bits...

cenit:
@Neumann-A have you ever tried running a "debug"-only triplet?
Any idea if it can be solved somehow easily? It seems to me that the fundamental design of vcpkg is very against a debug-only triplet, while at the same time it's nowhere explicitly said, on the contrary it seems that single-config builds should be supported

Neumann-A:
@cenit No. But I know that I wrote debug only unuseable code in the past. The main problem is a missing abstraction around build types which makes dealing with them so annoying. There is so much copypasta around buildtypes..... #7733 was closed because I did not take the time to merge master again.

cenit:
The main problem that I see is that the include folder at the moment cannot come from the debug at all.
If you move it afterwards, the CMake-generated config files would be broken.
If you keep it there, it cannot work.
If you want to define from the beginning, you cannot because the only general way to pilot packages is through a global prefix, which includes everything.

I don't know if I am clear enough. But my point is that it might be conceptually broken at the moment. And I fear that's the reason @ras0219 has still not answered...

JackBoosY:
Ping @ras0219 @ras0219-msft for response.

cenit:
https://discord.com/channels/400588936151433218/687365466422902841/821871684377116722

> Sorry for the looong poll on this: I believe debug only is very much not in the short-term plan for vcpkg. The best approach for that if you really wanted would be to do a release-only and set up your Release settings to actually be debug settings (though this will not trigger certain port-specific debug behavior). We have some very long term thoughts about how to improve this, but it’s very far down the priority stack compared to other things.

PR #16130, opened 2021-02-08
[(z_)vcpkg_prettify_command(_line)] Scripts Tree Audit

strega-nil:
Audit `vcpkg_prettify_command` to be in line with #16055 

* Since it's internal, rename to `z_vcpkg_prettify_command_line`
* Change it from a macro to a function
* Change how the function takes arguments, from taking two variable names, to taking a variable name output variable and the command line as the remaining arguments

Additionally, this adds the `z_vcpkg_escape_argv` helper macro from #16061, except that it's been renamed and modified slightly; #16061 should depend on this PR and use this new macro.


scripts/ports.cmake
```diff
@@ -122,9 +122,12 @@ if(CMD MATCHES "^BUILD$")
     include(vcpkg_install_nmake)
     include(vcpkg_install_qmake)
     include(vcpkg_internal_get_cmake_vars)
-    include(vcpkg_prettify_command)
     include(vcpkg_replace_string)
     include(vcpkg_test_cmake)
+
+    include("${SCRIPTS}/cmake/z_vcpkg_prettify_command_line.cmake")
+    include("${SCRIPTS}/cmake/z_vcpkg_function_arguments.cmake")
```
Neumann-A, line 129: is there a reason not to use the same style as above?
Meaning:
`include(z_vcpkg_prettify_command_line)`

scripts/cmake/vcpkg_execute_required_process.cmake
```diff
@@ -104,7 +101,7 @@ Halting portfile execution.
             file(TO_NATIVE_PATH "${LOG}" NATIVE_LOG)
             list(APPEND STRINGIFIED_LOGS "    ${NATIVE_LOG}\n")
         endforeach()
-        vcpkg_prettify_command(vcpkg_execute_required_process_COMMAND vcpkg_execute_required_process_COMMAND_PRETTY)
+        z_vcpkg_prettify_command_line(vcpkg_execute_required_process_COMMAND_PRETTY ${vcpkg_execute_required_process_COMMAND})
```
Neumann-A, line 104:
```suggestion
        z_vcpkg_prettify_command_line(${CMAKE_CURRENT_FUNCTION}_COMMAND_PRETTY ${${CMAKE_CURRENT_FUNCTION}_COMMAND})
```
I think this looks cleaner.

scripts/cmake/z_vcpkg_function_arguments.cmake
```diff
@@ -0,0 +1,47 @@
...
+    else()
+        # vcpkg bug
+        message(FATAL_ERROR "z_vcpkg_function_arguments: invalid arguments (${ARGV})")
+    endif()
+
+    set("${OUT_VAR}")
+
+    # this allows us to get the value of the enclosing function's ARGC
+    set(z_vcpkg_function_arguments_ARGC_NAME "ARGC")
+    set(z_vcpkg_function_arguments_ARGC "${${z_vcpkg_function_arguments_ARGC_NAME}}")
```
Neumann-A, line 40: magic? Is this actually documented somewhere in cmake?

scripts/cmake/z_vcpkg_function_arguments.cmake
```diff
@@ -0,0 +1,47 @@
...
+    foo(${ARGS})
+    ...
+endfunction()
+```
+#]===]
+macro(z_vcpkg_function_arguments OUT_VAR)
+    if("${ARGC}" EQUAL 1)
+        set(z_vcpkg_function_arguments_FIRST_ARG 0)
+    elseif("${ARGC}" EQUAL 2)
+        set(z_vcpkg_function_arguments_FIRST_ARG "${ARGV1}")
```
Neumann-A, line 30:
```suggestion
    if(${ARGC} EQUAL 1)
        set(z_vcpkg_function_arguments_FIRST_ARG 0)
    elseif(${ARGC} EQUAL 2)
        set(z_vcpkg_function_arguments_FIRST_ARG "${ARGV1}")
```

do you need the quotes here ?

scripts/cmake/z_vcpkg_function_arguments.cmake
Neumann-A: Ah so since in a macro ARGC is not a variable it does not get replaced here and since the second line is not directly dereferencing ARGC it gets the variable ARGC from the enclosing function. Very funny. should be more clearly be documented by cmake.

scripts/cmake/vcpkg_execute_required_process.cmake
Neumann-A: but probably requires more changes. I personally like to define a variable `_PREFIX` which I pass into `cmake_parse_arguments` and use that everywhere for function variables. In my eyes it makes it clearer which variables are passed into a function. I didn't notice it here that `vcpkg_execute_required_process` is actually the prefix of `cmake_parse_arguments`. So you can just ignore the above suggestion.

scripts/ports.cmake
strega-nil: I plan on changing all of the includes to use this form.

scripts/cmake/vcpkg_execute_required_process.cmake
strega-nil: This kind of stuff will come in future audit PRs; this isn't an audit of `vcpkg_execute_required_process`, so the only changes I'm making are to make sure it works with the new function.

scripts/cmake/z_vcpkg_function_arguments.cmake
strega-nil: Probably, but yeah.

scripts/cmake/z_vcpkg_function_arguments.cmake
strega-nil: Technically, no, but I don't want to have even a whiff of double-expansion.

Side note, the reason that this doesn't use `if (ARGC EQUAL 1)` is that that will use the ARGC of the containing function.

Neumann-A: > Side note, the reason that this doesn't use if (ARGC EQUAL 1) is that that will use the ARGC of the containing function.

Yeah according to the macro docs `ARGC` is not a variable in a macro. CMake is basically doing string replacement with anything which is `${ARGC}` within a macro.

strega-nil: Yeah, exactly :)

docs/maintainers/internal/z_vcpkg_function_arguments.md
```diff
@@ -0,0 +1,26 @@
+# z_vcpkg_function_arguments
```
ras0219-msft, line 1: Needs an internal use disclaimer.

scripts/cmake/z_vcpkg_function_arguments.cmake
```diff
@@ -0,0 +1,47 @@
...
+        message(FATAL_ERROR "z_vcpkg_function_arguments: invalid arguments (${ARGV})")
+    endif()
+
+    set("${OUT_VAR}")
+
+    # this allows us to get the value of the enclosing function's ARGC
+    set(z_vcpkg_function_arguments_ARGC_NAME "ARGC")
+    set(z_vcpkg_function_arguments_ARGC "${${z_vcpkg_function_arguments_ARGC_NAME}}")
+
+    math(EXPR z_vcpkg_function_arguments_LAST_ARG "${z_vcpkg_function_arguments_ARGC} - 1")
```
ras0219-msft, line 42: This is possibly benign, but note that ARGC could be 0 which makes LAST_ARG `-1`.

scripts/cmake/z_vcpkg_prettify_command_line.cmake
```diff
@@ -0,0 +1,30 @@
...
+* `scripts/cmake/vcpkg_execute_required_process.cmake`
+* `scripts/cmake/vcpkg_execute_required_process_repeat.cmake`
+#]===]
+
+function(z_vcpkg_prettify_command_line OUT_VAR)
+    set(OUTPUT "")
+    z_vcpkg_cmake_function_arguments(ARGUMENTS 1)
+    foreach(v IN LISTS ARGUMENTS)
+        if(v MATCHES "( )")
+            list(APPEND OUTPUT "\"${v}\"")
```
ras0219-msft, line 24:
```suggestion
        if(v MATCHES " ")
            string(REPLACE "\"" "\\\"" v "${v})
            list(APPEND OUTPUT "\"${v}\"")
```
Though this is minor.

ras0219-msft (approved):

scripts/cmake/z_vcpkg_function_arguments.cmake
strega-nil: Yeah, this is fine; since `RANGE` is inclusive, `[0, 0]` will give `0` still. We need it to be `[0, -1]`.

Issue #15452, opened 2021-01-05
vcpkg integrate install implementation


jmoguillansky-gpsw:
I have a question regarding vcpkg integrate install:
Can you please clarify what it means when it says "Linking will be handled automatically"?
What is the internal logic?
Does it add the library as a linker arg to all projects globally?  
Does it add the library search path globally?

JackBoosY:
For cmake project, vcpkg will generate the vcpkg toolchain for use in custom cmake projects.
For msvc project, according to the configuration (x64/x86), vcpkg will automatically add _VCPKG_ROOT/installed/TRIPLET/include_ to the `Include Directories` and add all libraries in the _VCPKG_ROOT/installed/TRIPLET/lib_ / _VCPKG_ROOT/installed/TRIPLET/debug/lib_ to the `Additional Dependencies`.

jmoguillansky-gpsw:
"add all libraries in the VCPKG_ROOT/installed/TRIPLET/lib": this adds all libraries as linker args for all MSVC projects?
This seems a bit dangerous:
1) not all projects require linking against the library.  
Linking against unnecessary libs slows down link time and adds unnecessary dependencies. 
If you're building a shared library, it will load unnecessary additional libs at runtime?  

From my perspective: 
vcpkg integrate install should work in a similar manner as when we install packages on Linux via a pkg-manager:
install the includes, libs, pkg-config files, etc. to a system-wide search path. 
Let the user specify the libraries to link against, either explicitly or indirectly via pkg-config or a cmake configuration file?

JackBoosY:
@jmoguillansky-gpsw I partly agree with you. vcpkg is currently not perfect for the integrated msvc project.
However, adding unneeded libraries to additional dependencies will not cause build problems, nor will they cause usage problems.

Since vcpkg does not currently provide a file list for each installed port, we cannot know what the library corresponding to the included header file should be when integrating into the msvc project, nor can we know what the library of the dependent port is.

jmoguillansky-gpsw:
@JackBoosY we shouldn't infer which lib to link against based on the header files.  The user should define this either explicitly, or indirectly via some higher-level mechanism like pkg-config or cmake toolchains.  
I think this current logic can very quickly cause problems.  Suppose the user installs 100 packages using vcpkg and then does vcpkg integrate install.  All MSVC projects will automatically link against 100 libs installed via vcpkg?  
[...]

PR #16137, opened 2021-02-09
[shaderc] Use build type to build targets

### LINKED ISSUE #16122: vcpkg integrate install logic is incorrect
**Describe the bug**
vcpkg integrate install does not apply the correct logic.  
It automatically adds all libraries under installed/x64-windows/lib as linker arguments
Some packages install multiple versions of libraries.
The user is only supposed to link against one specific version of the library, as specified via a pkg-config file or cmake toolchain file. 
Otherwise we get linker errors about a symbol already defined

**Environment**
- OS: Windows
- Compiler: MSVC 2019

**To Reproduce**
vcpkg install shaderc:x64-windows
vcpkg integrate install

**Expected behavior**
The current logic for vcpkg integrate install is not correct.
VCPKG cannot just add all libraries in installed/x64-windows/lib as linker inputs.
The user should specify the linker inputs either explicitly or via some tool such as pkg-config --libs or via a cmake toolchain file.

**Failure logs** 
untitled:3408: 2>    shaderc.lib(shaderc.cc.obj) : error LNK2005: shaderc_compile_into_spv already defined in shaderc_shared.lib(shaderc_shared.dll)

**Additional context**

JackBoosY:
Currently, shaderc has generated dynamic libraries, static libraries and static libraries containing all dependent binaries at the same time.
This PR will:
- Determine whether to generate a dynamic library or a static library according to `BUILD_SHARED_LIBS`.
- Disable generating static libraries containing all dependent binaries.

Partly solved #15452.
Fixes #16122.


ports/shaderc/portfile.cmake
```diff
@@ -1,12 +1,16 @@
...
     OUT_SOURCE_PATH SOURCE_PATH
     REPO google/shaderc
     REF 05c766a1f4615ee37d37d09271aaabca30ffd293
     SHA512 329697e8e23d619313440d57ef740a94c49d13533e1b8734fc8ff72fd5092c2addabb306f64cb69160fa5fee373a05ba39a5ee6d92d95e5e2e9c7ec96a51aadc
     HEAD_REF master
     PATCHES 
-    	"disable-update-version.patch"
-        "fix-install.patch"
+    	disable-update-version.patch
+        fix-install.patch
```
NancyLi1013, line 12:
```suggestion
        disable-update-version.patch
        fix-install.patch
```

ports/shaderc/portfile.cmake
```diff
@@ -31,21 +33,23 @@ vcpkg_add_to_path(PREPEND "${PYTHON3_EXE_PATH}")
...
 
-#Safe to remove as libs are static
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/bin)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/bin)
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 
-# Handle copyright
-file(COPY ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/shaderc)
-file(RENAME ${CURRENT_PACKAGES_DIR}/share/shaderc/LICENSE ${CURRENT_PACKAGES_DIR}/share/shaderc/copyright)
+file(INSTALL "${SOURCE_PATH}/LICENSE" DESTINATION "${CURRENT_PACKAGES_DIR}/share/shaderc" RENAME copyright)
```
NancyLi1013, line 55:
```suggestion
file(INSTALL "${SOURCE_PATH}/LICENSE" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME copyright)
```

NancyLi1013 (changes requested):

NancyLi1013 (approved):

JackBoosY:
Waiting for merge #16138.

ras0219-msft:
This LGTM

jmoguillansky-gpsw:
Hi @JackBoosY I updated vcpkg to the latest version.  It seems that only static shaderc was built.
How do I build the shared lib?
Am I supposed to set BUILD_SHARED_LIBS environment variable or something?

jmoguillansky-gpsw:
Seems that this PR disables shaderc shared library build?

#Note: glslang and spir tools doesn't export symbol and need to be build as static lib for cmake to work

This part doesn't make sense.  If glslang and spir-tools require static shaderc then they can link against the static shaderc, that doesn't mean that shaderc doesn't support shared library build?

JackBoosY:
@jmoguillansky-gpsw For building shared, use command `./vcpkg install shaderc`.
You should use the right triplet to build it.

jmoguillansky-gpsw:
> @jmoguillansky-gpsw For building shared, use command `./vcpkg install shaderc`.
> You should use the right triplet to build it.

Hi @JackBoosY ,
Yes I tried that: vcpkg install shaderc:x64-windows, but it still only built the static lib?

JackBoosY:
@jmoguillansky-gpsw  Ops, sorry, I forgot we force it to build as a static library.
I will investigate later.

ras0219-msft:
The issue is not that glslang and spir-tools require static shaderc -- rather, the problem is that glslang and spir-tools can only be built statically, which means building a shared library for shaderc is _extremely_ suspect (it will contain a duplicate copy of the glslang/spir-tools code along with the main application binary as well as any other consumers). Shaderc consumes glslang, not the other way around.

Therefore, until glslang and spir-tools can be built dynamically, we do not want to build shaderc dynamically.

jmoguillansky-gpsw:
Shaderc upstream supports shared lib build, and its shipped as part of vulkan sdk.  There's no issue with shaderc , the tool just decides to use the static lib, thats all

jmoguillansky-gpsw:
I think there's a broader discussion with regards to vcpkg architecture.  Each module doesn't know how it's used, only what it can build.  Some executables may choose to always use static lib, some shared lib, etc.  From my perspective the vcpkg architecture should accomodate these cases.  
One option:
The module should build all supported configurations (static, shared, etc).  
Vcpkg integrate install can have a param of default lib type (e.g. shared or static) the same way that a linker (ld on linux) has a param to specify default lib type.

jmoguillansky-gpsw:
Regarding vcpkg integrate install:
Maybe one option to consider is:
let the user navigate to a project and type something like "vcpkg integrate <module name>" and then vcpkg will integrate the module into the project.
This gives the user convenience, but perhaps safer than making all libs globally available for all msvc projects,etc.

JackBoosY:
glslang and spir-tools only depend on the static library, not dynamic.

PR #16142, opened 2021-02-09
[boost] build boost on windows x86 with /safeseh

autoantwort:
**Describe the pull request**

- What does your PR fix? Build boost on windows x86 with `/safeseh`. Otherwise you get linker errors if you build your application with `/safeseh` but boost is not build with `/safeseh`.

- Which triplets are supported/not supported? Have you updated the CI baseline? No Changes. Tested with x86-windows-static locally.

### DIFF
--- a/ports/boost-modular-build-helper/boost-modular-build.cmake
+++ b/ports/boost-modular-build-helper/boost-modular-build.cmake
@@ -307,6 +307,11 @@ function(boost_modular_build)
         list(APPEND B2_OPTIONS address-model=64 architecture=power)
     else()
         list(APPEND B2_OPTIONS address-model=32 architecture=x86)
+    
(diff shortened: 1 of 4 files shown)


ras0219-msft:
Thanks for the PR!

I strongly suspect you're looking for custom triplets and specifically https://github.com/microsoft/vcpkg/blob/master/docs/users/triplets.md#vcpkg_cxx_flags to change the flags used for a library.

If I'm in error and you think that boost should always be built with `/safeseh` for all users in all circumstances, let me know and I'll be happy to reopen the PR for discussion :)

autoantwort:
Hello, 
imho boost should be build always with `/safeseh`. It improves the security of the resulting application (see [here](https://stackoverflow.com/questions/25081033/what-safesehno-option-actually-do)) and has no downsides.

autoantwort:
@ras0219-msft ping

autoantwort:
@ras0219-msft ping

PR #16167, opened 2021-02-10
[vcpkg] resolve VcpkgEnabled default value conflict

### LINKED ISSUE #16165: How do I set "Use Vcpkg" to "NO" in Visual Studio property page widely or in a CMake manner?
Hi, vcpkg Team

I write C++ in Visual Studio, by first write CMakeLists.txt then generate .sln file, without using vcpkg.

Yesterday I  just installed vcpkg with Visual Studio integration via:
```
vcpkg integrate install
```

After the integration installation, there is `vcpkg` item in existing/new created Visual Project's property page, and `Use Vcpkg` is "YES" on default widely:

image

I know vcpkg is powerful and convenient from the readme page, however it may mess up my dependencies 

For example,
>compiling OpenCV imgcodecs module with zlib static library from source, however it automatically linking to vcpkg installed zlib dll file, which is not as expected.
(https://github.com/opencv/opencv/issues/17051)

My question: I would like to still keep vcpkg Visual Studio integration item/page in project property, and **is it possible to set `Use Vcpkg` to "NO" widely? And is it possible to just set it to "NO" for one specific project in a CMake manner? i.e. passing one option to disable / not use `vcpkg` temporarily? If possible then how to config that?**

Thanks.

zchrissirhcz:
**Resolve VcpkgEnabled default value conflict**

- What does your PR fix? Fixes #16165 

- Which triplets are supported/not supported? Have you updated the CI baseline? No


ras0219-msft:
Thanks for the PR!

Blocked on an answer for https://github.com/microsoft/vcpkg/issues/16165#issuecomment-777840155

strega-nil-ms:
This does seem like an issue, but this is definitely not the correct fix for the issue, since it breaks usage of MSBuild + vcpkg integration. We recommend using one of the stated workarounds from #16165.

Thank you for the PR, and for using vcpkg!

PR #16169, opened 2021-02-10
[proj4] Allow enabling tiff support for static builds

rhuijben:
**Describe the pull request**

- What does your PR fix?
Before this patch building proj in static or static-md triplets always disabled tiff support unconditionally. This patch moves the enable/disable tiff support to the already declared tiff feature flag.

Enabling/disabling tiff support enables/disables certain grid shift operations in proj[4], so with this enabled higher precision transformations may be used for several operations.

This patch leaves curl disabled, so the tiff flag doesn't allow using the CDN for transforms directly, unless a custom handler is installed (uncommon) or the data is available/cached locally (common).

Nothing changed.

### DIFF
--- a/ports/libspatialite/portfile.cmake
+++ b/ports/libspatialite/portfile.cmake
@@ -14,6 +14,8 @@ vcpkg_extract_source_archive_ex(
 )
 
 if (VCPKG_TARGET_IS_WINDOWS)
+  # TODO: This should be updated to use the knowledge from the base packages
+  #       instead of manually maintaining this knowledge here
   if(VCPKG_CRT_LINKAGE STREQUAL dynamic)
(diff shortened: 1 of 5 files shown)


rhuijben:
Looks like this approach breaks libspatialite, and by that the dependencies of libspatialite. Will need additional work. 

And using tiff as 'feature' is not recommended anyway. In general libspatialite would want to use the improved accuracy so I'm now working on a patch to fix libspatialite to find the dependencies in the proper way.

dg0yt:
What was actually breaking with static builds? Did it miss the extra libraries it should link?

rhuijben:
Yes, several dependencies miss their references. Fixes are combined with the proj 8.0.0 update patch.

PR #16172, opened 2021-02-10
[bfgroup-lyra] Install CMake Targets

Hoikas:
Per a report in discord, the `bfgroup-lyra` port is not installing its CMake targets. This fixes that issue.


ras0219 (approved):

ras0219-msft:
Thanks for the fix!

PR #16173, opened 2021-02-10
[vcpkg integrate install] Allow setting the vcpkg installed dir

strega-nil:
Still to do: this breaks the no-op case where `vcpkg install` should not be called; however, this was necessary because otherwise changing the current installed dir would not work. cc @ras0219 for any ideas?

### DIFF
--- a/scripts/buildsystems/msbuild/vcpkg-general.xml
+++ b/scripts/buildsystems/msbuild/vcpkg-general.xml
@@ -55,8 +55,8 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <StringProperty Name="VcpkgInstalledDir" DisplayName="vcpkg installed directory" Category="General" Subtype="folder" Visible="true"
-                  Description="The vcpkg installed directory: this is the place where vcpkg places all of the ports for installation.">
+  <StringProperty Name="VcpkgInstalledDir" DisplayName="Installed Directory" Category="General" Subtype="folder" Visible="true"
+                  Description="The location where headers and binaries will be consumed from. In manifest mode, this directory will be created and populated based on vcpkg.json.">
   </StringProperty>
 
   <BoolProperty Name="VcpkgUseStatic" DisplayName="Use Static Libraries" Category="Conditional" Default="false"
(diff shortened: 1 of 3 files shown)


scripts/buildsystems/msbuild/vcpkg-general.xml
```diff
@@ -55,11 +55,8 @@
     </StringProperty.DataSource>
   </StringProperty>
 
-  <StringProperty Name="VcpkgCurrentInstalledDir" DisplayName="Package install directory" Category="General" Subtype="folder" Visible="false"
-                  Description="Defines the direct path to the installed Vcpkg packages.">
-    <StringProperty.DataSource>
-      <DataSource Persistence="ProjectFile" Label="Vcpkg" HasConfigurationCondition="false" />
-    </StringProperty.DataSource>
+  <StringProperty Name="VcpkgInstalledDir" DisplayName="vcpkg installed directory" Category="General" Subtype="folder" Visible="true"
```
ras0219, line 58: The display name is in the context of the "Vcpkg" tab, so it does not need to be specifically named "vcpkg installed directory". Additionally, it should be Title Case to match other settings.

`Installed Directory`

scripts/buildsystems/msbuild/vcpkg-general.xml
```diff
@@ -55,11 +55,8 @@
...
+                  Description="The vcpkg installed directory: this is the place where vcpkg places all of the ports for installation.">
```
ras0219, line 59:
```suggestion
                  Description="The location where headers and binaries will be consumed from. In manifest mode, this directory will be created and populated based on vcpkg.json.">
```

scripts/buildsystems/msbuild/vcpkg-general.xml
ras0219, line 58: Please document this new public parameter in

https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md#msbuild-properties

ras0219 (approved):
Worked with @strega-nil to fix the issue :) LGTM

ras0219 (approved):

Issue #16188, opened 2021-02-12
[scripts-audit] Work list for stabilization of the scripts tree


strega-nil:
See PR #16055 

- [ ] `vcpkg_acquire_msys`:
  - [ ] Most of vcpkg_acquire_msys needs to become a tool port, except for the content of the mirror list.
  - [ ] Calls to vcpkg_acquire_msys that call PACKAGES are deprecated; the tool port will use NO_DEFAULT_PACKAGES DIRECT_PACKAGES
- [ ] `vcpkg_add_to_path`:
  - [ ] audit
- [ ] `vcpkg_apply_patches`:
  - [ ] audit
  - [ ] deprecate, add `z_vcpkg_apply_patches` as an internal function
- [ ] `vcpkg_build_cmake`:
  - [ ] extract to ports
  - [ ] audit
- [ ] `vcpkg_build_gn`:
  - [ ] audit
  - [ ] deprecate in favor of `vcpkg_build_ninja`
- [ ] `vcpkg_build_make`:
  - [ ] extract to ports
  - [ ] audit
  - [ ] remove/deprecate `ENABLE_INSTALL` (depending on whether it's used) (see also `INSTALL_TARGET`)
  - [ ] look and see if `MAKEFILE` is used usefully vs `SUBPATH`; if not, deprecate/remove depending on whether it's used
- [ ] `vcpkg_build_msbuild`:
  - [ ] deprecate
  - [ ] audit
  - [ ] update modern ports to use `vcpkg_install_msbuild()`
- [ ] `vcpkg_build_ninja`:
  - [ ] extract to ports
  - [ ] audit
- [ ] `vcpkg_build_nmake`:
  - [ ] extract to ports
  - [ ] audit
- [ ] `vcpkg_build_qmake`:
  - [ ] extract to ports
  - [ ] audit
- [ ] `vcpkg_buildpath_length_warning`:
  - [ ] audit
- [ ] `vcpkg_check_features`:
  - [ ] audit
- [ ] `vcpkg_check_linkage`:
  - [ ] audit
- [ ] `vcpkg_clean_executables_in_bin`:
  - [ ] deprecate, rename to `z_vcpkg_clean_executables_in_bin`
  - [ ] audit
- [ ] `vcpkg_clean_msbuild`:
  - [ ] extract to port
  - [ ] audit
- [ ] `vcpkg_common_definitions`:
  - [ ] audit
- [x] `vcpkg_common_functions`:
  - already deprecated
[...]

PR #16189, opened 2021-02-12
[scripts-audit] vcpkg_add_to_path

strega-nil:
See PR #16055, Issue #16188


vicroms (approved):

strega-nil:
Merging as "the only failures are also failures in the most recent CI run".

PR #16195, opened 2021-02-12
[z_vcpkg_prettify_command_line] write to output variable

strega-nil:


scripts/cmake/z_vcpkg_prettify_command_line.cmake
```diff
@@ -29,5 +29,6 @@ function(z_vcpkg_prettify_command_line OUT_VAR)
             list(APPEND OUTPUT "${v}")
```
ras0219-msft, line 29: This is still using `OUTPUT`.

ras0219-msft (changes requested):
Please ensure the results work with a failing command line.

ras0219-msft (approved):

PR #16202, opened 2021-02-13
[quill] Uncomment QUILL_FMT_EXTERNAL macro

RT2Code:
As we build using the vcpkg fmt port instead of the one bundled in Quill, this macro needs to be defined.

### DIFF
--- a/ports/quill/portfile.cmake
+++ b/ports/quill/portfile.cmake
@@ -20,9 +20,7 @@ vcpkg_configure_cmake(
 vcpkg_install_cmake()
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/quill)
 
-file(READ ${CURRENT_PACKAGES_DIR}/include/quill/TweakMe.h QUILLCONFIG)
-string(REGEX REPLACE "// #define QUILL_FMT_EXTERNAL" "#define QUILL_FMT_EXTERNAL" QUILLCONFIG "${QUILLCONFIG}")
-file(WRITE ${CURRENT_PACKAGES_DIR}/include/quill/TweakMe.h "${QUILLCONFIG}")
+vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/include/quill/TweakMe.h "// #define QUILL_FMT_EXTERNAL" "#define QUILL_FMT_EXTERNAL")
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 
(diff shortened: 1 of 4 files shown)


ports/quill/portfile.cmake
```diff
@@ -22,9 +18,11 @@ vcpkg_configure_cmake(
 )
 
 vcpkg_install_cmake()
-
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/quill)
 
+file(READ ${CURRENT_PACKAGES_DIR}/include/quill/TweakMe.h QUILLCONFIG)
+string(REGEX REPLACE "// #define QUILL_FMT_EXTERNAL" "#define QUILL_FMT_EXTERNAL" QUILLCONFIG "${QUILLCONFIG}")
+file(WRITE ${CURRENT_PACKAGES_DIR}/include/quill/TweakMe.h "${QUILLCONFIG}")
```
NancyLi1013, line 25: Could you please use `vcpkg_replace_string() ` here?

NancyLi1013 (changes requested):

ports/quill/portfile.cmake
RT222: Sure. I didn't know about this function, it's far cleaner.

ras0219-msft (approved):
Thanks for the PR! LGTM once CI passes.

NancyLi1013:
The failures on x86-windows:
```
Found the following errors:
Error: While reading versions for port quill from file: C:\a\1\s\versions\q-\quill.json
       File declares version `1.6.1#1` with SHA: aa542e38f0f2480e495b785840e919d46a1d58d5
       But local port with the same verion has a different SHA: a29a0dd1467a5224f7838e8b1c9f63d263702e87

           vcpkg x-add-version quill
```

Please run the command `vcpkg x-add-version quill` and submit the changes to solve this.

NancyLi1013 (approved):

ras0219-msft:
Thanks for the fix!

odygrd:
@RT222 I recently got this bug report

I am not sure if that PR has broken it or not. As I explained in the issue , quill has logic in cmake for defining `QUILL_FMT_EXTERNAL`  https://github.com/odygrd/quill/blob/master/CMakeLists.txt#L127

Why is this macro required ?

RT222:
Thanks for alerting me about this, here's my answer : https://github.com/odygrd/quill/issues/94#issuecomment-808824535

PR #16238, opened 2021-02-15
[ryu] added ryu_printf

georg-emg:
This patch updates the port files to build and install the library `ryu_printf` in addition to the base library `ryu`. The new library is a completely independent build, so that the existing library is not affected in any way.

_Note:_ The library `generic_128` was not added, because it requires a 128 bit integer type and thus cannot be built under Windows.


ports/ryu/CONTROL
```diff
@@ -1,5 +1,6 @@
 Source: ryu
 Version: 2.0-2
+Port-Version: 1
```
ras0219-msft, line 3:
```suggestion
Version: 2.0
Port-Version: 3
```

versions/r-/ryu.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "b48e39f7ff56d96b9815e2f4377e582d156ba4fa",
+      "version-string": "2.0-2",
+      "port-version": 1
+    },
```
ras0219-msft, line 7: When changing the port-version to 3, this entry needs to be removed.

ras0219-msft (changes requested):
Thanks for the PR!

LGTM after the version change.

versions/r-/ryu.json
georg-emg: Just trying to understand your request.

If I understand correctly, your concern is that the version number in the CONTROL file (2.0-2) is incorrect and does not correspond to the git tag listed in the portfile (v2.0). You suspect, and I agree with you, that whoever updated the CONTROL file before me mistakenly added "-1" and "-2" to the version, instead of adding a Port-Version entry. So in reality, my change would be port version 3 of ryu version 2.0, and not port version 1 of ryu version 2.0-2.

So basically, I see three ways to fix this:

1. use `Port-Version: 3` and `Version: 2.0`, as you suggested,
2. use `Port-Version: 1` and `Version: 2.0`, because that would be the next port version in line for v2.0, or
3. use `Port-Version: 3` and `Version: 2.0`, but also correct the whole version file, replacing `Version: 2.0-1` with `Port-Version: 1`/`Version: 2.0`, and replacing `Version: 2.0-2` with `Port-Version: 2`/`Version: 2.0` for the earlier commits.

I'll go with whatever you think is best, I just wanted to make sure.

Also, while I'm at it, should I replace the CONTROL file with a proper manifest file?

JackBoosY: Since you doesn't update ryu version, we should keep `Version` as `2.0`, and update the `Port-Version`.
And since the `Port-Version` is incorrect in the previous PR, it should be `3`.
So please use the first way.

We prefer to use manifest file.

georg-emg: Ok, I replaced the CONTROL file with a manifest with port-version 3 and version 2.0, as you requested.

ras0219-msft:
Looks great, thanks for the PR!

PR #16245, opened 2021-02-16
[magnum,-plugins] Fix deployment of plugin types and incorrect removal of plugin lib dir on unix

Squareys:
Hi all,

this PR fixes a couple of issues with the deployment of plugins:

- SceneConverter and ShaderConverter plugins were not deployed
- "MagnumTrade.dll" that was added in since original creation of magnumdeploy.ps1 allows better detection of when certain plugin types need to be deployed
- On Unix (support for which was added since original creation of magnumdeploy.ps1 also), deletion of the lib/magnum(-d) directories causes issues, hence now wrapped in `if(WIN32)`

Best,
Jonathan


PhoebeHui (approved):
@Squareys, thanks for the PR!

PhoebeHui:
The gsoap failures doesn't relate to this issue, it would be fixed by https://github.com/microsoft/vcpkg/pull/16270. I will rerun the CI test after the PR merged.

ras0219-msft:
This looks great to me, thanks for the fix!

PR #16246, opened 2021-02-16
[halide] Update Halide to 11.0.1

### LINKED ISSUE #16243: [Halide] update to 11.0.0
Library name: Halide

New version number: 11.0.0

I am happy to report that the Halide team has released a new major version, 11.0.0! This is a smaller release than last time and should only require a minimal change to the portfile. It should be enough to update the GitHub fetching information and remove the line containing `-DWITH_APPS=NO` (the option was removed, so this will throw a warning and do nothing anyway).

alexreinking:
Updates Halide to 11.0.1. The new Direct3D 12 on ARM support seems to require LLVM to always be built with ARM and x86 support. If this changes in a patch version, we can remove those LLVM backends from `Build-Depends`.

The `WITH_APPS` variable was removed (because the apps were moved entirely to an external project), so we no longer set it here.

- What does your PR fix? **Fixes #16243**

- Which triplets are supported/not supported? Have you updated the CI baseline? **No changes**. We could try removing `halide:x64-windows-static=fail` from the baseline. I forget exactly why it's there, but I think it might be due to either an MSVC compiler bug or an incompatibility with our autoschedulers.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? **Yes**, to the same extent as before.


alexreinking:
I opened https://github.com/halide/Halide/pull/5745 to fix build errors discovered here. I'll update this PR once that's merged.

NancyLi1013:
Hi @alexreinking 

Thanks for updating this port.

It seems all checks have passed now except for the regressions caused by `gsoap `that are not related with this PR.

So is there any work needed to be done? Also, I want to confirm with you the current status on `halide:x64-windows-static`.

Are you clear it about this point? Since it is still set as fail in ci.baseline.txt file. So I'm not sure if it can build successfully on the latest version.

alexreinking:
I'm also not sure about `halide:x64-windows-static`. I forget why it was failing in the first place. I wonder if it had to do with the autoschedulers. Couldn't hurt to try.

We're going to release 11.0.1 with the changes I made in response to prior build failures, so let's hold off merging until we tag that release.

NancyLi1013 (approved):

alexreinking:
This is ready to merge unless you still want to test `halide:x64-windows-static`.

ras0219-msft:
Awesome, thanks!

PR #15605, opened 2021-01-12
[android] vcpkg_configure_make correct set flags for android build

xandox:
vcpkg_configure_make properly set CFLAGS, CXXFLAGS, LDFLAG, CC and CXX env variables now for android cross compilation. 

- Which triplets are supported/not supported?  Mostly android triplets, 
- Have you updated the CI baseline? - no

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
yes

PR #16260, opened 2021-02-17
[vcpkg] vcpkg_configure_meson: cross-compile fixes for Android

jwtowner:
This PR includes two fixes to the meson cross-compile config.

The first change fixes the regex used to partition linker flag options so that it supports double-hyphen prefixes. Without this change, passing a linker flag such as "--target=armv7a-linux-androideabi21" will end up being passed incorrectly as "- -target=armv7a-linux-androideabi21".

The second change ensures that `vcpkg_internal_meson_generate_cross_file` always generates a cross file when the target platform is either Android or iOS, even if the host and target CPU architectures are the same. For example, when cross compiling harfbuzz:x64-android on a Windows 64-bit (x86-64) host, `vcpkg_internal_meson_generate_cross_file` normally does not generate a cross file because the host and target CPU architectures are equal, and this in turn causes Meson to attempt to run the generated x64-android sanity checking binaries as if they were Windows binaries, which causes the configuration to fail altogether.

One question, is there a better way to check if `VCPKG_CMAKE_SYSTEM_NAME` does not match the host system, and in a way that will be future proofed for other cross-compile target systems other than Android and iOS?

This PR, in tandem with PR #15605, allows for harfbuzz (and potentially other ports which also use meson) to be built for arm-android, arm64-android, x86-android and x64-android. Despite harfbuzz:*-android also relying on #15605, this PR should be safe to merge independently of it.


scripts/cmake/vcpkg_configure_meson.cmake
```diff
@@ -128,7 +128,7 @@ function(vcpkg_internal_meson_generate_native_file_config _config) #https://meso
     else()
         set(LINKER_FLAGS_${_config} "${VCPKG_DETECTED_CMAKE_STATIC_LINKER_FLAGS_${_config}}")
     endif()
-    string(REGEX REPLACE "( |^)(-|/)" ";\\2" LINKER_FLAGS_${_config} "${LINKER_FLAGS_${_config}}")
+    string(REGEX REPLACE "( |^)(--?|/)" ";\\2" LINKER_FLAGS_${_config} "${LINKER_FLAGS_${_config}}")
```
Neumann-A, line 131: Use `vcpkg_internal_meson_convert_compiler_flags_to_list` instead

scripts/cmake/vcpkg_configure_meson.cmake
```diff
@@ -316,7 +316,7 @@ function(vcpkg_internal_meson_generate_cross_file_config _config) #https://meson
...
```
Neumann-A, line 319: same as above

scripts/cmake/vcpkg_configure_meson.cmake
jwtowner: Fixed in 8242f9e

scripts/cmake/vcpkg_configure_meson.cmake
jwtowner: Fixed in 8242f9e

PhoebeHui (approved):

ras0219-msft:
Awesome, thanks for the fixes!

PR #16293, opened 2021-02-19
[URGENT] [vcpkg] fix script broken audit

cenit:
Otherwise after #16061 you cannot find anymore packages installed through vcpkg

@strega-nil @ras0219-msft

### DIFF
--- a/scripts/buildsystems/vcpkg.cmake
+++ b/scripts/buildsystems/vcpkg.cmake
@@ -468,19 +468,16 @@ if(NOT Z_VCPKG_ROOT_DIR)
     z_vcpkg_add_fatal_error("Could not find .vcpkg-root")
 endif()
 
-# NOTE: _VCPKG_INSTALLED_DIR cannot be removed without tool changes.
-if(NOT DEFINED VCPKG_INSTALLED_DIR)
-    if(DEFINED _VCPKG_INSTALLED_DIR)
-        set(VCPKG_INSTALLED_DIR "${_VCPKG_INSTALLED_DIR}")
-    elseif(VCPKG_MANIFEST_MODE)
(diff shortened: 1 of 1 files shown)


strega-nil (approved):

strega-nil:
Approved for now, but we should fix this more correctly.

strega-nil:
thanks @cenit

cenit:
Fastest PR merged on vcpkg?
I think I can set a record here 🤣 
I hope it's working :) but as you say, yes a cleaner approach will be necessary.

strega-nil:
Yeah, I'll open a new PR tomorrow; this was simply a case of too many changes to keep track of. I really think I need to work on that...

cenit:
I don't use discord, but I will try my best to start using it next time (hoping it does not happen). 
In any case, we should use CI to help us more. 
Just install a dummy project in the e2e tests that try to use vcpkg as a downstream project, both in a classic mode and in the manifest one.

Neumann-A:
> I'll open a new PR tomorrow; this was simply a case of too many changes to keep track of. I really think I need to work on that...

as I said earlier. Clean it up first and chunk it into smaller parts or move parts of it into functions. The toolchain is currently a mess with manifest options being scattered at the beginning and manifest logic being somewhere in the middle. 

I would like something like this in maybe separate cmake files:
- general variable setup 
- default target selection
- vcpkg path setup
- manifest options/variables & logic
- cmake path setup
- cmake function overrides (which itself could be split into the different functions)
- the rest

This would make reading and understanding the `vcpkg.cmake` so much easier.

strega-nil:
@Neumann-A you can argue with Robert about that; I'm not very opinionated either way. I don't think that breaking it up would have helped with this specific thing tho.

PR #16302, opened 2021-02-19
[bext-di] Rename `boost-di` to `bext-di`

kris-jusiak:
Problem:
- `boost-di` is not an official Boost library and it shouldn't use boost name.

Solution:
- Rename it to `di` instead.


ports/di/vcpkg.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "boost-di",
+  "name": di",
```
vicroms, line 2:
```suggestion
  "name": "di",
```

Missing an opening `"`.

vicroms (changes requested):
We agree with distinguishing it from official Boost libraries, but just `di` is too much of a generic name.
 Also, we would want to make it so that the name doesn't match a `boost*` regex. 

What about renaming this `extboost-di` instead?

ports/di/vcpkg.json
krzysztof-jusiak: I see, however, it can't have boost in the name, what about `dependency-injection` then?

PhoebeHui: How about bext-di?

krzysztof-jusiak: I'm okay with bext-di :+1:

PhoebeHui: @kreuzerkrieg, could you please update the port name to 'bext-di' and continue this PR?

PhoebeHui: BTW,  you can format the vcpkg.json file by command './vcpkg format-manifest' ./ports/bext-di/vcpkg.json.

kreuzerkrieg: hehe, how come my name suddenly popped up here? :)
will take a look at the evening

PhoebeHui: Sorry for my mistakes!

ahmedyarub:
There are still some references within vcpkg to boost-di. I don't think that you have renamed it everywhere.
Ex: https://github.com/microsoft/vcpkg/blob/master/versions/baseline.json#L539

PR #16303, opened 2021-02-19
[vcpkg.cmake] add VCPKG_INSTALLED_DIR as a settable option

strega-nil:
This is something that #16061 tried to do, but screwed up; see #16293 . This one fixes that.


Neumann-A:
You need to also fix the 68 matches for `_VCPKG_INSTALLED_DIR` in `/ports/*`. Otherwise you break the wrappers and so on.

strega-nil:
@Neumann-A how about the new version, which includes back-compat considerations?

ras0219-msft (approved):

ras0219-msft:
ping @strega-nil that this LGTM

strega-nil-ms:
Thanks robert!

PR #16313, opened 2021-02-20
[vcpkg_configure_make][vcpkg_build_make] Use prefix without DESTDIR

ras0219-msft:
**This PR has been superceded by https://github.com/microsoft/vcpkg/pull/16775 and remains open as a draft for the other improvements to be picked out**

Because vcpkg requires packages to be relocatable, there's no functional difference between `DESTDIR` and `--prefix`. By assigning `--prefix` to `packages\`, we can simplify the build process and avoid doubling path lengths.

### DIFF
--- a/ports/openmpi/opal_wrapper.rel.sh
+++ b/ports/openmpi/opal_wrapper.rel.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "$0" )" && pwd )"
+export OPAL_PREFIX="$( cd "$DIR/../../.." && pwd )"
+exec -a "$0" "$DIR/opal_wrapper_real" "$@"
--- a/ports/apr-util/CONTROL
+++ b/ports/apr-util/CONTROL
@@ -1,6 +1,0 @@
-Source: apr-util
-Version: 1.6.1
-Port-Version: 3
-Homepage: https://apr.apache.org/
-Description: Apache Portable Runtime (APR) project  mission is to create and maintain software libraries that provide a predictable and consistent interface to underlying platform-specific implementation
-Build-Depends: expat, apr, openssl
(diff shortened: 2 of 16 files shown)


JackBoosY:

ras0219:
@cenit As part of this, I've run into the non-relocatability of openmpi -- I believe I've built a solution using wrapper scripts, but if you have time I'd appreciate a second opinion.

https://github.com/microsoft/vcpkg/pull/16313/files#diff-f81251de4b55e635fe558a125065447ca8c81e9f5e916f6ff6341a3d9996a082

cenit:
> @cenit As part of this, I've run into the non-relocatability of openmpi -- I believe I've built a solution using wrapper scripts, but if you have time I'd appreciate a second opinion.
> 
> 
> 
> https://github.com/microsoft/vcpkg/pull/16313/files#diff-f81251de4b55e635fe558a125065447ca8c81e9f5e916f6ff6341a3d9996a082

thanks for asking. Tomorrow i will look into it :)

ports/openmpi/opal_wrapper.rel.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
cenit, line 1: first thing i notice, bash is "too much" on many hpc nodes (and also this shabang is not posix). The script should just require /bin/sh

tomorrow i will test your scripts but the workaround looks ok. 
in any case, you are right, open-mpi was not relocatable. My bad, very sorry for that.

ras0219: I started with `/bin/sh`, but I couldn't find documentation for the `exec -a <progname>` style (everyone assumes bash :p).

cenit: i can test easily dash, which is not the original sh but is a very nice modern clone ;)

JackBoosY:

PR #16323, opened 2021-02-20
[docs] Add CMake Integration to manifests.md

ras0219:


docs/users/manifests.md
```diff
@@ -2,35 +2,32 @@
...
+installed tree is not associated with an individual project, it's similar to tools like `brew` or `apt`, except that the
+installed tree is vcpkg-installation-local, rather than global to a system or user.
+
+In manifest mode, an installed tree is associated with a particular project rather than the vcpkg installation. The set
+of installed ports is controlled by editing the project's "manifest file", and the installed tree is placed in the
+project directory or build directory. This mode acts more similarly to language package managers like Cargo, or npm. We
+recommend using this manifest mode whenever possible, because it allows one to encode a project's dependencies
+explicitly in a project file, rather than in the documentation, making your project much easier to consume.
+
+Manifest mode is in beta, but one can use it from the CMake or MSBuild integration, which will be stable when used via
```
Hoikas, line 17:
```suggestion
Manifest mode is in beta, but it can be used from the CMake or MSBuild integration, which will be stable when used via
```

Change to active voice to match the surrounding text.

strega-nil (approved):

PR #5914, opened 2019-04-02
[fix][jsonnet] Improving jsonnet build

andreadps:
- linking md5.cpp into the library, like the original Makefile: https://github.com/google/jsonnet/blob/master/Makefile#L89
    - otherwise, build would always create a static `libmd5.a` that could conflict with other md5 libraries
- renaming targets so we do not end up with `liblibjsonnet.so`

PR #16335, opened 2021-02-21
[jsonnet] Incorporate md5.cpp into jsonnet library

tetsuh:
Resolve the issue of libjsonnet.a not including md5.cpp, previously discussed in #5914.

This is also a problem in the original [jsonnet](https://github.com/google/jsonnet/) repository, where the make build includes md5.cpp in libjsonnet, but the CMake build separates it into libjsonnet and libmd.

This is better solved with upstream, but I would like to use jsonnet sooner, and only vcpkg developers care about CMakeLists.txt..., so please allow patches to be added to vcpkg.  
Since vcpkg already applies patches to CMakeLists.txt for Visual Studio build, this doesn't really change.

One thing to note is the mixing of licenses.
[jsonnet](https://github.com/google/jsonnet/) and [third_party/md5](https://github.com/google/jsonnet/tree/master/third_party/md5) are under different licenses. jsonnet is under the apache license, while md5.cpp is under a different license belonging to RSA Data Security, Inc.  
However, since Linux distributions (including debian) offer libjsonnet as a mixed license as well, this should not be a problem.

This patch is expected to work triplet-independently.

### DIFF
--- a/ports/jsonnet/0004-incorporate-md5.patch
+++ b/ports/jsonnet/0004-incorporate-md5.patch
@@ -1,0 +1,42 @@
+diff --git a/core/CMakeLists.txt b/core/CMakeLists.txt
+index d4e77a8..5d36314 100644
+--- a/core/CMakeLists.txt
++++ b/core/CMakeLists.txt
+@@ -15,7 +15,8 @@ set(LIBJSONNET_HEADERS
+     static_error.h
+     string_utils.h
+     unicode.h
+-    vm.h)
++    vm.h
++    ../third_party/md5/md5.h)
+ 
+ set(LIBJSONNET_SOURCE
+     desugarer.cpp
+@@ -26,12 +27,13 @@ set(LIBJSONNET_SOURCE
+     pass.cpp
+     static_analysis.cpp
+     string_utils.cpp
+-    vm.cpp)
++    vm.cpp
++    ../third_party/md5/md5.cpp)
+ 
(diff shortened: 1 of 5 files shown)


tetsuh:
Oh, versioning...
Should I commit versions update from the result of  vcpkg x-add-version  ?

```diff
diff --git a/versions/baseline.json b/versions/baseline.json
index 6a2259ff5..be2748343 100644
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -2686,7 +2686,7 @@
     },
     "jsonnet": {
       "baseline": "0.16.0",
-      "port-version": 1
+      "port-version": 2
     },
     "jwt-cpp": {
       "baseline": "0.4.0",
diff --git a/versions/j-/jsonnet.json b/versions/j-/jsonnet.json
index 6af3ca65d..29ee5da44 100644
--- a/versions/j-/jsonnet.json
+++ b/versions/j-/jsonnet.json
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "53efaebae6e8131ee43960d0d58a5f110298a9c1",
+      "version-string": "0.16.0",
+      "port-version": 2
+    },
     {
       "git-tree": "77e5fa13bd74db46bf2626d82e8b9f88805031b6",
       "version-string": "0.16.0",
```

ras0219-msft:
This LGTM, thanks for the PR!

PR #16339, opened 2021-02-21
[psimd] add a new port

luncliff:
### What does your PR fix?

There was no port request for this project.

This is one of the 3rd party libraries for [the PyTorch project](https://github.com/pytorch/pytorch/blob/master/.gitmodules). The PR will be used for future support of the `libtorch` port.

### Which triplets are supported/not supported? Have you updated the CI baseline?

This port is a header-only library, so every triplet must be available.

### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

The library doesn't have a tag/release.
 
* Manifests and CONTROL files
    * [x] `vcpkg format-manifest`
* Versioning
    * [x] Follow common conventions for the "version" field  
      [... we use the date that the commit was accessed by you, formatted as `YYYY-MM-DD`.](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/manifest-files.md#version-fields)

PR #16340, opened 2021-02-21
[fp16] create a new port

luncliff:
### What does your PR fix?

There was no port request for this project.

This is one of the 3rd party libraries for [the PyTorch project](https://github.com/pytorch/pytorch/blob/master/.gitmodules). The PR will be used for future support of the `libtorch` port.

This port is a header-only library, so every triplet must be available.

The library doesn't have a tag/release.
 
* PR Structure
    * [x] Make separate Pull Requests per port.  **The PR requires #16339**
* Manifests and CONTROL files
    * [x] `vcpkg format-manifest`
* Versioning
    * [x] Follow common conventions for the "version" field  
      [... we use the date that the commit was accessed by you, formatted as `YYYY-MM-DD`.](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/manifest-files.md#version-fields)


luncliff:
This PR will be rebased after #16339 is merged. To prevent merge, it will remain in the draft mode.

ports/fp16/portfile.cmake
```diff
@@ -0,0 +1,20 @@
+
```
NancyLi1013, line 1:
```suggestion
```

ports/fp16/vcpkg.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "fp16",
+  "version-string": "2021-02-21",
```
NancyLi1013, line 3:
```suggestion
  "version-date": "2021-02-21",
```

We prefer to use `version-date` in this case. Please see https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions

versions/f-/fp16.json
```diff
@@ -0,0 +1,9 @@
+{
+  "versions": [
+    {
+      "git-tree": "e4e945808018cb1bbc0a54c8d8f186e61bd97bc1",
+      "version-string": "2021-02-21",
```
NancyLi1013, line 5:
```suggestion
      "version-date": "2021-02-21",
```
The same as above.

ports/fp16/vcpkg.json
```diff
@@ -0,0 +1,9 @@
...
+  "description": "Conversion to/from half-precision floating point formats",
```
NancyLi1013, line 4:
```suggestion
  "description": "Header-only library for conversion to/from half-precision floating point formats",
```
It would be better to add `header-only library` here.

NancyLi1013 (changes requested):

ports/fp16/vcpkg.json
luncliff: Updated in 4cdaa9d. Thanks for the point!

ports/fp16/vcpkg.json
luncliff: Applied the suggestion in 4cdaa9d.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for adding this new port @luncliff.

PR #16375, opened 2021-02-23
[boost] /safeseh

autoantwort:
**Describe the pull request**

- What does your PR fix?  
 Builds boost on windows x86 with `/safeseh`. Otherwise you get linker errors if you build your application with `/safeseh` but boost is not build with `/safeseh`.

- Should boost always be built with `/safeseh` for all users in all circumstances?  
  Yes! It improves the security of the resulting application (see [here](https://stackoverflow.com/questions/25081033/what-safesehno-option-actually-do)) and has no downsides.

### DIFF
--- a/ports/boost-modular-build-helper/boost-modular-build.cmake
+++ b/ports/boost-modular-build-helper/boost-modular-build.cmake
@@ -307,6 +307,11 @@ function(boost_modular_build)
         list(APPEND B2_OPTIONS address-model=64 architecture=power)
     else()
         list(APPEND B2_OPTIONS address-model=32 architecture=x86)
+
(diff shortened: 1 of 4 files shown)


PhoebeHui:
Related to old PR https://github.com/microsoft/vcpkg/pull/16142

ras0219-msft (approved):
Sorry for missing the pings in the other PR; thanks for opening a new one :)

Looking on the boost mailing list, it seems this is appropriate to always apply. Thanks for the PR!

cenit:
should it be promoted to a toolchain flag?

autoantwort:
> should it be promoted to a toolchain flag?

Maybe not an bad idea. Maybe create a Issue or PR to test if the packages will build with it.

PR #16377, opened 2021-02-23
[scripts-audit] CMake buildsystem!!!

strega-nil:
See PR #16055, issue #16188

Depends on PR #16419

Part of the "merge 2021-02-26 EOD" group.

cc @ras0219, @Neumann-A for review


ports/fmt/CONTROL
```diff
@@ -1,5 +1,6 @@
 Source: fmt
 Version: 7.1.3
-Port-Version: 0
+Port-Version: 1
 Homepage: https://github.com/fmtlib/fmt
 Description: Formatting library for C++. It can be used as a safe alternative to printf or as a fast alternative to IOStreams.
+Build-Depends: vcpkg-cmake-buildsystem
```
Neumann-A, line 6: isn't `vcpkg-cmake` enough? (or use `vcpkg-scripts-cmake`).
I don't feel like `buildsystem` is the right descriptor here.

ports/fmt/portfile.cmake
```diff
@@ -1,3 +1,5 @@
+include(vcpkg-cmake-buildsystem)
```
Neumann-A, line 1: please avoid this somehow. (e.g. by passing the dependency?)

ports/fmt/portfile.cmake
```diff
@@ -6,16 +8,17 @@ vcpkg_from_github(
     HEAD_REF master
     PATCHES fix-warning4189.patch
 )
-vcpkg_configure_cmake(
+vcpkg_cmake_buildsystem_configure(
```
Neumann-A, line 11: If you need another name why not `vcpkg_cmake_configure`. 
Also scripts need to be versioned so having `vcpkg_configure_cmake` use `cmake_language(CALL <command> [<arg>...])` and have command be something like `vcpkg_configure_cmake_<version>`. The reason behind this is that probably multiple versions of the scripts are required.

Neumann-A: scripts somehow does not work well with the single version behavior vcpkg currently has if you want to support multiple versions of ports which might depend on different script versions.

ports/fmt/portfile.cmake
```diff
@@ -40,7 +43,7 @@ if(VCPKG_LIBRARY_LINKAGE STREQUAL dynamic)
 endif()
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 
-vcpkg_fixup_cmake_targets()
+vcpkg_cmake_buildsystem_fixup_targets()
```
Neumann-A, line 46: Although targets are predominantly generated by cmake there is a high possibility that meson will also generate configs/targets. I also saw autoconf generate a config for something. So the name here is wrong.

ports/fmt/CONTROL
strega-nil: I'm not sure, I just kinda chose a name because I needed to; please post your thoughts on a better name.

ports/fmt/portfile.cmake
strega-nil: I don't really want to avoid this, and I honestly can't think of how?

ports/fmt/portfile.cmake
strega-nil: This does not make any sense to me.

ports/fmt/portfile.cmake
strega-nil: that's fair enough.

ports/fmt/portfile.cmake
Neumann-A: The `Build-Depends:` already has the information for the buildsystem. vcpkg needs to pass the information (+feature info of depends) to cmake any way so that some downstream ports might react to the linkage difference of different depends/features. If you have the dep info in the cmake script available you can just hide the `include` here away. 

also if you only use `include(vcpkg-cmake-buildsystem)` here users are able to override `CMAKE_MODULE_PATH` in a triplet.

ports/fmt/portfile.cmake
Neumann-A: port1#V1 depends on scripts#V4; port2#V15 depends on scripts#V2; 
port1#V1 does not build with scripts#V2 and port2#V15 does not build with scripts#V4 (because somebody broke backcompat) for some reason (port2#V50 does but that does not matter here.)

scripts are a bit different to normal port dependencies, in my opinion. You can have multiple version of scripts at the same time if you version the folders|files|functions without bigger problems.

strega-nil:
Closing in favor of a single PR that contains all of these PRs. (to avoid destroying the CI system)

ports/vcpkg-cmake/vcpkg_cmake_configure.cmake
```diff
@@ -0,0 +1,391 @@
...
+        set(targetting_uwp ON)
+    endif()
+
+    if(host_architecture STREQUAL "x86")
+        # Prebuilt ninja binaries are only provided for x64 hosts
+        set(ninja_can_be_used OFF)
+        set(ninja_host OFF)
+    elseif(targetting_uwp)
+        # Ninja and MSBuild have many differences when targetting UWP, so use MSBuild to maximize existing compatibility
+        set(ninja_can_be_used OFF)
```
JackBoosY, line 129: That breaks the behavor, see https://github.com/microsoft/vcpkg/issues/19818

PR #16378, opened 2021-02-23
[boost-context] fix linux -> windows cross compile

### LINKED ISSUE #16352: [boost-context] build failure (linux -> windows cross compile)
**Host Environment**
- OS: Linux (Docker Container with [mxe](https://github.com/mxe/mxe))
- Compiler: x86_64-w64-mingw32.shared.posix-gcc (GCC) 8.4.0

**To Reproduce**
1. Use the mxe environment + vcpkg with master of vcpkg-tool (for "host": true / #15424 ) + boost-build patches
1. ./vcpkg install boost-context --triplet=x64-linux-windows-static
2. Get error

**Failure logs** 
Not important, but: [install-x64-linux-windows-static-rel-out.log](https://github.com/microsoft/vcpkg/files/6018923/install-x64-linux-windows-static-rel-out.log)

**Solution**
You have to pass `abi=ms binary-format=pe` to b2 when compiling `boost-context`, but not when compiling other boost packages.  
So maybe we need an additional parameter for the function `boost_modular_build`.

autoantwort:
**Describe the pull request**

- What does your PR fix? Fixes #16352

### DIFF
--- a/ports/boost-context/b2-options.cmake.in
+++ b/ports/boost-context/b2-options.cmake.in
@@ -1,0 +1,7 @@
+
+if(@VCPKG_TARGET_IS_WINDOWS@)
+    list(APPEND B2_OPTIONS
+        abi=ms
+        binary-format=pe
+    )
+endif()
--- a/ports/boost-context/portfile.cmake
+++ b/ports/boost-context/portfile.cmake
@@ -15,7 +15,15 @@ file(WRITE "${SOURCE_PATH}/build/Jamfile.v2" "${_contents}")
 file(COPY "${CURRENT_INSTALLED_DIR}/share/boost-config/checks" DESTINATION "${SOURCE_PATH}/build/config")
 
 include(${CURRENT_INSTALLED_DIR}/share/boost-build/boost-modular-build.cmake)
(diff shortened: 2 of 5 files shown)


cenit:
usually boost port files are auto generated. it might be better to fix the generating script in order to not loose your mods the next update

autoantwort:
Yeah the script https://github.com/microsoft/vcpkg/blob/master/scripts/boost/generate-ports.ps1#L185 uses the `b2-options.cmake.in` file and creates the same code as I

PhoebeHui (approved):
LGTM, thanks for the PR!

ras0219-msft:
This PR LGTM, however should this fragment always be applied to all boost builds? Or is boost-context the only library that cares about `abi=ms`?

ras0219-msft (approved):

autoantwort:
> This PR LGTM, however should this fragment always be applied to all boost builds? Or is boost-context the only library that cares about `abi=ms`?

No this is a boost-context only option. See [here](https://github.com/boostorg/context/blob/develop/build/Jamfile.v2#L87).

ras0219-msft:
LGTM, thanks again!

PR #16309, opened 2021-02-19
[scripts audit] vcpkg_copy_pdbs

strega-nil:
See [scripts-audit], issue #16188 

[scripts-audit]: https://github.com/microsoft/vcpkg/blob/master/docs/specifications/scripts-extraction.md

Part of the "merge 2021-02-26 EOD" group.

PR #16385, opened 2021-02-24
[vcpkg baseline] [itpp] Fix include directory

PhoebeHui:
Relate to https://github.com/microsoft/vcpkg/pull/16309

Fix the failures:
FFT_INCLUDES (ADVANCED)
   used as include directory in directory E:/vcpkg/clean/vcpkg/buildtrees/itpp/src/b46d048fa7-9444ef0e36/itpp
   used as include directory in directory E:/vcpkg/clean/vcpkg/buildtrees/itpp/src/b46d048fa7-9444ef0e36/itpp


JackBoosY (approved):

ports/itpp/vcpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "itpp",
+  "version-semver": "4.3.1",
+  "port-version": 6,
+  "description": "IT++ is a C++ library of mathematical, signal processing and communication classes and functions. Its main use is in simulation of communication systems and for performing research in the area of communications.",
+  "homepage": "http://itpp.sourceforge.net"
```
strega-nil, line 6:
```suggestion
  "homepage": "http://itpp.sourceforge.net",
  "dependencies": [
    "fft"
  ]
```

I think this is necessary, right?

ports/itpp/fix_include.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/itpp/CMakeLists.txt
+@@ -42,8 +42,10 @@ include_directories ( ${CMAKE_BINARY_DIR}
+                       ${CMAKE_SOURCE_DIR}
+                       ${BLAS_INCLUDES}
+                       ${LAPACK_INCLUDES}
+-                      ${FFT_INCLUDES}
+                     )
++if(FFT_FOUND)
++    include_directories(${FFT_INCLUDES})
++endif()
```
strega-nil, line 13:
```suggestion
+if(FFT_FOUND)
+    include_directories(${FFT_INCLUDES})
+else
+    message(FATAL_ERROR "failed to find FFT")
+endif()
```

this should fail if FFT is not found, so that the resulting package doesn't depend on whether FFT is found or not.

strega-nil (changes requested):

ports/itpp/vcpkg.json
PhoebeHui: @strega-nil, we don't have fft library in vcpkg, so I haven't add it to the dependency.

ports/itpp/fix_include.patch
PhoebeHui: It already check the fft library and give a waring in the main CMakeLists.txt file which under the root directory of itpp, so I don't this is needed.

ports/itpp/portfile.cmake
```diff
@@ -10,6 +10,7 @@ vcpkg_from_sourceforge(
        msvc2013.patch
        fix-uwp.patch
        fix-linux.patch
+       fix_include.patch
```
ras0219, line 13: Instead of this patch, we should disable optional dependencies:
```cmake

    OPTIONS
        -DCMAKE_DISABLE_FIND_PACKAGE_LAPACK=ON
        -DCMAKE_DISABLE_FIND_PACKAGE_FFT=ON
        -DCMAKE_DISABLE_FIND_PACKAGE_BLAS=ON
```

On my machine, these options cause the build to succeed without this patch.

ras0219 (changes requested):

ports/itpp/portfile.cmake
PhoebeHui: @ras0219, thanks for your review!

strega-nil (approved):

ras0219-msft:
Merging this without a full CI success due to infrastructure issues.

PR #16396, opened 2021-02-24
[boost] fix cross compiling from linux -> windows

autoantwort:
Fixes the boost build if you are cross compile from linux -> windows and you are using #15424/microsoft/vcpkg-tool#5


autoantwort:
Covered in #16479

PR #16401, opened 2021-02-25
[breakpad] add support for Android triplets

jwtowner:
This changes vcpkg's custom CMakeLists.txt file for Breakpad so that it correctly builds libbreakpad_client for Android. It follows the same source file layout for Android as seen in [./android/google_breakpad/Android.mk](https://github.com/google/breakpad/blob/main/android/google_breakpad/Android.mk). Additionally, the -D_LIBCPP_VERSION preprocessor definition was removed, since this is defined when including libc++ headers.

Linux and Windows target triplets are unaffected by this change. This has been tested with Android NDK r21e and Android NDK r22.

### DIFF
--- a/ports/breakpad/CMakeLists.txt
+++ b/ports/breakpad/CMakeLists.txt
@@ -42,6 +42,8 @@ if(NOT CMAKE_SYSTEM_NAME STREQUAL Android)
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
             $<INSTALL_INTERFACE:include>
     )
+
+    set(TARGETS libbreakpad)
 endif()
 
 # libbreakpad_client target
--- a/ports/breakpad/CMakeLists.txt
+++ b/ports/breakpad/CMakeLists.txt
@@ -11,7 +11,6 @@ add_definitions(
     -D_CRT_SECURE_NO_WARNINGS
     -D_CRT_SECURE_NO_DEPRECATE
     -D_CRT_NONSTDC_NO_DEPRECATE
-    -D_LIBCPP_VERSION
 )
 
 set(CMAKE_DEBUG_POSTFIX d)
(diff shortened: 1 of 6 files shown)


JackBoosY:
Please get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=49827&view=artifacts&pathAsName=false&type=publishedArtifacts).

jwtowner:
So, since `-D_LIBCPP_VERSION` was removed in order for toolchains that actually use Clang/LLVM with libc++ to work correctly, I had to come up with another solution to get the preprocessor check in `src/common/unordered.h` compiling again for all targets.

MSVC++ 2013 (`_MSC_VER >= 1800`) added `unordered_map`, but does not define `__cplusplus` to the expected C++11 value of `201103L` in order to not break backwards compatibility. Only by specifying the `/std:` command line switch with `c++14` or later in MSVC++ 2017 (`_MSC_VER >= 1911`) has any effect on `__cplusplus` (see [here](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/)).

So for maximum compatibility, and without changing the pre-existing behavior around the `_LIBCPP_VERSION` macro in breakpad, I've set the preprocessor check for C++11's unordered_map to: `(__cplusplus >= 201103L) || defined(_LIBCPP_VERSION) || (defined(_MSC_VER) && (_MSC_VER >= 1800))`.

JackBoosY (approved):

ports/breakpad/CMakeLists.txt
```diff
@@ -20,42 +19,70 @@ string(COMPARE EQUAL "${CMAKE_BUILD_TYPE}" "Release" DEFAULT_INSTALL_HEADERS)
...
-        $<INSTALL_INTERFACE:include>
-)
+    target_include_directories(libbreakpad
+        PUBLIC
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
+            $<INSTALL_INTERFACE:include>
+    )
+endif()
 
-set(TARGETS libbreakpad)
```
ras0219-msft, line 46: Based on my reading, this no longer installs `libbreakpad` on all non-android platforms.

ras0219-msft (changes requested):

jwtowner:
Looks like there's a couple of unrelated regressions in gdal:x86-windows and opencv4:x86-windows.

ras0219-msft (approved):
Thanks!

To avoid the android sources list falling out of date, I've switched to actually regexing the sources list from the `Android.mk` file itself. Unless you see any issues, this final iteration LGTM

jwtowner:
@ras0219-msft 
> Thanks!
> 
> To avoid the android sources list falling out of date, I've switched to actually regexing the sources list from the `Android.mk` file itself. Unless you see any issues, this final iteration LGTM

Nice, looks good to me.

PR #16420, opened 2021-02-25
[python3] Add feature deprecated-win7-support.

Hoikas:
The PSF has a [policy](https://www.python.org/dev/peps/pep-0011/#microsoft-windows) for dropping support for any Microsoft OS that is beyond end of support. As of Python 3.,9, Windows 7 support was dropped, preventing use the of the python3 port on any Windows 7 machine. The support drop was relatively soft, so this feature restores support for Windows 7 behind and off-by-default feature. The included patch forward-ports the fix for [bpo-39401](https://bugs.python.org/issue39401) from the Python 3.8 branch due to the Python 3.9 version requiring Windows 8. See python/cpython#18234 whose content is included in the `0007-restores-support-for-windows-7.patch`, replacing the content of python/cpython#18231.

cc @ras0219-msft re discussion about this in discord last week.


NancyLi1013 (approved):

ras0219-msft (approved):
I reverted changes to the intermediate patch files which were otherwise unaffected -- other than that this LGTM.

PR #16434, opened 2021-02-26
[libpq] fix filenames (linux -> windows cross compile)

autoantwort:
Linux distinguish between upper and lower case.


NancyLi1013:
Hi @autoantwort 

Thanks for your PR.

Could you please help describe simply what problem you have encountered?

Thanks.

autoantwort:
The portfile wants to apply the path `patches/windows/Solution_RELEASE.patch`, but the filename is `ports/libpq/patches/windows/Solution_Release.patch`. Then the build fails because the file `patches/windows/Solution_RELEASE.patch` is not found. On Windows this does work, because the windows file system does not care about upper and lower case, but the linux filesystem distinguish between upper and lower case.

ras0219-msft:

PR #16461, opened 2021-02-28
[liblsl] Update to 1.14.0

chausner:
This updates liblsl to the latest version 1.14.0.

It also fixes an issue that prevented the previous ports from compiling on Linux and probably MacOS. So I removed the corresponding entries from the CI baseline assuming that it now builds fine on both platforms. I successfully tested the port on Windows and Linux but have no way to test on MacOS.

I once again tried to get static builds working but it appears bigger changes need to be made in the original repository so I kept it disabled for now.


chausner:
FYI The fix to make it compile also on non-Windows platforms was to simply remove the quotation marks where `-Dlslgitrevision=` is defined.

ports/liblsl/CONTROL
```diff
@@ -1,3 +1,4 @@
 Source: liblsl
-Version: 1.13.1
+Version: 1.14.0
+Homepage: https://github.com/sccn/liblsl
 Description: C++ lsl library for multi-modal time-synched data transmission over the local network
```
JackBoosY, line 4:
```suggestion
Description: C++ lsl library for multi-modal time-synched data transmission over the local network
Supports: !(static & windows)
```

JackBoosY (changes requested):

versions/l-/liblsl.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "0d55980b8a8da6eeea79d0fde8882ed33abbac52",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "4ade6d7cc872bf269aef8cae340180adf38d4900",
```

JackBoosY (changes requested):

JackBoosY (approved):

chausner:
@JackBoosY Thanks a lot for your refactorings, it's much cleaner now. 👍 Just tested again on Linux and Windows, and looks good to me overall.

chausner:
One more note: liblsl bundles pugixml but it can be turned off by setting `LSL_BUNDLED_PUGIXML=OFF`. I already tested with using the pugixml dependency in vcpkg and it compiles fine. However, I decided against using the vcpkg dependency for that since liblsl can only be built as a dynamic library at the moment (so there should not be any problem if the user links against pugixml in their own application again). Also, using the bundled version of pugixml is probably safer than some potentially incompatible version that comes in vcpkg.

I am not sure what the general recommendation is, but if you prefer to use the vcpkg dependency instead of the bundled one, I am happy to change it.

ras0219-msft (changes requested):
Thanks for the PR, overall it looks great!

Our policy is that we must use the vcpkg copy of pugixml: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-vendored-dependencies

LGTM once this is addressed :)

chausner:
> Our policy is that we must use the vcpkg copy of pugixml: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-vendored-dependencies

Alright, I made the necessary changes in 8932039aa91489da055f4d8a7da8652deb7e16b5.

What about Boost which is currently bundled in parts in liblsl, see https://github.com/sccn/liblsl#boost? In contrast to pugixml, it cannot be disabled easily so if we also want to replace it with the Boost vcpkg package, we need to resort to patching, I'm afraid.

versions/l-/liblsl.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "766843e1e973aa6a87e5d4aa8386e45906544fb8",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "5be8e82edc3261bef4114e428191a4ccc9271892",
```

JackBoosY (changes requested):

JackBoosY (approved):

JackBoosY:
@chausner Yep, we should patch its source to use boost components also.

chausner:
> @chausner Yep, we should patch its source to use boost components also.

I looked into this a bit but it seems non-trivial to do. They took a copy of Boost and changed all namespaces to "lslboost". So to make it work with the original Boost headers, we would have to patch the liblsl code base to use the original names everywhere. I am not sure if it is worth the effort. There also does not seem to be an easy way to do this in CMake without resorting to platform-dependent commands like `sed` or iterating through all files and doing a string-replace.

In any case, since the bundling of Boost in liblsl is independent of this package update (Boost has always been bundled with liblsl), I would suggest to merge this PR for now so that we have at least the latest version of the library in vcpkg. The patching to remove the bundled Boost copy could be done separately if anybody feels like it (but it will probably not be me).

JackBoosY:
> > @chausner Yep, we should patch its source to use boost components also.
> 
> I looked into this a bit but it seems non-trivial to do. They took a copy of Boost and changed all namespaces to "lslboost". So to make it work with the original Boost headers, we would have to patch the liblsl code base to use the original names everywhere. I am not sure if it is worth the effort. There also does not seem to be an easy way to do this in CMake without resorting to platform-dependent commands like `sed` or iterating through all files and doing a string-replace.
> 
> In any case, since the bundling of Boost in liblsl is independent of this package update (Boost has always been bundled with liblsl), I would suggest to merge this PR for now so that we have at least the latest version of the library in vcpkg. The patching to remove the bundled Boost copy could be done separately if anybody feels like it (but it will probably not be me).

Accepted.

ras0219-msft:
This PR LGTM, thanks for detaching pugixml!

PR #16492, opened 2021-03-02
[Azure pipeline] Upload file list logs for unstable testing

PhoebeHui:
'File list' logs is very important for investigating the conflicts issue in unstable testing, so upload the logs as well.

The Changes from https://github.com/microsoft/vcpkg/blob/master/scripts/azure-pipelines/windows/azure-pipelines.yml


ras0219-msft:
LGTM

PR #16494, opened 2021-03-02
[proj4] Update to 8.0.0

### LINKED ISSUE #16906: FindProj not working as expected
so i've got a problem at #14628 .
Proj is currently at version 7.2.0 and it should be supported as stated in https://proj.org/development/cmake.html to use `find_package(PROJ REQUIRED)` however when i'm using it like that I'm getting the following error: 
```
CMake Error at D:/dev/vcpkg_mapnik/scripts/buildsystems/vcpkg.cmake:861 (_find_package):
  By not providing "FindPROJ.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "PROJ", but
  CMake did not find one.

  Could not find a package configuration file provided by "PROJ" with any of
  the following names:

    PROJConfig.cmake
    proj-config.cmake

  Add the installation prefix of "PROJ" to CMAKE_PREFIX_PATH or set
  "PROJ_DIR" to a directory containing one of the above files.  If "PROJ"
  installed.
  cmake/MapnikFindPackage.cmake:18 (find_package)
  CMakeLists.txt:209 (mapnik_find_package)

```
If I'm building proj without vcpgk this will work, so there have to be a bug in the portfile of proj?

rhuijben:
**Describe the pull request**

- What does your PR fix?
Upgrade proj[4] to 8.0.0

Minimal change to bump the version. Removes the legacy (proj4 like) api, and therefore fixes some dependencies.

Originally this pull request contained some work to unconditonally enable tiff support, but this part is now removed. This same support is now independently(?) added to [#20443](https://github.com/microsoft/vcpkg/pull/20443) by @dg0yt.

This pull request (now open for half a year) re-opens the proj project for bumping to more recent versions (2021-11-15: last version 8.2.0)

All existing triplets supported (Windows, Linux, OS/X, UWP)

### DIFF
--- a/ports/mapnik/proj-find-fix.patch
+++ b/ports/mapnik/proj-find-fix.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 47bbaa1..43f384d 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -304,7 +304,7 @@ if(USE_WEBP)
+ 
+ if(USE_PROJ)
+     #https://proj.org/development/cmake.html
+-    mapnik_find_package(PROJ ${PROJ_MIN_VERSION} QUIET)
++    mapnik_find_package(PROJ REQUIRED)
+     # currently the cmake files are not installed, when installing proj via apt-get. So search via pkg-config
+     if(NOT PROJ_FOUND)
+         message(STATUS "PROJ not found via FindPROJ. Searching via pkg-config...")
--- a/ports/proj4/portfile.cmake
+++ b/ports/proj4/portfile.cmake
@@ -1,9 +1,9 @@
-set(VERSION 7.2.1) # Variable used for pc file
+set(VERSION 8.0.0) # Variable used for pc file
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO OSGeo/PROJ
     REF "${VERSION}"
-    SHA512 e6e77266dcd70c939c16667c916cccab8de161221d2ef600cfca43382f50da2dc8d790561556b4416adbb4ac6fba939004e0cc936c278e0e808dc3566e9a70d4
+    SHA512 353f0e14ba6c11bfcec8a30b3f322842e8a3103546e0fd40871e341c859ca7ae2b9411425be5b588eb45cc5716255d2ea81d6357407bcfb9506a0edb161376f5
     HEAD_REF master
     PATCHES
         fix-filemanager-uwp.patch
--- a/ports/vtk/portfile.cmake
+++ b/ports/vtk/portfile.cmake
@@ -29,6 +29,7 @@ vcpkg_from_github(
         d107698a.patch
         fix-gdal.patch
         missing-limits.patch # This patch can be removed in next version. Since it has been merged to upstream via https://gitlab.kitware.com/vtk/vtk/-/merge_requests/7611
+        UseProj5Api.patch # Allow Proj 8.0+ (commit b66e4a7, backported). Should be in soon after 9.0.3
 )
 
 # =============================================================================
--- a/ports/mapnik/portfile.cmake
+++ b/ports/mapnik/portfile.cmake
@@ -7,6 +7,8 @@ vcpkg_from_github(
     REF 14f913d6ab3b0903dd36a1cb2d22f7d5493b8bb8
     SHA512 f90762594d46946ddc512bb19b21c4d6a2f1ce81b7500a326ad512fae3a3f77e49ef3eb727ff8f98a31596e4132528212e0fa146e2eee0a9965a16551cfd0386
     HEAD_REF master
+    PATCHES
+        proj-find-fix.patch # Quiet search with version range somehow fails for proj8+
 )
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
--- a/ports/mapnik/vcpkg.json
+++ b/ports/mapnik/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "mapnik",
   "version-date": "2021-11-15",
+  "port-version": 1,
   "description": "Mapnik is an open source toolkit for developing mapping applications.",
   "homepage": "https://github.com/mapnik/mapnik",
   "supports": "!(static & windows)",
--- a/ports/proj4/vcpkg.json
+++ b/ports/proj4/vcpkg.json
@@ -1,7 +1,6 @@
 {
   "name": "proj4",
-  "version-semver": "7.2.1",
-  "port-version": 7,
+  "version-semver": "8.0.0",
   "description": "PROJ.4 library for cartographic projections",
   "homepage": "https://github.com/OSGeo/PROJ",
   "dependencies": [
--- a/ports/vtk/UseProj5Api.patch
+++ b/ports/vtk/UseProj5Api.patch
@@ -1,0 +1,149 @@
+From 0325638832e35c8c8c6fc96e2c1d887aeea3dd43 Mon Sep 17 00:00:00 2001
+From: Julien Schueller <schueller@phimeca.com>
+Date: Mon, 8 Mar 2021 10:57:46 +0100
+Subject: [PATCH] Geovis: Use proj>=5 api
+
+Closes #18130
+---
+ Geovis/Core/vtkGeoProjection.cxx    | 17 ++++++++++++++++-
+ Geovis/Core/vtkGeoTransform.cxx     | 28 ++++++++++++++--------------
+ ThirdParty/libproj/vtk_libproj.h.in |  7 +------
+ 3 files changed, 31 insertions(+), 21 deletions(-)
+
+diff --git a/Geovis/Core/vtkGeoProjection.cxx b/Geovis/Core/vtkGeoProjection.cxx
+index 7ff6526a5d..0a0d06eba1 100644
+--- a/Geovis/Core/vtkGeoProjection.cxx
++++ b/Geovis/Core/vtkGeoProjection.cxx
+@@ -121,7 +121,11 @@ vtkGeoProjection::~vtkGeoProjection()
+   this->SetPROJ4String(nullptr);
+   if (this->Projection)
+   {
++#if PROJ_VERSION_MAJOR >= 5
++    proj_destroy(this->Projection);
++#else
+     pj_free(this->Projection);
++#endif
+   }
+   delete this->Internals;
+   this->Internals = nullptr;
+@@ -185,13 +189,21 @@ int vtkGeoProjection::UpdateProjection()
+ 
+   if (this->Projection)
+   {
++#if PROJ_VERSION_MAJOR >= 5
++    proj_destroy(this->Projection);
++#else
+     pj_free(this->Projection);
++#endif
+     this->Projection = nullptr;
+   }
+ 
+   if (this->PROJ4String && strlen(this->PROJ4String))
+   {
++#if PROJ_VERSION_MAJOR >= 5
++    this->Projection = proj_create(PJ_DEFAULT_CTX, this->PROJ4String);
++#else
+     this->Projection = pj_init_plus(this->PROJ4String);
++#endif
+     if (!this->Projection)
+     {
+       vtkErrorMacro("Cannot set projection with string " << this->PROJ4String);
+@@ -234,8 +246,11 @@ int vtkGeoProjection::UpdateProjection()
+       stringHolder[i] = param.str();
+       pjArgs[3 + i] = stringHolder[i].c_str();
+     }
+-
++#if PROJ_VERSION_MAJOR >= 5
++    this->Projection = proj_create_argv(PJ_DEFAULT_CTX, argSize, const_cast<char**>(pjArgs));
++#else
+     this->Projection = pj_init(argSize, const_cast<char**>(pjArgs));
++#endif
+     delete[] pjArgs;
+   }
+   this->ProjectionMTime = this->GetMTime();
+diff --git a/Geovis/Core/vtkGeoTransform.cxx b/Geovis/Core/vtkGeoTransform.cxx
+index 5c2c74279d..1c99b6b11b 100644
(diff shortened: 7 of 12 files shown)


rhuijben:
With this release the base package was fixed to properly link curl in static mode.

Some of the depending packages have explicit rules on how to link against proj4 that don't include somewhat optional dependencies that are still required for full reprojection accuracy.

At compile time the code needs to know if it uses the TIFF component (as storage for gridshifts) or not, so this can't be a VCPKG feature knob if I read the code correctly.

I think we should enable the TIFF component unconditionally in the base port and perhaps add a slave port for PROJ4 without TIFF-gridshift support. I will see if I can fix the depending ports

ports/proj4/portfile.cmake
```diff
@@ -18,6 +18,7 @@ if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
   set(TOOL_NAMES cct cs2cs geod gie proj projinfo projsync)
 else()
   set(VCPKG_BUILD_SHARED_LIBS OFF)
+  set(EXTRA_FEATURES tiff ENABLE_TIFF tools BUILD_PROJSYNC tools ENABLE_CURL)
```
NancyLi1013, line 21: If you want to keep these extra features enabled on static build, it seems that there is no difference between dynamic and static. 

So you can handle these features as normal features.

rhuijben:
This should fix the build for libgeotiff and libspatialite for shared library builds. Next step would be checking static and static-md builds.

rhuijben:
The vtk port still depends on the legacy proj4 api that has been deprecated since proj 6.0, would have been removed by proj 7.0 and has finally been removed since proj 8.0.

Not sure what to do here. vtk probably needs a dependency on proj 7.2.1

NancyLi1013:
Seems need to merge master to resolve the conflicts. @rhuijben

meastp:
I'm eagerly awaiting this version update. :) Consider updating to v8.0.1 while you're at it, which [was released yesterday](https://github.com/OSGeo/PROJ/releases/tag/8.0.1).

dg0yt:
How about finally splitting port proj4 (legacy API) from port proj (modern API)?

rhuijben:
> I'm eagerly awaiting this version update. :) Consider updating to v8.0.1 while you're at it, which [was released yesterday](https://github.com/OSGeo/PROJ/releases/tag/8.0.1).

I just committed proj 8.0.1 support to my vcpkg overlay repository on  https://github.com/ampscm/vcpkg_additions

$ vcpkg upgrade --overlay-ports ..\vcpkg_additions\ports proj4:x64-windows-static-md --no-dry-run

rhuijben:
@NancyLi1013 What do we need to do with the VTK dependency that still assumes that it should use the proj4 apis that were deprecated since proj6 and now removed.

VTK has its own old version of proj4 included in its source, and as they recommend using that version there is no one working on fixing their support for newer proj versions.

rhuijben:
@dg0yt, @NancyLi1013 I like your suggestion on splitting between legacy [proj4] compatible and moving to the actual project name [proj] for future versions, but I think that decision should be taken by the VCPKG maintainers.

dg0yt:
PROJ 8 should also bring a pc file for pkg-config.

JackBoosY:
> @NancyLi1013 What do we need to do with the VTK dependency that still assumes that it should use the proj4 apis that were deprecated since proj6 and now removed.
> 
> VTK has its own old version of proj4 included in its source, and as they recommend using that version there is no one working on fixing their support for newer proj versions.

Therefore, I think we should update proj4 and vtk together.

dg0yt:
VTK has been updated: https://gitlab.kitware.com/vtk/vtk/-/issues/18130

NancyLi1013:
@rhuijben 

Could you please resolve the conflicts in this PR?

rhuijben:
Conflicts resolved, bump of vtk added, but I don't expect this to resolve all issues as the fix doesn't appear to be in the vtk release yet.

Fix is in https://github.com/Kitware/VTK/commit/b66e4a74dde620e4db1b789da55efb8433bd35c8

ports/proj4/portfile.cmake
```diff
@@ -14,29 +14,25 @@ vcpkg_from_github(
...
   database BUILD_PROJ_DATABASE
   tools BUILD_CCT
   tools BUILD_CS2CS
   tools BUILD_GEOD
   tools BUILD_GIE
   tools BUILD_PROJ
   tools BUILD_PROJINFO
-  ${EXTRA_FEATURES}
+  tools BUILD_PROJSYNC
+  tools ENABLE_CURL
```
dg0yt, line 33: I don't think `ENABLE_CURL` should be mixed with feature `tools` (even if this was the case before this PR). Most tools can perfectly be used offline. I suggest to add a new feature `net` or  `online`. I would use the same feature name for gdal and other related ports.

ports/vtk/portfile.cmake
```diff
@@ -132,14 +132,15 @@ vcpkg_download_distfile(QT_NO_KEYWORDS_PATCH
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO Kitware/VTK
-    REF 96e6fa9b3ff245e4d51d49f23d40e9ad8774e85e # v9.0.1
-    SHA512 0efb1845053b6143e5ee7fa081b8be98f6825262c59051e88b2be02497e23362055067b2f811eff82e93eb194e5a9afd2a12e3878a252eb4011a5dab95127a6f
+    REF 5e2ca5f054de37638e7473148fddebf9a40f8c98 # v9.0.3
+    SHA512 072380730160851f066845647e42bc1676dbe1ce8f958c9a14d8ed93e41a7564eb0c74339a824a51c53caa66fe420211774a335802e138b5869a550049c714ad
```
dg0yt, line 136: I'm not sure if it is a good idea to interfer with the other PR which updates VTK together with related ports?

ports/proj4/portfile.cmake
rhuijben: Nothing changed here. Online behavior needs explicit enabling at runtime anyway, but disabling this (as was done by the previous version bump) breaks the VCPKG guidelines that these flags should not change the ABI.

Without these flags it is not possible to apply gridshifts, even when they are required by the transforms.

ports/vtk/portfile.cmake
rhuijben: Updating VTK in sync with this change was explicitly requested. It is a minor version bump, and adds the fix for proj. Other PRs can be easily fixed.

(The only other PR that I see appears to add Paraview specific versioning inside the primary vcpkg repository.... I thought that was what overlays were invented for😕😕)

rhuijben:
One remaining failure: libspatialite on linux fails to build now.

ports/proj4/portfile.cmake
dg0yt: Okay, ABI change is bad, indeed. But:

Does `ENABLE_CURL` really create a different ABI, or just add additional interfaces? 
(If port A is build with `proj[core]`, won't it run against `proj[net]`?)

If yes, bundling `ENABLE_CURL` with the `tools` feature doesn't help: If `ENABLE_CURL` changes ABI,  the ABI would now change with feature `tools`.

NancyLi1013:
The failures caused by `libspatialite `on x64-linux:

```
./../src/atialite-5-7ecbdc35b2.clean/configure: line 14240: 0: command not found
./../src/atialite-5-7ecbdc35b2.clean/configure: line 14333: 0: command not found
```

rhuijben:
And this last version fixes libspatialite on linux for me, by tweaking the configure script and adding miniunzip as dependency on a minimalistic linux such as wsl 2.

rhuijben:
Slowly getting there... spatialite-tools now also fixed on Linux and most likely OS/X.

rhuijben:
@NancyLi1013 That last pushed version fixes spatialite-tools on linux for me locally, and most errors on AZP are burries in a logfile that I don't know how to access. Can you get access to the error for me?

dg0yt:
You get the build failure logs from the build's "artifacts" on AZP. On the failure log artifact line, there is vertical three buttons icon on the right side which lets you download the log.
In the run's overview, the artifacts hide under "Related", "xxx published".

dg0yt:
I don't see my previous comment as being resolved:

> Does ENABLE_CURL really create a different ABI, or just add additional interfaces?
> (If port A is build with proj[core], won't it run against proj[net]?)
> 
> If yes, bundling ENABLE_CURL with the tools feature doesn't help: If ENABLE_CURL changes ABI, the ABI would now change with feature tools.

This is relevant for this PR, because it extents the coupling of tools and curl from dynamic to static linkage.

You mentioned that this was discussed before. Do you have a reference where I can start reading?

NancyLi1013:
@rhuijben Sorry for the late response. Could you please resolve the conflicts?

NancyLi1013:
@rhuijben 

rhuijben:
> @rhuijben
> 
> Is work still being done for this PR?

I'm still working on this.. but given the responses here the pull request appears to be too big. I have to fix most dependencies again as those were bumped while this was open, and still rely on the old proj apis and dependencies.

For proper geometric transforms we need to enable tiff support, but that changes dependency building, etc.

I'll try to get things fixed... probably as two separate pull requests... soon, but this is no longer relevant for my dayjob so I can't give it the same priority as when I opened this issue (when it was for my previous employer).

dg0yt:
Thanks @rhuijben. As long as switching to PROJ 8 is blocked by vtk (needing patches), we can at least start improving feature and dependency structure for the current proj version, independent of your PR.

rhuijben:
Restarted work on a new clean version of the patch, for now without the enabling of libtiff... but with the recent changes on master that fixed library information for projects that properly use this. This should eventually fix the libtiff dependency issue.

dg0yt:
> Restarted work on a new clean version of the patch, for now without the enabling of libtiff... but with the recent changes on master that fixed library information for projects that properly use this. This should eventually fix the libtiff dependency issue.
> 
> proj is currently at 8.1.1 with 8.2 due soon.

Can you update the blocking dependending ports separately first? This will conflict with the ongoing improvements in proj. My PRs are stuck with slow reviews: #20443 needs #20480.

rhuijben:
> Can you update the blocking dependending ports separately first? The change to proj will conflict with the ongoing improvements in proj. My PRs are stuck with slow reviews: #20443 needs #20480.

Same problem here. This request has been open for half a year now. Luckily many other ports have been fixed by now. Moving to a minimalistic patch with those dependencies fixed now and then another pull request may enable the tiff support.

I hope this minimal approach at least gets us to 8.0.0 (and with minimal changes upon that to latest released).

rhuijben:
This should fix the patch on all platforms, but the x64_windows buildbot appears to be stuck downloading some patches.

rhuijben:
No changes, just force pushed in attempt to get through the issue that opencv is stuck downloading raw patchfiles from github.

dg0yt (approved):
Thanks!

NancyLi1013 (approved):

NancyLi1013:
Thanks for your efforts and patience. @rhuijben

Have you tested these features in this PR?

rhuijben:
Yes @NancyLi1013 , I tested the features of proj. I haven't tested it in the depending projects but depend on the proj version compatibility there. The vtk changes were tested in the vtk project itself as the patch(commit) is directly extracted from their git repository.

I still need the tiff support in my own builds for 'good' accuracy, but that part of the original patch is now handled by a separate patch from @dg0yt which I will follow closely. (Feel free to ask if you need help somewhere @dg0yt. My original patch for other projects is still in my vcpkg fork on github)

NancyLi1013:
Thanks for your above information. @rhuijben Unfortunately, there are some new changes merged to master. So could you please resolve the conflicts?

rhuijben:
@NancyLi1013 thanks for the heads up. I just rebased on master. It looks like the Cairo part of the vcpkg patch was merged upstream, so I just had to shrink the patch to just the proj part to get things working again.

nirvn:
@rhuijben , out of curiosity, why not going for PROJ 8.1.1?

rhuijben:
> @rhuijben , out of curiosity, why not going for PROJ 8.1.1?

Other packages may depend on explicit versions of proj. 

And getting the dependencies fixed against >= 8.0.0 to allow future further updates is of higher priority than getting to the best version.

This patch gets us to 8.0.0 and then patches to 8.0.1, 8.1.0 and 8.1.1 will be trivial to get merged later on.

(The last version is 8.2.0, but that version asserts in debug mode in my personal testing setup... So I wouldn't recommend using that version until 8.2.1 is released)

dg0yt:
I think it would be okay to jump to 8.0.1 immediately. But in any case, we need to the vtk patch finally merged. Other port updates are much easier.

nirvn:
@rhuijben , I wouldn't recommend 8.2.0 either. In any case, big thanks for your work here.

rhuijben:
> I think it would be okay to jump to 8.0.1 immediately. But in any case, we need to the vtk patch finally merged. Other port updates are much easier.

Good point. If I have to rebase it again I will bump to 8.0.1. With a bit of luck this last one might get through this week though.

rhuijben:
@NancyLi1013 Testing is all green now. Only change vs yesterday is that the patch file to make mapnik find proj is now smaller.

NancyLi1013 (approved):

ports/mapnik/proj-find-fix.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 47bbaa1..43f384d 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -304,7 +304,7 @@ if(USE_WEBP)
+ 
+ if(USE_PROJ)
+     #https://proj.org/development/cmake.html
+-    mapnik_find_package(PROJ ${PROJ_MIN_VERSION} QUIET)
++    mapnik_find_package(PROJ REQUIRED)
```
mathisloge, line 10: This is strange but ok for mapnik which vcpkg.

I'll investigate that upstream why and what fails. Maybe some version thingy in the config file of proj is requiring an exact match. 

Thanks for making the proj8 update!

PR #16499, opened 2021-03-02
Fix CURL filesystem deadlock with corporate proxy

### LINKED ISSUE #16476: Failed to take the filesystem lock on /home/USER/vcpkg/.vcpkg-root, need to pkill curl to unlock
**Describe the bug**
When running vcpkg install it works (exit code 0). When running it again, it says:

`Failed to take the filesystem lock on /home/USER/vcpkg/.vcpkg-root`

**Environment**
- OS: Ubuntu 18.04
- Corporate Proxy

**To Reproduce**
Have an install only install local overlay ports and a corporate proxy blocking access to the outside internet. However vcpkg shouldn't need internet access for any of these if the resources are local.

1. ./vcpkg install
2. ./vcpkg install

**Expected behavior**
Lock is given up after install finishes.

**Failure logs** 
None

**Additional context**
1. Running `pkill curl` allows executing vcpkg again, but only once. `lsof curl | rg -i vcpkg` shows that curl seems to be hanging here and keeping the file system locked.
2. I did use `-disableMetrics` when bootstrapping vcpkg. So it shouldn't be any telemetrics. But maybe this flag is ignored.
3. vcpkg was built from latest tag:
```bash
$ vcpkg version
Vcpkg package management program version 2020.06.15-unknownhash
```

SoftwareApe:
The CURL call to the metrics  endpoint doesn't have a timeout. When executing this behind a corporate proxy that blocks this access the filesystem is in a deadlock until curl is killed. By adding a 200ms timeout we give ample time for CURL to finish its request, while also not stopping usage inside local network.

**Describe the pull request**

- What does your PR fix? Fixes #16476 

- Which triplets are supported/not supported? Have you updated the CI baseline? Affects non-Windows users. No update to CI baseline.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? AFAIK, yes, but please edit the PR as necessary if there's something that doesn't conform to your guidelines.

### DIFF
--- a/toolsrc/src/vcpkg/metrics.cpp
+++ b/toolsrc/src/vcpkg/metrics.cpp
@@ -498,7 +498,7 @@ namespace vcpkg::Metrics
 #else
         auto escaped_path = Strings::escape_string(fs::u8string(vcpkg_metrics_txt_path), '\'', '\\');
         const std::string cmd_line = Strings::format(
-            R"((curl "https://dc.services.visualstudio.com/v2/track" -H "Content-Type: application/json" -X POST --tlsv1.2 --data '@%s' >/dev/null 2>&1; rm '%s') &)",
+            R"((curl --max-time 0.2 "https://dc.services.visualstudio.com/v2/track" -H "Content-Type: application/json" -X POST --tlsv1.2 --data '@%s' >/dev/null 2>&1; rm '%s') &)",
             escaped_path,
             escaped_path);
         System::cmd_execute_clean(cmd_line);


ras0219:
Thanks for the PR!

Additionally, we should explicitly release the filesystem lock before spawning the subprocess since there's no need to block another vcpkg invocation in that case. This would enable us to increase the timeout to 3-5 seconds, which should work better on slower connections.

Could you make this PR against the new source location: https://github.com/Microsoft/vcpkg-tool ?

SoftwareApe:
@ras0219 thank you for the quick response. Yes higher timeout may be better if the lock can be released separately. I didn't take too much time to go through the codebase, just grepped for curl. 🙂

JackBoosY:
@SoftwareApe Please make a same PR to https://github.com/microsoft/vcpkg-tool/pulls.
Thanks.

SoftwareApe:
@JackBoosY I created https://github.com/microsoft/vcpkg-tool/pull/22 as requested. I didn't look for how to release the filesystem lock earlier. I also increased the timeout 3s as discussed.

PR #16500, opened 2021-03-02
[gdal] Add Static Library Dependencies

n4z4m3:
Add geos.lib, shell32.lib, and ole32.lib so that downstream projects do not need to explicitly include them


nazame:
If somebody could figure out why this is failing the x86_windows check that would be great. It is not failing for me when I build locally. Could it just be timing out? Because it was failing on the x64_windows and x64_windows static before I updated the CONTROL file and it re-did the checks. So I don't think there is a real problem.

I clicked on through the error details and it gets to this:

_Found the following errors:
Error: While reading versions for port gdal from file: C:\a\1\s\versions\g-\gdal.json
       Version `3.1.3#3` was not found in versions file.
       Run:
           vcpkg x-add-version gdal
       to add the new port version._

What does that mean? Is that x-add-version something I am supposed to do?

Thanks in advance.

NancyLi1013:
Hi @nazame 

Thanks for your PR.

For the failure on x86-windows, you can run `.\vcpkg x-add-version gdal
` and then submit the changes for versions to solve.

nazame:
Thank you @NancyLi1013 

Is that x-add-versions a new command that is required for contributing changes? 
I guess I need to remember that one for next time.

NancyLi1013:
Yes, it is a new one. You need to run it every time when contributing.

Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#update-the-version-files-in-versions-of-any-modified-ports

https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.reference.md#updating-the-version-files

NancyLi1013 (approved):

nazame:
I just realized that this also resolves issue #15526

nazame:
Oops. I accidentally clicked the close button. How do I unclose this?

ports/gdal/portfile.cmake
```diff
@@ -88,7 +88,7 @@ if (VCPKG_TARGET_IS_WINDOWS)
       ${NMAKE_OPTIONS}
       GDAL_HOME=${CURRENT_PACKAGES_DIR}
       CXX_CRT_FLAGS=${LINKAGE_FLAGS}
-      PROJ_LIBRARY=${PROJ_LIBRARY_REL}
+      "PROJ_LIBRARY=${PROJ_LIBRARY_REL} shell32.lib ole32.lib"
```
JackBoosY, line 91: Can you double confirm this?
What happens when you don’t add it?

JackBoosY (changes requested):

ports/gdal/portfile.cmake
nazame: If I don't add this, then my downstream projects require me to add it. This was the main reason for this pull request. That and the geos.lib

JackBoosY: Can you please provide the failure logs?

nazame: I am sorry. I may be causing confusion by using the incorrect terminology. When I say I see the error in "downstream" projects what I mean are in projects that use the gdal library. There are no failures in compiling the gdal vcpkg package or any other package.

The problem is due to some change in the proj library (see this https://github.com/OSGeo/gdal/issues/2743)

When I compile my project with Visual Studio 2019 in Windows and I do not include ole32.lib as an Additional Dependency in the linker settings, I get the following error.

LNK2001	unresolved external symbol __imp_CoTaskMemFree	

This started happening with the previous updates to the vcpkg library when it went from 2.4.x to 3.1.x which also updated its required version of the proj library.

Note that I am not including  shell32.lib here because I have a lot of projects that use gdal and I am sure that I was getting a different linkage error with that missing as well, but I can't find the couple projects that had that error, But maybe it was actually due to some other feature I was using in that project that required it, so maybe we can remove shell32.lib from my change here. I can always add it back later if I find that it is indeed needed for some functionality within gdal/proj.

Also note that technically you can use the option in Visual Studio to "inherit from parent or project defaults" for the Additional Dependencies which will include ole32.lib and shell32.lib but I figure that not everybody using the gdal library will know to do this even though it is the default setting. Instead they may face this linker error and may have a difficult time figuring out how to resolve it.

I also figured that a few lines down from this one in the portfile.cmake shell32.lib is already included with PG_LIB (which is another reason why I am second guessing using shell32.lib here)

So in conclusion, I would be OK removing shell32.lib, but I really do believe that ole32.lib is helpful for others who may not know how to resolve that linker issue.

JackBoosY: image
Seems these 2 system libraries are not required.

JackBoosY:
And please notice #16088 and #15808 makes almost same changes.

ports/gdal/portfile.cmake
nazame: Actually, I forgot to mention that this is for building in static mode so you are right that the DLL does not have the dependency.

And it is definitely a dependency in the static version although it is really a dependency of a dependency (proj.lib) so maybe the gdal vcpkg package is the wrong place to implement a fix since it is possibly more of a workaround. Maybe the proj library is the proper place.

I know it is a dependency because the only thing that changed for my projec to make the issue appear was when I updated my gdal vcpkg which is a dependency of my project (links with gdal.lib statically) and did a recompile. That is when the linking error appeared and I tracked it down to ole32.lib.

I am fine with leaving this change out because as I said I already know how to resolve it in my project by linking with ole32.lib, but I figured it might be helpful to others.

So I will make the change to remove these .lib dependencies

JackBoosY: If this only happens on static build. I think that's okay.

JackBoosY (approved):

JackBoosY:
Can you please solve the file conflicts?

Thanks.

nazame:
> Can you please solve the file conflicts?
> 
> Thanks.

@JackBoosY  I noticed that there have been changes to master since originally forked which are causing the conflicts. The biggest one is that the port has been modified to use the new vcpkg.json instead of CONTROL which is a good thing.
But I am wondering in this case what is the best way to resolve the conflicts. I was thinking that I could re-base. Is that appropriate? Or should I just manually make this branch match the master by deleting the CONTROL, and adding the vcpkg.json and manually making the other changes that are in master?

JackBoosY:
@nazame Rebase to master, apply your changes would be okay.

JackBoosY (approved):

versions/g-/gdal.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "5d3e68869075ed189fc5b23c2ee85efc4295bd5a",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "1195f30c63703eca670c1675a24dcabf36a5f954",
```

JackBoosY (changes requested):

JackBoosY (approved):

ras0219-msft:
This looks like a nice improvement for now, thanks @nazame! Hopefully in the future, we can handle this with pkg-config :)

nazame:
Are all these changes because I didn't do the force push after re-basing?

vicroms:
Probably because I've been merging PRs to master that affected `gdal` as well.

JackBoosY:
Wow, the PR contains too many changes, can you split them?

nazame:
> Wow, the PR contains too many changes, can you split them?

@JackBoosY I agree it contains too many changes. The problem came because there were conflicts and I thought the way to resolve them would be to re-base the branch with master. However, I didn't to the force push after rebasing which caused all these changes to be pulled in from master.
Is there a way to undo this besides creating a new PR?

JackBoosY:
@nazame  You can revert to an earlier commit I think.

ytbigvinny27 (approved):
🤚🏾

nazame:
I wish to close this pull request since the latest update to the gdal package resolves part of it (the geos dependencies) and the other part (the proj shell32.lib and ole32.lib dependencies) are not really necessary. If you us the Additional Dependencies option in MS Visual Studio they are added already and this is the default for a new project in Visual Studio anyway.

PR #16523, opened 2021-03-03
[scripts audit] msbuild buildsystem

strega-nil:
See issue #16188


Neumann-A:
Missing: A call to `vcpkg_internal_get_cmake_vars`
MSBuild needs to learn how to use the flags set by the CMake toolchain variables like `make` and `meson`. 
One way to achieve this is by setting the appropriate compiler flags in the [environment](https://docs.microsoft.com/de-de/cpp/build/reference/cl-environment-variables?view=msvc-160) (something similar exist for the linker)

ports/vcpkg-msbuild/vcpkg_msbuild_install.cmake
```diff
@@ -0,0 +1,323 @@
...
+    [INCLUDES_DIRECTORY <path-to-include-dir>]
+
+    [RELEASE_CONFIGURATION <configuration>]
+    [DEBUG_CONFIGURATION <configuration>]
+    [OPTIONS <option>...]
+    [OPTIONS_RELEASE <option>...]
+    [OPTIONS_DEBUG <option>...]
+
+    [PLATFORM_VERSION <platform-version>]
+    [PLATFORM_ARCHITECTURE <architecture>]
```
ras0219-msft, line 20:
```suggestion
    [PLATFORM <msbuild platform>]
```
Platform is a technical term in MSBuild, that has special significance for the built-in VC targets.

strega-nil:
@Neumann-A I have no idea how to munge the `VCPKG_DETECTED_` variables into msbuild; would you mind looking into this?

Neumann-A:
@strega-nil: I also don't have any idea how to do it. But a quick google search gave me that: https://stackoverflow.com/questions/15141429/how-to-set-preprocessordefinitions-as-a-task-propery-for-the-msbuild-task
So you probably want to add a `cmake_toolchain.props.in` which gets configured using the `VCPKG_DETECTED_` variables and then passed via `/p:ForceImportBeforeCppTargets=cmake_toolchain.props` to msbuild?. It should probably be fine to only forward those as `AdditionalCompileOptions` and `AdditionalLinkOptions`

strega-nil-ms:
Closed to switch to the new scripts audit strategy (see issue #16188)

PR #16526, opened 2021-03-04
[poco]  Remove dependency port openssl for feature netssl on windows

### LINKED ISSUE #16456: [poco] OpenSSL support incorrectly configured
****

I think that the configuration for OpenSSL when enabling `netssl` is incorrect.  It turns on ENABLE_NETSSL_WIN which causes the NetSSL_Win library to be used.  This has an API which is different from what's documented in the pocoproject online documentation.  I don't know what the correct thing to do here but if `netssl` is supposed to use OpenSSL and now depends on the openssl port then I think it should not use NetSSL_Win.  It's possible that this is a new behavior since Poco 1.10.

**Proposed solution**
I think that `ENABLE_NETSSL_WIN` should be `OFF` and that `FORCE_OPENSSL` should be `ON` when the feature `netssl` is turned on.

If instead the `netssl` feature is supposed to really enable NETSSL_WIN on Windows then perhaps there should be a different feature called `openssl`.

**Additional context**
Not turning off `ENABLE_NETSSL_WIN` and leaving `FORCE_OPENSSL` to the default value of `OFF` leads to compilation errors in code which depends on the OpenSSL interface of the Poco NetSSL library.

JackBoosY:
Since poco's features netssl and openssl conflict in windows, according to our policy, delete these features.

Closes #16456.


Neumann-A:
With that logic you need to remove the MPI/parallel feature from hdf5 since it conflicts with the cpp feature. 
@ras0219-msft: I think you should really think about implementing "one of" selections for features. There are simply too many use cases for them and I don't think annoying people by using/writing overlays is the right thing to do (I know it breaks CI but than CI has to do a better Job). 
I have seen this problem know in:
- poco
- generalized blas port
- generalized lapack port
- generalized ssl port
- user discussion in discord where this would have solved everyhing
- hdf5
- probably other ports using MPI which is only interesting for scientific/hpc work.

ports/poco/vcpkg.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "poco",
+  "version-string": "1.10.1",
+  "port-version": 4,
+  "description": "Modern, powerful open source C++ class libraries for building network and internet-based applications that run on desktop, server, mobile and embedded systems.",
+  "homepage": "https://github.com/pocoproject/poco",
+  "supports": "!uwp",
+  "dependencies": [
+    "expat",
+    "openssl",
```
ras0219-msft, line 10: The netssl feature should be restored with a dependency on `openssl (!windows)`

ras0219-msft (changes requested):

ras0219-msft:
To restate the actual issue: `poco[netssl]` had a spurious dependency on openssl on Windows, even though it was not being built with the openssl-based networking backend. This caused some user confusion, plus unnecessary dependencies. The simplest fix that also maintaining overall status-quo of the port is to simply make the dependency platform-specific.

POCO _does_ have a flag to force using openssl on windows, however this is isn't an option we've ever exposed. As @Neumann-A mentioned, this is an XOR property which per policy should not be implemented using "features" and instead should be controlled by the user using overlays.

versions/p-/poco.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "3db1261f5c75dd98c729edad677d5ae9f7db1e69",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "9eb4e090e1f6a427cf05cfa75359c8e8b491a186",
```

PhoebeHui (approved):

vicroms:
@JackBoosY 

Can you please take care of the conflicts?

vicroms (approved):

versions/p-/poco.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "9eb4e090e1f6a427cf05cfa75359c8e8b491a186",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "e9c709300900158cdfcad75965e85b04413c8d5b",
```

JackBoosY:
@vicroms  Done.

PR #16527, opened 2021-03-04
[argparse] Update to the latest commit

### LINKED ISSUE #16517: [argparse] "present" method to check for optional arguments not available in vcpkg version
****

The argparse library provides a method to get a std::optional when querying a optional argument without default values. This method is not present in the version I get from vcpkg, which makes dealing with optional arguments cumbersome. 

**Proposed solution**
I would like an update to the argparse version such that I can use the feature named above. 

---

**Additional context**
---

JonLiu1993:
Fix #16517

Update argparse to the latest commit that can use std::optional method.

Note: no feature need to test


NancyLi1013 (approved):

ports/argparse/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "argparse",
-  "version-string": "2.1",
+  "version-date": "2021-03-04",
```
ras0219-msft, line 3:
```suggestion
  "version-semver": "2.2-predate.20210304",
```

ras0219-msft (changes requested):
Let's use semver prereleases for this.

ports/argparse/vcpkg.json
JonLiu1993: Thanks for your review. After I changed "version-date": "2021-03-04" to "version-semver": "2.2-predate.20210304", the format of the json file was wrong, when I used the command ./vcpkg When format-manifest ./ports/argparse/vcpkg.json, the following error was reported: $ (vcpkg.json):'version-semver' text was not a semantic version, so I changed it back

PR #16536, opened 2021-03-04
[ginkgo] Create new port

upsj:
This PR adds a port for the [Ginkgo](https://github.com/ginkgo-project/ginkgo) numerical linear algebra library with support for its Reference, OpenMP and CUDA backends. It does have a HIP backend for ROCm GPUs as well, but since the HIP ecosystem can still be a bit fragile, I left it out of the port.

We support and test compilation as a static and dynamic library for Linux (GCC/Clang/Intel), Windows (MSVC, MinGW, Cygwin) and Mac (AppleClang).

As far as I can tell, there is only one other library called Ginkgo, which is a Go testing framework, so name collisions should not be an issue here.

### DIFF
--- a/ports/ginkgo/cmake-fixes.patch
+++ b/ports/ginkgo/cmake-fixes.patch
@@ -1,73 +1,44 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index e2413dd29e..c835b0dc6e 100644
+index e75f7d6..63685f2 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
-@@ -20,6 +20,13 @@ include(cmake/hip_path.cmake)
+@@ -20,6 +20,12 @@ include(cmake/hip_path.cmake)
  include(cmake/autodetect_executors.cmake)
  include(cmake/build_type_helpers.cmake)
-
+ 
 +if (MSVC)
-+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
-+endif()
-+if (MINGW OR CYGWIN)
-+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
++    add_compile_options(/bigobj)
++elseif (MINGW OR CYGWIN)
++    add_compile_options("-Wa,-mbig-obj")
 +endif()
 +
  # Ginkgo configuration options
  option(GINKGO_DEVEL_TOOLS "Add development tools to the build system" OFF)
  option(GINKGO_BUILD_TESTS "Generate build files for unit tests" ON)
 diff --git a/cmake/GinkgoConfig.cmake.in b/cmake/GinkgoConfig.cmake.in
-index 961949822e..51a2e8a724 100644
+index 0348f95..3b8be0f 100644
 --- a/cmake/GinkgoConfig.cmake.in
 +++ b/cmake/GinkgoConfig.cmake.in
-@@ -140,8 +140,12 @@ set(GINKGO_OPENMP_LIBRARIES @OpenMP_CXX_LIBRARIES@)
+@@ -129,8 +129,6 @@ set(GINKGO_OPENMP_LIBRARIES @OpenMP_CXX_LIBRARIES@)
  set(GINKGO_OPENMP_FLAGS "@OpenMP_CXX_FLAGS@")
  
  # Provide useful HIP helper functions
 -include(${CMAKE_CURRENT_LIST_DIR}/hip_helpers.cmake)
 -include(${CMAKE_CURRENT_LIST_DIR}/windows_helpers.cmake)
-+if(GINKGO_BUILD_HIP)
-+    include(${CMAKE_CURRENT_LIST_DIR}/hip_helpers.cmake)
-+endif()
-+if (WIN32 OR CYGWIN)
-+    include(${CMAKE_CURRENT_LIST_DIR}/windows_helpers.cmake)
-+endif()
  
  # NOTE: we do not export benchmarks, examples, tests or devel tools
  #     so `third_party` libraries are currently unneeded.
 diff --git a/cmake/install_helpers.cmake b/cmake/install_helpers.cmake
-index 68a75a5f9a..74ae6f8b6b 100644
+index ba7ea3f..7cc3ba8 100644
 --- a/cmake/install_helpers.cmake
 +++ b/cmake/install_helpers.cmake
-@@ -2,28 +2,29 @@ include(CMakePackageConfigHelpers)
- include(GNUInstallDirs)
- 
- 
--set(GINKGO_INSTALL_INCLUDE_DIR "include")
--set(GINKGO_INSTALL_LIBRARY_DIR "lib")
--set(GINKGO_INSTALL_PKGCONFIG_DIR "lib/pkgconfig")
--set(GINKGO_INSTALL_CONFIG_DIR "lib/cmake/Ginkgo")
--set(GINKGO_INSTALL_MODULE_DIR "lib/cmake/Ginkgo/Modules")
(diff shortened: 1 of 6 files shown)


PhoebeHui:
@upsj, thanks for your contribution!

The x64-windows failed with following error in CI test.
```
D:\buildtrees\ginkgo\src\v1.3.0-eec5b92908.clean\core\matrix\dense.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

After you commit all changes, could you please execute the following command to update the baseline version and submmit the changes?

```
vcpkg x-add-version ginkgo
```

upsj:
@PhoebeHui Thanks, that's what I guessed as well. There are two possible solutions to this issue: 1. manually add the /bigobj flag to the builds (which is what we do in our CI setup, but I'm not sure that would be considered good practice) or 2. explicitly set symbol visibility, which would be a huge patch, since we don't have a release with this fix out yet. Do you have any policies for this situation?

PhoebeHui:
@upsj, you can pass the option via one of the following methonds in CMakeLists.txt file.

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")

Or

add_compile_options("/bigobj")

upsj:
@PhoebeHui Thanks for your help, I decided to add the flag in our CMakeLists.txt conditional on the compiler. The PR should now be ready to be reviewed.

JonLiu1993:
Hi @upsj ,An error occurred while testing feature: cuda , please take a look at this：
[config-x86-windows-out.log](https://github.com/microsoft/vcpkg/files/6099726/config-x86-windows-out.log)
There is also an error when testing feature: openmp:
[install-x86-windows-dbg-out.log](https://github.com/microsoft/vcpkg/files/6099766/install-x86-windows-dbg-out.log)

upsj:
@JonLiu1993 Thanks for the feedback. I have never tested CUDA with x86 before, did I understand correctly that x64 worked correctly? On the second error, MSVC's OpenMP support is quite outdated, so I guess the right way to approach this would be to disable this feature for windows?

JonLiu1993:
@upsj ，I tested the X64 triplet of two features at the same time, but there still seem to be errors，please take a look：
./vcpkg install ginkgo[cuda]:x64-windows
[config-x64-windows-out.log](https://github.com/microsoft/vcpkg/files/6099850/config-x64-windows-out.log)
./vcpkg install ginkgo[openmp]:x64-windows
[install-x64-windows-dbg-out.log](https://github.com/microsoft/vcpkg/files/6099864/install-x64-windows-dbg-out.log)

upsj:
@JonLiu1993 Can you give me some information on the software environment used? It looks like either CUDA isn't installed, or the CUDA MSVC integration doesn't work properly.

JonLiu1993:
@upsj ,My local environment is Windows 10，The cuda version is as follows：
PS F:\test-pr\2\vcpkg> nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2020 NVIDIA Corporation
Built on Mon_Nov_30_19:15:10_Pacific_Standard_Time_2020
Cuda compilation tools, release 11.2, V11.2.67
Build cuda_11.2.r11.2/compiler.29373293_0

upsj:
@JonLiu1993 That is the same environment I built in, and the configuration seems to work out of the box there. I would suspect that this is a build environment issue with the CUDA Visual Studio integration, see a [related SO discussion](https://stackoverflow.com/a/63607311/8217180). Though I am not sure how to proceed with this, since you probably don't want to modify your environment too much just for this specifically. Other packages maybe add specific workarounds for Windows CUDA support, but we just rely on CMake's CUDA support, which requires a correctly installed Visual Studio integration for the same build tools that are used by CMake.

PhoebeHui:
I also encountered same issue when test the features with x64-windows via vs2019 and cuda 10.2, I tried this in my 3 test machines, all failed with same failures, and I check the 'C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.2\bin' has been added to the PATH, and the 2 below enviroment variables are correctly set.

image


Failures:

```
CMake Error at third_party/CMakeLists.txt:2 (enable_language):
  No CMAKE_CUDA_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CUDACXX" or the CMake cache entry CMAKE_CUDA_COMPILER to the full
  path to the compiler, or to the compiler name if it is in the PATH.

```

@upsj, could you help double confirm if the feature 'cuda' test pass locally? if so, it will not block this PR.

For feature 'OpenMP', please disable it.

upsj:
@PhoebeHui There is something really weird going on, since a plain `cmake ..` build finds a CUDA compiler without issue, but the `vcpkg` invocation fails. I am investigating this further.

On OpenMP, is it possible to selectively disable a feature just for windows? It looks to me like only dependencies can be made conditional on the platform, but not features?
`cl` is the only major compiler to not support at least OpenMP 3.0, which has been released in 2008, so I would really like to keep the support for other compilers.

upsj:
@PhoebeHui Following up on my investigation, this is an [issue with CMake's `Ninja` generator](https://gitlab.kitware.com/cmake/cmake/-/issues/21908) and the fact that vcpkg overwrites the `Path` environment variable, so it can't be fixed in this setting directly. How should we proceed? Merge the project anyways and note that it doesn't support any features on Windows, or wait until this gets fixed in CMake upstream and the default version of CMake gets updated?

PhoebeHui:
@upsj, thanks for reporting the issue to cmake!

For feature 'OpenMP', since it doesn't work on windows, however, we don't have filed to indicate the supports for a feature, could you add a note in 'description' instead.

It's fine to note that it doesn't support any features on Windows, It should not block this PR, I will approve this PR after the baseline version issue fixed.

BTW, you can also use './vcpkg x-add-version --overwrite-version ginkgo' to fix the baseline version issue.

```
Error: While reading versions for port ginkgo from file: C:\a\1\s\versions\g-\ginkgo.json
       File declares version `1.3.0` with SHA: 9748b66062d3e27d5310f578953c6aa80adb034f
       But local port with the same verion has a different SHA: 2c686e4d669d1580c9cc84e808376d2189fee613

           vcpkg x-add-version ginkgo

       to add a new version
```

PhoebeHui (approved):

ras0219-msft:
Thanks for the PR! This looks really good, though I'd like to reduce the patches a bit more to get us as close to upstream as possible. No action required on your part.

I've pushed a reduction to the CMake changes which works on Linux and I've temporarily rolled back the windows-iterator patch so I can see the CI logs (since you've been rebase-pushing, I can't see precisely what the patch is fixing). If it is a warning, we'd prefer to fix it via turning off the warning instead of changing the code.

>As far as I can tell, there is only one other library called Ginkgo, which is a Go testing framework, so name collisions should not be an issue here.

There does appear to be https://github.com/OpenMandrivaAssociation/ginkgo/blob/4.0/ginkgo.spec, however it looks like this is an unmaintained GUI program and isn't suitable for packaging in vcpkg. Additionally, it looks like this is packaged in Gentoo as `ginkgo` as well, so using the name ginkgo here LGTM :)

upsj:
@ras0219-msft Thanks for the update! The windows patch is unfortunately necessary, since it broke due to a MSVC standard library update and an incomplete implementation of the RandomAccessIterator concept. The CMake changes could probably be minimized though, that is true. Do you squash PRs? If so, I will keep my history in the future.

ras0219-msft:
Yes, we squash PRs. Thanks for the info about Windows.

My reduced branch https://github.com/ras0219-msft/vcpkg/dev/roschuma/ginkgo appears to work -- could you `git pull https://github.com/ras0219-msft/vcpkg dev/roschuma/ginkgo` into your master to update this PR with those changes?

upsj:
Seems to work for me as well, thanks!

PR #16552, opened 2021-03-05
[vc] Update HEAD_REF and fix CMake configure options

bernhardmgruber:
Change the `HEAD_REF` of `vc` from `master` to `1.4`, because the development on `master` moved to a new repository focusing on `std::simd` (https://github.com/VcDevel/std-simd). The branch `1.4` contains the newest developments which are still the `vc` library. I have contacted the maintainer of vc and he now deleted the `master` branch.

Furthermore, the CMake configure options were not correctly passed. Please tell me if this is done correctly now!

With these changes I can now successfully run `vcpkg install vc --head` and `vcpkg install vc` with the triplet `x64-windows` defined via an environment variable.

### DIFF
--- a/ports/vc/CONTROL
+++ b/ports/vc/CONTROL
@@ -1,5 +1,6 @@
 Source: vc
 Version: 1.4.1
+Port-Version: 1
 Homepage: https://github.com/VcDevel/Vc
 Description: SIMD Vector Classes for C++ .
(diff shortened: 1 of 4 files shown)


ras0219-msft:
Awesome, this LGTM. Thanks for the fix!

PR #16035, opened 2021-02-04
[metrohash] support iOS triplets

luncliff:
### What does your PR fix?

This is an enhancement of #15715

### Which triplets are supported/not supported? Have you updated the CI baseline?

Support 4 ios triplets
* `arm64-ios`
* `arm-ios`
* `x64-ios`
* `x86-ios`

```console
$ ./vcpkg install metrohash:arm64-ios metrohash:arm-ios metrohash:x64-ios metrohash:x86-ios
...

The package metrohash:arm-ios provides CMake targets:

    find_package(metrohash CONFIG REQUIRED)
    target_link_libraries(main PRIVATE metrohash::metrohash)

The package metrohash:arm64-ios provides CMake targets:

    find_package(metrohash CONFIG REQUIRED)
    target_link_libraries(main PRIVATE metrohash::metrohash)

The package metrohash:x64-ios provides CMake targets:

    find_package(metrohash CONFIG REQUIRED)
    target_link_libraries(main PRIVATE metrohash::metrohash)

The package metrohash:x86-ios provides CMake targets:

    find_package(metrohash CONFIG REQUIRED)
    target_link_libraries(main PRIVATE metrohash::metrohash)

```

### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Yes

* Versioning
   * [x] `"port-version"` update
   * [x] Update the version files in `versions/`
* Code format
   * [x] Manifests

PR #15715, opened 2021-01-17
[metrohash] support Android triplet build

luncliff:
### What does your PR fix?

My team was using this library. Recently we found that build on Android reported errors for arch-specific intrinsic usage.  
This PR updates `portfile.cmake` and `CMakeLists.txt` to check variable `ANDROID` and handle the case.

`CONTROL` file is changed to `vcpkg.json` with new `"port-version"`.

### Which triplets are supported/not supported?

The following triplets should become available when the user followed docs/users/android.md correctly.

* `arm-android`
* `arm64-android`
* `x86-android`
* `x64-android`

### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Wish conversion from `CONTROL` to `vcpkg.json` was correct :)

PR #16553, opened 2021-03-05
[metrohash] support more triplets by excluding 128 CRC source

luncliff:
### What does your PR fix?

Make `metrohash` support more triplets, which are not available because of [`_mm_crc32_u64`](https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_crc32_u64) intrinsic

* Check `VCPKG_TARGET_ARCHITECTURE` to exclude source file that uses `_mm_crc32_u64`
* Lower the standard to C++ 11 so the compilers with lower versions can build this port.
* Minimize the CMake version constraint

#### Previous Work

* #16035 (triplet support - iOS)
* #15715 (triplet support - Android)

`metrohash` related failures are removed from ci.baseline.txt. ARM & UWP triplets become available because there is no more constraint of `_mm_crc32_u64`

* `arm64-windows`
* `arm64-uwp`
* `x64-uwp`

I thought we can create a patch for the header file, `metrohash.h`. It can add some macro checks like the following.  

```c++
// ...
#ifndef METROHASH_METROHASH_H
#define METROHASH_METROHASH_H

#include "metrohash64.h" // <-- maybe we can replace "" to <> in the patch...
#include "metrohash128.h"
#if defined(_M_X64) || defined(__x86_64__)
// use SSE4.2 intrinsic _mm_crc32_u64
#include "metrohash128crc.h"
#endif
#endif // #ifndef METROHASH_METROHASH_H
```

I decided to remain simple for now but I'd like to hear about replacing `vcpkg_replace_string` with a patch like it.


luncliff:
Feels like my team is the only one who is using this port 🤔

ports/metrohash/CMakeLists.txt
```diff
@@ -1,14 +1,15 @@
...
-            src/metrohash64.cpp
-            src/metrohash128.cpp
-            )
+    src/metrohash64.cpp
+    src/metrohash128.cpp)
 list(APPEND metro_headers src/metrohash.h src/metrohash64.h src/metrohash128.h)
 
-if(NOT ANDROID AND NOT IOS)
+if(VCPKG_TARGET_ARCHITECTURE MATCHES arm OR VCPKG_TARGET_ARCHITECTURE STREQUAL x86)
+    message(WARNING "MetroHash's 128 CRC requires SSE 4.2 intrinsic. The functions will be excluded.")
```
Hoikas, line 11: SSE 4.2 should work on x86.

luncliff: https://docs.microsoft.com/en-us/cpp/intrinsics/x86-intrinsics-list?view=msvc-160 I searched for the intrinsic but not it wasn't there. :(

Hoikas: I just checked the header, and it looks like `_mm_crc32_u64` is x64 only on MSVC. That being said, since this is a CMakeList, I think you should use something like [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) to try compiling a basic program, eg:

```C
#include <nmmintrin.h>
int main() {
    _mm_crc32_u64(0, 0);
    return 1;
}
```

and use the results from that to include or exclude the SSE4.2 features.

luncliff: I rarely used the function and I think it's really good case to try it. Thanks for the point!

PhoebeHui (changes requested):
@luncliff, thanks for the PR!

@ras0219-msft, could you help further review?

ras0219-msft:
Thanks for the PR!

I've pushed a change that includes @Hoikas's suggestion. @luncliff could you confirm that this still covers all your use cases?

ras0219-msft (approved):

luncliff:
@ras0219-msft Thanks for the 1c8c750, I will check the port within days. :)

luncliff:
Works as expected for `x86-android`, `x64-android`, `x86-ios`, and `x64-ios`. Great!

PhoebeHui (approved):

PR #16575, opened 2021-03-06
[wil] Updating WIL version to match its nuget package version

anaaru:
Updating wil vcpkg to match wil.nupkg version which is 1.20210204.0
Removed the fix-install-headers.patch since its no longer needed.


ports/wil/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: wil
-Version: 2020-05-19
+Version: 2021-02-04
 Port-Version: 1
```
Hoikas, line 3:
```suggestion
```

anaaru: Any particular reason to remove line-3?

Hoikas: When a port is updated to a new version, the port-version should be reset. See the [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#port-version) for more details. The general practice is to delete the entire line.

anaaru: Addressed in next iteration.

JonLiu1993: Hi @anaaru ,Thanks for posting this issue, @Hoikas gave a good suggestion, please modify the file before I can merge

JonLiu1993 (commented):
@Hoikas gave a good suggestion, please modify the file before I can merge

JonLiu1993 (approved):

ras0219-msft:
Thanks!

PR #16595, opened 2021-03-08
[nettle] fix build on apple arm macs

### LINKED ISSUE #16464: [nettle] fails to build an x64-osx triplet on Apple Silicon
**Host Environment**
- OS: macOS 11.2.2 (20D80) 
- Compiler: Apple clang version 12.0.0 (clang-1200.0.32.29)

**To Reproduce**
`./vcpkg install nettle`

**Failure logs** 
```
    Command failed: /bin/bash ./../src/48e146e54f-e59f545639.clean/configure --build=arm64-apple-darwin --host=x86_64-apple-darwin --disable-documentation --disable-openssl --disable-shared --disable-silent-rules --verbose --disable-shared --enable-static --prefix=/Users/cromo/Documents/Sourcecode/vcpkg/installed/x64-osx/debug --bindir=${prefix}/../tools/nettle/debug/bin --sbindir=${prefix}/../tools/nettle/debug/sbin --includedir=${prefix}/../include --datarootdir=${prefix}/share/nettle
```
`Sourcecode/vcpkg/buildtrees/nettle/x64-osx-dbg/config-x64-osx-dbg-err.log`:
```
Invalid configuration `arm64-apple-darwin': machine `arm64-apple' not recognized
configure: error: /bin/sh ./../src/48e146e54f-e59f545639.clean/./config.sub arm64-apple-darwin failed
```

**Additional context**
Notice that this is a x64 (Rosetta) build, but the `arm64-apple-darwin` triplet is being incorrectly passed to `configure`.

autoantwort:
**Describe the pull request**

- What does your PR fix? Fixes #16464

### DIFF
--- a/ports/nettle/portfile.cmake
+++ b/ports/nettle/portfile.cmake
@@ -113,13 +113,12 @@ else()
 
     if(VCPKG_TARGET_IS_OSX AND VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
         # fix for apple arm macs (see #16464)
-        set(BUILD_TRIPLET "--host=x86_64-apple-darwin")
+        set(VCPKG_MAKE_BUILD_TRIPLET "--host=x86_64-apple-darwin")
     endif()
 
     vcpkg_configure_make(
         SOURCE_PATH ${SOURCE_PATH}
         AUTOCONFIG
-        BUILD_TRIPLET ${BUILD_TRIPLET}
         OPTIONS
             --disable-documentation
             --disable-openssl
--- a/ports/nettle/portfile.cmake
+++ b/ports/nettle/portfile.cmake
@@ -111,9 +111,15 @@ else()
         set(OPTIONS --disable-shared)
     endif()
 
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
+        # fix for apple arm macs (see #16464)
+        set(BUILD_TRIPLET "--host=x86_64-apple-darwin")
+    endif()
+
     vcpkg_configure_make(
         SOURCE_PATH ${SOURCE_PATH}
         AUTOCONFIG
+        BUILD_TRIPLET ${BUILD_TRIPLET}
         OPTIONS
             --disable-documentation
             --disable-openssl
--- a/ports/nettle/vcpkg.json
+++ b/ports/nettle/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "nettle",
   "version-string": "3.6",
+  "port-version": 1,
   "description": "Nettle is a low-level cryptographic library that is designed to fit easily in more or less any context: In crypto toolkits for object-oriented languages (C++, Python, Pike, ...), in applications like LSH or GNUPG, or even in kernel space.",
   "homepage": "https://git.lysator.liu.se/nettle/nettle",
   "dependencies": [
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -4086,7 +4086,7 @@
     },
     "nettle": {
       "baseline": "3.6",
-      "port-version": 0
+      "port-version": 1
     },
     "networkdirect-sdk": {
       "baseline": "2.0.1",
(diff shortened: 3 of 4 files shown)


ports/nettle/portfile.cmake
```diff
@@ -111,9 +111,15 @@ else()
         set(OPTIONS --disable-shared)
     endif()
 
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
+        # fix for apple arm macs (see #16464)
+        set(BUILD_TRIPLET "--host=x86_64-apple-darwin")
```
Neumann-A, line 116: You know you can set `VCPKG_MAKE_BUILD_TRIPLET` in the triplet instead?

autoantwort: No, seems to be an better alternative.

ports/nettle/portfile.cmake
```diff
@@ -111,6 +111,11 @@ else()
         set(OPTIONS --disable-shared)
     endif()
 
+    if(VCPKG_TARGET_IS_OSX AND VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
+        # fix for apple arm macs (see #16464)
+        set(VCPKG_MAKE_BUILD_TRIPLET "--host=x86_64-apple-darwin")
```
Neumann-A, line 116: Isn't this an error in `_vcpkg_determine_autotools_host_arch_mac` ?

```
macro(_vcpkg_determine_autotools_host_arch_mac out_var)
    set(${out_var} "${VCPKG_DETECTED_CMAKE_HOST_SYSTEM_PROCESSOR}")
endmacro()
```

seems like it should be
```
macro(_vcpkg_determine_autotools_host_arch_mac out_var)
    set(${out_var} "${VCPKG_DETECTED_CMAKE_SYSTEM_PROCESSOR}")
endmacro()
```

Neumann-A: ah wait. it is in `_vcpkg_determine_autotools_target_arch_mac`

Neumann-A: https://github.com/microsoft/vcpkg/blob/c47216ac7904b09187c119c7e4d010dcf993e3d5/scripts/cmake/vcpkg_configure_make.cmake#L146-L160

needs to be fixed for your use case. don't add this extra arguments in every port

autoantwort: `_vcpkg_determine_autotools_host_arch_mac` is right. It created the following call: 
```
Command failed: /bin/bash ./../src/48e146e54f-e59f545639.clean/configure --build=arm64-apple-darwin --host=x86_64-apple-darwin ...
```
The Problem is, that the nettle configure tool can not handle `arm64-apple-darwin`

autoantwort: If you only pass `--host=x86_64-apple-darwin`, everything works. But this is a specific problem of the nettle configure tool. E.g. the icu configure tool can handle `--build=arm64-apple-darwin`

Neumann-A: why not remove 
https://github.com/microsoft/vcpkg/blob/c47216ac7904b09187c119c7e4d010dcf993e3d5/scripts/cmake/vcpkg_configure_make.cmake#L440
generally? Configure should be able to figure it out any way. Windows just needs it because configure is not common on windows and news some help in figuring out the correct one which everywhere works.

autoantwort: > Configure should be able to figure it out any way.

Yeah ok. Then we can remove it.

autoantwort: > Configure **should** be able to figure it out any way.

As you can see in #16596, gmp does not build without `--build=...` and I only have tested one package. Because of that I think it is better to stick with the solution of this PR.

Neumann-A: gmp on apple arm seems to be problematic anyway according to the release notes at https://gmplib.org/

autoantwort: But it currently builds (and it is a dependency of nettle)

ras0219-msft:
Waiting to merge https://github.com/microsoft/vcpkg/pull/16596 first

autoantwort:
Already fixed by #16596

Issue #12504, opened 2020-07-20
[vcpkg|manifest] Missing a way to pass --head to certain dependencies. 


Neumann-A:
The vcpkg.json should probably have an extra field `vcpkg_additional_args` or similar to pass additional vcpkg arguments to the build of some dependencies

strega-nil:
Is there any other reason you want this besides passing `--head`?

Neumann-A:
@strega-nil: currently cannot think of anything more. Just always want to install `eigen3 --head` and probably a few other dependencies. Especially if the ports in questions are your own ports via a overlay you might want to use `--head` for development without always changing the ref&hash

strega-nil:
Would it be okay to add something like a `"x-extra-arguments": [ "head" ]` to the dependency field?

strega-nil:
Or maybe in `vcpkg-configuration.json`, `x-dependency-arguments: { "eigen3": [ "head" ] }`

sylveon:
I would also want this, to always install the head of `wil` and `rapidjson`, however I feel that `vcpkg-configuration` is not a place where I'd want to put it in manifest mode, since it directly concerns my project, not vcpkg's configuration

strega-nil:
@sylveon `vcpkg-configuration.json` is the project configuration that doesn't belong in `vcpkg.json`: in other words, the stuff that people depending on you shouldn't care about.

sylveon:
My project won't build with the very old stable version of RapidJSON, so knowing that it's head should matter, no?

strega-nil:
it's one of those things that versioning should fix; using `head` is not supported and often doesn't work, and so it's not a part of the dependency interface of your library.

sylveon:
I've skimmed through the versioning docs and can't seem to find anything related to installing the head of a library. Is that feature not present at the moment?

strega-nil:
poke @vicroms ^

PR #16613, opened 2021-03-09
[vcpkg] Allow compilation of HEAD_REF-only ports in manifest mode

PazerOP:
**Requires https://github.com/microsoft/vcpkg-tool/pull/24 to work, otherwise it does nothing.**

Sort of addresses https://github.com/microsoft/vcpkg/issues/12504. More specifically, supports my desired use case:
* I have a registry with very, very frequently touched package. It does not make sense for me to version this package in any way, I am the only one using it, and I only ever want to use the latest version. Therefore, I have it [available on my vcpkg registry as a HEAD_REF-only port](https://github.com/PazerOP/vcpkg-registry/blob/397dc9e361f0727da46e45333940558c192f7b6c/ports/mh-stuff/portfile.cmake#L4).
* I have a project that consumes previously mentioned package, as well as many others from the official vcpkg registry. I do not want `--head` behavior for all those packages. 
* I want to use manifest mode, which means I cannot currently specify per-port build options (here's where #12504 comes into play)

With the given conditions, it is impossible to build my project. vcpkg will run the manifest install step, then hit my `--head`-requiring port, [and abort](https://github.com/microsoft/vcpkg/blob/0dc27b9b672cb9d069cf58c48f3ae3f7395d108b/scripts/cmake/vcpkg_from_github.cmake#L135).

I cannot use `--head`, because not all my vcpkg dependencies are built from source (discord-game-sdk). I also have no desire to be living on the bleeding edge for any third-party dependencies anyway. 

Theoretically, this change should not break any existing behavior (unless people were relying on vcpkg not working in this situation). It takes a previously non-working scenario (HEAD_REF-only port, manifests, and no `--head`) and makes it work in a (in my opinion) reasonable way.

- **What does your PR fix?** Somewhat related to #12504

- **Which triplets are supported/not supported? Have you updated the CI baseline?**
I do not believe this has any affect for CI pipelines, but I have not studied them in detail. Please correct me if this is wrong.

- **Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?** 
To the best of my understanding, yes.

### DIFF
--- a/scripts/cmake/vcpkg_from_github.cmake
+++ b/scripts/cmake/vcpkg_from_github.cmake
@@ -129,6 +129,11 @@ function(vcpkg_from_github)
         set(VCPKG_USE_HEAD_VERSION OFF)
     endif()
 
+    if(VCPKG_MANIFEST_MODE AND NOT VCPKG_USE_HEAD_VERSION AND NOT DEFINED _vdud_REF)
+        message(STATUS "Package specifies HEAD_REF but not REF, and --head was not passed to vcpkg. Manifest mode is active, so HEAD_REF is being used.")
+        set(VCPKG_USE_HEAD_VERSION ON)
+    endif()
+
     # Handle --no-head scenarios
     if(NOT VCPKG_USE_HEAD_VERSION)
         if(NOT _vdud_REF)


ras0219-msft:
We are _very_ hesistant to introduce any functionality that produces nondeterministic behavior, especially without loud, obnoxious messages for/from users.

Have you considered just defining `VCPKG_USE_HEAD_VERSION` yourself in your port?
```
set(VCPKG_USE_HEAD_VERSION ON)
```
Note: this is _not_ a supported scenario, but it is a hack that may give you what you want.

ras0219-msft:
Closing for now, let me know if the above solution doesn't work for you!

PR #16627, opened 2021-03-09
[host dependencies] add support in the scripts

strega-nil:
This contains all the docs and scripts changes from #16479,
without any of the ports changes, for easier CR


BillyONeal (approved):

docs/users/manifests.md
```diff
@@ -323,6 +323,20 @@ automatically bootstrapped if missing and invoked to install your dependencies i
 All vcpkg-affecting variables must be defined before the first `project()` directive, such as via the command line or
 `set()` statements.
 
+#### `VCPKG_TARGET_TRIPLET`
+
+This variable controls which triplet dependencies will be installed for.
```
vicroms, line 328: This reads a little weird but I don't have any good suggestions.

```markdown 
This variable controls triplet selection. Dependencies will be installed for the specified triplet.
```

vicroms (approved):

docs/users/manifests.md
strega-nil: I think this is fine for now; if we find a better wording later we can add it.

PR #16628, opened 2021-03-09
Port for PresentMon, windows-only ETW swap-chain presentation capture and analysis tool

aggieNick02:


scripts/ci.baseline.txt
```diff
@@ -1262,6 +1262,10 @@ pqp:x64-uwp=fail
 proj4:arm64-windows=fail
 proj4:arm-uwp=fail
 proj4:x64-uwp=fail
+presentmon:arm-uwp=fail
+presentmon:x64-linux=fail
+presentmon:x64-osx=fail
+presentmon:x64-uwp=fail
```
JackBoosY, line 1268: Does the official not support this?

JackBoosY (changes requested):

ports/presentmon/portfile.cmake
```diff
@@ -0,0 +1,27 @@
...
+vcpkg_install_msbuild(
+    SOURCE_PATH ${SOURCE_PATH}
+    PROJECT_SUBPATH PresentMon.sln
+    PLATFORM ${PLATFORM}
+)
+file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+file(GLOB PRESENTMON_EXE ${CURRENT_PACKAGES_DIR}/tools/${PORT}/PresentMon-*.exe)
+file(GLOB PRESENTMONTESTS_EXE ${CURRENT_PACKAGES_DIR}/tools/${PORT}/PresentMonTests-*.exe)
+file(RENAME ${PRESENTMON_EXE} ${CURRENT_PACKAGES_DIR}/tools/${PORT}/PresentMon.exe)
+file(REMOVE ${PRESENTMONTESTS_EXE})
```
JackBoosY, line 27: Are these executable files examples or tools?

JackBoosY (changes requested):

ports/presentmon/portfile.cmake
aggieNick02: PresentMon-*.exe is the main tool, and the project names it PresentMon-{Version}-{architecture}.exe, so I rename it appropriately. PresentMonTests-*.exe is an executable used in running a test suite. Unfortunately right now the *.sln file does not offer a configuration that omits the test project.

scripts/ci.baseline.txt
aggieNick02: So the tool is 100% windows only, so linux and osx definitely do need to be in there as fail. But I double checked and it does support arm/arm64. I can get it to build fine by opening the .sln manually and building. But when I try to build for any arm triplet from vcpkg, I get this:
```
Unsupported toolchain combination. Target was: arm64 but supported ones were:
x86,amd64,x86_amd64,x86_arm,amd64_x86,amd64_arm
```
Any idea if there's something I need to configure/do with vcpkg to get building arm/arm64 to work so I can update the ci failures appropriately?

Honestly, I don't quite know the difference between arm64-windows/arm64-uwp and arm-windows/arm-uwp, so I'm not sure which one corresponds to the sln config that works - in the vs sln, the platforms are listed as arm and arm64. I'm guessing those are the arm*-windows and not arm*-uwp triplets, but I can find out once I get vcpkg working for arm

aggieNick02: After learning a bit more, I'm pretty certain the console apps being built on arm/arm64 are arm-windows and arm64-windows, and not uwp. So I think the ci.baseline.txt is right.

I was able to build arm-windows just fine. But not arm64-windows, which gives the error message I posted above.

ports/presentmon/portfile.cmake
JackBoosY: Please use `vcpkg_copy_tools` and `file(RENAME)` instead of these, see [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_copy_tools.md).

scripts/ci.baseline.txt
JackBoosY, line 1268:
```suggestion
```

ports/presentmon/vcpkg.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "presentmon",
+  "version-string": "1.6.0",
+  "description": "PresentMon is a tool to capture and analyze ETW events related to swap chain presentation on Windows."
```
JackBoosY, line 4:
```suggestion
  "description": "PresentMon is a tool to capture and analyze ETW events related to swap chain presentation on Windows.",
  "supports": "windows & !uwp"
```

JackBoosY (changes requested):

ports/presentmon/portfile.cmake
aggieNick02: I can definitely change to those if that is preferred, but to do that I'll have to switch from using `vckpg_install_msbuild` to `vcpkg_build_msbuild`. which says it is deprecated in the docs. (`vcpkg_install_msbuild` automatically finds and pushes all `*.exe` files to tools.) Is that what I should do?

versions/p-/presentmon.json
```diff
@@ -0,0 +1,9 @@
+{
+  "versions": [
+    {
+      "git-tree": "19487b2a240d83297fcb1c7ed9f61a594b6488bd",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "a485296485c861110ed43f424e758c46976bd6fb",
```

ports/presentmon/portfile.cmake
JackBoosY: I don't think we need to use `vcpkg_copy_tools` by switching to `vcpkg_buid_msbuild`.

JackBoosY (changes requested):

ports/presentmon/portfile.cmake
aggieNick02: Sorry for the delay on this. I was out from work for a bit.

I'm a little confused on the whole `vcpkg_copy_tools`/`vcpkg_install_msbuild`/`vcpkg_build_msbuild`; my apologies for not understanding. If I switch to `vcpkg_build_msbuild` as requested, then I'll be using a method whose documentation says it is deprecated and says to use `vckpg_install_msbuild` instead. Is that documentation wrong and `vcpkg_build_msbuild` is still ok to use in new ports?

JackBoosY: `vcpkg_build_msbuild` is almost deprecated, we should use vcpkg_install_msbuild to install the binary as much as possible.
For tools, if they are not installed, we should use `vcpkg_copy_tools`.

aggieNick02: Ok, I'm happy to change it to whatever is recommended. Right now my PR uses `vcpkg_install_msbuild`, but then has to do a rename and a remove because we end up with a file in tools that needs to be renamed, and another that is unwanted as a tool and needs to be deleted.

Is what is preferred here that I instead use `vcpkg_build_msbuild`, perform the rename of `PresentMon-*.exe` in its output location, and then use `vcpkg_copy_tools` to copy the already renamed `PresentMon.exe` to the tools directory?

JackBoosY:
Please use [this](https://github.com/microsoft/vcpkg/files/6227551/presentmon.zip) changes.

aggieNick02:
Thank you for the continued discussion/teaching and suggested commit.

I looked at the changes in the zip file, but the approach there confuses me. It switches to doing a cmake based build; using cmake may avoid any issues with the vcpkg approach requiring renames, but it also seems like something that will be brittle moving forward.

The presentmon project is maintained entirely as a visual studio vcxproj-based build. If we use this cmake approach in vcpkg, then updating the port for a new presentmon release is much more likely to be non-trivial. The `fix-tool-build.patch` may break, new or removed source files will require updating `CMakeLists.txt`, etc. Is there a reason to prefer the cmake-based approach?

I'm happy to modify my original commits as needed; to do that it would be helpful to understand some of the "why" behind the problems with the original commits. Namely, the original feedback was

> Please use `vcpkg_copy_tools` and `file(RENAME)` instead of these, see documentation.

Was the problem with my original commit the `RENAME` or `REMOVE` operations done after `vcpkg_install_msbuild`? If I know what the problem is and some of the why behind it, I can try to address it.

JackBoosY:
@aggieNick02 
The reason for switching to cmake is that cmake is more maintainable and can automatically set compilation options, and we can directly declare and install header files in CMakeLists.txt.
I found that using msbuild only generated the release library but not the debug library, so I switched to cmake.

aggieNick02:
@JackBoosY

I'm sorry, but I'm still very confused, and feel like we're getting further from resolving issues with the PR.

I tried the files provided that do the cmake-based build. With it, PresentMon.exe is no longer installed in tools, which was the whole reason I was adding this port in the first place. Additionally, since this library is generally used as a tool and not a library, honoring the configuration in the visual studio projects seems more relevant than being able to adapt to different compilation options or install header files. Next, maybe I have something configured wrong, but I didn't need to use the patch file at all to get things to build.

Also, when building via the visual studio based approach, both the debug and release library are present for me, so I'm not sure if I'm just looking at that wrong.

I could be misunderstanding what vcpkg is supposed to be used for; in this case, I was using it as a nice way to deal with a dependency on the binary output of a visual studio solution hosted on github. All I really want to do is build the visual studio solution exactly as it is on the project's github page, and grab the output tool. Is there something wrong with that or with my initial commits/approach, and if so, details on what is wrong would be really helpful. If using vcpkg like this isn't what it is for, I can just close the PR and stick with using my port via --overlay-ports; presentmon is a pretty popular tool so I thought others might appreciate being able to grab it this way as well.

Last, it is confusing that vcpkg has commands to build visual studio projects but using them seems discouraged. I'd hope the documentation would make that clear if it was the case.

Thanks for your help.

ports/presentmon/portfile.cmake
```diff
@@ -0,0 +1,27 @@
+vcpkg_fail_port_install(ON_TARGET "linux" "osx" "uwp" "ios" "android" "freebsd")
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO GameTechDev/PresentMon
+    REF 6ddc9e15d2ef169cdce954b589c1ba190b3a25bd # 1.6.0
+    SHA512 2522b0e3218d4a6588531a09bc82631f14ad05c20f4560fe0574f00f2f5eece114ae04320f920eb52ba64173cea5cdf15bb223b7395c3782e4a6465afb5d9bec
+    HEAD_REF main
+)
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
JackBoosY, line 10: I think any port should have header files where the user can get the symbol definition.

ports/presentmon/portfile.cmake
```diff
@@ -0,0 +1,27 @@
...
+# A bit silly - this is not honored, but we make it dynamic so that vcpkg_install_msbuild does not disable whole program optimization (which it does if static)
+set(VCPKG_LIBRARY_LINKAGE dynamic)
```
JackBoosY, line 12: Should use `vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)` instead.

ports/presentmon/portfile.cmake
JackBoosY: We can use `vcpkg_copy_tools(TOOL_NAMES PresentMon-1.6.0-${PLATFORM} SEARCH_DIR ${CURRENT_PACKAGES_DIR}/tools/PresentMon)` first, then rename this executable.

JackBoosY (changes requested):

JackBoosY:
We accept to use msbuild to generate port binaries, but we must solve some compilation options. It will be more convenient to use cmake, and you can export cmake configuration files to make it easier for users.

ports/presentmon/portfile.cmake
aggieNick02: So the point of my port is to just make the tool available. I don't think there are any clients that use header files, but I'm reaching out to the tool author to verify.

ports/presentmon/portfile.cmake
aggieNick02: Since my goal is to build this as a tool only, I actually want it built with dynamic linkage regardless of the triplet's setting. I realize this doesn't make sense if there are people actually using the header files/libraries generated, but it does make sense if the only thing used from the project is an exe. I'll post back when I hear from the tool author.

aggieNick02:
> We can use `vcpkg_copy_tools(TOOL_NAMES PresentMon-1.6.0-${PLATFORM} SEARCH_DIR ${CURRENT_PACKAGES_DIR}/tools/PresentMon)` first, then rename this executable.

I'm confused by this. Why would we copy PresentMon-1.6.0.${PLATFORM} from ${CURRENT_PACKAGES_DIR}/tools/PresentMon to the same directory (you've set `SEARCH_DIR` and `DESTIANTION` to the same directory)? Wouldn't that be a no-op?

aggieNick02:
@JackBoosY 
I heard back quickly from the presentmon developers, and they *do* have clients that use just the library and headers. So your pushback was warranted, thank you for it.

Let me take another look at the cmake approach you provided. You can ignore my recent comments/questions from about 15 minutes ago, and I'll reach back out when I have something new ready. Thanks again.

ports/presentmon/portfile.cmake
aggieNick02: Please ignore. Looking into your cmake approach.

ports/presentmon/portfile.cmake
aggieNick02: Please ignore. Looking into your cmake approach.

aggieNick02:
I took the CMake based zip you provided and played a bit with it. I needed to make a few minor changes to get things building/installed right. Here were the changes I made. I've also committed the new files for your review.

In `CMakeLists.txt`:

- Added `target_compile_definitions` to define `UNICODE` for both projects . This prevents needing to use the patch to call the ASCII versions of various functions instead of the version that chooses Ascii/Wide based on `UNICODE` .
- Changed `install(TARGETS PresentData RUNTIME DESTINATION bin)` in `if (BUILD_TOOLS)` to `install(TARGETS PresentMon RUNTIME DESTINATION bin)` , as I'm pretty sure that is what is meant, and is needed.

Removed `fix-tool-build.patch`

In `portfile.cmake`
- Removed reference to patch file
- added `if("tools" IN_LIST FEATURES)
    vcpkg_copy_tools(TOOL_NAMES presentmon AUTO_CLEAN)
endif()` so that presentmon.exe is copied to `tools` when the feature is on

Thanks for your help in guiding me through this. Just let me know if there is anything in my changes that needs to be reconsidered.

JackBoosY (approved):

JackBoosY:
LGTM.

PR #16632, opened 2021-03-10
 [vcpkg] (Implementation) Possibility to control the compiler hash for binary caching

### LINKED ISSUE #16631: [vcpkg] Possibility to control the compiler hash for binary caching
****

This issue describes the context and motivation of a pull request I am about to open.

Binary caching uses hashes to make as likely as possible that, if cached binaries are used, they have compatible ABI. One of those hashes comes from the compiler. Actually, the whole compiler binary file is hashed:

```cmake
file(SHA1 "${CMAKE_CXX_COMPILER}" CXX_HASH)
```

This of course contributes to the goal of making binary reused ABI-safe. I think this should remain the default vcpkg behavior.

That being said, within my organisation, this behavior removes much of the benefits of using (Nuget) binary caching, which we will probably need if we are to use vcpkg. We are ~20 C++ developers and each of us has a more or less different setup, which is also a bit different from our CI setup. If the binary caching works, it can take 8 minutes to a developer to build our main repository. If the binary caching doesn't work, it can take more than an hour. It means for instance that updating MSVC, could suddenly have a ~1h cost.

With our current setup, we use binaries compiled with different MSVC versions on Windows and clang versions on Linux. We are aware of the risk, but in more than 10 years, we've yet to identify a bug related to this risk. I think that they're might be other organisations with our use cases, so it would be nice to have an option to control how the compiler hash is computed.

**Proposed solution**
Instead of getting into discussion about what is safe to do and what is unsafe, I suggest that we use optional environment variables `VCPKG_CXX_COMPILER_HASH_CMAKE_EXPRESSION` and `VCPKG_C_COMPILER_HASH_CMAKE_EXPRESSION`. This would let users decide the risk they are comfortable to live with. If the variables are not defined, the default behavior remains.

All the solutions I considered maintain the current behavior as the default since it is the safest. They'd used environment variables to enable the more unsafe alternatives. For instance

- Allowing the compiler hash to be fixed (`VCPKG_SKIP_COMPILER_HASH`), which would be unsafe, yet could be reasonable for some use cases.
- Allowing the compiler hash to be just based on `${CMAKE_CXX_COMPILER_ID}` or on `${CMAKE_CXX_COMPILER_ID}-${CMAKE_CXX_COMPILER_VERSION}`.

This is because there were many alternatives that I thought about using a CMake expression. This way, users have control over the risk.

**Additional context**
I'm about to open a pull request implementing this (12-lines) feature

klalumiere:
**Describe the pull request**

- What does your PR fix? Fixes #16631

Use optional environment variables `VCPKG_CXX_COMPILER_HASH_CMAKE_EXPRESSION` and `VCPKG_C_COMPILER_HASH_CMAKE_EXPRESSION` to control the compiler hash. If the variables are not defined, the default behavior remains.

Example of value for `VCPKG_CXX_COMPILER_HASH_CMAKE_EXPRESSION`:

```shell
export VCPKG_CXX_COMPILER_HASH_CMAKE_EXPRESSION='${CMAKE_CXX_COMPILER_ID}'
```

### DIFF
--- a/ports/soci/export-include-dirs.patch
+++ b/ports/soci/export-include-dirs.patch
@@ -1,0 +1,53 @@
+diff --git a/cmake/SociBackend.cmake b/cmake/SociBackend.cmake
+index 442d81d..ef56422 100644
+--- a/cmake/SociBackend.cmake
++++ b/cmake/SociBackend.cmake
+@@ -159,8 +159,13 @@ macro(soci_backend NAME)
+           ${THIS_BACKEND_HEADERS})
+ 
+         target_link_libraries(${THIS_BACKEND_TARGET}
++          PUBLIC
+           ${SOCI_CORE_TARGET}
+           ${THIS_BACKEND_DEPENDS_LIBRARIES})
++          
++        target_include_directories(${THIS_BACKEND_TARGET}
++          PUBLIC
++          ${THIS_BACKEND_DEPENDS_INCLUDE_DIRS})
+ 
+         if(WIN32)
+           set_target_properties(${THIS_BACKEND_TARGET}
+@@ -197,8 +202,14 @@ macro(soci_backend NAME)
+ 
+         # Still need to link the libraries for tests to work
+         target_link_libraries (${THIS_BACKEND_TARGET_STATIC}
++          PUBLIC
+           ${THIS_BACKEND_DEPENDS_LIBRARIES}
+         )
++        
++        target_include_directories(${THIS_BACKEND_TARGET_STATIC}
++          PUBLIC
++          ${THIS_BACKEND_DEPENDS_INCLUDE_DIRS}
++        )
+ 
+         set_target_properties(${THIS_BACKEND_TARGET_STATIC}
+           PROPERTIES
+diff --git a/src/core/CMakeLists.txt b/src/core/CMakeLists.txt
+index 3916cce..d537613 100644
+--- a/src/core/CMakeLists.txt
++++ b/src/core/CMakeLists.txt
+@@ -64,6 +64,7 @@ if (SOCI_SHARED)
+   target_include_directories(${SOCI_CORE_TARGET}
+       PUBLIC
+       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../../include>
++      $<INSTALL_INTERFACE:include>
+   )
+ 
+ endif()
+@@ -96,6 +97,7 @@ if (SOCI_STATIC)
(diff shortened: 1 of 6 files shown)


PhoebeHui:
@klalumiere, thanks for the PR!

@ras0219 @ras0219-msft, could you help review?

ras0219-msft:
Thanks for the PR!

First, the current implementation will only work on non-Windows platforms -- on Windows, we clean out the environment variables in order to produce more reliable builds between machines, so these variables will need to be explicitly leaked in via either adding to our "clean environment" list or the user also explicitly using `VCPKG_KEEP_ENV_VARS`.

Second, the entire need for running `detect_compiler` is to determine the compiler hashes, so if those are being specified in the environment we should avoid running it entirely. Instead, the user should provide all the information that we would have detected from that step (including compiler id and version). Additionally, this information must be part of the triplet instead of the environment in order to properly support multiple triplets (like host dependencies) in a single build.

Therefore, to directly fix these issues, I would suggest:
1. Make these triplet properties
2. Require all 6 values to be set if any are set (`{compiler hash, compiler id, compiler version}` x `{C, CXX}`)
3. Replace the [detect_compiler invocation implementation](https://github.com/microsoft/vcpkg-tool/blob/fc5bc1d7706106cd72b7175d6508ee40efb7c3e3/src/vcpkg/build.cpp#L572) to skip that step entirely if the information was provided as part of the triplet.

Alternatively, we could expose (effectively) `--feature-flags=-compilertracking` as a triplet setting, which would omit the detect compiler step entirely and rely on the hash of the triplet file + toolchain to distinguish package ABI. In this case, the user could add a comment to their triplet file containing any information they want to distinguish (such as compiler version). I think this is a more promising approach because it has the least interface churn.

klalumiere:
> First, the current implementation will only work on non-Windows platforms -- on Windows, 

Yeah I had thought about this: the plan was indeed to use `VCPKG_KEEP_ENV_VARS`.

> Alternatively, we could expose (effectively) --feature-flags=-compilertracking as a triplet setting, which would omit the detect compiler step entirely and rely on the hash of the triplet file + toolchain to distinguish package ABI. 

This is a great idea! I'll try to implement it soon. Thanks! :slightly_smiling_face:

ras0219-msft:
Moving this to draft for now; please feel free to mark it ready for review or close when you feel it's appropriate :)

klalumiere:
@ras0219-msft 
Actually, I've just realized, if I want to implement your suggestion, I'll need to make a pull request in [vcpkg-tool](https://github.com/microsoft/vcpkg-tool), not here, right?

Issue #16586, opened 2021-03-07
[flashlight-cpu/onednn] build fail


walenty500:
**Host Environment**

    OS: Ubuntu 18.04 
    Compiler: gcc

**To Reproduce**

`./vcpkg install flashlight-cpu`

**Failure logs**

```
CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:144 (message):
    Command failed: /home/krzysztof/vcpkg/downloads/tools/cmake-3.19.2-linux/cmake-3.19.2-Linux-x86_64/bin/cmake --build . --config Debug --target install -- -v -j9
    Working Directory: /home/krzysztof/vcpkg/buildtrees/onednn/x64-linux-dbg
    See logs for more information:
      /home/krzysztof/vcpkg/buildtrees/onednn/install-x64-linux-dbg-out.log

Call Stack (most recent call first):
  scripts/cmake/vcpkg_build_cmake.cmake:105 (vcpkg_execute_build_process)
  scripts/cmake/vcpkg_install_cmake.cmake:45 (vcpkg_build_cmake)
  ports/onednn/portfile.cmake:20 (vcpkg_install_cmake)
  scripts/ports.cmake:139 (include)


Error: Building package onednn:x64-linux failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `./vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: onednn:x64-linux
  Vcpkg version: 2021-01-13-unknownhash

Additionally, attach any relevant sections from the log files above.
```

Previously had problems with the same build, but with MLK (had to change MLK_REQUIRED_VERSION at ports/intel-mkl/portfile.cmake from 20200000 to 202101)
 
When doing ./vcpkg update :

`Using local portfile versions. To update the local portfiles, use `git pull`.
No packages need updating.`

When doing git pull : Already up to date

Attaching install-x64-linux-dbg-out.log:

[install-x64-linux-dbg-out.log](https://github.com/microsoft/vcpkg/files/6098141/install-x64-linux-dbg-out.log)

JackBoosY:
Please check if the memory and hard disk are free.

Thanks.

walenty500:
Thank you for the reply!

As it turned out the problem was with my memory size. I had only 8GB of RAM and 2GB of swap, so i changed the swapfile's size to 8GB, and after this didn't work i resized it to 16GB and it worked... 

[...]

PR #16633, opened 2021-03-10
[vcpkg_execute_build_process] Add warning message when memory or hard disk space is insufficient

JackBoosY:
When building a port in linux, occasionally there will be insufficient memory or disk, we should extract this information and display it in the console.

Related #16586.

### DIFF
--- a/scripts/cmake/vcpkg_execute_build_process.cmake
+++ b/scripts/cmake/vcpkg_execute_build_process.cmake
@@ -133,7 +133,7 @@ function(vcpkg_execute_build_process)
                 endif()
             endwhile()
         elseif(out_contents MATCHES "fatal error: ld terminated with signal 9 [Killed]")
-            message(WARNING "Insufficient memory or hard disk space is detected, please make sure there is enough space and try again.")
+            message(WARNING "ld was terminated with signal 9 [killed], please ensure your system has sufficient hard disk space and memory.")
         endif()
 
         if(error_code)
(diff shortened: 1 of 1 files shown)


scripts/cmake/vcpkg_execute_build_process.cmake
```diff
@@ -132,6 +132,8 @@ function(vcpkg_execute_build_process)
                     break()
                 endif()
             endwhile()
+        elseif(out_contents MATCHES "fatal error: ld terminated with signal 9 [Killed]")
+            message(WARNING "Insufficient memory or hard disk space is detected, please make sure there is enough space and try again.")
```
ras0219-msft, line 136: Signal 9 is a _very_ generic kill signal on linux, and cannot be assumed to mean "out of memory or hard disk". We can only suggest to the user that "ld was terminated with signal 9 [killed] -- ensure the system has sufficient hard disk space and memory".


https://stackoverflow.com/questions/16338884/what-does-exited-abnormally-with-signal-9-killed-9-mean#:~:text=Signal%209%20means%20that%20the%20application%20needs%20to,it%20property,%20then%20the%20process%20continues%20to%20live.

ras0219-msft (changes requested):

scripts/cmake/vcpkg_execute_build_process.cmake
JackBoosY, line 136:
```suggestion
            message(WARNING "ld was terminated with signal 9 [killed], please ensure your system has sufficient hard disk space and memory.")
```

PhoebeHui (approved):

Issue #16597, opened 2021-03-08
[darknet] build failure


ADerSchreiber:
**Host Environment**
- OS: Windows 10 Home
- Compiler: revision

**To Reproduce**
`.\vcpkg install darknet[opencv-base,cuda,cudnn]:x64-windows`

**Failure logs** 
CMake Error: Problem with archive_write_header(): Can't create '\\?\D:\OneFolder\AnotherFolder\vcpkg\downloads\tools\msys2\969c0913b9df89e1.tmp\usr\bin\gawk.exe'
CMake Error: Current file: usr/bin/gawk.exe
CMake Error: Problem extracting tar: D:/OneFolder/AnotherFolder/vcpkg/downloads/msys-gawk-5.1.0-1-x86_64.pkg.tar.xz

**Console output**

PS D:\OneFolder\AnotherFolder\vcpkg> .\vcpkg install darknet[opencv-base,cuda,cudnn]:x64-windows
Computing installation plan...
The following packages will be built and installed:
    darknet[core,cuda,cudnn,opencv-base]:x64-windows -> 2021-01-21
  * ffmpeg[avcodec,avdevice,avfilter,avformat,avresample,core,gpl,postproc,swresample,swscale]:x64-windows -> 4.3.1#11
  * libjpeg-turbo[core]:x64-windows -> 2.0.6
  * liblzma[core]:x64-windows -> 5.2.5#2
  * libpng[core]:x64-windows -> 1.6.37#14
  * libwebp[core,nearlossless,simd,unicode]:x64-windows -> 1.1.0#1
  * opencv[core,dnn,ffmpeg,jpeg,png,quirc,tiff,webp]:x64-windows -> 4.5.1
  * opencv4[core,dnn,ffmpeg,jpeg,png,quirc,tiff,webp]:x64-windows -> 4.5.1
  * protobuf[core]:x64-windows -> 3.14.0#1
  * pthreads[core]:x64-windows -> 3.0.0-6
  * quirc[core]:x64-windows -> 1.1#2
  * stb[core]:x64-windows -> 2020-09-14
  * tiff[core]:x64-windows -> 4.1.0#1
  * zlib[core]:x64-windows -> 1.2.11#9
Additional packages (*) will be modified to complete this operation.
Detecting compiler hash for triplet x64-windows...
Could not locate cached archive: C:\Users\Anonymous\AppData\Local\vcpkg\archives\f4\f44896aae58174d960a566c50389c807fc32909e.zip
[...]

PR #16635, opened 2021-03-10
[darknet] Update hash value for features

NancyLi1013:
**Describe the pull request**

- What does your PR fix? Fixes #https://github.com/microsoft/vcpkg/issues/16597#issuecomment-793726468

The hash values for feature `weights `and `weights-train` have changed. So update them to the correct ones.

```
-- Downloading https://github.com/AlexeyAB/darknet/releases/download/darknet_yolo_v4_pre/yolov4-tiny.weights -> darknet-cache/yolov4-tiny.weights...
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:105 (message):

          File path: [ F:/tool/vcpkg/downloads/temp/darknet-cache/yolov4-tiny.weights ]
      Expected hash: [ 804ca2ab8e3699d31c95bf773d22f901f186703487c7945f30dc2dbb808094793362cb6f5da5cd0b4b83f820c8565a3cba22fafa069ee6ca2a925677137d95f4 ]
        Actual hash: [ 7d4d9fe150f9fe3ea7d2310f1445fe983b31fbf06d301c70ecfe00e8559e6f1bf940198c2dd55db772238f23ea0092fb6553558e5414f3ee173b8b28e53c5b54 ]

  environment variables to

  scripts/cmake/vcpkg_download_distfile.cmake:195 (test_hash)
  ports/darknet/portfile.cmake:28 (vcpkg_download_distfile)
```

```
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:105 (message):

          File path: [ F:/tool/vcpkg/downloads/temp/darknet-cache/yolov4-tiny.conv.29 ]
      Expected hash: [ 318e47f4bdf43b7f4eff8f3669bc9ba66cd7bd8ffb31df5bc1978682c85fec8e63a8349958022fd933cc676cbf5241953f2181bf4d1789f7cf9d371e012e3e49 ]
        Actual hash: [ d145deb80aea3d422752e784323d89919281ed7a37d15f939f5b680ff62e502072a30e074c34a09026669126d360c7d6817ea1afa53cfa53c56bd1e16333a602 ]

  environment variables to

  scripts/cmake/vcpkg_download_distfile.cmake:195 (test_hash)
  ports/darknet/portfile.cmake:71 (vcpkg_download_distfile)
```


JackBoosY (approved):

cenit:
thanks.
I started some work https://github.com/microsoft/vcpkg/pull/15942 but then i got too busy. 
Your fix is very urgent

ras0219-msft:
This (darknet changing the published weights files) seems fundamentally unsustainable -- we cannot package unstable files like this in a public vcpkg while providing our commitments to reliability, robustness, and repeatability for users.

I believe the best we can do given the circumstances is to instead provide a checked-in script (.bat and/or .sh) in the package that is a list of `curl` calls to download the files. These will not require hash checks because users would be explicitly running this script as a procedure outside vcpkg; the files would not be part of any vcpkg package and thus their instability would not violate our policies.

@cenit Can you help describe how these files are used? Does having a script like I've described help or should we just cut the downloading of weights altogether?

andrewchoi5 (approved):

cenit:
> This (darknet changing the published weights files) seems fundamentally unsustainable -- we cannot package unstable files like this in a public vcpkg while providing our commitments to reliability, robustness, and repeatability for users.
> 
> I believe the best we can do given the circumstances is to instead provide a checked-in script (.bat and/or .sh) in the package that is a list of `curl` calls to download the files. These will not require hash checks because users would be explicitly running this script as a procedure outside vcpkg; the files would not be part of any vcpkg package and thus their instability would not violate our policies.
> 
> @cenit Can you help describe how these files are used? Does having a script like I've described help or should we just cut the downloading of weights altogether?

hi Robert
i completely understand vcpkg reasoning. At the same time weight files are never changed by us and are stable. In the recent darknet PR they are oscillating between two sets of hashes without any change on the server side not any update to the file itself. To be honest i don’t know what’s happening on github file attached to the release!

it would be a pity to loose the weight feature, which is really used by darknet users. vcpkg is also our preferred install mode on all operating systems, and so it would be really sad to say goodbye to them. At the same time, without a proper investigation on what’s going on, i can also understand your point of view...

ras0219-msft:
That's why I am proposing a compromise solution: we can provide a shell script that will download the weight files for users, which they could potentially invoke as part of their project configure or build (something like a FetchContent). I agree it would be best if the files were stable and vcpkg could reliably download them.

Depending on how much you'd like to cater to CMake, we could even provide some kind of cmake fragment that contains `FetchContent()` commands and is used either via `include()` or something like `find_package(darknet-weights)`.

cenit:
I can download those files within our own cmakelists.txt and use features to enable the download or not. But then it should be disabled on vcpkg because you never wanted any download performed automatically by ports (proxy etc, you wanted total control), so what’s the best solution?

should vcpkg override the file(DOWNLOAD ...) mechanism?

finally: do you have any idea why hashes are changing and flipping (not randomly, only a couple of hashes ever appear) every month for github-hosted files? i was thinking about abandoning that hosting but alternatives are not so reliable too

ras0219-msft:
I don't know what would be causing the github-hosted files to change unless the author is re-uploading them -- GitHub has stated (iirc) that assets uploaded to a release should never change.

My proposed solution is that the port would install a file such as `share/darknet-weights/darknet-weights-config.cmake`:
```cmake
# share/darknet-weights/darknet-weights-config.cmake
function(darknet_weights_fetch)
  FetchContent(...)
endfunction()
```
The user would then do something like this in their code:
```cmake
# CMakeLists.txt
find_package(darknet)
find_package(darknet-weights)
darknet_weights_fetch()
```
which would download the weights dynamically during configure. Since the download isn't happening during the vcpkg build and won't be part of the installed/ tree, it isn't a correctness problem if it isn't stable.

PhoebeHui:
Related to PR https://github.com/microsoft/vcpkg/pull/15942

cenit:
in PR #15942 i will remove weights and weights-train feature due to these problems. Users will download those files with a ps1 scripts afterwards

NancyLi1013:
Closing this PR in flavor of PR #15942, thanks for your help @cenit.

PR #16646, opened 2021-03-11
[sqlite3] Embed dllimport in header

ras0219-msft:
Sqlite3 uses a macro `SQLITE_API` on all external APIs as an injection point for dllimport/dllexport. This PR embeds the correct declspec according to the build configuration.


NancyLi1013:
Need to merge master to solve the conflicts here. @ras0219-msft

NancyLi1013 (approved):

strega-nil-ms:
Thanks robert!

PR #16390, opened 2021-02-24
Update taglib to 1.12.0

pdehne:
**Describe the pull request**

- What does your PR fix? Fixes #

Updates the port to 1.12.0

- Which triplets are supported/not supported? Have you updated the CI baseline?

No changes related to triplets or CI.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Yes.

PR #16266, opened 2021-02-17
[taglib] Upgrade to 1.12

uklotzde:
- Replace CONTROL with vcpkg.json
- Use Git tag as REF
- Use official version number (without the date suffix)
- Link official homepage instead of GitHub project page
- Add license string according to GitHub repo
  https://github.com/taglib/taglib/issues/999
- Add missing vcpkg_fixup_pkgconfig() in portfile

PR #16649, opened 2021-03-11
[taglib] Upgrade from 1.12-beta-2 to 1.12

uklotzde:
Follow-up of:
 - #16266 (original)
 - #16390 (merged beforehand)

Original changes:
- Replace CONTROL with vcpkg.json
- Use Git tag as REF
- Link official homepage instead of GitHub project page
- Add license string according to GitHub repo
- Add missing vcpkg_fixup_pkgconfig() in portfile

Additional changes:
- Switch to semantic versioning ("version-semver")
- Use VCPKG_LIBRARY_LINKAGE instead of VCPKG_CRT_LINKAGE to decide about static/dynamic build (https://github.com/microsoft/vcpkg/pull/16390/files#r591427047)
- Revert disabling of warnings in portfile.cmake and instead patch the original CMake build file. This ensures that the warnings are re-enabled when a fix from upstream becomes available.

### DIFF
--- a/ports/taglib/vcpkg.json
+++ b/ports/taglib/vcpkg.json
@@ -1,0 +1,10 @@
+{
+  "name": "taglib",
+  "version-semver": "1.12.0",
+  "description": "TagLib Audio Meta-Data Library",
+  "homepage": "https://taglib.org/",
+  "license": "LGPL-2.1 OR MPL-1.1",
+  "dependencies": [
+    "zlib"
+  ]
+}
--- a/ports/taglib/CONTROL
+++ b/ports/taglib/CONTROL
@@ -1,5 +1,0 @@
-Source: taglib
-Version: 1.12.0-20210123
-Description: TagLib Audio Meta-Data Library
-Homepage: https://github.com/taglib/taglib
-Build-Depends: zlib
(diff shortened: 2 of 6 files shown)


JonLiu1993 (approved):

uklotzde:
Please confirm that the versioning is correct after switching to *semver*. I don't want to mess up any existing dependencies.

ports/taglib/vcpkg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "taglib",
+  "version-semver": "1.12.0",
```
Be-ing, line 3: @JonLiu1993 is `port-version: 1` needed?

uklotzde: According to the docs 1.12.0-yyyymmdd precedes 1.12.0. But someone else should confirm this.

https://vcpkg.readthedocs.io/en/latest/specifications/versioning/#2-specifying-package-versions

ras0219-msft:
👍 Looks great to me, thanks for the update PR!

ports/taglib/vcpkg.json
JonLiu1993: @Be-ing ,If we update the version, we will delete the port-version. If we modify the BUG or other changes, we will add the port-version: 1. If the port-version already exists, we will increment it by 1 each time we modify it.

PR #16651, opened 2021-03-11
[drogon] Update to 1.4.1

an-tao:


ports/drogon/CONTROL
```diff
@@ -1,8 +1,5 @@
 Source: drogon
-Version: 1.3.0
+Version: 1.4.1
 Homepage: https://github.com/an-tao/drogon
 Description:Drogon: A C++14/17 based HTTP web application framework running on Linux/macOS/Unix/Windows
-Build-Depends: trantor, zlib, jsoncpp, libmariadb (!osx), libmariadb[iconv] (osx), libpq, sqlite3, brotli, libuuid (!windows)
-
-Feature: ctl
```
JonLiu1993, line 7: Hi @an-tao ,Why did you delete this feature?

ports/drogon/portfile.cmake
```diff
@@ -1,46 +1,42 @@
...
+  an-tao/drogon
+  REF
+  v1.4.1
+  SHA512
+  8611c18e65229095f5443f10c87e91593d619bc3a2d47da6d19d501a64be1aba12a813e44ccfc1c10179e5b0fa10121e8806955c0091258992501b88fa50d939
+  HEAD_REF
+  master
+  PATCHES
+  vcpkg.patch
+  resolv.patch)
```
JonLiu1993, line 14: @an-tao ，Please restore the format here.

ports/drogon/CONTROL
an-tao: I think drogon_ctl is needed in almost all scenarios. 
Actually, some people complained to me that drogon_ctl is missing after installing drogon in the latest version of vcpkg.

ports/drogon/portfile.cmake
```diff
@@ -1,46 +1,42 @@
...
-    PREFER_NINJA
-    OPTIONS
-        -DBUILD_EXAMPLES=OFF
-        ${FEATURE_OPTIONS}
-)
+vcpkg_configure_cmake(SOURCE_PATH
+                      ${SOURCE_PATH}
+                      PREFER_NINJA
+                      OPTIONS
+                      -DBUILD_EXAMPLES=OFF)
```
JonLiu1993, line 20: @an-tao ，Please keep the format here the same as before.

ports/drogon/CONTROL
JonLiu1993: Ok that's great

an-tao:
@JonLiu1993 If some dependencies update after the PR and break the build of drogon, what can I do to avoid this situation?

JonLiu1993:
> @JonLiu1993 If some dependencies update after the PR and break the build of drogon, what can I do to avoid this situation?

I'll fix it

JonLiu1993 (approved):

an-tao:
> If a new problem arises in response to this situation, we will solve it, you know this cannot be avoided

I see, thanks for your reply.

ports/drogon/CONTROL
ras0219-msft: I think we should keep this as a feature, but make it a default:
```
Default-Features: ctl
```
This is useful for cases like cross-compiling to android or ios where creating an executable may not be possible.

ras0219-msft (changes requested):
Thanks for the PR! LGTM once the feature is moved to default.

JonLiu1993 (approved):

ras0219-msft:
Hi @an-tao, thanks for the PR!

Obviously needing to directly link lower level postgresql libraries is not great, so I've fixed the underlying libpq port and reverted that part of the drogon patches here: https://github.com/ras0219-msft/vcpkg/commits/drogon. Could you please pull those additional commits into this PR?

Additionally, I noticed that drogon performs some automatic detection with a boost polyfill fallback -- this hidden optional dependency causes a bunch of issues in a package manager so I've taken the simplest route for now and ensured the build will fail if that path would be taken.

an-tao:
> Hi @an-tao, thanks for the PR!
> 
> Obviously needing to directly link lower level postgresql libraries is not great, so I've fixed the underlying libpq port and reverted that part of the drogon patches here: https://github.com/ras0219-msft/vcpkg/commits/drogon. Could you please pull those additional commits into this PR?
> 
> Additionally, I noticed that drogon performs some automatic detection with a boost polyfill fallback -- this hidden optional dependency causes a bunch of issues in a package manager so I've taken the simplest route for now and ensured the build will fail if that path would be taken.

Thank you very much, I have added them to this PR, you solved a very annoying problem for me.

ras0219-msft:
Thanks so much for the PR!

PR #16659, opened 2021-03-11
[qt5-base] make libpq an optional dependency

autoantwort:
**Describe the pull request**

- What does your PR fix? 
Make `libpq` an optional dependency. Reason: `libpq` does not build when cross compiling from linux -> windows and does not build with mingw and adding support for these two cases is a lot of work, additionally only a few users need the libpq support in qt5-base.

### DIFF
--- a/ports/qt5-base/portfile.cmake
+++ b/ports/qt5-base/portfile.cmake
@@ -45,6 +45,7 @@ else()
     )
 endif()
 
+set(WITH_PGSQL OFF)
 if("postgresqlplugin" IN_LIST FEATURES)
     set(WITH_PGSQL ON)
 endif()
--- a/ports/qt5-base/portfile.cmake
+++ b/ports/qt5-base/portfile.cmake
@@ -45,6 +45,11 @@ else()
     )
 endif()
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        postgresqlplugin WITH_PGSQL
+)
+
 include(qt_port_functions)
 include(configure_qt)
 include(install_qt)
@@ -121,12 +126,15 @@ list(APPEND CORE_OPTIONS
     -system-doubleconversion
     -system-sqlite
     -system-harfbuzz
-    -sql-psql
     -icu
     -no-angle)      # Qt does not need to build angle. VCPKG will build angle!
 
 if(QT_OPENSSL_LINK)
     list(APPEND CORE_OPTIONS -openssl-linked)
+endif()
+
+if(WITH_PGSQL)
+    list(APPEND CORE_OPTIONS -sql-psql)
 endif()
 
 if(VCPKG_TARGET_IS_WINDOWS)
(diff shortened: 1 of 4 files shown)


ports/qt5-base/portfile.cmake
```diff
@@ -45,6 +45,11 @@ else()
     )
 endif()
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        postgresqlplugin WITH_PGSQL
+)
```
Neumann-A, line 51: `vcpkg_check_features` is mainly meant for cmake based ports. Was the function changed to actually set `WITH_PGSQL` in the scope of the portfile?

autoantwort: Ok, i didn't know that. I thought it is good style to use `vcpkg_check_features` instead of `if("..." IN_LIST FEATURES)`. 

> Was the function changed to actually set WITH_PGSQL in the scope of the portfile?

I used it like in the examples, so I thought this woulds

Neumann-A: Better use the IN_LIST variant and set WITH_PGSQL there.

autoantwort: Ok

ports/qt5-base/portfile.cmake
```diff
@@ -45,6 +45,10 @@ else()
     )
 endif()
 
+if("postgresqlplugin" IN_LIST FEATURES)
```
Neumann-A, line 48:
```suggestion
set(WITH_PGSQL)
if("postgresqlplugin" IN_LIST FEATURES)
```

need to make sure nobody is messing with it in the triplet

autoantwort: Ok. This could be fixed for all other variables also.

autoantwort:
the `x86_windows` pipeline error looks like an error in the pipeline and not in the port

PhoebeHui:
The openmvg:x86-windows failed with following failures, I have rerun the CI testing.
```
\main_svgSample.dir/manifest.res" failed (exit code 1201) with the following output:
LINK : fatal error LNK1201: error writing to program database 'D:\buildtrees\openmvg\x86-windows-dbg\Windows-AMD64-Debug\main_svgSample.pdb'; check for insufficient disk space, invalid path, or insufficient privilege
```

PhoebeHui (approved):
Thanks for the PR!

ports/qt5-base/portfile.cmake
Neumann-A: yeah looks like `QT_OPENSSL_LINK` and `QT_MYSQL_PLUGIN` should use set and not `option`. For consistency you probably want to rename this variable to `QT_PGSQL_PLUGIN `

ports/qt5-base/portfile.cmake
ras0219-msft: `vcpkg_check_features()` will define variables in the parent:

https://github.com/microsoft/vcpkg/blob/fced4bef1606260f110d74de1ae1975c2b9ac549/scripts/cmake/vcpkg_check_features.cmake#L219

If `PREFIX` is passed, then it will define them as `$PREFIX_$VAR`, otherwise it will directly define `$VAR`.

ports/qt5-base/portfile.cmake
```diff
@@ -121,14 +126,17 @@ list(APPEND CORE_OPTIONS
...
     -icu
     -no-angle)      # Qt does not need to build angle. VCPKG will build angle!
 
 if(QT_OPENSSL_LINK)
     list(APPEND CORE_OPTIONS -openssl-linked)
 endif()
 
+if(WITH_PGSQL)
+    list(APPEND CORE_OPTIONS -sql-psql)
+endif()
```
ras0219-msft, line 138: Should this be
```suggestion
if(WITH_PGSQL)
    list(APPEND CORE_OPTIONS -sql-psql)
else()
    list(APPEND CORE_OPTIONS -no-sql-psql)
endif()
```
?

autoantwort: The default mode is auto, if it detects a libpq it will use it. I have now applied your change, so that you only use libpq if you requested it.

ports/qt5-base/portfile.cmake
ras0219-msft: Agreed that these additional changes would be good, but don't seem critical enough to block merging this PR.

ras0219-msft:
LGTM, thanks for the improvement!

PR #16478, opened 2021-03-01
[yasm-tool] Migrate to Host Dependencies

strega-nil:
This allows mpg123, nettle, and gmp to support non-x86 windows manifest mode

fixes #15087 

~~Depends on #16627~~

~~Depends on #16660~~

PR #16660, opened 2021-03-11
[ci.baseline.txt] Add 'host' type

strega-nil:
port=host means "skip on non-host triplets"

this also pulls out baseline parsing from the behavior that the
baseline represents, which means we're not duplicating
extremely complex code twice.

Taken from work in #16478


strega-nil:
Closed in favor of adding a new supports key.

Issue #9404, opened 2019-12-21
[New Port Request] websocketpp-standalone


aaslane:
Library name: websocketpp-standalone

Library description:  Library that implements RFC6455 The WebSocket Protocol using standalone version of ASIO

Source repository URL: https://github.com/zaphoyd/websocketpp

Project homepage (if different from the source repository):

Anything else that is useful to know when adding (such as optional features the library may have that should be included):

This is effectively the same as the websocketpp package, except the build-depends has to change boost-asio -> asio

This removes the dependency on the boost libs for those that do not require them.

NancyLi1013:
If you wish to add this port in the future, you can reopen this issue, but we're closing it for now.

PR #16675, opened 2021-03-12
[websocketpp] Add features to build without Boost

joe-noel-dev:
This creates features for websocketpp so that the consumer can choose the ASIO library and whether openssl and zlib are used (since these are all optional).

This should be a transparent change for existing users.

### DIFF
--- a/ports/websocketpp/vcpkg.json
+++ b/ports/websocketpp/vcpkg.json
@@ -1,0 +1,21 @@
+{
+  "name": "websocketpp",
+  "version": "0.8.2",
+  "port-version": 1,
+  "description": "Library that implements RFC6455 The WebSocket Protocol",
+  "homepage": "https://github.com/zaphoyd/websocketpp",
+  "documentation": "http://docs.websocketpp.org/",
+  "default-features": [
+    "recommended"
+  ],
+  "features": {
+    "recommended": {
+      "description": "Use recommended dependencies",
+      "dependencies": [
+        "boost-asio",
+        "openssl",
+        "zlib"
+      ]
+    }
+  }
+}
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -6234,7 +6234,7 @@
     },
     "websocketpp": {
       "baseline": "0.8.2",
-      "port-version": 0
+      "port-version": 1
     },
     "wepoll": {
       "baseline": "1.5.8",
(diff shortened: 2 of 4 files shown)


joefocusrite:
This has been requested in issue #9404

PhoebeHui:
@joefocusrite, thanks for the PR!

Could you update the baseline version via the following command, and then summit the changes?

./vcpkg x-add-version websocketpp --overwrite-version

versions/baseline.json
```diff
@@ -6218,7 +6218,7 @@
     },
     "websocketpp": {
       "baseline": "0.8.2",
-      "port-version": 0
+      "port-version": 2
```
PhoebeHui, line 6221:
```suggestion
      "port-version": 1
```

PhoebeHui (changes requested):

PhoebeHui:
@joefocusrite, could you execute './vcpkg x-add-version websocketpp --overwrite-version' and summit the changes?

```
Error: While reading versions for port websocketpp from file: C:\a\1\s\versions\w-\websocketpp.json
       File declares version `0.8.2#1` with SHA: 02d3d2638964b690dafd98788c4a035c11611077
       But local port with the same verion has a different SHA: 0ca3e13e38a50cb0da259618ab3bd19a8eaace05

           vcpkg x-add-version websocketpp
       to add a new version.

    vcpkg x-add-version --all
```

joefocusrite:
Hi @PhoebeHui. Thanks for getting back to me! I've just re-run that command and pushed 🙂

PhoebeHui (approved):
Thanks for your updates!

ras0219-msft (changes requested):
Thanks for the PR!

Because this is a header-only library, the `asio` feature provides little value to users since it is off by default (they can just directly depend on asio instead).

Additionally, for the same reason, I would suggest bundling all these features together into a single feature called "recommended" which depends on all three:
```json
"features": {
  "recommended": {
    "description": "Use recommended dependencies",
    "dependencies": [ "boost-asio", "openssl", "zlib" ]
  }
}
```

ports/websocketpp/vcpkg.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "websocketpp",
+  "version-string": "0.8.2",
```
ras0219-msft, line 3:
```suggestion
  "version": "0.8.2",
```

ras0219-msft:
LGTM, thanks!

PR #16693, opened 2021-03-14
[libbacktrace] new port

### LINKED ISSUE #15774: [New Port Request] libbacktrace
Library name: libbacktrace

Library description: "Sample uses would be to print a detailed backtrace when an error occurs or to gather detailed profiling information."

Source repository URL: https://github.com/ianlancetaylor/libbacktrace

No. But the lib can be used for [boost stacktrace to get meaningful stacktraces on any compiler on POSIX, or MinGW, or MinGW-w64](https://www.boost.org/doc/libs/1_75_0/doc/html/stacktrace/configuration_and_build.html)

autoantwort:
**Describe the pull request**

- _What does your PR fix?_ 
Fixes #15774

- _Which triplets are supported/not supported? Have you updated the CI baseline?_
   Only gcc and clang are supported

- _Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?_
Yes (there are no guidelines for handling licenses, I have copied from other ports)


scripts/ci.baseline.txt
```diff
@@ -544,6 +544,13 @@ leptonica:x64-uwp=fail
 leptonica:arm-uwp=fail
 leveldb:arm-uwp=fail
 leveldb:x64-uwp=fail
+libbacktrace:arm64-windows=fail
+libbacktrace:arm-uwp=fail
+libbacktrace:x64-uwp=fail
+libbacktrace:x64-windows=fail
+libbacktrace:x64-windows-static=fail
+libbacktrace:x64-windows-static-md=fail
+libbacktrace:x86-windows=fail
```
PhoebeHui, line 553: Could you use 'supports' field in vcpkg.json instead?

See https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md#supports

PhoebeHui (changes requested):
@autoantwort, thanks for your contribution!

scripts/ci.baseline.txt
autoantwort: Can I say `{ "supports": "!msvc" }`? Because mingw and clang on windows are supported

PhoebeHui: I see, thanks for the information! since we don't support 'msvc', I think it's fine to skip the triplets in ci.baseline.txt.

autoantwort: Ok 👍

PhoebeHui (approved):
LGTM！

ports/libbacktrace/portfile.cmake
```diff
@@ -0,0 +1,19 @@
+if (CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+    vcpkg_fail_port_install(ALWAYS MESSAGE "The library relies on the C++ unwind API defined at https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html This API is only provided by GCC and clang. You are using MSVC.")
+endif()
```
ras0219-msft, line 3: This should be changed to be a warning instead of a fatal error -- additionally the `CMAKE_C_COMPILER_ID` is not available in `portfile.cmake` so this condition can currently never be triggered.

Instead, I would suggest:
```suggestion
if(VCPKG_TARGET_IS_WINDOWS)
    message("libbacktrace cannot be built using MSVC on Windows due to relying on the C++ unwind API https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html")
endif()
```

ras0219-msft (changes requested):
LGTM with the change!

ras0219-msft:
Thanks!

PR #16694, opened 2021-03-14
[vcpkg] Ignore toolsrc build files

autoantwort:
currently files like `toolsrc/build.rel/*.o` are shown if you do `git status`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -286,6 +286,7 @@ __pycache__/
 /installed*/
 /packages/
 /scripts/buildsystems/tmp/
+/toolsrc/*
 #ignore custom triplets
 /triplets/*
 #add vcpkg-designed triplets back in


StarGate-One:
@autoantwort /cc: @BillyONeal 
- New version (bootstrap): The VERSION.txt file should be the only file present in the `[vcpkg-root]/toolsrc directory` (no sub-directories).
- Old version (bootstrap): The build sub-directories should be deleted after a successful build, compile, link and copy of the binary to [vcpkg-root].

BillyONeal (approved):
Thanks!

ras0219-msft:
While this change is harmless, @autoantwort can you indicate the _specific_ steps in the current product that lead to producing `build.rel` files in the toolsrc directory?

autoantwort:
Steps:
1. `git clone git@github.com:microsoft/vcpkg.git`
2. `cd vcpkg`
3. `git checkout a9a636c1a6af4cbfa0427762f9258a76348a0a0d` // an old version where the toolsrc was in the vckpg repo
4. `./bootstrap-vcpkg.sh`
5. `git checkout master` // you use vcpkg since more than 2 months and update now to a new version
6. `git status` // shows untracked files in `toolsrc/build.rel/`

So this happens if you use vcpkg more than two months and update now to a new version. A common scenario, I would say.

StarGate-One:
Very odd indeed?
I can replicate on OpenSUSE running on WSL2 and Hyper-V.

ras0219-msft:
I don't think we should make this change to hide these files from the user -- they can and should be deleted at the user's earliest convenience and all future bootstraps will no longer create those files. For example, adding them to `.gitignore` prevents them from being cleaned via `git clean`. It would require a more powerful `git clean` operation to purge them, which would also purge installed, downloads, and many other important directories.

PR #16711, opened 2021-03-15
[asio] update to 1.18.1

### LINKED ISSUE #16692: [asio] update to 1.18.1
Library name: asio

New version number: 1.18.1

- Enabled support for UNIX domain sockets on Windows. From Windows 10, UNIX domain sockets (a.k.a "local" sockets) are supported on Windows, with the exception of the connect_pair function (which will fail with an operation_not_supported error).

- Added executor-converting construction and assignment to ip::basic_resolver.

- Added compatibility between polymorphic executors and the (deprecated) handler invocation hook.

- Added the experimental::as_single completion token adapter. The as_single completion token adapter can be used to specify that the completion handler arguments should be combined into a single argument. For completion signatures with a single parameter, the argument is passed through as-is. For signatures with two or more parameters, the arguments are combined into a tuple. The as_single adapter may be used in conjunction with use_awaitable and structured bindings as follows:

```cpp
auto [e, n] = co_await socket.async_read_some(
    asio::buffer(data), as_single(use_awaitable));
```

Alternatively, it may be used as a default completion token like so:
```cpp
using default_token = as_single_t<use_awaitable_t<>>;
using tcp_socket = default_token::as_default_on_t<tcp::socket>;
// ...
awaitable<void> do_read(tcp_socket socket)
{
  // ...
  auto [e, n] = co_await socket.async_read_some(asio::buffer(data));
  // ...
}
```

- Added support for MSG_NOSIGNAL on more platforms by using _POSIX_VERSION to detect whether it is supported.

- Added the ability to compile using libpthread on Windows.
- Added workarounds for the Intel C++ compiler.
- Added more support for detecting and optimising for handlers that have no custom executor.
- Reduced lock contention for timer cancellation on Windows.
- Reinstated a previously removed null-pointer check, as it had a measurable impact on performance.
- Fixed the executor concept to test for a const-qualified execute().
- Fixed any_executor support for builds without RTTI support.
- Fixed the thread_pool unit test to work without RTTI support.
- Fixed C++20 coroutines compatibility with clang on Windows.
- Fixed some compatibility issues with Windows Runtime.
- Fixed shadow name warnings caused by addition of asio::query.
- Fixed a "logical ‘or’ of equal expressions" warning on linux.
- Fixed a benign switch fallthrough warning.
- Added missing push/pop_options.hpp includes.
- Suppressed zero-as-null-pointer-constant warnings.
- Fixed a comma-operator warning.
- Updated the documentation to clarify when the select reactor is used on Windows.
- Fixed potential ambiguity caused by any_executor comparisons and conversion.
- Added detection of non-experimental C++20 coroutines on MSVC 19.8.
- Fixed compatibility with uClibc.
- Fixed strand<> adaptation of Networking TS executors when targeting older C++ versions or less conformant compilers.

JonLiu1993:
Fix #16692 

Update asio to the latest version 1.81.1

- x86-windows
- x64-windows
- x64-windows-static


NancyLi1013 (approved):

versions/a-/asio.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "ef4724218afa535cc2bec662e7c59ada0c9162c0",
+      "version-string": "1.18.1",
```
ras0219-msft, line 5:
```suggestion
      "version": "1.18.1",
```

ras0219-msft (changes requested):

ras0219-msft:
LGTM

Issue #9272, opened 2019-12-11
[ffmpeg] add amd/amf, qsv hard encoding and openh264 encoding support


fawdlstty:

MicroYY:
Anything update? Still no qsv support

NancyLi1013:
@mcmtroffaes

Would you like to help add these features for `ffmpg`?

Thanks in advance.

mcmtroffaes:
Currently a bit too swamped with other things to take a lead on this.

I lack the hardware to test amf or qsv. For openh264, https://github.com/cisco/openh264 needs to be added as a port first. Judging from the readme, this will be non-trivial for vcpkg, requiring msvc to be called from an msys2 shell (similar to what we currently do to compile ffmpeg itself). Happy to help with reviewing and testing where I can.

NancyLi1013:
Thanks for your help and clarifications about this.

We might delay to implement these features.

shahneil:
For AMF, it looks like https://github.com/GPUOpen-LibrariesAndSDKs/AMF needs to be added as a port first, as the AMF framework headers are needed to configure ffmpeg with `--enable-amf` (see https://www.ffmpeg.org/general.html#toc-AMD-AMF_002fVCE).

Since AMF/VCE is Windows-only (see the ffmpeg link above), I believe a new Windows-only ffmpeg feature `amf` would need to be added that includes the `--enable-amf` option during ffmpeg compilation.

If anyone can provide further guidance that would be great, as I'm quite new to vcpkg. I'll try to create a port for the AMF headers in the meantime.

xiaozhuai:
Any progress? Still no qsv support.

PR #16714, opened 2021-03-15
[openh264] Add openh264 port

RichLogan:
Adds OpenH264 port to vcpkg (possibly of help to #9272).

Waiting for CI as I don't have a windows machine in front of me.


ports/openh264/portfile.cmake
```diff
@@ -0,0 +1,40 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO cisco/openh264
+    REF f15f940425eebf24ce66984db2445733cf500b7b
```
NancyLi1013, line 4: I noticed there was released version. So why do you not use the latest commit `50a1fcf70fafe962c526749991cb4646406933ba` that tag v2.1.1 has?

ports/openh264/vcpkg.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "openh264",
+  "version-string": "1.0.0",
```
NancyLi1013, line 3:
```suggestion
  "version": "2.1.1",
```
This might be `version` here. Please see the doc https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions.

scripts/cmake/vcpkg_find_acquire_program.cmake
```diff
@@ -337,14 +337,14 @@ function(vcpkg_find_acquire_program VAR)
...
     set(NOEXTRACT true)
     set(PROGNAME gas-preprocessor)
-    set(SUBDIR "b5ea3a50")
+    set(SUBDIR "4daa6115")
     set(REQUIRED_INTERPRETER PERL)
     set(SCRIPTNAME "gas-preprocessor.pl")
     set(PATHS ${DOWNLOADS}/tools/gas-preprocessor/${SUBDIR})
     set(_vfa_RENAME "gas-preprocessor.pl")
-    set(URL "https://raw.githubusercontent.com/FFmpeg/gas-preprocessor/b5ea3a50ed991e6a3218e89402a8162c73f59cb2/gas-preprocessor.pl")
+    set(URL "https://raw.githubusercontent.com/FFmpeg/gas-preprocessor/4daa611556a0558dfe537b4f7ad80f7e50a079c1/gas-preprocessor.pl")
```
NancyLi1013, line 345: Why do you update this tool here?

NancyLi1013 (changes requested):

ports/openh264/portfile.cmake
RichLogan: Meson builds (without which building would be complicated) seem to be broken on that release unfortunately and were fixed relatively recently. I figured to just use the latest commit instead.

scripts/cmake/vcpkg_find_acquire_program.cmake
RichLogan: Has vcpkg perl's been updated recently? The current gas-preprocessor did not work for me due to warnings that recently became errors. Ref: https://github.com/FFmpeg/gas-preprocessor/commit/9a769cfd8332638f97d551840844123e7bf1534d

RichLogan: On this topic, is there not a better way of splitting up find_acquire_program? Changing that caused a rebuild of many many ports despite a much smaller amount actually using `gas-preprocessor`. Is that because everyone who uses `vcpkg_find_acquire_program` at all needs to rebuild?

NancyLi1013: What do you mean splitting up `find_acquire_program`? Sorry, I'm not clear.

ports/openh264/portfile.cmake
NancyLi1013: Thanks for your clarification.

ports/openh264/vcpkg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "openh264",
+  "version": "2.1.1",
```
NancyLi1013, line 3: Since you used the latest commit instead of a released version here, it might be better to use a date as version.

Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#version and https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions
```suggestion
  "version-date": "2021-03-16",
```

NancyLi1013 (changes requested):

scripts/cmake/vcpkg_find_acquire_program.cmake
RichLogan: My aplogies, it seems my assumption of what was going on isn't quite right.

I was wondering why so many ports are being built on CI despite a very small change. I thought it was because I had made a change to `vcpkg_find_acquire_program` it was rebuilding every port that used that function despite the fact that I only changed the `GASPREPROCESSOR` part of that function which is barely used by any ports. On review I'm not sure if that's actually what's going on or not.

NancyLi1013: For CI test, it will run every ports that are related with the changes, such as dependent ports.

For the tool update, I think it's necessary if this port requires the new version.

ports/openh264/portfile.cmake
```diff
@@ -0,0 +1,40 @@
...
+    vcpkg_add_to_path(${NASM_EXE_PATH})
+endif()
+
+# Gas.
+if ((VCPKG_TARGET_ARCHITECTURE STREQUAL "arm" OR VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64") AND VCPKG_TARGET_IS_WINDOWS)
+    vcpkg_find_acquire_program(GASPREPROCESSOR)
+    foreach(GAS_PATH ${GASPREPROCESSOR})
+        get_filename_component(GAS_ITEM_PATH ${GAS_PATH} DIRECTORY)
+        vcpkg_add_to_path(${GAS_ITEM_PATH})
+    endforeach(GAS_PATH)
```
NancyLi1013, line 26: I'm not sure why you used `foreach` function here. Can you make a simple description?

RichLogan: That gives multiple paths, one for perl and one for the scripts itself: 

```cmake
# message(${GASPREPROCESSOR})
/vcpkg/downloads/tools/perl/5.30.0.1/perl/bin/perl.exe;/vcpkg/downloads/tools/gas-preprocessor/4daa6115/gas-preprocessor.pl
```

NancyLi1013: Thanks for your clarification about this point.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your PR @RichLogan.

scripts/cmake/vcpkg_find_acquire_program.cmake
ras0219-msft: We're working on breaking this helper up into individual ports as part of our ongoing script stabilization work; we've recently added "Host dependencies" (https://github.com/microsoft/vcpkg/blob/master/docs/users/host-dependencies.md) which gives us a lot more power in this area.

ras0219-msft (approved):
LGTM once the CI passes, thanks for the PR!

PR #16725, opened 2021-03-16
[openmama] Modify the output path of the common header file

Cheney-W:
**Describe the pull request**
In CI Test, workflow installation failed with following errors:
```
error C2079: 'list' uses undefined struct 'list_head'
error C2198: 'list_for_each': too few arguments for call 
error C2143: syntax error: missing ';' before '{'
```
This issue occurs because there is a list.h file generated by openmama and saved in the vcpkg\installed{triplets}\include folder, it conflict with the list.h file of workflow. As a result, the structure and macro defined in the header file list.h(workflow) become undefined during compilation.
After communicatied with upstream, I moved these common header files into the wombat folder and modified header files those include the common header files.

BTW, update the REPO and homapage from OpenMAMA/OpenMAMA to finos/OpenMAMA, because OpenMAMA/OpenMAMA is no longer found.

### DIFF
--- a/ports/openmama/portfile.cmake
+++ b/ports/openmama/portfile.cmake
@@ -2,8 +2,8 @@ vcpkg_find_acquire_program(FLEX)
 
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
-    REPO OpenMAMA/OpenMAMA
-    REF c4925ee103add1a51c1d27be45b46d97af347f36 # https://github.com/OpenMAMA/OpenMAMA/tree/OpenMAMA-6.3.1-release
+    REPO finos/OpenMAMA
+    REF c4925ee103add1a51c1d27be45b46d97af347f36 # https://github.com/finos/OpenMAMA/releases/tag/OpenMAMA-6.3.1-release
     SHA512 e2773d082dd28e073fe81223fc113b1a5db7cd0d95e150e9f3f02c8c9483b9219b5d10682a125dd792c3a7877e15b90fd908084a4c89af4ec8d8c0389c282de2
     HEAD_REF next
 )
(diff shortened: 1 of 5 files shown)


ports/openmama/portfile.cmake
```diff
@@ -41,7 +41,20 @@ endif()
...
 
-# Vcpkg does not like this header name and shouldn't be required anyway, so remove it
-file(REMOVE "${CURRENT_PACKAGES_DIR}/include/platform.h")
+foreach(OPENMAMA_ROOT_HEADER destroyhandle.h platform.h list.h lookup2.h property.h timers.h wlock.h windows)
+    if(EXISTS "${CURRENT_PACKAGES_DIR}/include/${OPENMAMA_ROOT_HEADER}")
+        file(RENAME "${CURRENT_PACKAGES_DIR}/include/${OPENMAMA_ROOT_HEADER}" "${CURRENT_PACKAGES_DIR}/include/wombat/${OPENMAMA_ROOT_HEADER}")
+    endif()
+endforeach()
+
+if(EXISTS "${CURRENT_PACKAGES_DIR}/include/mama/integration/transport.h")
```
ras0219-msft, line 50: These files should always exist iiuc; we shouldn't need `if(EXISTS)`.

ras0219-msft (changes requested):
Please also validate that you can `#include` every installed file in the `include/` folder and successfully compile (no need to link).

ports/openmama/portfile.cmake
Cheney-W: Thanks for your review.

Cheney-W:
@ras0219-msft 
There are more than 300 header files generated by openmama, most of which have nothing to do with this modification. I have verified the header file that was modified and other header files affected by this modification, and they can be compiled normally. So I want to confirm with you that other header files still need to be verified?

NancyLi1013 (approved):

ras0219-msft:
Please create one .cpp file that includes every header and try to compile that file.

```powershell
ls -recurse .\packages\openmama_x64-windows\include\* | % { "#include <$_>" -replace "\\","/" } | out-file -enc ascii test.cpp
```

Cheney-W:
I have tested all of the generated header files of openmama, and them built failed with following errors:
error

I also tested the files of previous versions of openmama and the files manually downloaded from https://github.com/finos/OpenMAMA, and above errors still exist. 
So these errors have nothing to do with this PR.

fquinner:
Hi it's the OpenMAMA maintainer here and yes this looks good to me. The fact is most users won't actually be using those headers anyway and in future I'll try and get these moved into a subdirectory by default. When that time comes I'll update this workaround in VCPKG but for now yes this looks perfectly fine to me.

NancyLi1013 (approved):

PR #16730, opened 2021-03-16
[libsodium] Rework CMake build to match autoconf

### LINKED ISSUE #14134: [libsodium] use of AVX512 is not portable
**Describe the bug**
We are using vcpkg to build dependencies for an application deployed as a Docker container. Our CI build machines have the AVX512 extensions, but the machines we deploy to often don't. The libsodium port [detects the availability of the AVX512 extensions](https://github.com/microsoft/vcpkg/blob/4aa08938a22f7cb90842280571005c4a637e7098/ports/libsodium/CMakeLists.txt#L358), and libsodium itself seems to perform runtime detection of CPU extensions when you call `sodium_init()`, but I suspect these runtime checks only protect against explicit use of AVX512 instructions. I believe by setting the compiler flag `-mavx512f`, this allows GCC to use these instructions in generated code. The runtime checks libsodium performs won't help here, as far as I'm aware. Ultimately, the end result is our application failing, signalling an illegal instruction when it ultimately encounters one of these AVX512 instructions while running on a machine without support for them.

I am not sure what the right course of action is. I suppose I could add an overlay port where I patch the `CMakeLists.txt` file to not light up AVX512. That would solve my immediate problem, but I think there is a bigger issue here. I am sure some users would be happy to always have the AVX512 instructions enabled if it fits their deployment model. This does speak to broader issue around portability. I'm sure libsodium isn't the only project making use of these instructions.

**Environment**
- build: machine with AVX512
- runtime: machine without AVX512

thomasgt:
**Describe the pull request**

The goal of this PR is to get this port's CMake buildsystem to match the behaviour of the project's autoconf buildsystem. The autoconf build is careful to ensure that CPU intrinsics are only enabled for translation units that do runtime detection of available CPU features. This PR achieves this by separating the library into CMake object libraries that can each have separate compile flags. I also used CMake interface libraries to define groups of common compiler flags.

I also tried using some of the new `vcpkg_*` functions in the portfile and converted CONTROL file over to a manifest file. I was somewhat confused by the original `ports/libsodium/sodiumConfig.cmake.in` file, so I converted it over to match the one I saw on the sqlite3 port.

- What does your PR fix? Fixes #14134

I've tested `x64-windows` and `x64-linux`. I used this repo to test the consumption of the package: https://github.com/pathcore/vcpkg-libsodium-avx512

I hope so! I tried to modernize the port using some of the new features so it is possible I didn't follow all the new recommendations.


ports/libsodium/vcpkg.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "libsodium",
+  "version-string": "1.0.18",
+  "port-version": 4,
+  "description": "A modern and easy-to-use crypto library",
+  "homepage": "https://github.com/jedisct1/libsodium",
+  "dependencies": [
+    { "name": "vcpkg-cmake", "host": true },
+    { "name": "vcpkg-cmake-config", "host": true }
```
JackBoosY, line 9:
```suggestion
+    {
+      "name": "vcpkg-cmake",
+      "host": true
+    },
+    {
+      "name": "vcpkg-cmake-config",
+      "host": true
+    }
```

JackBoosY (approved):

JackBoosY:
Thanks for improve it!

thomasgt:
I found a tool to inspect an elf and print the x86 extensions used within [1]. I ran it on a binary we're generating internally to confirm that the x86 extensions are actually being used.
```
➜  ~/.cargo/bin/elfx86exts <redacted>
MODE64 (call)
SSE2 (pxor)
SSE1 (movaps)
CMOV (cmovs)
AVX (vmovdqa)
NOVLX (vpxor)
SSSE3 (pshufb)
SSE41 (pinsrb)
AVX2 (vinserti128)
CPU Generation: Unknown
```

I then took this binary and ran it on a machine that didn't support SSSE3, SSE41, and AVX2 and it still worked, so I'm pretty confident that this new build system lights up the x86 extensions in a portable way.

[1] https://github.com/pkgw/elfx86exts

thomasgt:
There was a legitimate error in the windows-static builds that I addressed last night with https://github.com/microsoft/vcpkg/pull/16730/commits/71fddf76cb153b35946bb026b8f1dadb3f778c38, but I'm also seeing some errors from the various builds along the lines of...
```
Convert-Path: /agent/_work/1/s/scripts/azure-pipelines/analyze-test-results.ps1:424
Line |
 424 |  … t_hash = build_test_results( Convert-Path "$logDir\$($triplet).xml" )
     |                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path '/agent/_work/1/a/xml-results/x64-linux.xml'
     | because it does not exist.

##[error]PowerShell exited with code '1'.
##[error]PowerShell wrote one or more lines to the standard error stream.
##[error]Convert-Path: /agent/_work/1/s/scripts/azure-pipelines/analyze-test-results.ps1:424
Line |
 424 |  … t_hash = build_test_results( Convert-Path "$logDir\$($triplet).xml" )
     |                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path '/agent/_work/1/a/xml-results/x64-linux.xml'
     | because it does not exist.
```

Is there a recommended way to reproduce this locally?

versions/l-/libsodium.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "b59d9a70ee8f259b7aa153109d26f9288cf5bacd",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "f4afd75ee5bc2d93aae58edb7800b2254ef07287",
```

JackBoosY (changes requested):

JackBoosY (approved):

ports/libsodium/CMakeLists.txt
```diff
@@ -48,58 +49,273 @@ configure_file(
...
-    ) 
-endif ()
-
-if (MSVC)
-    enable_language(RC)
-
-    list(APPEND sodium_sources
-        builds/msvc/resource.rc
-    )
-endif ()
```
PhoebeHui, line 74: I suggest to revert this change here and remove linux related changes, and then you can use the following code instead.

```
 target_sources(${PROJECT_NAME}
        PRIVATE
        ${sodium_sources}
        ${sodium_headers})
```

ports/libsodium/CMakeLists.txt
```diff
@@ -48,58 +49,273 @@ configure_file(
...
+            ${PROJECT_SOURCE_DIR}/src/libsodium/crypto_core/ed25519/ref10/fe_25_5/base.h
+            ${PROJECT_SOURCE_DIR}/src/libsodium/crypto_core/ed25519/ref10/fe_51/constants.h
+            ${PROJECT_SOURCE_DIR}/src/libsodium/crypto_core/ed25519/ref10/fe_51/fe.h
+            ${PROJECT_SOURCE_DIR}/src/libsodium/crypto_core/ed25519/ref10/fe_51/base2.h
+            ${PROJECT_SOURCE_DIR}/src/libsodium/crypto_core/ed25519/ref10/fe_51/base.h
+        
+            ${PROJECT_SOURCE_DIR}/builds/msvc/resource.h
+
+            ${PROJECT_SOURCE_DIR}/builds/msvc/resource.rc
+    )
```
PhoebeHui, line 317: If you check the source and header files, it includes all sources and headers, and what you removed above also do the same thing.

The disadvantage is more than 200 files listed here, if the Upstream add new files, we may miss the new changes in future.

PhoebeHui (changes requested):
Thanks for the PR!

ports/libsodium/CMakeLists.txt
thomasgt: If I use a GLOB_RECURSE to collect the source files on Linux and MacOS, I won't be able to differentiate between the files that need different build flags. Much of the complexity of the Linux/MacOS builds is related to the fact that only some of the files can have x86 extensions enabled during compilation.

ports/libsodium/CMakeLists.txt
thomasgt: In the case of the Windows builds, you are correct. I could use a GLOB_RECURSIVE to collect all the .c and .h files. I'm happy to implement that change, but there will still be a lot of differences between the Windows and Linux/MacOS builds.

thomasgt: The Linux/MacOS builds will still be somewhat tedious to update on a new upstream release, but I tried to structure the changes so that someone can follow along with the upstream's autoconf files. Given that the project requires different groups of objects be built with different flags, I can't think of any other approach.

PhoebeHui (approved):

@vicroms, could you help further review?

Issue #14890, opened 2020-12-02
eglext_angle.h not copied at install


mediabuff:
include\EGL\eglext_angle.h  should be copied to the install\... \include like eglext.h

NancyLi1013:
Hi @mediabuff 

Thanks for your posting this issue.

It seems that there is no eglext_angle.h in current `angle `version. We need to update angle to the latest to add this header.

mediabuff:
Also, angle_util.lib is not built - as this library is required to run angle samples

NancyLi1013:
Hi @robmikh @driver1998 @isanych 

Would you like to help update `angle `to the latest? I checked the latest source codes and found there were a lot of changes. Since I'm not familiar with this port. So could anyone help?

Thanks in advance.

robmikh:
Unfortunately I no longer use ANGLE for my cross-platform projects (I've moved onto [wgpu](https://github.com/gfx-rs/wgpu)), so I don't have a good way to validate a potential port update. I can't make promises about my schedule and when I may have time for this, but having a good way to validate changes would help me or whomever picks this up.

NancyLi1013:
@robmikh 

Thanks for your response. It doesn't matter even though you have no enough time to do this. Can you give some advice on the new version update? Maybe I can try to do some work.

hanseuljun:
I submitted a pull request that includes eglext_angle.h. The issue was caused by all header files with "EGL" in their name being excluded at the end of ports/angle/CMakeLists.txt.

Since I worked on this one before reading this issue, this PR does not solve the angle_util.lib issue or points at the up-to-date version of ANGLE.

hanseuljun:
This PR also does not let you call eglCreateDeviceANGLE directly. At least in my machine for x64-windows.

The way I currently work with it is through using eglGetProcAddress. I first call
PFNEGLCREATEDEVICEANGLEPROC createDeviceANGLE = (PFNEGLCREATEDEVICEANGLEPROC)eglGetProcAddress("eglCreateDeviceANGLE");
then use createDeviceANGLE instead of eglCreateDeviceANGLE.

Neumann-A:
@LilyWangLL Close. The header is copied according to the file list.

LilyWangLL:
Now `eglext_angle.h` has installed to include folder of `angle`, so I am closing this issue for now.

PR #16733, opened 2021-03-17
[angle] Include EGL/eglext_angle.h in ANGLE

hanseuljun:
**Describe the pull request**
Stops excluding eglext_angle.h from ANGLE. #14890 

eglext_angle.h includes entry points to angle-specific EGL extension functions and this PR exposes them.

I tested it on x64-windows and arm64-uwp.

I think so. I updated the version of the port at least.

### DIFF
--- a/ports/angle/CMakeLists.txt
+++ b/ports/angle/CMakeLists.txt
@@ -391,6 +391,8 @@ add_library(libEGL
     "src/libEGL/libEGL.cpp"
     "src/libEGL/libEGL.rc"
     "src/libEGL/resource.h"
+    "src/libEGL/egl_loader_autogen.cpp"
+    "src/libEGL/egl_loader_autogen.h"
 )
 target_link_libraries(libEGL PRIVATE angle::common angle::libANGLE libGLESv2)
 target_include_directories(libEGL PUBLIC "$<INSTALL_INTERFACE:include>" "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>")
@@ -469,7 +471,9 @@ if(NOT DISABLE_INSTALL_HEADERS)
         PATTERN "*.h"
         PATTERN "*.inc"
         PATTERN "GLSLANG" EXCLUDE
-        PATTERN "EGL" EXCLUDE
+        PATTERN "egl.h" EXCLUDE
+        PATTERN "eglext.h" EXCLUDE
+        PATTERN "eglplatform.h" EXCLUDE
         PATTERN "KHR" EXCLUDE
         PATTERN "export.h" EXCLUDE
     )
--- a/ports/angle/CONTROL
+++ b/ports/angle/CONTROL
@@ -1,5 +1,6 @@
 Source: angle
-Version: 2020-05-15-2
+Version: 2020-05-15
+Port-Version: 2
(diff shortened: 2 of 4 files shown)


ports/angle/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: angle
-Version: 2020-05-15-1
+Version: 2020-05-15-2
```
NancyLi1013, line 2:
```suggestion
Version: 2020-05-15
Port-Version: 2
```
Please use `Port-Version` to describe the changes for the port. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md#port-version

NancyLi1013 (changes requested):

NancyLi1013:
For the failure on x86-windows, please run `vcpkg x-add-version angle` and submit the changes to solve it.
```
Error: While reading versions for port angle from file: C:\a\1\s\versions\a-\angle.json
       Version `2020-05-15-2` was not found in versions file.
       Run:

           vcpkg x-add-version angle

       to add the new port version.
```
Note: Please update the changes first, then run the above command.

ports/angle/CMakeLists.txt
```diff
@@ -469,7 +471,9 @@ if(NOT DISABLE_INSTALL_HEADERS)
         PATTERN "*.h"
         PATTERN "*.inc"
         PATTERN "GLSLANG" EXCLUDE
-        PATTERN "EGL" EXCLUDE
+        PATTERN "egl.h" EXCLUDE
+        PATTERN "eglext.h" EXCLUDE
+        PATTERN "eglplatform.h" EXCLUDE
```
NancyLi1013, line 476: Why do you exclude these heads?

ports/angle/CONTROL
hanseuljun: I will!

ports/angle/CMakeLists.txt
hanseuljun: This is because those 3 header files are included in egl-registry, a port angle depends on in vcpkg. So, avoiding duplicate installations. This is handled in a different way for GLES by the last lines of angle/portfile.cmake. I think sticking to one method would be better but was too scared of breaking things to actually make that happen...

hanseuljun:
I ran vcpkg x-add-version angle but an error again... I guess it has to do with the previous version being 2020-05-15-1 not 2020-05-15...? Of course it's just a guess...

NancyLi1013:
Since you have updated versions, need to run `vcpkg x-add-version --overwrite-version angle` to overwrite the changes.

ports/angle/CMakeLists.txt
NancyLi1013: Thanks for your details. Will this affect the port of usage or other ports that depend on it?

hanseuljun: I don't think so. When I checked the angle/include directory installed by vcpkg to the google's repo, the only difference in terms of header files was eglext_angle.h.

hanseuljun:
Ran x-add-version with the overwrite-version flag. Some new hash was assigned so it seems promising!

NancyLi1013 (approved):

ras0219-msft:
LGTM, thanks for the fix!

PR #16742, opened 2021-03-17
[glad] Add features to support different configurations.

### LINKED ISSUE #10098: [glad] Support for building GLES specific headers should be added
Glad supports generating bindings for a multitude of specs, profiles, and apis and by default, it just builds `gl=latest, profile=compatibility`.
Support should also be added for generating the `gles=latest` APIs and for targeting the `core` profile.

**Proposed solution**
Glad should be updated to have a `gles` feature.
I'm unsure how you would expose functionality such that `gl=latest` could be avoided (generating only `gles` is valid) or how the profile could be toggled.
Namely because `vcpkg install` doesn't look to have a mechanism for specifying options like that.

Create a separate port specific to GLES.

**Additional context**
Happy to implement the feature if we can come to an agreement about how it should be done.

### LINKED ISSUE #16027: [glad] Allow building with support for extensions.
As also stated in issue #10098, glad allows for a variety of customization. However, the current port only uses the default settings. This greatly limits is use besides the core functionality. I encountered the problem that some defines and functions are simply missing, when trying to use GL extensions. 

In OpenGL extensions are not mandatory and should be detected at runtime. glad already implements such functionality, but the extensions it checks are configurable. This way it can be made a little bit more lightweight by omitting unused checks when initializing it.

**Proposed solution**

Include all extensions by default. This has the downside of increasing the static library and header file sizes and also introduces a minor one-time runtime cost when initializing the glad loader. However, this is the most simple way of making GL extensions through glad accessible using vcpkg. 

**Alternatives**

Using a custom glad configuration, when GL extensions are required. However, this would defeat the purpose of using a package manager in my opinion. Also I could write a custom loader for the extensions only, but then I could simply drop glad straight away.

A more advanced alternative would be to make extensions configurable, similar to what is proposed in the issue linked above (it regards the configuration profile). I do think, however, that this is overkill, since one has to write code that checks if an extension is available on a given hardware in the first place anyway. 

**Additional context**

[This comment](https://github.com/microsoft/vcpkg/issues/10098#issuecomment-607204171_) by @MrROBUST also describes the same problem, however I think it deserves its own issue, since the proposed solution can be implemented independently from the issue.

crud89:
This PR improves the glad port by updating to the most recent version and adding functionality. Furthermore it addresses the issues #16027 and #10098. Overall, it improves the port in multiple aspects:

- It bumps the used version to *0.1.34*.
- It replaces the control file with a proper vcpkg.json manifest file.
- The features `wgl`, `glx` and `egl` can be used to control which specifications glad loads (see [this PR](https://github.com/Dav1dde/glad/pull/316)). It is possible to combine multiple specifications, for example by `vcpkg install glad[egl,wgl] --recurse`. Note that the `gl` specification is always required and implicitly generated by glad (hence there is no feature for it).
- By specifying the `extensions` feature, it can automatically import all extensions (resolves #16027). The default behavior still omits the extensions, as it is done currently.
- It is possible to specify the API version (resolves #10098). The default behavior is to not import extensions from a certain API version. It can be explicitly set to `-latest` for each API by `vcpkg install glad[extensions,gl-api-latest,gles1-api-latest,gles2-api-latest,glsc2-api-latest] --recurse`. <s>Furthermore, the `core-profile` feature will disable the compatibility profile and will make glad use the core profile instead.</s> <sup>[1](https://github.com/microsoft/vcpkg/pull/16742#discussion_r597378211)</sup>
- The generation of the loader code can be disabled using the `no-loader` feature.

Overall this PR makes it possible to use almost all configuration capabilities glad provides, except two:
- It is not possible to omit the KHR header (which could break the build and is most certainly not wanted).
- It is not possible to specify individual extensions. The `extensions` feature imports all available extensions if specified. It could be argued that this can be enabled by default, however I decided against it to not deviate from the current default behavior.


Aschratt:
Windows builds appear to fail in CI, because CMake is unable to find Python 3. Not sure what to do about this. 🤔

ports/glad/portfile.cmake
```diff
@@ -3,16 +3,117 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
...
+    set(DISABLE_LOADER ON)
+else()
+    set(DISABLE_LOADER OFF)
+endif()
+
+if ("extensions" IN_LIST FEATURES)
+    set(WITH_EXTENSIONS ON)
+else()
+    set(WITH_EXTENSIONS OFF)
+endif()
```
PhoebeHui, line 42: Could you please use [vcpkg_check_features](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_check_features.md) instead?

ports/glad/portfile.cmake
```diff
@@ -3,16 +3,117 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
...
+    SHA512 f6a8ba7d0d09b89c23b6f76962d3e6eef1babc8e1a659e238d30e143eb33ccba424957e5a6d46d99a714bfa2967523b193586d0ff24e29ad8d86c92c9faf9c02
     HEAD_REF master
     PATCHES encoding.patch
 )
 
+if("core-profile" IN_LIST FEATURES)
+    set(GLAD_PROFILE "core")
+else()
+    set(GLAD_PROFILE "compatibility")
+endif()
```
PhoebeHui, line 16: This breaks the policy https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#features

PhoebeHui (changes requested):
For the python issue, it may need to patch the https://github.com/Dav1dde/glad/blob/master/CMakeLists.txt#L27 to use the one provided by vcpkg.

See https://github.com/microsoft/vcpkg/blob/master/ports/python3/usage

PhoebeHui:
@ras0219-msft, could you please help further review?

Aschratt:
Thank you for reviewing @PhoebeHui! 

> For the python issue, it may need to patch the https://github.com/Dav1dde/glad/blob/master/CMakeLists.txt#L27 to use the one provided by vcpkg.
> 
> See https://github.com/microsoft/vcpkg/blob/master/ports/python3/usage

You are right. The portfile already sets the `PYTHON_EXECUTABLE` variable, however glad tries to overwrite it. I will try to add a patch to remove lines 26-35 from the CMakeLists.txt, since they are not required when building from vcpkg.

```cmake
# Find the python interpreter, set the PYTHON_EXECUTABLE variable
if (CMAKE_VERSION VERSION_LESS 3.12)
    # this logic is deprecated in CMake after 3.12
    find_package(PythonInterp REQUIRED)
else()
    # the new hotness.  This will preferentially find Python3 instead of Python2
    find_package(Python)
    set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
endif()
```

I will also take a look at the other changes you requested and update the PR accordingly. Regarding the `core-profile`, I will output a message to use overlay ports if required, as suggested in the document you've linked.

ports/glad/portfile.cmake
Aschratt: Resolved in b770ded0b32f49b1ad2ed0afd61ee0b0fdbb1a27.

ports/glad/portfile.cmake
Aschratt: Resolved in e7359d3ad3ae6631e5faedd331d2883161c3f28f.

Aschratt:
I've added a patch to remove the python detection from the *CMakeLists.txt* file, so the CI should not fail under windows anymore (See d56840aa8fbf4285875abc8974eccfdd5eab554e).

Also I've added dependencies to `vcpkg-cmake` and `vcpkg-cmake-config` ports and changed the portfile accordingly in 662e0d5016bad51a9bc55f57df9b28e2c600bbfa in order to resolve deprecation warnings for `vcpkg_cmake_*` commands.

ports/glad/vcpkg.json
```diff
@@ -0,0 +1,125 @@
...
+      "description": "Imports extensions from OpenGL ES 2 specification version 3.0."
+    },
+    "gles2-api-31": {
+      "description": "Imports extensions from OpenGL ES 2 specification version 3.1."
+    },
+    "gles2-api-32": {
+      "description": "Imports extensions from OpenGL ES 2 specification version 3.2."
+    },
+    "gles2-api-latest": {
+      "description": "Imports extensions from latest OpenGL ES 2 specification."
```
ras0219-msft, line 107:
```suggestion
      "description": "Imports extensions from latest OpenGL ES 2 specification.",
      "dependencies": [ { "name": "glad", "features": ["gles-api-32"], "default-features": false } ]
```

ports/glad/vcpkg.json
```diff
@@ -0,0 +1,125 @@
+{
+  "name": "glad",
+  "version-string": "0.1.34",
+  "port-version": 2,
```
ras0219-msft, line 4:
```suggestion
```
Since this is a source update, we conventionally reset the port version back to 0

ports/glad/vcpkg.json
```diff
@@ -0,0 +1,125 @@
+{
+  "name": "glad",
+  "version-string": "0.1.34",
```
ras0219-msft, line 3:
```suggestion
  "version": "0.1.34",
```
Using a more intelligent version scheme enables our new versioning feature to work better

ras0219-msft (changes requested):
Overall, looks great! Thanks for the PR!

I've made a few suggestions that should be addressed before merging.

ports/glad/portfile.cmake
```diff
@@ -3,41 +3,140 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
...
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        no-loader   GLAD_NO_LOADER
+        extensions  GLAD_ALL_EXTENSIONS
+)
+
+set(GLAD_SPEC "gl")
+
+if("egl" IN_LIST FEATURES)
+    set(GLAD_SPEC "${GLAD_SPEC},egl")
```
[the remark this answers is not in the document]
Aschratt: Resolved in d89b96b83f5b74c59435efa69b216d177b2e4e40.

ports/glad/vcpkg.json
```diff
@@ -0,0 +1,125 @@
...
+    },
+    "glsc2-api-latest": {
+      "description": "Imports extensions from latest OpenGL SC API specification."
+    },
+    "glx": {
+      "description": "Use `glx` spec instead of `gl`."
+    },
+    "no-loader": {
+      "description": "Disable generation of loader logic."
+    },
```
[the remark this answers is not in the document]
Aschratt: Resolved in 8c11a11fd5f55aff05af6881b25e692a10ee0e7c.

I've also added `loader` to the list of default features, since this glad also generates it by default.

ports/glad/vcpkg.json
Aschratt: Resolved with a8ab1fc9f0182fa5a37f66603db75d0b5d4e55aa and 4585c6dc64204440c6593829c5d3acc6f1e2a5e1.

Aschratt:
@JackBoosY Since you've re-added <kbd>author-response</kbd>, did I miss something?

JackBoosY (approved):

JackBoosY:
Nope, waiting for test all features.

PhoebeHui:
@Aschratt, appreciate your effort very much!

Test feature glad[glx]:x64-windows, and it failed with following failure, could you please take a look?

E:\vcpkg\16742\vcpkg\buildtrees\glad\x64-windows-dbg\include\glad/glad_glx.h(24): fatal error C1083: Cannot open include file: 'X11/X.h': No such file or directory

Aschratt:
@PhoebeHui thanks for the review! *glx* adds support for the X Window System, which is not available under Windows and UWP (afaik). Similar, *wgl* should not work under Linux/Mac. Is it possible to make this explicit somehow?

PhoebeHui:
@Aschratt, thanks for confirming the issue! Could you add that information to feature description?

Aschratt:
@PhoebeHui done. 🙂 Let me know if there's anything else that could be done!

PhoebeHui (approved):

vicroms:
Thanks for PR!

PR #16758, opened 2021-03-18
[docs] Rename docs/index.md to docs/README.md

autoantwort:
When I search for documentation, I open the docs folder in github, but then I don't see the topics of the help pages. I first have to open the `index.md` file. With this change I directly see the content of the `index.md` file without opening it.
See [here](https://github.com/autoantwort/vcpkg/tree/rename-to-README/docs) for reference.


PhoebeHui (approved):
LGTM, thanks for the PR!

ras0219-msft:
We name the file `index.md` for compatibility with `MkDocs` (https://www.mkdocs.org/); to rename this we would need to also add a mkdocs config file.

autoantwort:
> We name the file `index.md` for compatibility with `MkDocs` (https://www.mkdocs.org/); to rename this we would need to also add a mkdocs config file.

I don't think so. From [here](https://www.mkdocs.org/user-guide/writing-your-docs/#index-pages):
> Many repository hosting sites provide special treatment for README files by displaying the contents of the README file when browsing the contents of a directory. Therefore, MkDocs will allow you to name your index pages as README.md instead of index.md. In that way, when users are browsing your source code, the repository host can display the index page of that directory as it is a README file. However, when MkDocs renders your site, the file will be renamed to index.html so that the server will serve it as a proper index file.
>
> If both an index.md file and a README.md file are found in the same directory, then the index.md file is used and the README.md file is ignored.

If I understand correctly, you can name the file `index.md` or `README.md`

ras0219-msft:
Awesome, just a reading comprehension fail on my part. Thanks!

PR #16583, opened 2021-03-07
[QtWebEngine] - Fix build failure

heydojo:
This commit fixes an issue where the build fails on X86_64 Windows 10 Pro with a message similar to: 'PATCHES no such argument' printed to the console output.

It also makes a few small changes for sanity. Build trees for QtWebEngine come in at between 40GB-50GB and the minimum recommended RAM size for building Chromium according to google is 8GB. The port shouldn't be building debug by default due to it's size and the fact that you likely only want debug if you are a Qt developer or are working on vcpkg. The options "" is intentionally left blank due to the fact that it allows the string `-webengine-proprietary-codecs` to be easily be added in case you have licences for that kind of thing. Without that option enabled, many html5 videos do not work in QtWebEngine. Maybe adding a configuration switch for proprietary-codecs is something to think about for the future?

Enjoy!

From the build output here:

`Running configuration tests...
Checking for architecture supported... yes
Checking for bison... H:/vcpkg/installed/x64-windows/tools/qt5/mkspecs/features/qt_configure.prf:413: Unexpected return value from test 'ensurePathEnv': H:/vcpkg/buildtrees/qt5-webengine/src/gnuwin32/bin/bison.exe.
yes
Checking for submodule initialized... yes
Checking for build path without whitespace... yes
Checking for platform supported... 
Checking for gperf... H:/vcpkg/installed/x64-windows/tools/qt5/mkspecs/features/qt_configure.prf:413: Unexpected return value from test 'ensurePathEnv': H:/vcpkg/buildtrees/qt5-webengine/src/gnuwin32/bin/gperf.exe.
yes
Checking for flex... H:/vcpkg/installed/x64-windows/tools/qt5/mkspecs/features/qt_configure.prf:413: Unexpected return value from test 'ensurePathEnv': H:/vcpkg/buildtrees/qt5-webengine/src/gnuwin32/bin/flex.exe.
yes
Checking for python2... H:/vcpkg/downloads/tools/python/python-2.7.16-x64/python.exe
Checking for 64bit compiler... yes
Checking for winversion... yes
Checking for jumbo build merge limit... 0
Checking for d-bus... no
Checking for fontconfig... no
Checking for libdrm... no
Checking for system ninja... yes
Checking for xcomposite... no
Checking for xcursor... no
Checking for xi... no
Checking for xtst... no
Checking for embedded build... no
Checking for node.js... no
Done running configuration tests.

Configure summary:

Qt WebEngine Build Tools:
  Use System Ninja ....................... yes
[...]

PR #16770, opened 2021-03-18
[qt5-webengine] Fix issues

Neumann-A:
changes from #16583

 - add feature "proprietary-codecs"
 - fix build error due to internal changes how options are passed.
 - try to fail early on too long build paths instead of late

- [x] wait for CI
- [x] Run `vcpkg x-add-version --all`


ports/qt5-webengine/portfile.cmake
```diff
@@ -1,7 +1,13 @@
...
-#set(VCPKG_BUILD_TYPE release) #You probably want to set this to reduce build type and space requirements
-message(STATUS "${PORT} requires a lot of free disk space (>300GB), ram (>32 GB) and time (>4h per configuration) to be successfully build.\n\
--- As such ${PORT} is not properly tested.\n\
+string(LENGTH "${CURRENT_BUILDTREES_DIR}" buildtrees_path_length)
+if(buildtrees_path_length GREATER 15 AND CMAKE_HOST_WIN32)
+        message(FATAL_ERROR "${PORT}'s buildsystem uses very long paths and will fail on your system.\n"
+            "We recommend moving vcpkg to a short path such as 'C:\\src\\vcpkg' or using the subst command."
+        )
+endif()
+set(VCPKG_BUILD_TYPE release) #You probably want to set this to reduce build type and space requirements
```
Neumann-A, line 8:
```suggestion
#set(VCPKG_BUILD_TYPE release) #You probably want to set this to reduce build type and space requirements
```

scripts/ci.baseline.txt
```diff
@@ -1321,7 +1321,7 @@ qt5-webengine:x64-osx=fail
 # D:\buildtrees\qt5-webengine\x64-windows-dbg\src\core\debug\gen\net/third_party/quiche/src/quic/core/proto/cached_network_parameters.pb.h(17):
 # fatal error C1189: #error:  This file was generated by an older version of protoc which is
 # Succesful built requires protobuf to be installed after qt5-webengine not before. Otherwise the build picks up the wrong headers from inside vcpkg.
-qt5-webengine:x64-windows=skip
+# qt5-webengine:x64-windows=skip
```
Neumann-A, line 1324:
```suggestion
qt5-webengine:x64-windows=skip
```

JackBoosY:
Seems CI passed.

ports/qt5-webengine/portfile.cmake
```diff
@@ -1,7 +1,13 @@
 vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
+string(LENGTH "${CURRENT_BUILDTREES_DIR}" buildtrees_path_length)
+if(buildtrees_path_length GREATER 35 AND CMAKE_HOST_WIN32)
+        message(FATAL_ERROR "${PORT}'s buildsystem uses very long paths and will fail on your system.\n"
+            "We recommend moving vcpkg to a short path such as 'C:\\vcpkg' or using the subst command."
+        )
+endif()
```
ras0219-msft, line 7:
```suggestion
vcpkg_buildpath_length_warning(35)
```

ras0219-msft (changes requested):
Thanks for the PR! LGTM once we use the official helper.

If you would like it to be a fatal error, we should rely on the official helper to provide the message and then emit the fatal error afterwards.

ports/qt5-webengine/portfile.cmake
Neumann-A: The helper is used inside of `qt_submodule_installation`

Neumann-A: What I need here is a `vcpkg_buildpath_length_error(35)` and not a warning. Building 2h and getting a long filepath error (file not found although it is there) is not really a good user experience. 
That is why I added it here explicitly.

ras0219-msft: Should qt_submodule_installation be modified to terminate then?

Neumann-A: > Should qt_submodule_installation be modified to terminate then?

Most qt modules are fine on longer paths and I think nobody actually knows the correct number to put into the warning. 
In Qt6 for example I only know of qtquick3d having this issue and it builds rather fast compared to qt5-webengine.

JackBoosY:
Depends on #16904

JackBoosY (approved):

PR #16774, opened 2021-03-18
[ci,vcpkg-cmake-*] remove host-only nature of vcpkg-cmake-* ports

strega-nil:


ports/vcpkg-cmake-config/portfile.cmake
```diff
@@ -1,6 +1,9 @@
+#[===[
+# uncomment when issue #16773 fixed
 if(NOT TARGET_TRIPLET STREQUAL _HOST_TRIPLET)
     message(FATAL_ERROR "vcpkg-cmake-config is a host-only port; please mark it as a host port in your dependencies.")
```
ras0219-msft, line 4: I think we should keep this but as a warning (to help new port authors). In the future, we can then promote it to error-only-in-ci.

ras0219-msft (approved):

ras0219-msft:
Awesome, thanks for quickly getting this fixed ❤️

Issue #16658, opened 2021-03-11
shaderc missing shared library


jmoguillansky-gpsw:
**Describe the bug**
As a follow up to this issue: https://github.com/microsoft/vcpkg/issues/16122, seems that shaderc shared library is no longer built?

**Environment**
- OS: Windows
- Compiler: MSVC 2019

**To Reproduce**
1. ./vcpkg install shaderc:x64-windows
2. See error
-- Note: shaderc only supports static library linkage. Building static library."

**Expected behavior**
Shaderc vcpkg used to support both static and shared lib build, now it seems that only static lib is supported?

**Failure logs** 
-(please attached failure logs)

**Additional context**

NancyLi1013:
Hi @jmoguillansky-gpsw 

Thanks for posting this issue.
Currently `shaderc` only supports static build. Please see 

```
#Note: glslang and spir tools doesn't export symbol and need to be build as static lib for cmake to work
vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
https://github.com/microsoft/vcpkg/blob/master/ports/shaderc/portfile.cmake#L1-L2

Generally, vcpkg supports both static and dynamic builds. But for some ports, they only support static or dynamic build, which is related with the ports themselves.

So the current behavior is expected.

jmoguillansky-gpsw:
Hi, I saw that comment.  Shaderc upstream supports shared library build.  Its not a shaderc issue.  In the most recent commit vcpkg disabled shaderc shared library build.  Its a vcpkg issue

dg0yt:
This should be revisited. Probably it never was a problem for non-MSVC, and now there is also dllexport:
https://github.com/google/shaderc/blob/aaa44b544909600381e1a180074ed7f544e48410/libshaderc/include/shaderc/visibility.h#L21-L25

PR #16781, opened 2021-03-19
[shaderc] Re-fix installation

JackBoosY:
- Enable dynamic build and re-fix the dependency relationship.
- Remove the util installation patch since we don't need it (added in https://github.com/microsoft/vcpkg/pull/18979).
- Disable build sample.
- Re-factory portfile.cmake.

Related: #16137.


JackBoosY:
@jmoguillansky-gpsw  Can you please test this PR?

Thanks.

NancyLi1013:
Related issue #16658

NancyLi1013 (approved):

ras0219-msft (changes requested):
Please see https://github.com/microsoft/vcpkg/pull/16137#issuecomment-802952196

jmoguillansky-gpsw:
> @jmoguillansky-gpsw Can you please test this PR?
> 
> Thanks.

Hi @JackBoosY I tested your branch, I was able to build shaderc shared lib successfully.
Thanks, much appreciated!

JackBoosY:
@ras0219-msft glslang and spir-tools use some symbols that are not exported in shaderc, so if we need to build them, we must first build static shaderc.

ras0219-msft:
>glslang and spir-tools use some symbols that are not exported in shaderc

shaderc _depends_ on glslang and spirv-tools; therefore they cannot be depending on symbols exported by shaderc:

```
# ports/control/shaderc

Build-Depends: glslang, spirv-tools
```

Therefore, if glslang and spirv-tools can only be built statically, shaderc should also be built statically. Dynamic libraries should (generally) not depend on static libraries.

jmoguillansky-gpsw:
> The reason I need to use shaderc shared library is because visual studio toolchain doesn't let you mix /MT (multithreaded) and /MD (multithreaded DLL) so if I have a shared library that links against shaderc, it needs to link against the shared lib of shaderc if I understand correctly?

I see there's an option: "DSHADERC_ENABLE_SHARED_CRT", maybe it's possible to use the shaderc static lib if I set this option?
But I don't think vcpkg exposes this option, maybe I need to build shaderc manually from source?

CaptainHorse:
Any updates on this? Would love to use shaderc thru vcpkg as it has issues with path size limits in my project.

JackBoosY:
@CaptainHorse  Will continue to handle this PR later.

versions/s-/shaderc.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "acddbab84ab19cc4d1e04ecb5277d3969cf9b093",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "f6a82ca83afbf20c08d6d001f133cd5fd2672108",
```

JackBoosY:
Requires https://github.com/microsoft/vcpkg/pull/19219.

JackBoosY:
Linux regression related to https://github.com/KhronosGroup/SPIRV-Tools/issues/1569.

PhoebeHui:
The related upstream issue on linux has been fixed, please reopen and continue this PR if you have time.

PR #16804, opened 2021-03-20
[blend2d] Update to beta_2021-03-17

ZeeWanderer:
Updated blend2d to beta_2021-03-17. Updated local asmjit to 2021-03-19.
versioning scheme changed from `version-string` to `version-date`
Fixed facepalm `BLEND2D_BUILD_NO_JIT` usage in portfile. Probably forgot to rename after blend2d changed variable naming to `BLEND2D_NO_JIT`.


JonLiu1993 (approved):

ports/blend2d/portfile.cmake
```diff
@@ -21,8 +21,8 @@ if(NOT BLEND2D_BUILD_NO_JIT)
   vcpkg_from_github(
     OUT_SOURCE_PATH ASMJIT_SOURCE_PATH
```
ras0219-msft, line 22: Noting here that asmjit is available as a vcpkg port: https://github.com/microsoft/vcpkg/blob/master/ports/asmjit/CONTROL

By policy we would _really_ like this to stop being embedded here, however that won't block this update PR.

ports/blend2d/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blend2d",
-  "version-string": "beta_2021-01-02",
+  "version-string": "beta_2021-03-17",
```
ras0219-msft, line 3:
```suggestion
  "version-date": "2021-03-17",
```
It looks like this is a rolling release, so we should use a date scheme

versions/b-/blend2d.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "a73748cae8a849c48f8a749e2ff9b3c85a572d43",
+      "version-string": "beta_2021-03-17",
```
ras0219-msft, line 5:
```suggestion
      "version-date": "2021-03-17",
```

ras0219-msft (changes requested):
LGTM after the version scheme change, thanks for the PR!

ports/blend2d/portfile.cmake
ZeeWanderer: blend2d depends on asmjit CMakeLists directly. I could try to patch it but that would complicate this port by a lot.
https://github.com/blend2d/blend2d/blob/master/CMakeLists.txt#L373

ras0219-msft:
Thanks!

PR #16806, opened 2021-03-20
[sail] Update 0.9.0-pre11 -> 0.9.0-pre12

HappySeaFox:
**Describe the pull request**

Numerous improvements, see https://github.com/smoked-herring/sail/releases/tag/v0.9.0-pre12

Everything except UWP.  CI baseline not updated.

Yes.

**The port has been tested locally.**


JackBoosY:
Already test usage.

JackBoosY (approved):

ports/sail/portfile.cmake
```diff
@@ -42,5 +42,14 @@ else()
...
 endif()
 
+# Move C++ configs
+file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/share/${PORT}c++)
+
+file(GLOB SAIL_CPP_CONFIGS "${CURRENT_PACKAGES_DIR}/share/${PORT}/SailC++*")
+foreach(SAIL_CPP_CONFIG IN LISTS SAIL_CPP_CONFIGS)
+    get_filename_component(SAIL_CPP_CONFIG_NAME "${SAIL_CPP_CONFIG}" NAME)
+    file(RENAME ${SAIL_CPP_CONFIG} ${CURRENT_PACKAGES_DIR}/share/${PORT}c++/${SAIL_CPP_CONFIG_NAME})
+endforeach()
```
ras0219-msft, line 52: Could this simply be
```
vcpkg_fixup_cmake_targets(TARGET_PATH share/${PORT}c++)
```
?

ports/sail/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sail",
-  "version-string": "0.9.0-pre11",
+  "version-string": "0.9.0-pre12",
```
ras0219-msft, line 3:
```suggestion
  "version-semver": "0.9.0-pre12",
```

ras0219-msft (commented):
Thanks for the PR!

ports/sail/portfile.cmake
smoked-herring: Unfortunately `vcpkg_fixup_cmake_targets` doesn't help with this use-case. I need to move *some* cmake configs (C++-specific only), not all.

BTW, it would be great to have a function to do that. Like when a port stores its cmake configs for multiple targets in the same directory and we need to split them per-target.

versions/s-/sail.json
```diff
@@ -1,5 +1,15 @@
...
+    {
+      "git-tree": "5494cae319135c06b4ce8d6150c90e148fb6c924",
+      "version-semver": "0.9.0-pre12",
+      "port-version": 1
+    },
+    {
+      "git-tree": "2df563059b9c2c71ecb2ac1356b5756ac1ce864a",
+      "version-semver": "0.9.0-pre12",
+      "port-version": 0
+    },
```
ras0219-msft, line 12:
```suggestion
```
Each PR must only add up to one version of a library (we squash when merging, so all other version will not be available).

ras0219-msft (changes requested):

smoked-herring:
@ras0219-msft Thanks for the review!

ras0219-msft:
Thanks again!

PR #16822, opened 2021-03-22
[bext-ut] create a new port

### LINKED ISSUE #16691: [New Port Request] ut
Library name: ut

Library description: UT: C++20 μ(micro)/Unit Testing Framework

Source repository URL: https://github.com/boost-ext/ut

In the READNE of this repository:

[Optional] CMake integration

This project provides a CMake config and target. Just load ut with find_package to import the boost::ut target. Linking against this target will add the necessary include directory for the single header file. This is demonstrated in the following example.
```cmake
find_package(ut REQUIRED)
add_library(my_test my_test.cpp)
target_link_libraries(my_test PRIVATE boost::ut)
```

luncliff:
### What does your PR fix?

Fixes #16691

This is a header-only library, so every triplet should be available.

Wish I worked correctly :)

#### Naming
* See #16302. This port is named `bext-ut`


luncliff:
The error looks trivial, requesting for `/azp run`. :D

ports/ut/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "ut",
```
ras0219-msft, line 2:
```suggestion
  "name": "bext-ut",
```
For libraries from `boost-ext`, we are using the `bext-` prefix: https://github.com/microsoft/vcpkg/pull/16302

ras0219-msft (changes requested):
`ut` is _very_ short, so we'd like to avoid using such a name unless it is an extremely wide-spread association. We've named `boost-ext/di` `bext-di` in another issue, so I believe that's the right way to name this library as well.

Thanks for the PR!

ports/ut/vcpkg.json
luncliff: Applied the feedback in 745ff1f, ed56237. Thanks for the point!

JackBoosY (approved):

ports/bext-ut/portfile.cmake
```diff
@@ -0,0 +1,29 @@
...
+    SOURCE_PATH ${SOURCE_PATH}
+    PREFER_NINJA
+    OPTIONS
+        -DBOOST_UT_BUILD_BENCHMARKS=OFF
+        -DBOOST_UT_BUILD_EXAMPLES=OFF
+        -DBOOST_UT_BUILD_TESTS=OFF
+)
+vcpkg_install_cmake()
+vcpkg_copy_pdbs()
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ut)
```
ras0219-msft, line 19:
```suggestion
vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ut TARGET_PATH share/ut)
```
I assume that the expected call is `find_package(ut CONFIG REQUIRED)`? This will currently create `share/bext-ut/...`.

ras0219-msft (changes requested):

ports/bext-ut/portfile.cmake
JackBoosY:
```
bext-ut:/include/boost/ut.hpp
bext-ut:/share/bext-ut/copyright
bext-ut:/share/bext-ut/ut-config.cmake
bext-ut:/share/bext-ut/vcpkg_abi_info.txt
```

ras0219-msft (approved):
Thanks!

vicroms (approved):

ras0219-msft (approved):

JackBoosY:
Ready to merge.

PR #16036, opened 2021-02-04
[cairo] Update to 1.17.2

NancyLi1013:
**Describe the pull request**

- What does your PR fix? Fixes #
https://github.com/microsoft/vcpkg/pull/5646#issuecomment-772316916

Update to latest stable version 1.17.2

Note: All features have passed with x86-windows.

PR #16825, opened 2021-03-22
[cairo,cairomm] add fontconfig and quartz feature and update cairo version

abique:


JackBoosY (changes requested):

abique:
@JackBoosY Thank you very much for your feedback. I've updated the pull request accordingly.

JackBoosY:
Please get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=50919&view=artifacts&pathAsName=false&type=publishedArtifacts).

abique:
To me it looks line a problem in the other packages.

abique:
I've made fontconfig a feature, that way it can be disabled.

abique:
> Also please run command `vcpkg x-add-version --all` then commit the changes.

Thank you for that, this is now addressed.

JackBoosY:
Please get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=51088&view=artifacts&pathAsName=false&type=publishedArtifacts).
The osgearth regression is not related to this PR.

Waiting for #16904 merged.

abique:
I had to rework cairomm because it was trying to compile quartz on Linux and Windows which obviously will fail.

JackBoosY:
Get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=51195&view=artifacts&pathAsName=false&type=publishedArtifacts).

ports/cairo/vcpkg.json
```diff
@@ -13,10 +14,15 @@
     "freetype"
   ],
   "features": {
+    "fontconfig": {
+      "description": "build with fontconfig",
+      "dependencies": [
+        "fontconfig"
+      ]
+    },
```
Hoikas, line 22: This should be added to the default-features, otherwise users will find their feature set changing unexpectedly.

abique: We want fontconfig only on Linux or if x11 is enabled right?
There is no point in enabling fontconfig on windows or osx

JackBoosY: Therefore, it should be:
```json
        {
          "name": "fontconfig",
          "platform": "linux"
        }
```

abique: Thank you, I've done that.

abique: Actually this is illegal to put such a thing in default features:
```./vcpkg format-manifest ports/cairo/vcpkg.json 
Failed to parse manifest file: /home/abique/develop/vcpkg/ports/cairo/vcpkg.json
Errors occurred while parsing /home/abique/develop/vcpkg/ports/cairo/vcpkg.json
    $.default-features[1]: mismatched type: expected an identifier
```

JackBoosY:
Please use command `./vcpkg format-manifest --all` then commit changes.

ports/cairo/vcpkg.json
abique: I've made fontconfig a default feature as you requested.

ports/cairomm/vcpkg.json
```diff
@@ -0,0 +1,78 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg/master/scripts/vcpkg.schema.json",
+  "name": "cairomm",
+  "version-string": "1.16.0",
+  "port-version": 6,
```
JackBoosY, line 5:
```suggestion
```

Since this port is updating, please remove this.

ports/cairo/vcpkg.json
```diff
@@ -10,13 +11,19 @@
...
   ],
   "default-features": [
+    "fontconfig",
     "freetype"
   ],
   "features": {
+    "fontconfig": {
+      "description": "build with fontconfig",
+      "dependencies": [
+        "fontconfig"
```
JackBoosY, line 21:
```suggestion
        {
          "name": "fontconfig",
          "platform": "linux"
        }
```

JackBoosY (changes requested):

abique:
LGTM!

JackBoosY (approved):

JackBoosY:
I hope my colleague will do further review.

Neumann-A:
this PR collides with #13100. Also could have switched the build to the native buildsystem via `vcpkg_configure_make` which will be required anyway.

abique:
> this PR collides with #13100. Also could have switched the build to the native buildsystem via `vcpkg_configure_make` which will be required anyway.
#13100 is massive, and it does not address the same issues. While this one has a minimum amount of changes and brings cairo into a better shape. You now can have freetype without using fontconfig, and it provide better support for quartz on mac.

I'd merge #16825 first.

Neumann-A:
Yeah but the question is if the native buildsystem supports the new features introduced here. If not this PR has to remove the features added here.

abique:
Of course you can disable fontconfig when using the configure script.

abique:
I've bumped cairo to 17.4 because 1.16.0 had a bunch of known bugs.

Hoikas:
A previous cairo update PR was shelved because 1.17.x are snapshots. https://github.com/microsoft/vcpkg/pull/16036#discussion_r570077146

abique:
I have a crash in cairo, which I'll confirm soon if it is fixed by 1.17.4.
I believe 1.16.0 to be less stable than 1.17.4 despite the snapshot denomination.

abique:
To be precise we're running into https://gitlab.freedesktop.org/cairo/cairo/-/issues/412

JackBoosY:
Should we apply [this patch](https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/44/diffs)?

abique:
After some testing we figured that even 1.17.4 did not fix our issues on macos.
We are going to check if switching from static linking to dynamic linking fixes it.
Sorry for the noise, I'll come back with the conclusion.
Through, archlinux is using 1.17.4 and if that version was broken we'd know it. I've checked the changelog and basically the big change is the new backend cogl. Appart from that it is mostly fixes.

JackBoosY:
> After some testing we figured that even 1.17.4 did not fix our issues on macos.
> We are going to check if switching from static linking to dynamic linking fixes it.
> Sorry for the noise, I'll come back with the conclusion.
> Through, archlinux is using 1.17.4 and if that version was broken we'd know it. I've checked the changelog and basically the big change is the new backend cogl. Appart from that it is mostly fixes.

So should we wait for upstream fix this issue completely?

abique:
> 
> 
> > After some testing we figured that even 1.17.4 did not fix our issues on macos.
> > We are going to check if switching from static linking to dynamic linking fixes it.
> > Sorry for the noise, I'll come back with the conclusion.
> > Through, archlinux is using 1.17.4 and if that version was broken we'd know it. I've checked the changelog and basically the big change is the new backend cogl. Appart from that it is mostly fixes.
> 
> So should we wait for upstream fix this issue completely?

We don't understand why, but switching from static build to dynamic build seems to fix it.
We'll get back with more info.

abique:
We've been using 1.17.4 for some time now on Windows and Macos and it works like a charm. Plus Archlinux is using this version since a lot of time.

I think it is safe, and the right move to jump to this version.

abique:
As you don't feel excited about cairo-1.17.4 I'll revert to 1.16.0 so at least our work is not lost.

abique:
@PhoebeHui Hi,
Shall we merge it?

JackBoosY:
Ping @strega-nil for review and merge this PR.

strega-nil:
Thanks @abique :)

PR #16833, opened 2021-03-22
[vcpkg-gn wip] scripts audit

strega-nil:
This change makes v8 use the new vcpkg-gn scripts


Neumann-A:
this is also in need to generate build flags for gn from the cmake toolchain:

PR #16842, opened 2021-03-23
[workflow] update version of workflow port

dengjun101:
**Describe the pull request**

- What does your PR fix? Fixes # upgrade port version


ports/workflow/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "workflow",
-  "version-string": "0.9.3",
+  "version-string": "0.9.4",
```
ras0219-msft, line 3:
```suggestion
  "version": "0.9.4",
```
This looks like a numeric+dots version, so we should use that scheme to enable versioning support

ras0219-msft (changes requested):
Thanks for the PR! LGTM after the version scheme change.

Note that you'll need to re-run `vcpkg x-add-version --overwrite-version workflow` after changing the scheme.

ras0219-msft:
Thanks!

PR #15717, opened 2021-01-17
[mnn,flatbuffers] add a new port and patch

luncliff:
### Changes

Create a new port for https://github.com/alibaba/mnn. Focused general desktop build configurations.

##### ports/flatbuffers/fix-issue-6036.patch

Workaround for https://github.com/google/flatbuffers/issues/6036. The patch can be removed when the `flatbuffers` package gets the next release.

##### use-packages-and-install.patch
The project embedded several libraries that are available in VcPkg. It modifies the `CMakeLists.txt` files to use those packages.

##### ~~fix-dllexport.patch~~

Some symbols become duplicated when MNN.dll is being generated. The scripts of the project don't generate executables in the Windows platform, so it seems like this issue is not reported.  
~~This patch makes minimum changes to enable the `tools` feature.~~

Extracted all code changes to https://github.com/alibaba/MNN/pull/1391

#### Feature

Supports 5 features.

* `test`: Generate train/benchmark executables
* `tools`: Not available for Windows
* `opencl`, `vulkan`, `cuda`: API backends that use packages in VcPkg
* `metal`: API backend only for the Apple platform

#### Known Issues

`cuda` errors when Ninja in Windows environment. `PREFER_NINJA` won't be used in this case.  
Notice that CUDA installation may affect the build result.

Using the `metal` feature may require Xcode CLI tools configuration in the environment. Please reference 
https://github.com/gfx-rs/gfx/issues/2309#issuecomment-506130902

### Which triplets are supported/not supported?

`arm`, `uwp`, `ios`, `android` won't be considered in this PR. (Probably in the next port update?)

* `x64-windows`
* `x86-windows`
* `x64-osx`
* `x64-linux`

### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Checked!

PR #7380, opened 2019-07-23
[WIP] [flatbuffers] Add support for iOS

Helco:
This is a fix for flatbuffers on iOS using the alcroito's work of PR #6275.
Neither the compiler nor the flathash executable should be compiled, same as with UWP.

This should be regarded as WIP, as I didn't got around testing it on hardware (hopefully in the next few days).
As to the location of the PR, I am also okay with this being merged into alcroito/ios_support first.

PR #16843, opened 2021-03-23
[flatbuffers] support iOS triplets

luncliff:
### What does your PR fix?

Exclude `flatc` build target which requires `BUNDLE DESTINATION`. (config error when iOS triplets)

I encountered this issue while trying `arm` triplets for #15717, and found an old issue #7380.  

4 iOS triplets should become available with this PR.

* arm64-ios
* arm-ios
* x64-ios
* x86-ios

Sure. I think this is the minimum change.


luncliff:
#15717 changes `Port-Version` of `flatbuffers`. To prevent a collision, changed to a draft.

NancyLi1013 (approved):

NancyLi1013:
Is this PR ready for review now @luncliff?

luncliff:
I think so.  
I don't have a test project which uses `flatbuffers` in iOS in my hand... but the change is small(just disables `flatc`), so it won't be a matter I think.

NancyLi1013:
Thanks for your feedback. 

Since these triplets are all community triplets, CI machine will not test these. Have you tested them locally?

luncliff:
> Thanks for your feedback.
> 
> Since these triplets are all community triplets, CI machine will not test these. Have you tested them locally?

Yes. Without the executable build target, the port only builds several source files.

[flatbuffers-ios.zip](https://github.com/microsoft/vcpkg/files/6195248/flatbuffers-ios.zip)

ras0219-msft:
LGTM, thanks for the PR!

PR #15792, opened 2021-01-21
[libarchive] update to 3.5.1

JonLiu1993:
Fix #14985

Update libarchive to the latest version 3.5.1

All features are tested successfully in the following triplets:
- x86-windows
- x64-windows
- x64-windows-static

PR #10770, opened 2020-04-10
[cmake] Add cmake port to test qt 

Neumann-A:
PR #10769 is included for static builds to work

~~cmake-gui in static builds will only work with Qt > 5.14 due to the automatic plugin linkage added to Qt 5.14. Seems like cmake forgot to link against and add sources to include the required qt plugins.~~

- [x] waiting for the qt 5.15 merge so i don't have to deal with the plugins in static builds ;)
- [x] retesting and updating

PR #10769, opened 2020-04-10
[libarchive, libuv]Fix static linkage

Neumann-A:
adds a wrapper to correctly link in dependencies for libarchive and libuv

PR #16845, opened 2021-03-23
[libarchive] update to 3.5.2

### LINKED ISSUE #14985: [libarchive] update to 3.5.0
Library name: libarchive

New version number: 3.5.0

https://github.com/libarchive/libarchive/releases/tag/v3.5.0

Thieum:
Fixes #14985

Based on #15792 . Took the changes, and applied them to the latest version of the codebase. 

Tested with windows-x86 and windows-x64 (got one error on install in x86 with libiconv, then one error with lzo, but after each retries I was a step further, and it finally installed everything correctly. x64 installed without error. First time install error?).


Thieum:
@JonLiu1993 the cmake build script located here: https://github.com/microsoft/vcpkg/tree/master/scripts/test_ports/cmake depends on libarchive. I'm not sure how to reproduce the script execution on my machine?

Thieum:
ah I think that would be ` .\vcpkg install cmake --overlay-ports=scripts/test_ports` from #10770 . Will try to find out what is happening.

Thieum:
I can reproduce but there's not much info I have the ability to work with so far:

```
[525/526] cmd.exe /C "cd /D D:\GitHub\vcpkg\buildtrees\cmake\x86-windows-dbg && D:\GitHub\vcpkg\buildtrees\cmake\x86-windows-dbg\bin\cmake.exe -P cmake_install.cmake"
FAILED: CMakeFiles/install.util 
cmd.exe /C "cd /D D:\GitHub\vcpkg\buildtrees\cmake\x86-windows-dbg && D:\GitHub\vcpkg\buildtrees\cmake\x86-windows-dbg\bin\cmake.exe -P cmake_install.cmake"
```
I tried to upgrade the package to cmake 3.20 in the test_ports, but I get the same error, so there must something broken between cmake and libarchive, but it's hard to tell what.

cmake specifies a version of 3.3.3 as a minimum in its cmake file, but the new version should be compatible afaik.

@JonLiu1993 any idea what I should try to have a better understanding of the issue?

ports/libarchive/portfile.cmake
```diff
@@ -3,14 +3,14 @@ vcpkg_fail_port_install(ON_TARGET "UWP")
...
     HEAD_REF master
     PATCHES
-        fix-buildsystem.patch
-        fix-dependencies.patch
-        fix-cpu-set.patch
-        disable-warnings.patch
+        #fix-buildsystem.patch
+        #fix-dependencies.patch
+        #fix-cpu-set.patch
+        #disable-warnings.patch
```
JonLiu1993, line 13: @Thieum ,If these patches are not useful, please delete them

JonLiu1993:
@Thieum ,I noticed the error reported by cmake. The current version of cmake is not compatible with the current version of libarchive. Maybe the _DCMAKE_USE_SYSTEM_LIBARCHIVE_ option in vcpkg/scripts/test_ports/cmake/portfile.cmake should be static.

ports/libarchive/portfile.cmake
Thieum: fixed in https://github.com/microsoft/vcpkg/pull/16845/commits/5989f92df9a34b9078efce1c29672d4e07897286

Thieum:
@JonLiu1993 still not working even with -DCMAKE_USE_SYSTEM_LIBRARIES=OFF . Tried with cmake 3.18.1 and 3.19.2.

Since the static builds are passing, could we force static linking in any way for this specific test?

Thieum:
Also, I don't know if it's intended but some dependencies are used as release version even in debug : 
image

Neumann-A:
> Also, I don't know if it's intended but some dependencies are used as release version even in debug :

that means the linkage is wrong. Just check the build logs.... and after that run configure with `--trace-expand` to figure out where it goes wrong. 

> And since we set DCMAKE_USE_SYSTEM_LIBARCHIVE off

from the error there is no reason to deactivate it. The error is in `cmake_install` which is probably due to your linkage issue. If you want to figure out what is going on. Try running the command in `vcpkg env` by hand

scripts/test_ports/cmake/portfile.cmake
```diff
@@ -16,7 +16,7 @@ vcpkg_configure_cmake(
     OPTIONS
         -DBUILD_TESTING=OFF
         #-DCMAKE_USE_SYSTEM_LIBRARIES=ON
-        -DCMAKE_USE_SYSTEM_LIBARCHIVE=ON
+        -DCMAKE_USE_SYSTEM_LIBARCHIVE=OFF # CMake is not compatible with libarchive 3.5.1
```
Neumann-A, line 19:
```suggestion
        -DCMAKE_USE_SYSTEM_LIBARCHIVE=ON
```
the build logs in 8a00f62c720795ea0587d68af072448afa6f7e61 doesn't support that.

Thieum:
> also your link issues are due to the one patch you removed....

Indeed! I'll recover them, and try again.

ports/libarchive/fix-buildsystem.patch
```diff
@@ -0,0 +1,163 @@
...
+ 
+ IF(HAVE_STRERROR_R)
+   SET(HAVE_DECL_STRERROR_R 1)
+@@ -2012,13 +2001,10 @@ CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/build/cmake/config.h.in
+ INCLUDE_DIRECTORIES(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
+ ADD_DEFINITIONS(-DHAVE_CONFIG_H)
+ 
+-# Handle generation of the libarchive.pc file for pkg-config
+-INCLUDE(CreatePkgConfigFile)
+-
```
Neumann-A: Why is this removed ?

ports/libarchive/fix-buildsystem.patch
```diff
@@ -0,0 +1,163 @@
...
+-CHECK_FUNCTION_EXISTS_GLIBC(wcrtomb HAVE_WCRTOMB)
+-CHECK_FUNCTION_EXISTS_GLIBC(wcscmp HAVE_WCSCMP)
+-CHECK_FUNCTION_EXISTS_GLIBC(wcscpy HAVE_WCSCPY)
+-CHECK_FUNCTION_EXISTS_GLIBC(wcslen HAVE_WCSLEN)
+-CHECK_FUNCTION_EXISTS_GLIBC(wctomb HAVE_WCTOMB)
++set(HAVE_WCRTOMB 1)
++set(HAVE_WCSCMP 1)
++set(HAVE_WCSCPY 1)
++set(HAVE_WCSLEN 1)
++set(HAVE_WCTOMB 1)
```
Neumann-A: Are the tests failing in the release build only? is there a way to  pass the set variables via `vcpkg_configure_cmake`?

ports/libarchive/fix-buildsystem.patch
```diff
@@ -0,0 +1,163 @@
...
+   IF(WIN32 AND NOT CYGWIN)
+-    #
+-    # Test if ZLIB_WINAPI macro is needed to use.
+-    #
+-    TRY_MACRO_FOR_LIBRARY(
+-      "${ZLIB_INCLUDE_DIR}" "${ZLIB_LIBRARIES}"
+-      RUNS
+-      "#include <zlib.h>\nint main() {uLong f = zlibCompileFlags(); return (f&(1U<<10))?0:-1; }"
+-      ZLIB_WINAPI)
++	SET(ZLIB_WINAPI OFF) # skip following test, it crashes with weird message box
```
Neumann-A: does it crash on x64-windows? test x64-windows-static instead. I assume it needs the dll to be on path.

ports/libarchive/fix-buildsystem.patch
Thieum: restored - it doesn't seem to break the build. As for the why, it was there since https://github.com/microsoft/vcpkg/commit/f99d143dc3e7cf9b07120256d2801da59653b486

ports/libarchive/fix-buildsystem.patch
Thieum: no it seems to fail on the debug build of cmake:

```
Building package cmake[core]:x86-windows...
-- Installing port from location: C:\a\1\s\scripts/test_ports\cmake
-- Downloading https://gitlab.kitware.com//cmake/cmake/-/archive/b7b0fb430397bafae4a7bd80b41d474c91a3b7f3/cmake-b7b0fb430397bafae4a7bd80b41d474c91a3b7f3.tar.gz -> cmake-cmake-b7b0fb430397bafae4a7bd80b41d474c91a3b7f3.tar.gz...
-- Extracting source D:/downloads/cmake-cmake-b7b0fb430397bafae4a7bd80b41d474c91a3b7f3.tar.gz
-- Using source at D:/buildtrees/cmake/src/4c91a3b7f3-acfeed9ef6.clean
-- Configuring x86-windows
-- Building x86-windows-dbg
CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:146 (message):
    Command failed: D:/downloads/tools/cmake-3.19.2-windows/cmake-3.19.2-win32-x86/bin/cmake.exe --build . --config Debug --target install -- -v -j17
    Working Directory: D:/buildtrees/cmake/x86-windows-dbg
      D:\buildtrees\cmake\install-x86-windows-dbg-out.log
```
Even the master branch behavior seems weird to me - the debug dll version of libarchive 3.4.3 depends on the release version of lzma and other dlls that are used as build options. Maybe it's normal, but that makes a complicated dependency graph.
image

Neumann-A: > Maybe it's normal, but that makes a complicated dependency graph.

No its not. You need to fix the linkage.

ports/libarchive/portfile.cmake
```diff
@@ -3,14 +3,14 @@ vcpkg_fail_port_install(ON_TARGET "UWP")
...
+    PATCHES 
+        disable-warnings.patch
         fix-buildsystem.patch
-        fix-dependencies.patch
         fix-cpu-set.patch
-        disable-warnings.patch
+        fix-dependencies.patch
 )
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
```
JonLiu1993, line 16: Please add "FEATURES" under the function "vcpkg_check_features"

JonLiu1993:
@Thieum ,Can you please resolve the conflicts against master?

Thieum:
@JonLiu1993 done!

mirostauder:
Bump.
Any progress?

Thieum:
@mirostauder I tried to understand the underlined linking issue, but without success so far. The current version seems to be problematic too, as it has release dependencies, in debug mode. I repackaged the result of this branch in a nuget package, and it works for my use cases. But I cannot find a way to have it working in the vcpkg context. Any help would be greatly appreciated!

xavier2k6:
Libarchive 3.5.2 is latest
https://github.com/libarchive/libarchive/releases

dg0yt:
Are you still working on this PR?
There is an issue with the libarchive .pc file, noticed in #20062.

JonLiu1993:
@Thieum , The Libarchive 3.5.2 is latest version, and there is an issue issue with the libarchive .pc file, Could you please help update the version to 3.5.2 and fix the error?

Thieum:
@JonLiu1993 I can try to update to 3.5.2, and see how it goes, but I never really understood the issues around linking I get with my different attempts so far.

dg0yt:
Regarding the errors building port cmake with static linkage: The vcpkg-cmake-wrapper uses `find_dependency`. This macro isn't well suited for wrappers (`return` statement, missing "Package File Interface Variables", cf. https://github.com/microsoft/vcpkg/pull/19692#discussion_r693559124). Maybe this is the culprit? CC author @cenit

cenit:
> CC author @cenit

??

the author of the wrapper is @Neumann-A in https://github.com/microsoft/vcpkg/pull/10769

in any case, i can try to offer my help fixing it

Thieum:
@dg0yt would find_package work better than find_dependency for the cases we need? The tiff package seems to work this way with a few common features (lzma, zstd, zlib)

dg0yt:
> would find_package work better than find_dependency for the cases we need? The tiff package seems to work this way with a few common features (lzma, zstd, zlib)

@Thieum *I'm not sure if it is the fix but* yes `find_package` is the right thing. It was even used in libarchive until https://github.com/microsoft/vcpkg/commit/f89336d703bf3ffd2aa4f832f214b42a11c7d044#diff-142052910b682b4133dda7472e82131058b2dabf70be4d052e62db1f8819a0ea
Disclaimer: I authored the tiff wrapper. The latest state of the art is waiting in #18473, https://github.com/microsoft/vcpkg/blob/f8deebc2ebbe6a17bfcda327724f32cece24134a/ports/tiff/vcpkg-cmake-wrapper.cmake.in. The lessons learned went to the wrapper documentation PR.

I try to build your PR and the cmake test port on linux right now, but I don't know if I succeed with all cmake dependencies.

ports/libarchive/portfile.cmake
```diff
@@ -77,6 +77,12 @@ configure_file("${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake.in" "${CURRE
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/share)
```
dg0yt, line 81: Gotcha! Here you are removing the wrapper which was installed a few lines above.
(Sorry, the content of the wrapper doesn't matter at this point.)

Thieum:
@dg0yt  Awesome it worked without those lines! Thanks a lot for your help!

ports/libarchive/portfile.cmake
Thieum: removed in https://github.com/microsoft/vcpkg/pull/16845/commits/1bd8cff5a10375cc2c9d06636cb5c0f8e7abceac

cenit:
> Regarding the errors building port cmake with static linkage: The vcpkg-cmake-wrapper uses `find_dependency`. This macro isn't well suited for wrappers (`return` statement, missing "Package File Interface Variables", cf. [#19692 (comment)](https://github.com/microsoft/vcpkg/pull/19692#discussion_r693559124)). Maybe this is the culprit? CC author @cenit

my bad, if find_package is the way to go, please fix it!

dg0yt:
> my bad, if find_package is the way to go, please fix it!

@cenit Well, at least `find_dependency` wasn't the culprit. 

I would suggest to merge this PR as is (update, no regression), and handle other issues in separate PRs:
 - pc file (already broken, to be used in #20062).
 - wrapper (already working, covered by usage in CI)

JonLiu1993 (approved):

JonLiu1993:
- x86-windows
- x64-windows
- x64-windows-static

BillyONeal:
Thanks for the update!

PR #16847, opened 2021-03-23
Compile for android including armv6.

talregev:
Compile for android including armv6.
I want to compile with ndk and vcpkg libs from vcpkg. 
I use help from here:
https://vcpkg.readthedocs.io/en/latest/users/android/
Also modify android.cmake to support armv6 compilation (from ndk13 - ndk16). I use ndk16b for my test. 
Also I am active on opensource that we compile libs for armv6 for old phones. 

--------------------------------------------------------
ndk 16b have a bug for armv6 (it compile it to armv7). so I use ndk 15c for armv6 testing.


triplets/community/arm-android.cmake
```diff
@@ -0,0 +1,4 @@
+set(VCPKG_TARGET_ARCHITECTURE arm)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE dynamic)
```
ras0219-msft, line 3:
```suggestion
set(VCPKG_LIBRARY_LINKAGE static)
```
Because we prefer static library linkage on Linux by default, we should do the same for android.

ras0219-msft (changes requested):

triplets/community/arm-android.cmake
talregev: I agree.

talregev: I also want to do more changes to the android.cmake file.

talregev: @ras0219-msft What is VCPKG_CRT_LINKAGE? should it also be static?

JackBoosY: `VCPKG_CRT_LINKAGE` is just for Windows.

talregev: Maybe I will remove `VCPKG_CRT_LINKAGE` from these triplets. 
I succeeded to compile rappture to armv6, but didn't succeeded to consume it (to compile this lib with another application).
When I will success, I will push the change you requested, and some more.
Thank you for your feedback.

talregev: @JackBoosY I update the PR.

talregev:
I want also support ANDROID_NATIVE_API_LEVEL from env variable to be able to compile to ANDROID_NATIVE_API_LEVEL=16.
Also check this with ndk 15c for armv6. and ndk21d for all other android.
Please review my PR. thank you.

triplets/community/arm-android.cmake
talregev: @ras0219-msft I update like you request, please review my PR. thank you.

talregev:
@JackBoosY Should I do something?

JackBoosY:
@talregev  Waiting for test finish.

JackBoosY:
Ping @christophe-calmejane for review this PR.

JackBoosY (approved):

talregev:
@JackBoosY Thank you

christophe-calmejane:
The changes looks fine to me, but one question though.
Why add another way to control the API version to be used (which takes precedence over the triplet variable)?
I'm just worried the env variable might someday be exported (same name than the cmake variable) without paying attention.

Maybe it's because the documentation does not specify that the API version can be changed from the triplet file using `VCPKG_CMAKE_SYSTEM_VERSION`?

talregev:
I don't want to put it VCPKG_CMAKE_SYSTEM_VERSION in the triple, because it will be just for our project, and it should be general. 
In android dev each project has his api that it want to support, and it should able to do it from outside with easy way.
In our opensource project we compile for armv6, arm, x86 to api 16 and arm64, x64 to api 21, but not all the projects do that.

talregev:
https://developer.android.com/ndk/guides/cmake#android_native_api_level
maybe `ANDROID_PLATFORM` as env var?

christophe-calmejane:
I think `ANDROID_NATIVE_API_LEVEL` is actually a better name than `ANDROID_PLATFORM`. It should be fine like this I think.

talregev:
great! thank you that you like it more :)

talregev:
Because it hard to change cmake flags from vcpkg. (only from profile.cmake) if I understand it correctly, and I don't want to change each of profile.cmake of the port.
I want to set from env var: ANDROID_ARM_NEON.
So I will add it to android.cmake, and the default will be true as it now.

Neumann-A:
The question here is:
Are all environment variables used in the android toolchain hashed or added to the binary caching as a hash? 
If not the current android toolchain terribly breaks binary caching. 
That is probably a case for @ras0219-msft to comment on.

scripts/toolchains/android.cmake
```diff
@@ -6,8 +6,13 @@ set(CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION clang CACHE STRING "")
 
 if (VCPKG_TARGET_TRIPLET MATCHES "^arm64-android")
     set(ANDROID_ABI arm64-v8a CACHE STRING "")
+elseif(VCPKG_TARGET_TRIPLET MATCHES "^armv6-android")
+    set(ANDROID_ABI armeabi CACHE STRING "")
 elseif(VCPKG_TARGET_TRIPLET MATCHES "^arm-android")
     set(ANDROID_ABI armeabi-v7a CACHE STRING "")
+    if(DEFINED ENV{ANDROID_ARM_NEON})
+        set(ANDROID_ARM_NEON $ENV{ANDROID_ARM_NEON} CACHE STRING "")
+    endif()
```
ras0219-msft, line 15:
```suggestion
elseif(VCPKG_TARGET_TRIPLET MATCHES "^arm-android")
    set(ANDROID_ABI armeabi-v7a CACHE STRING "")
elseif(VCPKG_TARGET_TRIPLET MATCHES "^armv7neon-android")
    set(ANDROID_ABI armeabi-v7a CACHE STRING "")
    set(ANDROID_ARM_NEON ON CACHE BOOL "")
```

https://developer.android.com/ndk/guides/cmake#android_abi

ras0219-msft: Then, also add an `armv7neon-android` community triplet.

ras0219-msft (changes requested):
This LGTM after my suggestion is applied.

I'm really not happy about string matching on the triplet name, but it is what we currently do.

scripts/toolchains/android.cmake
talregev: can we do it better name, just arm-neon?

talregev: it can be neon with armv8 and etc.

talregev: I need to add arm-neon cmake file.

talregev: done

strega-nil-ms:
Alright, thanks @talregev for working with us to get this done!

PR #16858, opened 2021-03-24
[ryu] ryuConfig.cmake now selects the correct libraries when cross compiling

georg-emg:
When using ryu in a cmake project, the correct library names are now used when cross compiling. Previously, ryuConfig.cmake checked the host system rather than the target system to decide whether to attempt to link with ryu.lib or libryu.a, which caused the wrong library names to be used when the host and target were different.


ports/ryu/ryuConfig.cmake
```diff
@@ -13,7 +13,7 @@ get_filename_component(ROOT "${CMAKE_CURRENT_LIST_FILE}" PATH)
 get_filename_component(ROOT "${ROOT}" PATH)
 get_filename_component(ROOT "${ROOT}" PATH)
 
-if (CMAKE_HOST_WIN32)
+if (CMAKE_SYSTEM_NAME STREQUAL Windows)
 set_library_target("RYU" "ryu" "${ROOT}/debug/lib/ryu.lib" "${ROOT}/lib/ryu.lib" "${ROOT}/include/")
 set_library_target("RYU" "ryu_printf" "${ROOT}/debug/lib/ryu_printf.lib" "${ROOT}/lib/ryu_printf.lib" "${ROOT}/include/")
```
ras0219-msft, line 18:
```suggestion
if (TRUE)
find_library(RYU_RELEASE_LIB ryu PATHS "${ROOT}/lib" NO_DEFAULT_PATH)
find_library(RYU_DEBUG_LIB ryu PATHS "${ROOT}/debug/lib" NO_DEFAULT_PATH)
find_library(RYUPF_RELEASE_LIB ryu_printf PATHS "${ROOT}/lib" NO_DEFAULT_PATH)
find_library(RYUPF_DEBUG_LIB ryu_printf PATHS "${ROOT}/debug/lib" NO_DEFAULT_PATH)
set_library_target("RYU" "ryu" "${RYU_DEBUG_LIB} "${RYU_RELEASE_LIB}" "${ROOT}/include/")
set_library_target("RYU" "ryu_printf" "${RYUPF_DEBUG_LIB} "${RYUPF_RELEASE_LIB}" "${ROOT}/include/")
```

ras0219-msft: This should work across all platforms, including cross compilation.

ras0219-msft (changes requested):

ras0219-msft:
Thanks for the PR! This LGTM once my suggestion is applied

ports/ryu/ryuConfig.cmake
georg-emg: Just a quick question, out of curiosity: why `if(TRUE)`?

JackBoosY (approved):

ports/ryu/ryuConfig.cmake
ras0219-msft: I was just trying indicate that it should always be applied; it's not needed to have `if(TRUE)`.

ras0219-msft:
Thanks for the PR!

PR #16882, opened 2021-03-25
[ffmpeg] Fix FindFFMPEG lib versions for building opencv[ffmpeg]

genevanmeter:
libavutil, libavdevice, etc. define their versions separate from FFMPEG main version in a version.h file located with each component headers.
This change fixes building opencv[ffmpeg] when it attempts to discover the individual versions of FFMPEG libs. Before this change opencv[ffmpeg] would build but the FFMPEG functionality was disabled since all the lib versions were set to 4.3.1 and didn't meet its requirements.


Cheney-W (changes requested):
1. Please modify the `Port-Version` of CONTROL to 13.
2. Please run `.\vcpkg.exe x-add-version ffmpeg` to update the json file.

Cheney-W:
About port's json file, each of your commit changes will modify its git tree value, which means you need to update json file after you have done all the changes and executed the git commit -m "XXX" command. The specific sequence is as follows:
```
Git checkout -b "xxx"
Modify the local files
Git diff
Git add .
Git commit -m "comments description"
.\vcpkg x-add-version {Port name}
Git add .
Git commit -m "comments description"
git push
```
If you haven't executed the `. \vcpkg x-add-version {port name}` on the current branch, you need to use it after your modification is completed and submitted.
If you have executed this command on the current branch, you need to execute the following command when you need to update the json file:
`.\vcpkg.exe x-add-version {port name} --overwrite-version`
So, for this PR, you need to execute .\vcpkg.exe x-add-version {port name} --overwrite-version again.

BTW, osgearth:x64-linux failure is not related with your PR, this is a CI issue, we are working on it.

mcmtroffaes:
This is great in principle, however I'm not in favour of hardcoding these versions. May I suggest instead to extract the versions from the source files (e.g. from the version.h file of each library, such as https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/version.h)?

Cheney-W (approved):

JackBoosY:
Please solve the file conflicts first.

Thanks.

PR #16905, opened 2021-03-26
[fontconfig] fixes for arm64-macos

abique:
References:
 - https://github.com/Homebrew/homebrew-core/pull/57999
 - https://github.com/Homebrew/homebrew-core/issues/57994

### DIFF
--- a/ports/fontconfig/macos_arm_fccache.patch
+++ b/ports/fontconfig/macos_arm_fccache.patch
@@ -1,0 +1,34 @@
+ src/fcatomic.h | 12 ++----------
+ 1 file changed, 2 insertions(+), 10 deletions(-)
+
+diff --git a/src/fcatomic.h b/src/fcatomic.h
+index cc28a883c13fadf2ad9900586a9cae3097354671..657a5f551adfc37415231b7002746ad6cd031c78 100644
+--- a/src/fcatomic.h
++++ b/src/fcatomic.h
+@@ -70,24 +70,16 @@ typedef LONG fc_atomic_int_t;
+ #elif !defined(FC_NO_MT) && defined(__APPLE__)
+
+ #include <libkern/OSAtomic.h>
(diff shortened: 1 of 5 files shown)


ports/fontconfig/macos_arm_fccache.patch
```diff
@@ -0,0 +1,43 @@
+From 6def66164a36eed968aae872d76acfac3173d44a Mon Sep 17 00:00:00 2001
+From: Nirbheek Chauhan <nirbheek@centricular.com>
+Date: Sat, 22 Feb 2020 03:57:50 +0530
+Subject: [PATCH] fcatomic: Fix EXC_BAD_ACCESS on iOS ARM64
+
+iPhone 2.1 was released a long time ago, and the macro for checking
+the target iOS SDK version has changed. We can simplify everything and
+do a very basic check.
+---
```
PhoebeHui, line 9:
```suggestion
```

PhoebeHui (changes requested):
@abique, thanks for the PR! 

Could you minimize the patch?

ports/fontconfig/macos_arm_fccache.patch
abique: Hi @PhoebeHui,

Thank you for reviewing the change.

Why would we minimize it?
I think it provides context which can help if there is a problem with this patch in the future.

Through if you insist, I don't mind applying your changes.

Regards,
Alex

abique: Hi @PhoebeHui,

Thank you for reviewing the change.

Why would we minimize it?
I think it provides context which can help if there is a problem with this patch in the future.

Through if you insist, I don't mind applying your changes.

Regards,
Alex

PhoebeHui (approved):
Thanks for your updates!

abique:
Thank you for the review :)

PR #16908, opened 2021-03-26
[xmlsec] add pkgconfig files

### LINKED ISSUE #16830: [xmlsec] Add missing pkgconfig file

xmlsec misses the pkgconfig file, which the upstream provides. This makes it problematic for software that doesn't use CMake. Moreover, if xmlsec was installed by the OS package manager (or Homebrew), it will get picked instead. 

**Proposed solution**
Add the `xmlsec.pc` and `xmlsec-openssl.pc` to installed files.

wrobelda:
**Describe the pull request**

- What does your PR fix? 
Fixes #16830 
Also fixes the library filename, which should actually be `libxmlsec1`. This follows the upstream naming scheme.
I believe this port in general should be renamed to `xmlsec1`, which is how vast majority of repositories call it: https://repology.org/project/xmlsec/versions . I did not, however, make that change as part of this PR.

No change

Yes

### DIFF
--- a/ports/xmlsec/CMakeLists.txt
+++ b/ports/xmlsec/CMakeLists.txt
@@ -3,9 +3,6 @@ project (xmlsec1 C)
 
 option(INSTALL_HEADERS_TOOLS "Install public header files and tools" ON)
 
-set(CMAKE_SHARED_LIBRARY_PREFIX)
-set(CMAKE_STATIC_LIBRARY_PREFIX)
-
 find_package(LibXml2 REQUIRED)
 find_package(OpenSSL REQUIRED)
 find_package(Iconv REQUIRED)
@@ -58,16 +55,16 @@ if(MSVC)
   add_compile_options(/wd4130 /wd4127 /wd4152)
 endif()
 
-add_library(libxmlsec1 ${SOURCESXMLSEC})
-add_library(libxmlsec1-openssl ${SOURCESXMLSECOPENSSL})
+add_library(xmlsec1 ${SOURCESXMLSEC})
+add_library(xmlsec1-openssl ${SOURCESXMLSECOPENSSL})
 
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/include include ${LIBXML2_INCLUDE_DIRS})
 
-target_link_libraries(libxmlsec1 PRIVATE
+target_link_libraries(xmlsec1 PRIVATE
     ${LIBXML2_LIBRARIES} OpenSSL::SSL
 )
-target_link_libraries(libxmlsec1-openssl PRIVATE
-    ${LIBXML2_LIBRARIES} OpenSSL::SSL libxmlsec1
+target_link_libraries(xmlsec1-openssl PRIVATE
+    ${LIBXML2_LIBRARIES} OpenSSL::SSL xmlsec1
 )
 
 add_compile_definitions(inline=__inline)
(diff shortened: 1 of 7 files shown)


wrobelda:
I believe this port should be renamed to `xmlsec1`, which is how vast majority of repositories call it: https://repology.org/project/xmlsec/versions

wrobelda:
I see those `osgearth` errors, but I can't reproduce them locally. Once it build successfully, the other time (after `rm -rf installed packages`) it fails with:

```
CMake Error at src/osgEarth/CMakeLists.txt:791 (PROTOBUF_GENERATE_CPP):
  Unknown CMake command "PROTOBUF_GENERATE_CPP".`
```

I don't see how these errors are related to this PR. I especially don't understand why that many ports are being rebuilt as part of CI/CD run, since none of them depend on `xmlsec`. In fact none of the ports in the repository depend on it. I suspect this must have to do with your CI/CD and that `osgearth` was recently modified, together with dependencies, and has not cached properly? Before you ask, the branch is up to date with `master`.

ports/xmlsec/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required (VERSION 3.8)
-project (xmlsec C)
+project (xmlsec1 C)
```
PhoebeHui, line 2: I don't think you need to change this, it still named 'xmlsec' in official https://github.com/lsh123/xmlsec.

ports/xmlsec/CMakeLists.txt
```diff
@@ -158,3 +161,22 @@ if(INSTALL_HEADERS_TOOLS)
...
+set(OPENSSL_LIBS "-lssl -lcrypto")
+set(XMLSEC_CORE_CFLAGS "${XMLSEC_CORE_CFLAGS} -DXMLSEC_DL_LIBLTDL=1 -I\${includedir}/xmlsec1 -DXMLSEC_CRYPTO_OPENSSL=1")
+set(XMLSEC_CORE_LIBS "-lxmlsec1 -lltdl")
+set(XMLSEC_OPENSSL_CFLAGS "${XMLSEC_OPENSSL_CFLAGS} -I\${includedir}/xmlsec1")
+set(XMLSEC_OPENSSL_LIBS "-L\${libdir} -lxmlsec1-openssl ${XMLSEC_CORE_LIBS} ${OPENSSL_LIBS}")
+
+configure_file(${PROJECT_SOURCE_DIR}/xmlsec.pc.in ${PROJECT_BINARY_DIR}/xmlsec1.pc @ONLY)
+configure_file(${PROJECT_SOURCE_DIR}/xmlsec-openssl.pc.in ${PROJECT_BINARY_DIR}/xmlsec1-openssl.pc @ONLY)
+install(FILES ${PROJECT_BINARY_DIR}/xmlsec1.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig/)
+install(FILES ${PROJECT_BINARY_DIR}/xmlsec1-openssl.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig/)
```
PhoebeHui, line 182: Does it reuire to rename xmlsec.pc.in to xmlsec1.pc.in, xmlsec-openssl.pc.in to xmlsec1-openssl.pc.in, I think we'd better to keep same with official.

PhoebeHui (changes requested):
@wrobelda, thanks for your PR!

cc @SvenPStarFinanz

ports/xmlsec/CMakeLists.txt
wrobelda: It's only an official "name" of the project, but internally all scripts and binaries actually use `xmlsec1`. 

See e.g.:
1) https://github.com/lsh123/xmlsec/blob/3e1b0076c1833112c0cda4de2e00423304245c48/configure.ac#L4
2) https://github.com/lsh123/xmlsec/blob/4b6ab2d86b71f8642f19ab3b7a0777984b6bce9a/xmlsec.spec.in
3) https://github.com/lsh123/xmlsec/blob/4b6ab2d86b71f8642f19ab3b7a0777984b6bce9a/Makefile.am

Also see:
https://repology.org/project/xmlsec/versions
https://packages.ubuntu.com/hirsute/amd64/xmlsec1/filelist
https://packages.ubuntu.com/hirsute/amd64/libxmlsec1-dev/filelist

ports/xmlsec/CMakeLists.txt
wrobelda: I does, unfortunately. This is what the upstream does and this is what other packages look for. See notes above, as well as e.g.: 
https://github.com/search?p=4&q=xmlsec1+filename%3A%2A.pc&type=Code

PhoebeHui (approved):

ports/xmlsec/CMakeLists.txt
```diff
@@ -58,20 +58,20 @@ if(MSVC)
   add_compile_options(/wd4130 /wd4127 /wd4152)
 endif()
 
-add_library(libxmlsec ${SOURCESXMLSEC})
-add_library(libxmlsec-openssl ${SOURCESXMLSECOPENSSL})
+add_library(libxmlsec1 ${SOURCESXMLSEC})
```
ras0219, line 61: Based on https://github.com/lsh123/xmlsec/blob/master/configure.ac#L2330, this should be

```suggestion
add_library(xmlsec1 ${SOURCESXMLSEC})
```
and the CMAKE_SHARED_LIBRARY_PREFIX / CMAKE_STATIC_LIBRARY_PREFIX should be left unset so it will produce:

> libxmlsec1.dylib
> libxmlsec1.a
> libxmlsec1.so
> xmlsec1.lib
> xmlsec1.dll

ras0219 (changes requested):

PhoebeHui (approved):

strega-nil-ms:
This looks great! Thanks @wrobelda :)

PR #16913, opened 2021-03-27
[nowide] Upgrade to 11.1.1

tats-u:
**Describe the pull request**

- What does your PR fix? Fixes #16912

- Which triplets are supported/not supported? Have you updated the CI baseline?
Same as the current version

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
Probably yes

PR #16910, opened 2021-03-27
[mpg123] Upgrade to 1.26.3-1

evpobr:
Fix invalid `MPG123_API_VERSION` value in `mpg123.h.in` for Windows platform. It was equal to `@API_VERSION@`, now it is read from `configure.ac` and set to correct value.

See also libsndfile/libsndfile#499.

PR #16879, opened 2021-03-25
[rapidcsv] new port

autoantwort:
**Describe the pull request**

- What does your PR fix? Fixes #16301

- Which triplets are supported/not supported? All

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? Yes

PR #16904, opened 2021-03-26
[vcpkg baseline][osgearth/gdal] Re-fix dependency gdal

JackBoosY:
The gdal requires freexl on non-Windows:
```
/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libgdal.a(ogrxlsdatasource.o): In function `OGRXLSDataSource::~OGRXLSDataSource()':
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:66: undefined reference to `freexl_close'
/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libgdal.a(ogrxlsdatasource.o): In function `OGRXLSDataSource::Open(char const*, int)':
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:147: undefined reference to `freexl_get_info'
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:152: undefined reference to `freexl_select_active_worksheet'
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:155: undefined reference to `freexl_get_worksheet_name'
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:160: undefined reference to `freexl_worksheet_dimensions'
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:171: undefined reference to `freexl_close'
/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libgdal.a(ogrxlsdatasource.o): In function `OGRXLSDataSource::GetXLSHandle()':
/mnt/vcpkg-ci/buildtrees/gdal/x64-linux-dbg/ogr/ogrsf_frmts/xls/ogrxlsdatasource.cpp:186: undefined reference to `freexl_open'
/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libgdal.a(ogrxlslayer.o): In function `OGRXLSLayer::DetectHeaderLine(void const*)':
[...]

PR #16915, opened 2021-03-27
[osgearth] fails in ci

autoantwort:
As you can see in #16914 and in #16879, #16913, #16910, ... osgearth fails in the ci, so set it to `fail` as a quickfix


PhoebeHui:
This issue has been fixed by PR https://github.com/microsoft/vcpkg/pull/16904, so close the PR!

PR #16927, opened 2021-03-28
[vcpkg] Convert env vars slash if host is Win32

huangqinjin:
Environment variables are host properties rather than target.


PhoebeHui (approved):
LGTM, thanks for the PR!

PR #16936, opened 2021-03-29
[ghc-filesystem] create a new port

luncliff:
### What does your PR fix?

This is a rework of #15031.  
Targeting the latest release of https://github.com/gulrak/filesystem/releases/tag/v1.5.4

The project is a header-only library, so every triplet must be available.

Yes.


ports/ghc-filesystem/portfile.cmake
```diff
@@ -0,0 +1,22 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO gulrak/filesystem
+    REF v1.5.4
```
NancyLi1013, line 4:
```suggestion
    REF v1.5.4
    HEAD_REF master
```

ports/ghc-filesystem/portfile.cmake
```diff
@@ -0,0 +1,22 @@
...
+vcpkg_configure_cmake(
+    SOURCE_PATH ${SOURCE_PATH}
+    PREFER_NINJA
+    OPTIONS
+        -DGHC_FILESYSTEM_BUILD_TESTING=OFF
+        -DGHC_FILESYSTEM_BUILD_EXAMPLES=OFF
+        -DGHC_FILESYSTEM_WITH_INSTALL=ON
+)
+vcpkg_install_cmake()
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ghc_filesystem TARGET_PATH share/${PORT})
```
NancyLi1013, line 17:
```suggestion
vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ghc_filesystem)
```
Since TARGET_PATH will default to `share/${PORT}`, it's unnecessary to add it here. Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_fixup_cmake_targets.md.

NancyLi1013 (changes requested):

luncliff:
Thanks for suggestions, @NancyLi1013. 👍   
I'd like to check if the library works for old NDK versions, which doesn't support C++ `<filesystem>` . I will mark "Ready for review" after then.

luncliff:
I checked the library and works quite well. I'm gonna like this one :D

ports/ghc-filesystem/portfile.cmake
```diff
@@ -0,0 +1,23 @@
...
+vcpkg_configure_cmake(
+    SOURCE_PATH ${SOURCE_PATH}
+    PREFER_NINJA
+    OPTIONS
+        -DGHC_FILESYSTEM_BUILD_TESTING=OFF
+        -DGHC_FILESYSTEM_BUILD_EXAMPLES=OFF
+        -DGHC_FILESYSTEM_WITH_INSTALL=ON
+)
+vcpkg_install_cmake()
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/ghc_filesystem)
```
NancyLi1013, line 18: These function have deprecated. Could you please update these to the new ones?
```
vcpkg_cmake_configure()
vcpkg_cmake_install()
vcpkg_cmake_config_fixup()
```
We prefer to use suggested functions for new ports. Please also pay attention to the differences between deprecated functions and the new ones.

Note: if we use new functions, vcpkg-cmake and vcpkg-cmake-config are also required to be added in dependency lists like this:
```
    {
      "name": "vcpkg-cmake",
      "host": true
    },
    {
      "name": "vcpkg-cmake-config",
      "host": true
    }
```

NancyLi1013 (changes requested):

ports/ghc-filesystem/portfile.cmake
luncliff: Oh. The time came! Thanks for the check. I will update in this weekend.

luncliff:
Applied the feedback in 852d0bb. Wish it looks good to you, @NancyLi1013.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your work @luncliff.

strega-nil:
great, thanks @luncliff :)

PR #16959, opened 2021-03-30
[vcpkg/make/scripts]only append the lib directories if they exist.

Neumann-A:
In #13100 I observed that configure was treating compiler warnings as errors on osx and thus configure determined some checks as no although they should be yes. This was observed in cairo:x64-osx in #13100 which always complained that it cannot link via `-pthread`. After this change configure passes correctly since cairo finally builds in #13100 on osx.


ras0219-msft:
LGTM, thanks!

PR #16978, opened 2021-03-31
Better handling of DOS attempts (on-sight connection's termination, improved events)

rafalsk:
That it to better handle situations described in here: [](https://www.reddit.com/r/gridnetproject/comments/md2a2m/denial_of_servicedos_attacks_weve_been_facing_and/)

The aim was to be able to abort a connection possibly as soon as the incoming connection had been accepted on the listening TCP socket. Thus we wanted to approach the usual way that to us would seem like

1)create ssh_bind

2)initialize ssh_bind_callbacks_struct with ssh_callbacks_init()

3)pass a pointer to a function with a footprint of void callbackConn (ssh_bind sshbind, void* userdata) to incoming_connection of ssh_bind_callbacks_struct which just got initialized (0ed out is seems)

call ssh_bind_set_callbacks() and with that set-up we would have expected for the callbackConn() function to fire.

It never happens. 'normally' we would have accepted the connection in such a function.

The usual accept_bind_accept works but as the practice shows, peers are able to exploit the timeout set (we could change the default timeout of infinity which as practice shows peers might exploit but not completing one of the handshake steps they participate in).

So with accept_bind_accept IF a timeout is set evil peers might still block the server-processing queue up to the timeout value; and if method like-keyboard interactive is enabled we cannot set the timeout to say 2sec as user wouldn't be able to type in credentials at all and if we set to to 10 seconds then the malicious attacker can block processing for all users for up to 10 seconds. By the way; when we disabled keyboard-interactive authentication, the keyboard interactive thing it the first to show up when connecting to server. ssh_bind_listen indeed doesn't seem to accomplish anything more than 'listening', no event processing logic seems to be invoked. ssh_bind_poll_callback never gets called.

We've added a couple of modifications that make the appropriate events fire and also making it possible to abort connection on sight without any reply at all.

### DIFF
--- a/ports/libssh/use-onconnection-bind-callback.patch
+++ b/ports/libssh/use-onconnection-bind-callback.patch
@@ -1,0 +1,108 @@
+diff --git a/include/libssh/server.h b/include/libssh/server.h
+index 41f89d5..aca6805 100644
+--- a/include/libssh/server.h
++++ b/include/libssh/server.h
+@@ -367,6 +367,44 @@ LIBSSH_API int ssh_channel_request_send_exit_signal(ssh_channel channel,
+
+ LIBSSH_API int ssh_send_keepalive(ssh_session session);
+
++/**
++     * @brief Accept an incoming ssh connection without initializing the session.
++     *
++     * @param  ssh_bind_o     The ssh server bind to accept a connection.
++     * @return client socket when a connection is established, or SSH_ERROR on failure.
++     */
++LIBSSH_API socket_t ssh_bind_accept_alone(ssh_bind ssh_bind_o);
++
++/**
++ * @brief Accept an incoming ssh connection on the given file descriptor
++ *        and initialize the session. Same as ssh_bind_accept_fd() with additional
++ *        freeing of memory associated with client socket should anything go wrong.
++ *
++ * @param  ssh_bind_o     The ssh server bind to accept a connection.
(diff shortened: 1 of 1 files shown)


cryptomilk:
Microsoft, please do not accept this pull request. This is not an upstream patch and hasn't been accepted upstream. It is possible that it will break the API/ABI in case the patch needs to be modified or will not be accepted upstream!

JackBoosY:
Since this PR modification involves source code, can you create a PR to upstream?

JackBoosY:
Ping @rafalsk for response.

rafalsk:
Well I vote this to be a considered a major security concern and addressed by vcpkg asap. It does not modify/break behavior of existing code. We've been faced by attacks targeting this rendering  the receiving socket completely unresponsive. All it takes to carry out a DOS attack is a single node with low capabilities. @cryptomilk while none of us had time to to pull this into your repository, you also didn't seem to be much concerned about the matter and the existing users should be informed. That's my viewpoint.

cryptomilk:
@rafalsk Please open a bug report or merge request upstream with the details what the issues are and what you are trying to fix. New features should be backed by test and a reproducer :-)

If you think there is a security bug, the process for reporting security bugs is documented here: https://www.libssh.org/development/security-process/

JackBoosY:

Issue #16991, opened 2021-03-31
[brotli] build failure when used with cargo-vcpkg


waych:
**Host Environment**
- OS: debian, ubuntu-latest
- Compiler: rust toolchain nightly 1.53 + gcc 10.2.1 + ld.bfd 2.35.2

**To Reproduce**
```
git clone https://github.com/Rust-SDL2/rust-sdl2.git
cd rust-sdl2
cargo install cargo-vcpkg
cargo vcpkg -v build
cargo build --examples --features "use-vcpkg static-link gfx image ttf mixer"
```

**Failure logs** 
`cargo vcpkg -v build` succeeds without error.  The `cargo build ...` command fails on final link however with the following:

```
  = note: /usr/bin/ld: /home/runner/work/rust-sdl2/rust-sdl2/target/vcpkg/installed/x64-linux/lib/libbrotlidec-static.a(decode.c.o): in function `DecodeLiteralB
lockSwitch':
          decode.c:(.text+0x863): undefined reference to `_kBrotliPrefixCodeRanges'
          /usr/bin/ld: decode.c:(.text+0x94f): undefined reference to `_kBrotliContextLookupTable'
          /usr/bin/ld: /home/runner/work/rust-sdl2/rust-sdl2/target/vcpkg/installed/x64-linux/lib/libbrotlidec-static.a(decode.c.o): in function `DecodeCommandBlockSwitch':
          decode.c:(.text+0xa7b): undefined reference to `_kBrotliPrefixCodeRanges'
          /usr/bin/ld: /home/runner/work/rust-sdl2/rust-sdl2/target/vcpkg/installed/x64-linux/lib/libbrotlidec-static.a(decode.c.o): in function `DecodeDistanceBlockSwitch':
          decode.c:(.text+0xc3b): undefined reference to `_kBrotliPrefixCodeRanges'
          /usr/bin/ld: /home/runner/work/rust-sdl2/rust-sdl2/target/vcpkg/installed/x64-linux/lib/libbrotlidec-static.a(decode.c.o): in function `ProcessCommands':
          decode.c:(.text+0x2319): undefined reference to `BrotliTransformDictionaryWord'
          /usr/bin/ld: /home/runner/work/rust-sdl2/rust-sdl2/target/vcpkg/installed/x64-linux/lib/libbrotlidec-static.a(decode.c.o): in function `SafeDecodeLiteralBlockSwitch':
[...]

PR #16992, opened 2021-03-31
[brotli] Fold common into dec and enc archives for static builds.

waych:
This change allows downstream consumers of the archives to not worry
about ordering the link correctly when using this package, which can be
problematic when --as-needed is passed to the linker and the order isn't
otherwise known.

Keeping the brotlicommon-static archive as is allows for probing of the
brotli libraries to remain symmetric between dynamic and static links.

Fixes static linking issue discussed in
https://github.com/microsoft/vcpkg/issues/16991 which is blocking
rust-sdl2 CI dependent upon vcpkg discussed here
https://github.com/Rust-SDL2/rust-sdl2/pull/1085 .

### DIFF
--- a/ports/brotli/fold-common-into-static-builds.patch
+++ b/ports/brotli/fold-common-into-static-builds.patch
@@ -1,0 +1,15 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 4ff3401..62d891f 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -171,8 +171,8 @@ endif()
+ 
+ set(BROTLI_STATIC_LIBS brotlicommon-static brotlidec-static brotlienc-static)
+ add_library(brotlicommon-static STATIC ${BROTLI_COMMON_C})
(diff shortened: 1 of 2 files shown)


Neumann-A:
äh no..... submit this upstream or even better just fix the link order.

waych:
Agreed that cargo-vcpkg would benefit from probing pkgconfig correctly. This seems like a simple enough workaround for at least rust-sdl2's case (or anything else that indirectly includes brotli in the cargo-vcpkg ecosystem) until lack of pkgconfig is fixed.

I was uncertain to what degree changes like this were being maintained within vcpkg, consider the creation of -static variant libs happening and being maintained here, and not upstream.

Neumann-A:
> I was uncertain to what degree changes like this were being maintained within vcpkg, consider the creation of -static variant libs happening and being maintained here, and not upstream.

That is incorrect, vcpkg just executes the upstream CMakeLists.txt with minor fixes to the pc files to reflect the library naming in the static case.

JackBoosY:
Please open a new PR to the upstream first.

Thanks.

waych:
Pursuing fixing this in vcpkg-rs. Thanks!

PR #17024, opened 2021-04-01
Libspatialite remove cflags from vcpkg_configure_make options

xandox:
Libspatialite remove cflags from vcpkg_configure_make options
This flag do not mentioned in sources 

- What does your PR fix? Fixes for PR #15605

- Which triplets are supported/not supported? same
- Have you updated the CI baseline? no

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? yes

### DIFF
--- a/ports/libspatialite/portfile.cmake
+++ b/ports/libspatialite/portfile.cmake
@@ -105,7 +105,6 @@ elseif (VCPKG_TARGET_IS_LINUX OR VCPKG_TARGET_IS_OSX) # Build in UNIX
       SOURCE_PATH ${SOURCE_PATH}
       AUTOCONFIG
       OPTIONS
-          "CFLAGS=-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H"
           "LIBS=-lpthread -ldl -lm -l${STDLIB}"
           "LIBXML2_CFLAGS=-I\"${CURRENT_INSTALLED_DIR}/include\""
           "--enable-rttopo"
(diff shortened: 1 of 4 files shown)


ports/libspatialite/portfile.cmake
```diff
@@ -105,7 +105,6 @@ elseif (VCPKG_TARGET_IS_LINUX OR VCPKG_TARGET_IS_OSX) # Build in UNIX
       SOURCE_PATH ${SOURCE_PATH}
       AUTOCONFIG
       OPTIONS
-          "CFLAGS=-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H"
```
JackBoosY, line 108: Any reason to remove this here?

JackBoosY (changes requested):

ports/libspatialite/portfile.cmake
xandox: This cflag option completely overwrite any cflag set by vcpkg_configure_make. Probably there should be some merge algorithm for system cflags and cflags from options, but it does not exists yet. This line does not affect at anything with current scripts but breakes new ones.

JackBoosY: I think we should add `-DACCEPT_USE_OF_DEPRECATED_PROJ_API_H` to compile options, can you patch it?

rhuijben:
The flag is no longer needed with current libspatialite versions that support newer proj versions. And once proj is bumped to 8.0 this will even cause an error, so +1 on removing the define.

rhuijben:
The flag used to enable the legacy api in the proj headers

JackBoosY (approved):

PR #17027, opened 2021-04-01
[google-cloud-cpp] update to the latest release (v1.26.0)

coryan:
Update google-cloud-cpp to v1.16.0, the latest release.

- What does your PR fix?

N/A

N/A

Yes.

### DIFF
--- a/ports/google-cloud-cpp/vcpkg.json
+++ b/ports/google-cloud-cpp/vcpkg.json
@@ -1,7 +1,6 @@
 {
   "name": "google-cloud-cpp",
   "version": "1.26.0",
-  "port-version": 1,
   "description": "C++ Client Libraries for Google Cloud Platform APIs.",
   "homepage": "https://github.com/googleapis/google-cloud-cpp",
   "license": "Apache-2.0",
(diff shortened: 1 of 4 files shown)


coryan:
The `x64_linux` error seems unrelated (maybe a ci script problem?).

versions/baseline.json
```diff
@@ -2281,8 +2281,8 @@
       "port-version": 4
     },
     "google-cloud-cpp": {
-      "baseline": "1.25.0",
-      "port-version": 3
+      "baseline": "1.26.0",
+      "port-version": 1
```
JonLiu1993, line 2285: @coryan ，When the port is upgraded, we should set the port-version to 0 or delete

ports/google-cloud-cpp/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "google-cloud-cpp",
-  "version": "1.25.0",
-  "port-version": 3,
+  "version": "1.26.0",
+  "port-version": 1,
```
JonLiu1993, line 4: @coryan ，When the port is upgraded, we should set the port-version to 0 or delete

coryan: Fixed. Please take another look.

ras0219-msft:
Thanks for the update!

PR #16577, opened 2021-03-06
[icu] add host depedency 

autoantwort:
**Describe the pull request**

- What does your PR fix? Fixes #16424

Same as #16433, but now with host support in vcpkg as base

PR #17037, opened 2021-04-02
[icu] support Mac/ARM build

luncliff:
### What does your PR fix? Fixes

Support `arm64` cross-compile in Mac OS `x86_64` host system.

* `arm64-osx`

The above triplet must become available when `uname` result is like below.

```console
user@host$ uname -msp
Darwin x86_64 i386
```

After changing `VCPKG_LIBRARY_LINKAGE` to `dynamic`, the build output should be arm64 shared.

```console
user@host$ ./vcpkg install icu:arm64-osx
    icu[core]:arm64-osx -> 67.1#9
Detecting compiler hash for triplet arm64-osx...
...

user@host$ file ./installed/arm64-osx/lib/libicuio.dylib 
./installed/arm64-osx/lib/libicuio.dylib: Mach-O 64-bit dynamically linked shared library arm64
```

Yeah.  
It was quite complicated to make it work. Wish this can save people's time.

### DIFF
--- a/ports/icu/CONTROL
+++ b/ports/icu/CONTROL
@@ -1,6 +1,6 @@
 Source: icu
 Version: 67.1
-Port-Version: 8
+Port-Version: 9
 Homepage: http://icu-project.org/apiref/icu4c/
 Description: Mature and widely used Unicode and localization library.
 Supports: !(arm|uwp)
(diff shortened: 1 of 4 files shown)


autoantwort:
Would be fixed by #16577 in a more general way :)

luncliff:
> Would be fixed by #16577 in a more general way :)

Oh, I did't look it in detail. Let me close this. :D

PR #17045, opened 2021-04-02
[vcpkg] Fix windres-rc wrapper

dg0yt:
**Describe the pull request**

- What does your PR fix? 
  It fixes logged error messages about not being able to execute `RCFLAGS`, and actually output the contents of this variable.
  **If the output of "RCFLAGS:..." is used, this will change behaviour.**

  AFAICT this is used for MSVC windows triplets, when building packages based on autotools, such as `libiconv`.

  No relevant guidelines for this script.

**Note:**
I'm not using MSVC, so **I couldn't test** with this toolchain.
Maybe this should be run with, or `set`, `-e`.

### DIFF
--- a/scripts/buildsystems/make_wrapper/windres-rc
+++ b/scripts/buildsystems/make_wrapper/windres-rc
@@ -1,7 +1,7 @@
 #! /bin/sh
 # Wrapper for windres to rc which do not understand '-i -o --output-format'.
 # cvtres is invoked by the linker
-scriptversion=2020-08-17.03; # UTC
+scriptversion=2021-04-02.18; # UTC
 
 
 nl='
@@ -58,8 +58,6 @@ func_file_conv ()
 # Adjust compile command to suit rc instead of windres
 func_windres_wrapper ()
 {
-  echo "FROM WINDRESWRAPPER FUNCTION:$@"
-  echo "RCFLAGS:$(RCFLAGS)"
   # Assume a capable shell
   in=
   out=


scripts/buildsystems/make_wrapper/windres-rc
```diff
@@ -59,7 +59,7 @@ func_file_conv ()
 func_windres_wrapper ()
 {
   echo "FROM WINDRESWRAPPER FUNCTION:$@"
-  echo "RCFLAGS:$(RCFLAGS)"
+  echo "RCFLAGS:$RCFLAGS"
```
Neumann-A, line 62:
```suggestion
```

scripts/buildsystems/make_wrapper/windres-rc
```diff
@@ -59,7 +59,7 @@ func_file_conv ()
 func_windres_wrapper ()
 {
   echo "FROM WINDRESWRAPPER FUNCTION:$@"
```
Neumann-A, line 61:
```suggestion
```

dg0yt:
There is a `scriptversion` at the top of the script. Does it need to be bumped?

Neumann-A:
>  Does it need to be bumped?
do as you like

dg0yt:
Done.

dg0yt:
Ping.

JackBoosY:
Ping @strega-nil @strega-nil-ms @BillyONeal for review this PR.

dg0yt:
FTR This also contributes to the performance issues with MSYS2, because *for every single compiler invocation*,
- it creates a subshell which probably involves the expensive implementation of `fork()` in MSYS2,
- it searches for the non-existing program called RCFLAGS.

ras0219-msft:

PR #17051, opened 2021-04-03
[x265] Allow release-only build.

matttyson:
When doing a release-only build, conditionally disable commands
for debug builds

When doing a release only build with the following triplet:

set(VCPKG_BUILD_TYPE release)

x265 will fail to install as it is assuming debug files exist to be copied.

This patch disables the copy commands for debug files when building in release mode

Note that this patch does not enable a debug-only build

### DIFF
--- a/ports/x265/CONTROL
+++ b/ports/x265/CONTROL
@@ -1,6 +1,6 @@
 Source: x265
 Version: 3.4
-Port-Version: 3
+Port-Version: 4
 Homepage: https://github.com/videolan/x265
 Description: x265 is a H.265 / HEVC video encoder application library, designed to encode video or images into an H.265 / HEVC encoded bitstream.
 Supports: !(uwp | arm)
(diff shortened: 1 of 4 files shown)


PhoebeHui (changes requested):
@matttyson, thanks for the PR!

Could you update the port version in CONTROL file and commit the changes? then update the baseline version by executing  './vcpkg x-add-version x265', and submit the changes?

matttyson:
@PhoebeHui Hi, I've made the requested changes, please let me know if that's OK.

Cheers,
Matt.

PhoebeHui (approved):

strega-nil:
@matttyson could you rebase on latest master?

matttyson:
@PhoebeHui @strega-nil I've rebuilt the patch on top of master

Cheers,
Matt.

strega-nil-ms:
Thanks @mattyson :)

PR #17076, opened 2021-04-04
Update libjuice to 0.7.1 to update libdatachannel

Nemirtingas:
**Describe the pull request**
Updates libjuice to 0.7.1 to update libdatachannel to 0.12.0.

### DIFF
--- a/ports/libjuice/vcpkg.json
+++ b/ports/libjuice/vcpkg.json
@@ -1,0 +1,14 @@
+{
+  "name": "libjuice",
+  "version-string": "v0.7.1",
+  "description": "The library is a simplified implementation of the Interactive Connectivity Establishment (ICE) protocol in C for POSIX platforms (including Linux and Apple macOS) and Microsoft Windows.",
+  "homepage": "https://github.com/paullouisageneau/libjuice",
+  "features": {
+    "nettle": {
+      "description": "Use nettle for HMAC computation instead of the Builtin",
+      "dependencies": [
+        "nettle"
+      ]
+    }
+  }
+}
--- a/ports/libjuice/vcpkg.json
+++ b/ports/libjuice/vcpkg.json
@@ -1,6 +1,6 @@
 {
   "name": "libjuice",
-  "version-string": "v0.7.1",
+  "version": "0.7.1",
   "description": "The library is a simplified implementation of the Interactive Connectivity Establishment (ICE) protocol in C for POSIX platforms (including Linux and Apple macOS) and Microsoft Windows.",
   "homepage": "https://github.com/paullouisageneau/libjuice",
   "features": {
--- a/versions/l-/libjuice.json
+++ b/versions/l-/libjuice.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "fd8be0f1271d562c1667cd64c2d700c272dc9ee0",
+      "version-string": "v0.7.1",
+      "port-version": 0
+    },
+    {
       "git-tree": "e7ef5f1f7d883e5f8a7dcb52ea5821730281e375",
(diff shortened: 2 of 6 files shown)


Nemirtingas:
I'll fix this, I forgot to update the control file.

ports/libjuice/vcpkg.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "libjuice",
+  "version-string": "v0.7.1",
```
JonLiu1993, line 3: @Nemirtingas ,Because its version is a progressive combination of consecutive numbers, we should use "version" instead of "version-string"

versions/l-/libjuice.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "f3e297509756bb3afa9577b985f33c26f6867f29",
+      "version-string": "v0.7.1",
```
JonLiu1993, line 5: we should use "version" instead of "version-string"

Nemirtingas: Hi thanks, should I put something like:
`"version": "0.7.1"` or `"version": 071` ?

Nemirtingas:
Hi thanks, should I put something like:
`"version": "0.7.1"` or `"version-semver": "0.7.1"` ?

https://github.com/microsoft/vcpkg/pull/17077#discussion_r607447541

I'm quite lost, version formats everywhere.

JonLiu1993:
> Hi thanks, should I put something like:
> `"version": "0.7.1"` or `"version-semver": "0.7.1"` ?
> 
> [#17077 (comment)](https://github.com/microsoft/vcpkg/pull/17077#discussion_r607447541)
> 
> I'm quite lost, version formats everywhere.

Here you should use "version", if you are confused about this, you can refer to the detailed meaning and difference [here](https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md)

Nemirtingas:
> 
> 
> > Hi thanks, should I put something like:
> > `"version": "0.7.1"` or `"version-semver": "0.7.1"` ?
> > [#17077 (comment)](https://github.com/microsoft/vcpkg/pull/17077#discussion_r607447541)
> > I'm quite lost, version formats everywhere.
> 
> Here you should use "version", if you are confused about this, you can refer to the detailed meaning and difference [here](https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md)

Ok then, but why does NancyLi1013 want to use version-semver while we have the same version scheme ? e.g. X.Y.Z ?

JonLiu1993:
> > > Hi thanks, should I put something like:
> > > `"version": "0.7.1"` or `"version-semver": "0.7.1"` ?
> > > [#17077 (comment)](https://github.com/microsoft/vcpkg/pull/17077#discussion_r607447541)
> > > I'm quite lost, version formats everywhere.
> > 
> > 
> > Here you should use "version", if you are confused about this, you can refer to the detailed meaning and difference [here](https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md)
> 
> Ok then, but why does NancyLi1013 want to use version-semver while we have the same version scheme ? e.g. X.Y.Z ?

Generally speaking, if every number of the version has meaning or there is an api change in the version update, you need to use version-semver. If the number of the version has no meaning, it is incremental or there is no API change, you need to use version, if it is generally a date. Need to use version-date

Nemirtingas:
> Generally speaking, if every number of the version has meaning or there is an api change in the version update, you need to use version-semver. If the number of the version has no meaning, it is incremental or there is no API change, you need to use version, if it is generally a date. Need to use version-date

Thanks for the explanation, I'll keep this in mind.

Issue #6045, opened 2019-04-11
Collection of common port issues


Neumann-A:
Trying to get a list together of common port issues which have been seen in the past and have been fixed or are still in need of a fix (maybe the fix is already known and it it just must be applied):

__Oberserved Problems?__
- Changing the debug name of a library, find_package broken: #6014 
- Changing the exported symbols: #6043 
- Wrong library linkage: #5543, #5540 
- packageConfig.cmake in wrong directory: #4783
- package*.cmake different for release and debug and release version is overwritten by debug version: (probably also mentioned in #5543 or add a better reference)

__What should new ports always test?__
(after of course building successfully)
- Port\Package should be correctly be found by `find_package`
- Linkage of packages should be correct for debug/release

Should I ping people for more visibility?

Please add comments so that I can extend the list (and some of you have probably more experience than me fixing ports).

vicroms:
There's also the case when consuming libraries cannot find a dependency due to `static` and `shared` libraries having different names. 

"Fixable" by using a `vcpkg-cmake-wrapper.cmake` file, e.g.: [rocksdb](https://github.com/Microsoft/vcpkg/blob/master/ports/rocksdb/vcpkg-cmake-wrapper.cmake).

Neumann-A:
@vicroms 
> "Fixable" by using a vcpkg-cmake-wrapper.cmake file, e.g.: rocksdb.

Please don't do that. Target RocksDB::rocksdb is probably intended to be always a static target while RocksDB::rocksdb-shared should always be a shared target. I had the same issue with hdf5 once and somebody thought it would be a good idea to redefine the targets. In x64-windows hdf5:hdf5-static was mapped to hdf5:hdf5-shared while in x64-windows-static it was the other way around. If you have problems with consuming packages not thinking about how to link correctly to rocksdb in all cases patch their behavior instead. If rocksdb is building shared as well as static targets (in e.g. x64-windows) do not deactivate that in vcpkg because of the missing target you generate with that. (Furthermore default behavior is unnecessary changed.)

vicroms:
@Neumann-A  thanks for the explanation. I haven't actually done it myself, it's just something I've noticed while reviewing PRs.

I agree, that we shouldn't change the default behavior of libraries, however producing both shared and static libs will fail during post-build validation. We could probably add a variable that should be explicitly set for libraries that produce both libraries, to bypass this post-build validation.

ras0219-msft:
[...]

PR #17078, opened 2021-04-04
[llvm] fix debug build

abique:
LLVM did no build in debug because the include directory was removed.
I'm not sure this behavior is correct (llvm port). Yet I did add the policy so it builds.

### DIFF
--- a/ports/llvm/portfile.cmake
+++ b/ports/llvm/portfile.cmake
@@ -254,6 +254,8 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "debug")
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/tools)
+    
+    set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
 endif()
 
 # LLVM still generates a few DLLs in the static build:
--- a/ports/llvm/CONTROL
+++ b/ports/llvm/CONTROL
@@ -1,6 +1,6 @@
 Source: llvm
 Version: 11.0.0
-Port-Version: 2
+Port-Version: 3
 Homepage: https://llvm.org/
 Description: The LLVM Compiler Infrastructure
 Supports: !uwp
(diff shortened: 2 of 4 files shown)


PhoebeHui:
@abique, could you report an issue for this? I'm unsure what failures that you encountered.

abique:
@PhoebeHui the build failed because the include directory war empty.

ports/llvm/portfile.cmake
```diff
@@ -254,10 +254,12 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "debug")
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
     file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/tools)
+    
+    set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
abique, line 258: This allows the debug build to complete with an empty include folder.

yurybura:
> @PhoebeHui the build failed because the include directory war empty.

Hello @abique, could you tell which directory is empty? Could you provide also install command to reproduce your issue?
All known empty directories should be removed. You can find the following code:
https://github.com/microsoft/vcpkg/blob/eb1dac2331f5f9a0e68546867c8b01fc0f5b7fd9/ports/llvm/portfile.cmake#L199-L203
https://github.com/microsoft/vcpkg/blob/eb1dac2331f5f9a0e68546867c8b01fc0f5b7fd9/ports/llvm/portfile.cmake#L209-L213

abique:
Build in debug, see the portfile it is doing:
`file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)`

PhoebeHui:
@abique, Removing debug/include folder is by design in vcpkg, it will check the folder in post-build validation step. When you use the port, it will use the headers from include folder.

```
Include files should not be duplicated into the /debug/include directory. If this cannot be disabled in the project cmake, use
```

abique:
In my case I did build the debug with a custom toolchain/triplet with only debug.

PhoebeHui:
@abique, this is a known common issue in vcpkg see https://github.com/microsoft/vcpkg/issues/6045.

I would suggest to use overlay or custom ports instead.

abique:
Is it unsupported to install only the debug version of a port?

PhoebeHui:
@abique, only install debug or release version of a port may all have issues, or not, it depends what the port that was used. We hope this issue would be solved in future, however, currently we would suggest to solve this issue by using overlay or custom ports.

abique:
I understand.
I close this issue.
Thank you :)

PR #17083, opened 2021-04-04
[vcpkg baseline][gdal] Fix configure error on OSX

JackBoosY:
Fixes #16990.

PR #17085, opened 2021-04-04
[vcpkg/scripts/make] add compiler tools to PATH

huangqinjin:

### DIFF
--- a/scripts/cmake/vcpkg_configure_make.cmake
+++ b/scripts/cmake/vcpkg_configure_make.cmake
@@ -329,15 +329,22 @@ function(vcpkg_configure_make)
         endmacro()
 
         set(CONFIGURE_ENV "V=1")
-        # Remove full filepaths due to spaces (compiler needs to be on path)
+        # Remove full filepaths due to spaces and prepend filepaths to PATH (cross-compiling tools are unlikely on path by default)
+        set(_toolpaths)
         set(progs VCPKG_DETECTED_CMAKE_C_COMPILER VCPKG_DETECTED_CMAKE_CXX_COMPILER VCPKG_DETECTED_CMAKE_AR
                   VCPKG_DETECTED_CMAKE_LINKER VCPKG_DETECTED_CMAKE_RANLIB VCPKG_DETECTED_CMAKE_OBJDUMP
                   VCPKG_DETECTED_CMAKE_STRIP VCPKG_DETECTED_CMAKE_NM VCPKG_DETECTED_CMAKE_DLLTOOL VCPKG_DETECTED_CMAKE_RC_COMPILER)
         foreach(prog IN LISTS progs)
             if(${prog})
+                get_filename_component(path "${${prog}}" DIRECTORY)
                 get_filename_component(${prog} "${${prog}}" NAME)
+                if(NOT path IN_LIST _toolpaths)
+                    list(APPEND _toolpaths ${path})
+                endif()
             endif()
         endforeach()
+        vcpkg_add_to_path(PREPEND ${_toolpaths})
+
         if (_csc_AUTOCONFIG OR _csc_USE_WRAPPERS) # without autotools we assume a custom configure script which correctly handles cl and lib. Otherwise the port needs to set CC|CXX|AR and probably CPP
             _vcpkg_append_to_configure_environment(CONFIGURE_ENV CPP "compile ${VCPKG_DETECTED_CMAKE_C_COMPILER} -E")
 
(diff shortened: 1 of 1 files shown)


JackBoosY:
cc @Neumann-A

SamuelMarks:
[offtopic]
So is this to compile the GNU `make` binary? - Or to take non CMake projects and enable them to work with vcpkg? - Or to aid in rewriting autotools incl. `make` projects into CMake ones?

PS: Looks like they fixed Windows again http://git.savannah.gnu.org/cgit/make.git/commit/?id=1ae90c030d5ffc79f74e3938880445a6b8b2412e
[/offtopic]

Neumann-A:
@SamuelMarks This: Or to take non CMake projects and enable them to work with vcpkg?

`vcpkg_configure_make` takes the cmake toolchain and setups everything so that configure/make works with the cmake toolchain.

Why is it assumed that the compiler is on the path:
**SPACES** 
autotools are really bad in handling spaces in paths. The normal way vcpkg is setup on windows uses vcvars which puts the compiler on the path. That is why it is assumed here.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -329,13 +329,15 @@ function(vcpkg_configure_make)
...
-        # Remove full filepaths due to spaces (compiler needs to be on path)
+        # Remove full filepaths due to spaces and prepend filepaths to PATH (cross-compiling tools are unlikely on path by default)
         set(progs VCPKG_DETECTED_CMAKE_C_COMPILER VCPKG_DETECTED_CMAKE_CXX_COMPILER VCPKG_DETECTED_CMAKE_AR
                   VCPKG_DETECTED_CMAKE_LINKER VCPKG_DETECTED_CMAKE_RANLIB VCPKG_DETECTED_CMAKE_OBJDUMP
                   VCPKG_DETECTED_CMAKE_STRIP VCPKG_DETECTED_CMAKE_NM VCPKG_DETECTED_CMAKE_DLLTOOL VCPKG_DETECTED_CMAKE_RC_COMPILER)
         foreach(prog IN LISTS progs)
             if(${prog})
+                get_filename_component(path "${${prog}}" DIRECTORY)
                 get_filename_component(${prog} "${${prog}}" NAME)
+                vcpkg_add_to_path(PREPEND ${path})
```
Neumann-A, line 340: this needs some kind of check if ${prog} is already on the path. Otherwise this adds the same path multiple times which will run into PATH length problems.

huangqinjin: done

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -329,15 +329,22 @@ function(vcpkg_configure_make)
...
         foreach(prog IN LISTS progs)
             if(${prog})
+                get_filename_component(path "${${prog}}" DIRECTORY)
                 get_filename_component(${prog} "${${prog}}" NAME)
+                if(NOT path IN_LIST _toolpaths)
+                    list(APPEND _toolpaths ${path})
+                endif()
             endif()
         endforeach()
+        vcpkg_add_to_path(PREPEND ${_toolpaths})
```
Neumann-A, line 346: This check is still incomplete PATH can still have `${_toolpaths}` already on the path. You need to use `find_program` with `NO_DEFAULT_PATH` and `PATHS ENV PATH` here and check weather or not the paths for the tool agree. If not you need to put the tool on the path.

huangqinjin: But how to check if the found program is exactly the same as prog? I expect that all the tools are in same path, so only one path is prepended to PATH.

huangqinjin: Simply compare two path strings. Works fine.

Neumann-A: > Simply compare two path strings. Works fine.

you need to compare the path string against the paths also contained in ENV{PATH} and on windows they use \ instead of /. So comparing them is not that easy.

huangqinjin: `get_filename_component` returns cmake path which uses forward slash /. And `prog` should also be cmake path since it is cmake detected.

huangqinjin: `find_program` also returns cmake path.

Neumann-A: I missed the `find_program` change. Nevermind.

JackBoosY (approved):

JackBoosY:
Waiting for #17083 merge.

PR #17088, opened 2021-04-04
[vcpkg/scripts/make] cross compile on Windows

huangqinjin:
Support Android triplets on Windows.

Tested ports: libbacktrace.

Depend on: ~#16927~ (Merged), #17084 , ~#17085~ (Merged), #17086

PR #17089, opened 2021-04-04
[vcpkg/scripts/make] avoid versioning for android

huangqinjin:


JackBoosY:
cc @Neumann-A @vicroms

Neumann-A:
Looks valid? Don't know any details about this. But we probably need some kind of `VCPKG_<variable>` to pass additional libtool flags?

JackBoosY:
Waiting for #17083 merge.

JackBoosY:
Please get failure logs [here](https://dev.azure.com/vcpkg/public/_build/results?buildId=51626&view=artifacts&pathAsName=false&type=publishedArtifacts).

huangqinjin:
This PR is only for android triplets, the failure log seems irrelevant.

JackBoosY:
Depends on https://github.com/microsoft/vcpkg/pull/17227

JackBoosY (approved):

dg0yt:
@huangqinjin Which ports did you test this with? The flag is passed into `configure` here. `configure` doesn't use `libtool`, it actually *generates* it. Now it is starting to create problems: #31332.

huangqinjin:
@dg0yt libbacktrace. Actually libtool is totally unusable currently. There is an unfinished work in https://github.com/microsoft/vcpkg/pull/17088. If `-avoid-version` is causing problem, please simply revert this commit.

dg0yt:
> libbacktrace. 

I will try this. Thanks.

> Actually libtool is totally unusable currently. There is an unfinished work in #17088. If `-avoid-version` is causing problem, please simply revert this commit.

Note that the static android triplets do work on windows now. I had to fix vpckg's use of msys2 for this.

huangqinjin:
> Note that the static android triplets do work on windows now.

Yeah, only shared libraries do not work, as libtool strips out `--target` link flags, that is what I fixed in https://github.com/microsoft/vcpkg/pull/17088 (though abandoned finally).

PR #17137, opened 2021-04-07
[vcpkg] Fix lib uuid handling for x64-mingw-dynamic

dg0yt:
**Describe the pull request**

- What does your PR fix?
  This PR fixes the build of gettext (and possibly other packages) on Windows when triplet, host triplet and cmake are mingw. Here, the cmake-detected standard libraries come with `-luuid`, not `uuid`, so it is not removed from LIBS by vcpkg. However, libtool refuses to create shared libraries for mingw because there is no shared library uuid. This change explicitly instructs libtool (or gcc) to statically link uuid. 

  These flags do not work with plain `ld`.

  -/-

### DIFF
--- a/scripts/cmake/vcpkg_configure_make.cmake
+++ b/scripts/cmake/vcpkg_configure_make.cmake
@@ -545,6 +545,11 @@ function(vcpkg_configure_make)
         list(TRANSFORM ALL_LIBS_LIST REPLACE "^(${_lprefix})" "")
     endif()
     list(JOIN ALL_LIBS_LIST " ${_lprefix}" ALL_LIBS_STRING)
+    if(VCPKG_TARGET_IS_MINGW AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+        # libtool must be told explicitly that there is no dynamic linkage for uuid.
+        # The "-Wl,..." syntax is understood by libtool and gcc, but no by ld.
+        string(REPLACE " -luuid" " -Wl,-Bstatic,-luuid,-Bdynamic" ALL_LIBS_STRING "${ALL_LIBS_STRING}")
+    endif()
 
     if(ALL_LIBS_STRING)
         set(ALL_LIBS_STRING "${_lprefix}${ALL_LIBS_STRING}")


Neumann-A:
uuid probably does not need to be removed if VCPKG_TARGET_IS_MINGW. uuid is just removed due to spurious CI errors where uuid seems not to be available

dg0yt:
Hm. I found these errors quite reproducible wih `x64-mingw-dynamic`, with some output like this:

~~~
version.lo xsize.lo osdep.lo intl-compat.lo windows-mutex.lo windows-rwlock.lo windows-recmutex.lo windows-once.lo libintl.res.lo
/bin/sh ../libtool  --tag=CC   --mode=link x86_64-w64-mingw32-gcc.exe -fvisibility=hidden -O3 -Wl,--export-all-symbols -liconv  -ladvapi32 -no-undefined -export-symbols-regex '^([^g]|g[^l]|gl[^w]|glw[^t]|glwt[^h]|glwth[^r]|glwthr[^e]|glwthre[^a]|glwthrea[^d]).*' -version-info 10:0:2 -rpath /D/a/1/s/installed/x64-mingw-dynamic/lib -LD:/a/1/s/installed/x64-mingw-dynamic/lib  -Wl,--disable-auto-import -o libintl.la  bindtextdom.lo dcgettext.lo dgettext.lo gettext.lo finddomain.lo hash-string.lo loadmsgcat.lo localealias.lo textdomain.lo l10nflist.lo explodename.lo dcigettext.lo dcngettext.lo dngettext.lo ngettext.lo plural.lo plural-exp.lo localcharset.lo threadlib.lo lock.lo relocatable.lo langprefs.lo localename.lo localename-table.lo log.lo printf.lo setlocale.lo setlocale-lock.lo setlocale_null.lo version.lo xsize.lo osdep.lo intl-compat.lo windows-mutex.lo windows-rwlock.lo windows-recmutex.lo windows-once.lo libintl.res.lo -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32

*** Warning: linker path does not have real file for library -luuid.
*** I have the capability to make that library automatically link in when
*** you link to this library.  But I can only do this if you have a
*** shared version of the library, which you do not appear to have
*** because I did check the linker path looking for a file starting
*** with libuuid and none of the candidates passed a file format test
*** using a file magic. Last file checked: C:/msys64/mingw64/x86_64-w64-mingw32/lib/libuuid.a
*** The inter-library dependencies that have been dropped here will be
*** automatically added whenever a program is linked with this library
*** or is declared to -dlopen it.

*** Since this library must not contain undefined symbols,
*** because either the platform does not support them or
*** it was explicitly requested with -no-undefined,
*** libtool will only create a static version of it.
~~~

I honestly couldn't take much out of the search engine results "linker path does not have real file for library" "luuid" "mingw" except for some issue related to static vs dynamic linking.

Neumann-A:
It basically means it cannot find uuid for some reason.

dg0yt:
Sure, `libtool` cannot find lib uuid, in terms of eventually achieving usage of some libuuid.dll at runtime. So this is the situation:
- These requests for uuid come not from the port but from cmake via `..._STANDARD_LIBRARIES`.
- It is libtool which insists on an import library for dynamic linkage, not a plain static library.
- In the case of `uuid.lib` (from cmake's `Modules/Platform/Windows-MSVC.cmake`), uuid is removed by vcpkg, due to "spurious CI errors where uuid seems not to be available".
- In the case of `-luuid` (from cmake's `Modules/Platform/Windows-GNU.cmake`), removing uuid is not considered appropriate.(?)

However, there is a pure static archive, reported as "Last file checked". And I eventually figured out how to make this acceptable for libtool.
~~~
    if(VCPKG_TARGET_IS_MINGW AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
        # no dynamic libuuid on MinGW
        string(REPLACE " -luuid" " -Wl,-Bstatic,-luuid,-Bdynamic" ALL_LIBS_STRING "${ALL_LIBS_STRING}")
    endif()
~~~
Will update the PR.

dg0yt:
>  -Wl,-Bstatic,-luuid,-Bdynamic

On the downside, this syntax is understood by gcc or libtool, but not by ld. So it depends on how the linker is invoked.

dg0yt:
As a heuristic for safely assuming usage of libtool, one might probably look for 
`REQUIRES_AUTOCONFIG OR EXISTS "${SRC_DIR}/build-aux/lt-main.sh"`.

Neumann-A:
I checked `Uuid.lib` in  `C:\Program Files (x86)\Windows Kits\10\Lib\10.0.19041.0\um\x64` and as it seems it is a static library. So this change might be generally valid for windows targets?

dg0yt:
> So this change might be generally valid for windows targets?

I cannot answer anything for MSVC. And even when the answer is yes, it probably needs different flags than for libtool.

dg0yt:
@JackBoosY I don't think the "requires:author-response" tag is still justified:

Other than in the initial form of the PR, libuuid is not removed but marked to be linked statically for mingw. This is tested with some packages. So initial concerns regarding the removal are resolved.

Formally, there remain concerns whether the syntax for linking statically is understood by all ports. But without CI support this cannot easily be answered upfront - someone has to stand up point to a port which breaks (for mingw) with this change.

Last not least, a question was raised about Uuid.lib in general. But this is out of scope for this PR/author.

JackBoosY:
@dg0yt I will handle this PR in next week.

JackBoosY:
Sorry for late, I will handle it this week.

dg0yt:
Now the original (overwritten) commit was duplicated in #17769 and #17780.  Can you please help to avoid more duplicate work?

JackBoosY:
libmicrohttpd:x64-osx
```
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
ld: symbol(s) not found for architecture x86_64
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
ld: symbol(s) not found for architecture x86_64
make[4]: *** [minimal_example] Error 1
make[4]: *** Waiting for unfinished jobs....
ld: symbol(s) not found for architecture x86_64
make[4]: *** [chunked_example] Error 1
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
      _osstatus_error in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
ld: symbol(s) not found for architecture x86_64
      _add_system_trust in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
make[4]: *** [benchmark] Error 1
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
      _osstatus_error in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
      _osstatus_error in libgnutls.a(certs.o)
make[4]: *** [benchmark_https] Error 1
  "_SecItemExport", referenced from:
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
      _add_system_trust in libgnutls.a(certs.o)
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
      _add_system_trust in libgnutls.a(certs.o)
ld: symbol(s) not found for architecture x86_64
Undefined symbols for architecture x86_64:
  "_SecCopyErrorMessageString", referenced from:
ld: symbol(s) not found for architecture x86_64
      _osstatus_error in libgnutls.a(certs.o)
ld: symbol(s) not found for architecture x86_64
  "_SecItemExport", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
  "_SecTrustSettingsCopyCertificates", referenced from:
      _add_system_trust in libgnutls.a(certs.o)
```
Is this related to your changes?

dg0yt:
> libmicrohttpd:x64-osx
> ...
> Is this related to your changes?

Obviously not: my change is within `if(VCPKG_TARGET_IS_MINGW...`.
But note #17636, #17653 for libmicrohttpd.

JackBoosY (approved):

JackBoosY:
Waiting for https://github.com/microsoft/vcpkg/pull/17853 merge.

dg0yt:
#17853 was replaced by #17653, and that one was merged now.

past-due:
I can confirm that this fixes the issue in my tests. 👍

past-due (approved):

Neumann-A:
Hmm could you test some port combination with a custom triplet with the following:
- 2 make based ports (A,B) 
- A depends on B 
- linkage of B was changed to static in the triplet
- linkage of A is dynamic.

I want to know if libtool in A also chokes on B in this case. We might need to simply ignore all libtool stupidity and allow passing of all possible lib files.

JackBoosY (approved):

dg0yt:
@Neumann-A Tested successfully with libmicrohttpd (A) in a modified triplet with static libiconv, gettext (B):
~~~
/bin/sh ../../libtool  --tag=CC   --mode=link x86_64-w64-mingw32-gcc.exe -fvisibility=hidden    -O3 -fno-strict-aliasing -export-dynamic -no-undefined  -Wl,--output-def,.libs/libmicrohttpd.def -XCClinker -static-libgcc -version-info 62:0:50 -LD:/msys64/home/dg0yt/vcpkg/installed/x64-mingw-mixed/lib  -o libmicrohttpd.la -rpath /D/msys64/home/dg0yt/vcpkg/installed/x64-mingw-mixed/lib libmicrohttpd_la-connection.lo libmicrohttpd_la-reason_phrase.lo libmicrohttpd_la-daemon.lo libmicrohttpd_la-internal.lo libmicrohttpd_la-memorypool.lo libmicrohttpd_la-mhd_mono_clock.lo libmicrohttpd_la-sysfdsetsize.lo libmicrohttpd_la-mhd_str.lo libmicrohttpd_la-mhd_sockets.lo libmicrohttpd_la-mhd_itc.lo libmicrohttpd_la-mhd_compat.lo libmicrohttpd_la-response.lo  libmicrohttpd_la-mhd_threads.lo  libmicrohttpd_la-postprocessor.lo libmicrohttpd_la-digestauth.lo libmicrohttpd_la-md5.lo libmicrohttpd_la-sha256.lo libmicrohttpd_la-basicauth.lo libmicrohttpd_la-base64.lo    libmicrohttpd_la-microhttpd_dll_res.lo -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -Wl,-Bstatic,-luuid,-Bdynamic -lcomdlg32 -ladvapi32 -lws2_32
libtool: link: x86_64-w64-mingw32-gcc.exe -shared  .libs/libmicrohttpd_la-connection.o .libs/libmicrohttpd_la-reason_phrase.o .libs/libmicrohttpd_la-daemon.o .libs/libmicrohttpd_la-internal.o .libs/libmicrohttpd_la-memorypool.o .libs/libmicrohttpd_la-mhd_mono_clock.o .libs/libmicrohttpd_la-sysfdsetsize.o .libs/libmicrohttpd_la-mhd_str.o .libs/libmicrohttpd_la-mhd_sockets.o .libs/libmicrohttpd_la-mhd_itc.o .libs/libmicrohttpd_la-mhd_compat.o .libs/libmicrohttpd_la-response.o .libs/libmicrohttpd_la-mhd_threads.o .libs/libmicrohttpd_la-postprocessor.o .libs/libmicrohttpd_la-digestauth.o .libs/libmicrohttpd_la-md5.o .libs/libmicrohttpd_la-sha256.o .libs/libmicrohttpd_la-basicauth.o .libs/libmicrohttpd_la-base64.o .libs/libmicrohttpd_la-microhttpd_dll_res.o   -LD:/msys64/home/dg0yt/vcpkg/installed/x64-mingw-mixed/lib -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -lcomdlg32 -ladvapi32 -lws2_32  -O3 -Wl,--output-def -Wl,.libs/libmicrohttpd.def -static-libgcc -Wl,-Bstatic -Wl,-luuid -Wl,-Bdynamic   -o .libs/libmicrohttpd-12.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker .libs/libmicrohttpd.dll.a
~~~
i.e. libtool changes the order of arguments, but nothing bad happens.

Based on my pre-vcpkg mingw experience, I did a more radical test now with the pristine x64-mingw-dynamic triplet:
Completely disable the passing of CMake's CMAKE_<LANG>_STANDARD_LIBRARIES via LIBS, instead of just fixing uuid. And it works for microhttpd:
~~~
/bin/sh ../../libtool  --tag=CC   --mode=link x86_64-w64-mingw32-gcc.exe -fvisibility=hidden    -O3 -fno-strict-aliasing -export-dynamic -no-undefined  -Wl,--output-def,.libs/libmicrohttpd.def -XCClinker -static-libgcc -version-info 62:0:50 -LD:/msys64/home/dg0yt/vcpkg/installed/x64-mingw-dynamic/lib  -o libmicrohttpd.la -rpath /D/msys64/home/dg0yt/vcpkg/installed/x64-mingw-dynamic/lib libmicrohttpd_la-connection.lo libmicrohttpd_la-reason_phrase.lo libmicrohttpd_la-daemon.lo libmicrohttpd_la-internal.lo libmicrohttpd_la-memorypool.lo libmicrohttpd_la-mhd_mono_clock.lo libmicrohttpd_la-sysfdsetsize.lo libmicrohttpd_la-mhd_str.lo libmicrohttpd_la-mhd_sockets.lo libmicrohttpd_la-mhd_itc.lo libmicrohttpd_la-mhd_compat.lo libmicrohttpd_la-response.lo  libmicrohttpd_la-mhd_threads.lo  libmicrohttpd_la-postprocessor.lo libmicrohttpd_la-digestauth.lo libmicrohttpd_la-md5.lo libmicrohttpd_la-sha256.lo libmicrohttpd_la-basicauth.lo libmicrohttpd_la-base64.lo    libmicrohttpd_la-microhttpd_dll_res.lo -lws2_32
libtool: link: x86_64-w64-mingw32-gcc.exe -shared  .libs/libmicrohttpd_la-connection.o .libs/libmicrohttpd_la-reason_phrase.o .libs/libmicrohttpd_la-daemon.o .libs/libmicrohttpd_la-internal.o .libs/libmicrohttpd_la-memorypool.o .libs/libmicrohttpd_la-mhd_mono_clock.o .libs/libmicrohttpd_la-sysfdsetsize.o .libs/libmicrohttpd_la-mhd_str.o .libs/libmicrohttpd_la-mhd_sockets.o .libs/libmicrohttpd_la-mhd_itc.o .libs/libmicrohttpd_la-mhd_compat.o .libs/libmicrohttpd_la-response.o .libs/libmicrohttpd_la-mhd_threads.o .libs/libmicrohttpd_la-postprocessor.o .libs/libmicrohttpd_la-digestauth.o .libs/libmicrohttpd_la-md5.o .libs/libmicrohttpd_la-sha256.o .libs/libmicrohttpd_la-basicauth.o .libs/libmicrohttpd_la-base64.o .libs/libmicrohttpd_la-microhttpd_dll_res.o   -LD:/msys64/home/dg0yt/vcpkg/installed/x64-mingw-dynamic/lib -lws2_32  -O3 -Wl,--output-def -Wl,.libs/libmicrohttpd.def -static-libgcc   -o .libs/libmicrohttpd-12.dll -Wl,--enable-auto-image-base -Xlinker --out-implib -Xlinker .libs/libmicrohttpd.dll.a
~~~
There is still `-lws2_32`. This appears in microhttpd's config.log first when testing for `-pthread`.

So why generally pass CMake's implicit libraries at all (for mingw)?

Neumann-A:
> ested successfully with libmicrohttpd (A) in a modified triplet with static libiconv, gettext (B):

either I am blind or cannot find it linking against iconv or intl in the call

> Completely disable the passing of CMake's CMAKE__STANDARD_LIBRARIES via LIBS, instead of just fixing uuid
So why generally pass CMake's implicit libraries at all (for mingw)?

These might be implicit on MINGW and note that we are not passing https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_IMPLICIT_LINK_LIBRARIES.html but 
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LIBRARIES.html

Passing those libraries is required for some ports on windows and all ports on uwp (and is different for UWP).

dg0yt:
> > ested successfully with libmicrohttpd (A) in a modified triplet with static libiconv, gettext (B):
> 
> either I am blind or cannot find it linking against iconv or intl in the call

Good point. 
(I picked a port that didn't even properly use its dependencies? 
Well I started with depend-info of all ports that use `vcpkg_configure_make` and checked the direct dependencies if they use `vcpkg_configure_make`, too. It turned out there are not so many pairs where both ports would actually use `vcpkg_configure_make` for mingw. And on my way I had to merge another pending PR (no wrappers for mingw), and I was reminded to not set VCPKG_DEFAULT_HOST_TRIPLET because this script would omit passing the autoconf triplet...
Maybe it is really to early to worry about mixing different linkage for mingw?)

> > Completely disable the passing of CMake's CMAKE__STANDARD_LIBRARIES via LIBS, instead of just fixing uuid
> > So why generally pass CMake's implicit libraries at all (for mingw)?
> 
> These might be implicit on MINGW and note that we are not passing https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_IMPLICIT_LINK_LIBRARIES.html but
> https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LIBRARIES.html

From a cmake user's POV, CMake (not the compiler) adds them implicitly. (CMake user experience.) 
I have difficult to find the "standard" which nominates just these libraries, but I'm not a windows expert.

> Passing those libraries is required for some ports on windows and all ports on uwp (and is different for UWP).

I don't want to dispute that it may be necessary for other triplets, possibly for all ports. But I wonder how many ports need these libs explicitly for *mingw*. Outside the vcpkg world, autotools with mingw has been used for a while, and I have difficulties to find references about these libraries being necessary to be given separately to `configure`. (Autotools user experience.)

dg0yt:
I regard the following as somewhat off-topic to the original PR (which is to make *-mingw-dynamic triplets usable at all), but I want to answer the question by @Neumann-A.

> > > tested successfully with libmicrohttpd (A) in a modified triplet with static libiconv, gettext (B):
> > 
> > 
> > either I am blind or cannot find it linking against iconv or intl in the call
> 
> Good point.

Actually I fell over the consequences of static linkage and transitive dependencies. The reported build didn't use gettext because libmicrohttd's `configure` only tested linking `installed/.../lib/libintl.a` without adding (static) liconv.
Passing `-liconv` in `LIBS` doesn't help because that test adds `installed/.../lib/libintl.a` after `$LIBS`.
Now intl and iconv are a bit special. But I'm not optimistic if it could work better in general, with or without use of pkgconfig.

> I want to know if libtool in A also chokes on B in this case.

As a simpler case, I looked at building dynamic gettext (=A) with static libiconv (=B).
Here, "*libtool in A also chokes on B*", so this should answer the question.
~~~
/bin/bash ../libtool  --tag=CC   --mode=link x86_64-w64-mingw32-gcc -fvisibility=hidden -g -Wl,--export-all-symbols -liconv  -ladvapi32 -no-undefined -export-symbols-regex '^([^g]|g[^l]|gl[^w]|glw[^t]|glwt[^h]|glwth[^r]|glwthr[^e]|glwthre[^a]|glwthrea[^d]).*' -version-info 10:0:2 -rpath /home/kpa/Projekte/vcpkg/vcpkg/installed/x64-mingw-mixed/debug/lib -L/home/kpa/Projekte/vcpkg/vcpkg/installed/x64-mingw-mixed/debug/lib  -Wl,--disable-auto-import -o libintl.la  bindtextdom.lo dcgettext.lo dgettext.lo gettext.lo finddomain.lo hash-string.lo loadmsgcat.lo localealias.lo textdomain.lo l10nflist.lo explodename.lo dcigettext.lo dcngettext.lo dngettext.lo ngettext.lo plural.lo plural-exp.lo localcharset.lo threadlib.lo lock.lo relocatable.lo langprefs.lo localename.lo localename-table.lo log.lo printf.lo setlocale.lo setlocale-lock.lo setlocale_null.lo version.lo xsize.lo osdep.lo intl-compat.lo windows-mutex.lo windows-rwlock.lo windows-recmutex.lo windows-once.lo libintl.res.lo -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -Wl,-Bstatic,-luuid,-Bdynamic -lcomdlg32 -ladvapi32

*** Warning: linker path does not have real file for library -liconv.
*** I have the capability to make that library automatically link in when
*** you link to this library.  But I can only do this if you have a
*** shared version of the library, which you do not appear to have
*** because I did check the linker path looking for a file starting
*** with libiconv and none of the candidates passed a file format test
*** using a file magic. Last file checked: /home/kpa/Projekte/vcpkg/vcpkg/installed/x64-mingw-mixed/debug/lib/libiconv.a
*** The inter-library dependencies that have been dropped here will be
*** automatically added whenever a program is linked with this library
*** or is declared to -dlopen it.

*** Since this library must not contain undefined symbols,
*** because either the platform does not support them or
*** it was explicitly requested with -no-undefined,
*** libtool will only create a static version of it.
~~~

leading to:
~~~
CMake Error at scripts/cmake/vcpkg_build_make.cmake:204 (message):
  libtool could not find a file being linked against!
~~~

I don't feel like "simply ignore all libtool stupidity and allow passing of all possible lib files" is really the right thing to do in general. After all it works fairly well for pure static or dynamic configurations (apart from uuid). User-defined mixed configurations are a different story.

Neumann-A:
> I don't feel like "simply ignore all libtool stupidity and allow passing of all possible lib files" is really the right thing to do in general.

The question is does libtool actually do anything relevant? If it only runs `objdump/file` to figure out if the libs have the correct architecture for linkage or it is just a better `find_library` which means it isn't required and libtool should just believe us.  

> After all it works fairly well for pure static or dynamic configurations (apart from uuid)

I read somewhere that uuid on windows is always a static library, so it makes sense that it errors in dynamic builds.
The fact that it might make problems with custom triplets means we might need to unconditionally add `lt_cv_deplibs_check_method=pass_all` (it already has problems for arm and uwp builds....) to the configure options.

JackBoosY:
Since `-luuid` is provided by cmake, is this a cmake bug?
I prefer to fix this issue in the toolchain file.

dg0yt:
> Since `-luuid` is provided by cmake, is this a cmake bug?

I don't think it can be counted as a cmake bug. IMO it is a convenience feature in cmake so that users don't have to explicitly specify frequently used (or actually required) libraries. It is used for linking executables and shared libraries, but cmake doesn't require these libraries to be shared libraries.
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_STANDARD_LIBRARIES.html

Now here we are dealing with autoconf and libtool. Generally, it is convenient and fair to forward the libraries proposed by cmake. I'm not aware of another source of libraries which are frequently used or required for building for a particular triplet. However, in the case of uuid, we are faced with the fact that libtool normally refuses to accept a static candidate library when creating a shared library. **This problem is specific to vcpkg** because noone else attempts to blindly forward cmake's list of libs to autotools.

I think what is needed is an "informed forwarding" based on the triplet. Options:
- Don't forward at all, and fix broken ports. (Should work for mingw because this is how autotools are normally used.)
- Don't forward what is known to break. (Here, it means omitting uuid. Proposed a few times.)
- Adjust the forwarded lib so that it is expected to work. (Current state of the PR.)

There is another option mentioned by @Neumann-A which is:
- Disable the libtool check and let libtool accept static libs when creating shared libs. (Should work for uuid, but has more general effects on linking.)

past-due:
Personally, I prefer the current state of this PR over disabling the libtool check altogether.

JackBoosY:
After some internal discussion, we think we can accept this PR.

strega-nil-ms:
Thanks @dg0yt :)

PR #17156, opened 2021-04-08
[PCL] add a way to disable SSE

fran6co:
PCL adds `-march=native` and other flags to the whole project making it harder to make binaries that work across different instruction sets. This PR add a feature flag to be able to disable SSE optimizations

### DIFF
--- a/ports/pcl/CONTROL
+++ b/ports/pcl/CONTROL
@@ -1,6 +1,6 @@
 Source: pcl
 Version: 1.11.1
-Port-Version: 2
+Port-Version: 3
 Homepage: https://github.com/PointCloudLibrary/pcl
 Description: Point Cloud Library (PCL) is open source library for 2D/3D image and point cloud processing.
 Build-Depends: eigen3, flann, qhull, libpng, boost-system, boost-filesystem, boost-thread, boost-date-time, boost-iostreams, boost-random, boost-foreach, boost-dynamic-bitset, boost-property-map, boost-graph, boost-multi-array, boost-signals2, boost-sort, boost-ptr-container, boost-uuid, boost-interprocess, boost-asio
@@ -32,3 +32,6 @@ Description: OpenGL support for PCL
 Feature: vtk
 Description: VTK-Visualizations support for PCL
 Build-Depends: vtk
+
+Feature: disable-sse
+Description: Disables SSE support
--- a/ports/pcl/portfile.cmake
+++ b/ports/pcl/portfile.cmake
@@ -29,15 +29,18 @@ if ("tools" IN_LIST FEATURES AND VCPKG_LIBRARY_LINKAGE STREQUAL static)
 endif()
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
-    openni2     WITH_OPENNI2
-    qt          WITH_QT
-    pcap        WITH_PCAP
-    cuda        WITH_CUDA
-    cuda        BUILD_CUDA
-    cuda        BUILD_GPU
-    tools       BUILD_tools
-    opengl      WITH_OPENGL
-    vtk         WITH_VTK
+    FEATURES
+        openni2     WITH_OPENNI2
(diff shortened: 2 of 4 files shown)


ports/pcl/portfile.cmake
```diff
@@ -40,6 +40,10 @@ vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     vtk         WITH_VTK
 )
 
+if("disable-sse" IN_LIST FEATURES)
+    list(APPEND FEATURE_OPTIONS -DPCL_ENABLE_SSE=OFF)
+endif()
```
NancyLi1013, line 45: You can add this feature in `vcpkg_check_features()` feature lists. This feature should be added to `INVERTED_FEATURES`.

Please also add keyword `FEATURES` to these current features to solve deprecated warnings. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_check_features.md

NancyLi1013 (changes requested):
Could you please also update Port-Version in CONTROL file?

NancyLi1013:
Hi @fran6co 

Thanks for your PR.

Please run `vcpkg x-add-version pcl` to solve the failures on x86-windows.
```
Error: While reading versions for port pcl from file: C:\a\1\s\versions\p-\pcl.json
       File declares version `1.11.1#2` with SHA: 965ac7e810b04459a8060a73273f924cf8d7f317
       But local port with the same verion has a different SHA: 7efe6bc9f411b5669ec8ae76407a1250318f3772

           vcpkg x-add-version pcl

       to add a new version.
```

Note: Please run the command after you submitted the changes.

ytbigvinny27 (approved):

ports/pcl/CONTROL
```diff
@@ -32,3 +32,6 @@ Description: OpenGL support for PCL
 Feature: vtk
 Description: VTK-Visualizations support for PCL
 Build-Depends: vtk
+
```
NancyLi1013, line 35: Please update `Port-Version: 2` as `Port-Version: 3`.

NancyLi1013 (changes requested):

NancyLi1013:
```
Error: While reading versions for port pcl from file: C:\a\1\s\versions\p-\pcl.json
       File declares version `1.11.1#3` with SHA: 0e13d4046052222f0135999a05aa72875d2647a4
       But local port with the same verion has a different SHA: b4a8098db5f2ca93a22cc1bfd94d2fe32ca9f539

           vcpkg x-add-version pcl

       to add a new version.
```
Please run `vcpkg x-add-version --overwrite-version pcl` to solve the failures on x86-windows.

ytbigvinny27 (approved):

NancyLi1013 (approved):

NancyLi1013:
Hi @fran6co 

Have you tested this new feature `disable-sse`?

fran6co:
Yes, I need it for the project i’m working on right now.

NancyLi1013:
LGTM now, thanks for your PR @fran6co.

strega-nil-ms:
I am concerned about this PR. I don't think this is the right way forward; features should be additive, and should not be used to disable features. This should likely be done with triplets instead... cc @ras0219, could you comment?

BillyONeal:
Dropping "status:reviewed" as there are outstanding comments.

NancyLi1013:
@fran6co Is work still being done for this Pr?

fran6co:
@NancyLi1013 there is no more work needed on this PR save for fixing the conflicts

NancyLi1013:
Thanks for your response. Please resolve the conflicts later.

BillyONeal:
I believe the problem is that we are still uncomfortable with this being controlled with the feature mechanism and would want to see this driven through triplets. It is a setting that needs to be consistent for the universe, not only for pcl.

fran6co:
Not sure how to do this with triplets, I'll close the PR for now then

PR #9221, opened 2019-12-05
[rtmidi] Fix configure CMakeLists.txt

LilyWangL:
Use CMakeLists.txt which is in source code to configure. It has different configurations for Windows, Linux or MacOS. Due to `HMIDIIN` `HMIDIOUT` only support desktop apps, rtmidi doesn't support target UWP. Please refer to the following docs: 
`HMIDIIN`: https://docs.microsoft.com/en-us/previous-versions/dd798460(v%3Dvs.85)
`HMIDIOUT`: https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-midioutshortmsg
There are code in source file RtMidi.cpp L2394.
```
struct WinMidiData {
  HMIDIIN inHandle;    // Handle to Midi Input Device
  HMIDIOUT outHandle;  // Handle to Midi Output Device
  DWORD lastTime;
  MidiInApi::MidiMessage message;
  LPMIDIHDR sysexBuffer[RT_SYSEX_BUFFER_COUNT];
  CRITICAL_SECTION _mutex; // [Patrice] see https://groups.google.com/forum/#!topic/mididev/6OUjHutMpEo
};
```
Related issue: #6993
There are no features of this port need to test.

PR #17159, opened 2021-04-08
[rtmidi] Add support for ALSA

joe-noel-dev:
The port description references ALSA, but it was turned off in the port CMake file. Since it is reasonable to expect it to be installed on a Linux distribution, all that is required is the feature to turn on the flag.


Hoikas:
This already defaults to `ON` upstream: https://github.com/thestk/rtmidi/blob/cc887191c3b4cb6697aeba5536d9f4eb423090aa/CMakeLists.txt#L39

joefocusrite:
@Hoikas 

Ah yes! is it better to remove the ` OPTIONS -DRTMIDI_API_ALSA=OFF` altogether then? This would mean that the port is not backwards compatible and the flag will be turned on for some users.

Hoikas:
Ugh, when I first read the diff, I didn't see that the portfile was turning it off. I assumed "but it was turned off in the port CMake file" meant the upstream CMakeLists.txt :neutral_face:.

It looks like ASLA was disabled in [this commit](https://github.com/microsoft/vcpkg/pull/9221/commits/02d33429253938d14b67a2e23bd66711ac3b6c3a), but I have no idea why. IMO, it would make sense to turn ALSA OFF only if `VCPKG_TARGET_IS_LINUX` is false, since upstream is just testing for non-Apple UNIXy OSes. That's similar to what the openal-soft port is [doing](https://github.com/microsoft/vcpkg/blob/master/ports/openal-soft/portfile.cmake#L44).

ytbigvinny27 (approved):

ytbigvinny27 (approved):

NancyLi1013:
@joefocusrite 

The feature `alsa ` build failed on x64-linux on my machine with the following errors:

config-x64-linux-dbg-err.log
```
CMake Error at CMakeLists.txt:118 (message):
  ALSA API requested but no ALSA dev libraries found
```
config-x64-linux-dbg-out.log
```
-- The CXX compiler identification is GNU 10.1.0
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1")
-- Checking for module 'jack'
--   No package 'jack' found
-- Could NOT find ALSA (missing: ALSA_LIBRARY ALSA_INCLUDE_DIR)

```
Does it work on your machine?

PhoebeHui:
@joefocusrite, the feature also test failed with x64-windows, could you please take a look?

```
-- Could NOT find ALSA (missing: ALSA_LIBRARY ALSA_INCLUDE_DIR) 
CMake Error at CMakeLists.txt:118 (message):
  ALSA API requested but no ALSA dev libraries found
```

joefocusrite:
@PhoebeHui @NancyLi1013 It was missing the OPTIONS keyword 🤦🏼 

Regarding @Hoikas [comment](https://github.com/microsoft/vcpkg/pull/17159#issuecomment-816357173). I'm happy to make it default ALSA on for Linux, or keep it as it is right now for backwards compatibility. Either way 🙂 .

NancyLi1013:
The same error as before on x64-linux. Seems `alsa `still cannot be found. 
After install `libasound2-dev` via `apt-get install libasound2-dev`, it can be built successfully.

See the reference https://askubuntu.com/questions/1281211/no-package-alsa-found-in-ubuntu-18-04

I agree with the suggestion from @Hoikas, since upstream also set `RTMIDI_API_ALSA `as on in Linux.

https://github.com/thestk/rtmidi/blob/master/CMakeLists.txt#L38-L40
```
    if(UNIX AND NOT APPLE)
      option(RTMIDI_API_ALSA "Compile with ALSA support." ON)
    endif()
```

joefocusrite:
@NancyLi1013 That makes sense. Would that be instead of the vcpkg feature or as well as the feature?

NancyLi1013:
> @NancyLi1013 That makes sense. Would that be instead of the vcpkg feature or as well as the feature?

Sorry, I'm not clear what you mean. But I think you need to do these changes in this PR.

Seems `alsa `only supports Linux. So we should add `platform: linux` to feature `alsa`. Also it would be better to add a message to remind users to install `libasound2`. Otherwise, it cannot be built.

PhoebeHui:
The 'platform' filed doesn't work with 'feature'. So the following message should be added:
1. The feature only works on linux.
2. It requires additional system package 'libasound2-dev' on linux.

@joefocusrite, could you address above suggestions?

PhoebeHui:
Pinging @joefocusrite for response. Is work still being done for this PR?

JonLiu1993:
@joefocusrite ,I’m going to close this PR because it’s been a long time since any update has been made. If this is something you’d like to see merged and work is being done, please reopen.

PR #17174, opened 2021-04-09
[xaudio2redist] Add new port

walbourn:
Adds a vcpkg port for the NuGet package Microsoft.XAudio2.Redist. This provides XAudio 2.9 in a form you can use on Windows 7 SP1, Windows 8.0, Windows 8.1, and Windows 10.

It does not support UWP or ARM/ARM64. You can use XAudio2.9 directly in those scenarios.

> None of the lib or dll names collide with the Windows SDK. The headers do, but they are in a ``include/xaudio2redist`` folder.

### DIFF
--- a/ports/xaudio2redist/portfile.cmake
+++ b/ports/xaudio2redist/portfile.cmake
@@ -1,0 +1,34 @@
+vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp" "linux" "osx")
+
+vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
+
+vcpkg_download_distfile(ARCHIVE
+    URLS "https://www.nuget.org/api/v2/package/Microsoft.XAudio2.Redist/1.2.6"
+    FILENAME "xaudio2redist.1.2.6.zip"
+    SHA512 f9b0bacb5787e0239cbf1ffed8e1e8896aa32f69755b084995b2fd1bfe4d49a876d9a69e26469e8779a461e8ff74dc9d5ea86e1ff045dfe2770dfb08b8ba16c3
+)
+
+vcpkg_extract_source_archive_ex(
+    OUT_SOURCE_PATH PACKAGE_PATH
+    ARCHIVE ${ARCHIVE}
+    NO_REMOVE_ONE_LEVEL
+)
+
+file(GLOB HEADER_FILES ${PACKAGE_PATH}/build/native/include/*.h)
+file(INSTALL ${HEADER_FILES} DESTINATION ${CURRENT_PACKAGES_DIR}/include/${PORT})
+
+file(COPY ${PACKAGE_PATH}/build/native/release/lib/${VCPKG_TARGET_ARCHITECTURE}/xaudio2_9redist.lib DESTINATION ${CURRENT_PACKAGES_DIR}/lib/)
+file(COPY ${PACKAGE_PATH}/build/native/debug/lib/${VCPKG_TARGET_ARCHITECTURE}/xaudio2_9redist.lib DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib/)
+
+if(VCPKG_CRT_LINKAGE STREQUAL "static")
+    file(COPY ${PACKAGE_PATH}/build/native/release/lib/${VCPKG_TARGET_ARCHITECTURE}/xapobaseredist.lib DESTINATION ${CURRENT_PACKAGES_DIR}/lib/)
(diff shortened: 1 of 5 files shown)


ports/xaudio2redist/portfile.cmake
```diff
@@ -0,0 +1,34 @@
...
+if(VCPKG_CRT_LINKAGE STREQUAL "static")
+    file(COPY ${PACKAGE_PATH}/build/native/release/lib/${VCPKG_TARGET_ARCHITECTURE}/xapobaseredist.lib DESTINATION ${CURRENT_PACKAGES_DIR}/lib/)
+    file(COPY ${PACKAGE_PATH}/build/native/debug/lib/${VCPKG_TARGET_ARCHITECTURE}/xapobaseredist.lib DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib/)
+else()
+    file(COPY ${PACKAGE_PATH}/build/native/release/lib/${VCPKG_TARGET_ARCHITECTURE}/xapobaseredist_md.lib DESTINATION ${CURRENT_PACKAGES_DIR}/lib/)
+    file(COPY ${PACKAGE_PATH}/build/native/debug/lib/${VCPKG_TARGET_ARCHITECTURE}/xapobaseredist_md.lib DESTINATION ${CURRENT_PACKAGES_DIR}/debug/lib/)
+endif()
+
+file(COPY ${PACKAGE_PATH}/build/native/release/bin/${VCPKG_TARGET_ARCHITECTURE}/xaudio2_9redist.dll DESTINATION ${CURRENT_PACKAGES_DIR}/bin/)
+file(COPY ${PACKAGE_PATH}/build/native/debug/bin/${VCPKG_TARGET_ARCHITECTURE}/xaudio2_9redist.dll DESTINATION ${CURRENT_PACKAGES_DIR}/debug/bin/)
```
PhoebeHui, line 32: @strega-nil, could you help confirm if this allowed in vcpkg?

PhoebeHui, line 32: @strega-nil, could you help confirm if this allowed in vcpkg?

ytbigvinny27 (approved):

ports/xaudio2redist/portfile.cmake
walbourn: The NuGet package provides a helper library ``xapobaseredist.lib`` which is a static library. There are both STATIC CRT and DYNAMIC CRT verisons.

The core XAudio2 library itself is always a DYNAMIC library so it must always have DLLs.

I added logic support to support a scenario where you asked for the STATIC_CRT here, but I couldn't find a standard triplet that allows DYNAMIC_ONLY with STATIC_CRT. The error msg from vcpkg said you can do that however.

ytbigvinny27 (approved):
Request new ports🤞🏾🤞🏾🖖🏾🤟🏾🔜🤌🏽

ports/xaudio2redist/portfile.cmake
walbourn: Rather than try to be clever here, I'll just include both helper libraries. They have file names that don't conflict and it's kind of an edge-case scenario anyhow.

ports/xaudio2redist/portfile.cmake
```diff
@@ -0,0 +1,31 @@
+vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp" "linux" "osx")
```
Hoikas, line 1: It might be better to just fail `if(NOT VCPKG_TARGET_IS_WINDOWS)`, considering that there are community targets (eg android, ios, and various BSDs) that you are not testing.

walbourn: I copied this from another port. It basically only supports Win32 desktop x86 and x64.

walbourn: I'll switch to this, although I see a lot of other ports doing what I did originally...

```
vcpkg_fail_port_install(ON_ARCH "arm" "arm64" ON_TARGET "UWP")

    message(FATAL_ERROR "${PORT} only supports Windows.")
endif()
```

PhoebeHui (approved):

walbourn:
No idea what the OSX failures have to do with this port...

PhoebeHui:
@walbourn, it's not relate to your changes, it failed with 'no space left in pipeline machine', I have rerun the PR.

walbourn:
OK. I've rebased against the latest master...

strega-nil:
cool, thanks @walbourn :)

PR #17175, opened 2021-04-09
[dxsdk-d3dx] Add new port

walbourn:
Adds a vcpkg port for the NuGet package Microsoft.DXSDK.D3DX. This provides the D3DX9, D3DX10, and D3DX11 utility libraries, their headers, and SHA-2 signed DLLs which can be redistributed side-by-side with the application per a new license.

Use of these legacy libraries is still not recommended for new projects, but this is a much more supportable way to use legacy D3DX libraries for existing projects that rely on it.

These headers are compatible with the modern Windows SDK, but are located in ``include/dxsdk-d3dx`` to isolate them.

### DIFF
--- a/ports/dxsdk-d3dx/CONTROL
+++ b/ports/dxsdk-d3dx/CONTROL
@@ -1,0 +1,5 @@
+Source: dxsdk-d3dx
+Version: 9.29.952.8
+Homepage: https://walbourn.github.io/legacy-d3dx-on-nuget/
+Description: Redistributable package for the legacy DirectX SDK's D3DX9, D3DX10, and/or D3DX11 utility libraries.
+Supports: windows & !arm & !uwp & !static
--- a/ports/dxsdk-d3dx/CONTROL
+++ b/ports/dxsdk-d3dx/CONTROL
@@ -1,5 +1,0 @@
-Source: dxsdk-d3dx
-Version: 9.29.952.8
-Homepage: https://walbourn.github.io/legacy-d3dx-on-nuget/
-Description: Redistributable package for the legacy DirectX SDK's D3DX9, D3DX10, and/or D3DX11 utility libraries.
-Supports: windows & !arm & !uwp & !static
--- a/ports/dxsdk-d3dx/portfile.cmake
+++ b/ports/dxsdk-d3dx/portfile.cmake
@@ -1,5 +1,7 @@
 vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp" "linux" "osx")
 
+message(WARNING "Use of ${PORT} is not recommended for new projects. See https://aka.ms/dxsdk for more information.")
+
 vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
 
 set(VCPKG_POLICY_ALLOW_OBSOLETE_MSVCRT enabled)
--- a/ports/dxsdk-d3dx/portfile.cmake
+++ b/ports/dxsdk-d3dx/portfile.cmake
@@ -1,4 +1,8 @@
-vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp" "linux" "osx")
+vcpkg_fail_port_install(ON_ARCH "arm" "arm64" ON_TARGET "UWP")
+
+if(NOT VCPKG_TARGET_IS_WINDOWS)
+    message(FATAL_ERROR "${PORT} only supports Windows.")
+endif()
 
 message(WARNING "Use of ${PORT} is not recommended for new projects. See https://aka.ms/dxsdk for more information.")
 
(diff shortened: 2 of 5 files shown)


walbourn:
Note I had to make use of ``set(VCPKG_POLICY_ALLOW_OBSOLETE_MSVCRT enabled)`` because these D3DX legacy DLLs use ``msvcrt.dll``.  The are binary identical to the binaries that shipped in the legacy DirectX SDK (June 2010). The only difference is that they have been resigned with SHA-2.

ytbigvinny27 (approved):

ports/dxsdk-d3dx/CONTROL
```diff
@@ -0,0 +1,5 @@
+Source: dxsdk-d3dx
+Version: 9.29.952.8
+Homepage: https://walbourn.github.io/legacy-d3dx-on-nuget/
+Description: Redistributable package for the legacy DirectX SDK's D3DX9, D3DX10, and/or D3DX11 utility libraries.
+Supports: windows & !arm & !uwp & !static
```
NancyLi1013, line 5: Can you use vcpkg.json instead of CONTROL here? 

Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/pr-review-checklist.md

For version field names, please see https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md and https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.reference.md

NancyLi1013 (changes requested):

ports/dxsdk-d3dx/CONTROL
walbourn: Is there a tool or script for converting existing CONTROL files to vcpkg.json? I hate writing json by hand...

ytbigvinny27 (changes requested):
~~~🤌🏽🤞🏾

ports/dxsdk-d3dx/CONTROL
walbourn: Also, ``9.29.952.8`` is the actual version number of the NuGet package. Since it contains the the bits from the legacy DirectX SDK which version ``9.29.952``, we used that number here.

walbourn: Converted to vcpkg.json

ports/dxsdk-d3dx/portfile.cmake
```diff
@@ -0,0 +1,40 @@
+vcpkg_fail_port_install(ON_ARCH "arm" ON_TARGET "uwp" "linux" "osx")
```
Hoikas, line 1: It might be better to just fail `if(NOT VCPKG_TARGET_IS_WINDOWS)` (in addition to the arm failure) considering that there are community targets (ios, android, various bsds) that you are not testing.

NancyLi1013: @walbourn 
Could you please help confirm this? Is `arm `supported on other platforms?

walbourn: This port only supports Win32 desktop x86 and x64.

walbourn: I'll switch to this, although I see a lot of other ports doing what I did originally...

```
vcpkg_fail_port_install(ON_ARCH "arm" "arm64" ON_TARGET "UWP")

    message(FATAL_ERROR "${PORT} only supports Windows.")
endif()
```

NancyLi1013 (approved):

walbourn:
No idea what the OSX failures have to do with this port...

Hoikas:
Looks like the disk space ran out on the macOS machine. Not related to this port.

NancyLi1013:
> Looks like the disk space ran out on the macOS machine. Not related to this port.

As @Hoikas said above, I will rerun this PR once the disk space problem solved.

NancyLi1013:
The failures caused by `popsift `will be fixed by #17277.

walbourn:
> The failures caused by `popsift `will be fixed by #17277.

OK. I've rebased against the latest master...

NancyLi1013:
LGTM now, thanks for adding this port @walbourn.

walbourn:
This port doesn't support OSX, so no idea why it's failing that check.

NancyLi1013:
For osx pipeline, there is something wrong with our CI test machine. We're looking into this. You can ignore this failure, which is not related with this PR. I will rerun it once solved.

Note: Please don't update changes in a PR unless it's necessary after we have added the label 'reviewed'.

walbourn:
> Note: Please don't update changes in a PR unless it's necessary after we have added the label 'reviewed'.

Sorry. I was applying feedback to this open PR that I had been given in another PR about the new [version semantics](https://devblogs.microsoft.com/cppblog/take-control-of-your-vcpkg-dependencies-with-versioning-support/).

walbourn:
@Hoikas is there something holding up this PR?

Hoikas:
Good question! Hopefully a maintainer will chime in if there are any outstanding other blockers. It LGTM.

strega-nil-ms:
Thanks @walbourn :), sorry it took a little while to merge

PR #16953, opened 2021-03-29
[libtasn1] Add Windows support, update to 4.17

wrobelda:
**Describe the pull request**
Enables windows build

- Which triplets are supported/not supported? Have you updated the CI baseline?

static-md build is unsupported. CI baseline updated.

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?

Yes.

PR #17215, opened 2021-04-11
[vcpkg/scripts/make] determine whether to use automake wrappers

huangqinjin:
Using `ar-lib` to wrap GNU `ar` will result in error
```
ar: invalid option -- 'L'
```

### DIFF
--- a/scripts/cmake/vcpkg_configure_make.cmake
+++ b/scripts/cmake/vcpkg_configure_make.cmake
@@ -264,6 +264,19 @@ function(vcpkg_configure_make)
 
     debug_message("REQUIRES_AUTOGEN:${REQUIRES_AUTOGEN}")
     debug_message("REQUIRES_AUTOCONFIG:${REQUIRES_AUTOCONFIG}")
+
+    if(CMAKE_HOST_WIN32 AND VCPKG_DETECTED_CMAKE_C_COMPILER MATCHES "cl.exe") #only applies to windows (clang-)cl and lib
+        if(_csc_AUTOCONFIG)
+            set(_csc_USE_WRAPPERS TRUE)
+        else()
+            # Keep the setting from portfiles.
+            # Without autotools we assume a custom configure script which correctly handles cl and lib.
+            # Otherwise the port needs to set CC|CXX|AR and probably CPP.
+        endif()
+    else()
+        set(_csc_USE_WRAPPERS FALSE)
+    endif()
+
     # Backup environment variables
     # CCAS CC C CPP CXX FC FF GC LD LF LIBTOOL OBJC OBJCXX R UPC Y 
     set(_cm_FLAGS AS CCAS CC C CPP CXX FC FF GC LD LF LIBTOOL OBJC OBJXX R UPC Y RC)
@@ -317,7 +330,7 @@ function(vcpkg_configure_make)
         endif()
         if(CMAKE_HOST_WIN32)
             set(APPEND_ENV)
-            if(_csc_AUTOCONFIG OR _csc_USE_WRAPPERS)
+            if(_csc_USE_WRAPPERS)
                 set(APPEND_ENV ";${MSYS_ROOT}/usr/share/automake-1.16")
                 string(APPEND APPEND_ENV ";${SCRIPTS}/buildsystems/make_wrapper") # Other required wrappers are also located there
             endif()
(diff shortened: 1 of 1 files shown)


scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -261,6 +257,13 @@ function(vcpkg_configure_make)
 
     debug_message("REQUIRES_AUTOGEN:${REQUIRES_AUTOGEN}")
     debug_message("REQUIRES_AUTOCONFIG:${REQUIRES_AUTOCONFIG}")
+
+    # Use autotools ar-lib and compile wrappers (only applies to windows cl and lib)
+    set(USE_WRAPPERS FALSE)
+    if (VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_MINGW)
+        set(USE_WRAPPERS TRUE)
+    endif()
```
Neumann-A, line 265: Not all ports require the wrapper on windows. Also it is possible better to check VCPKG_DETECTED_C_COMPILER

huangqinjin: Any harm that always use wrappers for windows? Actually it's best to invoke compiler with some options to check if wappers needed, just like what `configure` does to for option `-c` and `-o`. But I think this simple PR should work well with all ports in vcpkg currently or most future ports.

dg0yt:
Oh, I'm fighting with this too. I would have changes for the whole file, but there seem to be other PRs pending (targeting Android). The more minimal attempts:
https://github.com/dg0yt/vcpkg/commit/7bd9735b67c59bbd01ce6d4bfac3c7b5594ad673  (used during the last days)
https://github.com/dg0yt/vcpkg/commit/5d7f9ae8e1f1cf6399adf099e7a44f1770c0ea36   (just testing now)

huangqinjin:
@dg0yt Once the PRs targeting Android get approved and merged, I would try to make `CMAKE_HOST_WIN32`-only parts to be common for all hosts to enable cross compiling. This PR is also prepared for that.

scripts/cmake/vcpkg_configure_make.cmake
huangqinjin: Changed to honor setting from portfiles if C compiler is cl.exe

dg0yt:
There is only libtasn1 and libidn2 failing now, on x64_osc, in the gtkdocize step while running autoconf. Confirmed to be fixed by `brew install autoconf automake gtk-doc`.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -261,6 +261,19 @@ function(vcpkg_configure_make)
 
     debug_message("REQUIRES_AUTOGEN:${REQUIRES_AUTOGEN}")
     debug_message("REQUIRES_AUTOCONFIG:${REQUIRES_AUTOCONFIG}")
+
+    if(CMAKE_HOST_WIN32 AND VCPKG_DETECTED_CMAKE_C_COMPILER MATCHES "cl.exe") #only applies to windows cl and lib
```
Neumann-A, line 265: this also matches `clang-cl.exe` please update the comment

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -261,6 +261,19 @@ function(vcpkg_configure_make)
 
     debug_message("REQUIRES_AUTOGEN:${REQUIRES_AUTOGEN}")
     debug_message("REQUIRES_AUTOCONFIG:${REQUIRES_AUTOCONFIG}")
+
+    if(CMAKE_HOST_WIN32 AND VCPKG_DETECTED_CMAKE_C_COMPILER MATCHES "cl.exe") #only applies to windows cl and lib
+        if(_csc_AUTOCONFIG)
+            set(_csc_USE_WRAPPERS TRUE)
+        else()
```
Neumann-A, line 268: There is a high probability that there needs to be a way to disable the wrappers for cl

Neumann-A: nevermind I thought you  removed `USE_WRAPPERS` as an option and always turned on the wrappers (without using `AUTOCONFIG`). Probably had an old changeset in mind.

dg0yt:
This is mandatory for mingw on linux. Merge conflict resolution proposed at https://github.com/huangqinjin/vcpkg/pull/1. I hope it gets resolved before new conflicts arrive.

dg0yt:
Ping. Blocker for mingw on Windows.

wrobelda:
Seems that the issue I am having with Windows support for `libtasn1` is related: https://github.com/microsoft/vcpkg/pull/16953#issuecomment-820979602

dg0yt:
> Seems that the issue I am having with Windows support for `libtasn1` is related: [#16953 (comment)](https://github.com/microsoft/vcpkg/pull/16953#issuecomment-820979602)

I don't think so. Your issue is with windows triplets. This PR is for mingw triplets, and the libtasn1 issues on osx were not related to wrappers.

wrobelda:
> I don't think so. Your issue is with windows triplets. This PR is for mingw triplets, and the libtasn1 issues on osx were not related to wrappers.

Not sure why you bring up macOS – #16953 is specifically to fix Windows build, and indeed this PR does fix the `ar-lib` call for me.

JackBoosY (approved):

past-due (approved):
I can confirm that this fixes several issues I have encountered with mingw (clang environments). 👍

ras0219-msft:
This LGTM, thanks for the PR!

Issue #16924, opened 2021-03-28
Unable to access library after `integrate install` and/or `integrate project` in Visual Studio / Win10


zipzit:
**Describe the bug**
Visual Studio reports "cannot open source file "zmq.hpp"

**Environment**
- OS: 
- Edition	Windows 10 Home, Version 20H2, Installed on ‎6/‎11/‎2020,  OS build 19042.870
- Experience Windows Feature Experience Pack 120.2212.551.0
- Processor	Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz   3.40 GHz  Installed RAM	32.0 GB
- System type	64-bit operating system, x64-based processor
- Visual Studio Community 2019, Version 16.9.1
- Visual C++ 2019 -- 00435-60000-00000-AA449

VCPKG installed via: 
$ `git clone https://github.com/microsoft/vcpkg`
   ` .\vcpkg\bootstrap-vcpkg.bat`
   `.\vcpkg\vcpkg install cppzmq`
    `.\vcpkg\vcpkg integrate install`

Note: I installed VCPKG on my M drive.  VCPKG is installed no where else on this computer (typically I use virtual machines.. not this time...)  Operating system located on C drive.. 

Install reports success:
   ```
The package cppzmq:x86-windows provides CMake targets:
        find_package(cppzmq CONFIG REQUIRED)
        target_link_libraries(main PRIVATE cppzmq cppzmq-static)
```
Integrate Install also successful:
```
    PS M:\> .\vcpkg\vcpkg integrate install
    Applied user-wide integration for this vcpkg root.
      All MSBuild C++ projects can now #include any installed libraries.
      Linking will be handled automatically.   
      Installing new libraries will make them instantly available.
```
I can see that both CPPZMQ and its base ZEROMQ are successfully installed.  I can search VCPKG folder and see the .h files my program can't find.

PS M:\vcpkg> `./vcpkg list`
`cppzmq:x86-windows                                 4.7.1#1          lightweight messaging kernel, C++ bindings`
`zeromq:x86-windows                                 2019-09-20#2     The ZeroMQ lightweight messaging kernel is a lib...`

After an initial search I discovered [this posting,](https://github.com/microsoft/vcpkg/issues/9432)  I attempted to use `$ ./vcpkg integrate project`  From what I can see that was successful, too.  

```
    PS M:\vcpkg> `./vcpkg integrate project`
[...]

Issue #12357, opened 2020-07-10
[Feature] Default to x64-windows triplet on x64 Windows


VelocityRa:
Most people don't use 32-bit builds these days, so we have to always specify the triplet which is cumbersome, or set the env-var which is not elegant.

### Implementation
Instead of this:
https://github.com/microsoft/vcpkg/blob/22c8e3a23afc6be45020cd3f8c92bcea783ce8fa/toolsrc/src/vcpkg/triplet.cpp#L100
there should be a check to see if the host is 64-bit and use the `x64-windows` triplet if so.

`IsWow64Process` or `IsWow64Process2` can be used for this (will work even if vcpkg is compiled as 32-bit).

PhoebeHui:
Duplicated to issue https://github.com/microsoft/vcpkg/issues/1254 

Related PR https://github.com/microsoft/vcpkg/pull/11610

cloudhan:
if I have commit permission, I will change
https://github.com/microsoft/vcpkg/blob/22c8e3a23afc6be45020cd3f8c92bcea783ce8fa/toolsrc/src/vcpkg/triplet.cpp#L99-L100
to
```cpp
#if defined(_WIN64)
                return Triplet::X64_WINDOWS;
#elif defined(_WIN32)
                return Triplet::X86_WINDOWS;
```
right now and force push it...

See how many thumb ups in #1254

VelocityRa:
@cloudhan This is wrong, `_WIN64` will only check if vcpkg is compiled in a 64-bit environment, not if the host is a 64-bit operating system. If vcpkg itself was compiled as 32-bit on a 64-bit system it wouldn't work.

See my implementation recommendation above instead.

cloudhan:
@VelocityRa Good point, and seems the bootstrap script indeed build 32bit version of vcpkg.exe on a 64bit platform. Another feature request, here you go.

MarkIngramUK:
This has just caught me out on a fresh install (again). 64-bit by default on 64-bit machines seems like a sensible choice.

Issue #1254, opened 2017-06-09
Install x64 packages as default instead of x86


jasjuang:
I believe nowadays most computers are 64 bit. Is it possible to install x64 libraries by default?

For example:
``````
vcpkg.exe install gflags 
``````

This now install gflags_x86-windows, can we change it so it install gflags_x64-windows?

**Comment by KindDragon** [comment]
You can set system environment variable `VCPKG_DEFAULT_TRIPLET=x64-windows`

**Comment by jasjuang** [comment]
Thanks, it works for me

**Comment by MikeGitb** [comment]
I'm still thinking that x64 should be the default.

**Comment by Voskrese** [comment]
?
`set VCPKG_DEFAULT_TRIPLET=x64-windows`
?
`setx VCPKG_DEFAULT_TRIPLET=x64-windows`

**Comment by jasjuang** [comment]
@MikeGitb I agree x64 should be the default.
@Voskrese Go to control panel -> system -> environment variables to make it permanent

**Comment by kklouzal** [comment]
I second the notion that x64 needs to be default. At the very least put the information on how to do it in the readme.md

P.S.
I can't thank you guys enough for developing this package.

**Comment by VioletGiraffe** [comment]
Can't install x64 via `vcpkg` arguments? Only environment variables? Terribly inconvenient.

**Comment by ras0219-msft** [comment]
You can definitely choose x64 through command line arguments!

```
vcpkg install zlib:x64-windows
vcpkg install zlib openssl --triplet x64-windows
```

This thread is discussing how to make it the default.

VioletGiraffe:
@ras0219-msft: Now I understand my mistake, thanks a lot (and sorry).

Xis-Ytneves:
how is anyone to know about the environment variable without stumbling across this page?

rangershs:
maybe some users still use x86 machine..

Rogtesh:
Im new to this, how to make this defaults in Win10 Environment Variables? I dont want to mess it up and not sure where i put it. Ive added stuff to path before just never added something like this:  "VCPKG_DEFAULT_TRIPLET=x64-windows" to Environment Variables. Im only building in x64 i dont have any x86 compilers and such installed from what i know so it would be nice to have this set as default so i dont need to guess the package name with x64 in it.

drescherjm:
https://www.architectryan.com/2018/08/31/how-to-change-environment-variables-on-windows-10/

Const-me:
[...]

PR #17218, opened 2021-04-11
[vcpkg Readme] Added x64 triplet note for Windows Quick Start

MarkIngramUK:
**Describe the pull request**
Adds a note to the README.md about default installation being x86, and an example of how to target x64 instead.

- What does your PR fix? 
Not a fix, but inspired by #1254 and #12357

N/A

Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,18 @@ To install the libraries for your project, run:
 > .\vcpkg\vcpkg install [packages to install]
 ```
 
+Note: This will install x86 libraries by default. To install x64, run:
+
+```cmd
+> .\vcpkg\vcpkg install package:x64-windows
+```
+
+Or
+
+```cmd
+> .\vcpkg\vcpkg install [packages to install] --triplet=x64-windows
+```
+
 You can also search for the libraries you need with the `search` subcommand:
 
 ```cmd


PhoebeHui:
@MarkIngramUK, thanks for the PR!

This also resolve https://github.com/microsoft/vcpkg/issues/16924#issuecomment-808851643.

PhoebeHui (approved):

README.md
```diff
@@ -86,6 +86,18 @@ To install the libraries for your project, run:
 > .\vcpkg\vcpkg install [packages to install]
 ```
 
+Note: This will install x86 libraries by default. To install x64, run:
+
+```cmd
+> .\vcpkg\vcpkg install package:x64-windows
+```
+
+Or,
```
Hoikas, line 95:
```suggestion
Or
```

comma splice

MarkIngramUK: Done 👍

MarkIngramUK:
> @MarkIngramUK, thanks for the PR!
> 
> This also resolve [#16924 (comment)](https://github.com/microsoft/vcpkg/issues/16924#issuecomment-808851643).

Looking through that issue, it seems the reason x86 is still the default, is because Visual Studio also defaults to x86 for new projects. Is this still a requirement (keeping the defaults in-sync)? If not, I'd like to change (in a future PR) the default triplet to x64 on Windows, with the fix proposed in #12357. I won't work on it if the change isn't wanted though :-)

PhoebeHui:
@MarkIngramUK, adding a note to readme is fine, please don't change the default triplet to x64 in this PR, I think we will do that if we decide to change it.

PR #16740, opened 2021-03-17
[libpq] mingw support

longnguyen2004:
**Describe the pull request**

I don't have much CPU power to test, @autoantwort can you test for me?

- What does your PR fix? Fixes #16435
Add support for mingw in libpq
- Which triplets are supported/not supported? Have you updated the CI baseline?
All mingw triplets
- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
Yes

Issue #16647, opened 2021-03-11
[OpenSSL] build failure


HideakiAtsuyo:
A kind of suite of https://github.com/microsoft/vcpkg/issues/16539 =>(https://github.com/microsoft/vcpkg/issues/16539#issuecomment-794186324)

**Host Environment**
- OS: Windows 10
- Compiler: CMake from MinGW

**To Reproduce**

- Clone vcpkg
- Use the bootstrap bat file
- vcpkg install zlib

**Failure logs** 

`./vcpkg install cpr:x86-mingw-dynamic` => 
```
Computing installation plan...
The following packages will be built and installed:
    openssl[core]:x86-mingw-dynamic -> 1.1.1j
Detecting compiler hash for triplet x86-mingw-dynamic...
Could not locate cached archive: C:\Users\Administrator\AppData\Local\vcpkg\archives\44\4457ab51d2c026cbd11b5378477c72a8d9f92552.zip
Starting package 1/1: openssl:x86-mingw-dynamic
Building package openssl[core]:x86-mingw-dynamic...
-- Using community triplet x86-mingw-dynamic. This triplet configuration is not guaranteed to succeed.
-- [COMMUNITY] Loading triplet configuration from: C:\Dev\vcpkg\triplets\community\x86-mingw-dynamic.cmake
-- Downloading https://www.openssl.org/source/openssl-1.1.1j.tar.gz -> openssl-1.1.1j.tar.gz...
-- Extracting source C:/Dev/vcpkg/downloads/openssl-1.1.1j.tar.gz
-- Using source at C:/Dev/vcpkg/buildtrees/openssl/src/1.1.1j-ff1006361a.clean
-- Downloading https://repo.msys2.org/msys/x86_64/bash-4.4.023-2-x86_64.pkg.tar.xz -> msys-bash-4.4.023-2-x86_64.pkg.tar.xz...
-- Downloading https://repo.msys2.org/msys/x86_64/diffutils-3.7-1-x86_64.pkg.tar.xz -> msys-diffutils-3.7-1-x86_64.pkg.tar.xz...
-- Downloading https://repo.msys2.org/msys/x86_64/coreutils-8.32-1-x86_64.pkg.tar.xz -> msys-coreutils-8.32-1-x86_64.pkg.tar.xz...
[...]

Issue #14561, opened 2020-11-13
How do I tell whether your project is referring to minGW or mingw-w64


SlySven:
I recently got into a heated discussion with another member of the project I code for who is setting up what is labelled as using a `x64-mingw-dynamic` triplet as a GitHub Action. As the people over at [Software in the Public Interest Inc.](http://mingw.org/Welcome_to_MinGW_org) strongly insist:
> MinGW is a registered trademark of Software in the Public Interest Inc., registration number 86017856; it has been registered on behalf of MinGW.org, and its use by any other project is unauthorized.

I am also aware that their software (which includes `MSYS`) only targets the 32 Bit Windows OS but can run in 64 Bit ones.

However it has finally become clear to me that you (like myself 😁) are using the more modern, up to date and capable of targetting both 32 bit AND 64 bit Windows OS natively, fork that is Mingw-w64 and MSYS2 - homepage: http://mingw-w64.org/doku.php.

Is there anything that can be done to make this important distinction more clear?

PhoebeHui:
I have no idea about this, @strega-nil, could you please take a look?

strega-nil:
This is really weird. I don't know anything about this... cc @ras0219-msft

SlySven:
If it helps to clarify things - you seem to be using the *modern*/*up-to-date* fork that is Mingw-w64 [Wikipedia article](https://en.wikipedia.org/wiki/Mingw-w64) (which targets both 32 and 64 bit Windows directly) rather than the original MinGW [Wikipedia article](https://en.wikipedia.org/wiki/MinGW) that only targets 32-bit Windows but which can run on 64-bit Windows because the latter has Windows-On-Windows64 [Wikipedia article](https://en.wikipedia.org/wiki/WoW64).

Putting it simply, the ambiguity of your naming scheme(s) is confusing for those who know (and care) about the difference...!

*\<aside>It is my opinion that the *original* project does not like the *forked* one, probably (again in my opinion) as the latter is more popular - especially with those who want to make the most of 64-bit processors!\</aside>* 😜

longnguyen2004:
[...]

Issue #9068, opened 2019-11-21
[GDAL] On Linux completly broken


Neumann-A:
This port does not build correctly on linux. (even though it builds succesfully)
It does not pass any option of 3rd party libraries and the configure discovery of libraries of this port is terrible broken which can be seen by simply looking at the config-x64-linux-(rel|dbg)-out.log. 

As an example: gdal depents on sqlite3 but the configure log tells:
> checking for sqlite3_open in -lsqlite3... no

the reason why configure fails is due to undefined symbols due to not linking pthread and dl in the test (the script even sets LIBS='' for the test so i cannot pass those via the LIBS variable).  

To solve the problem all libraries must be passed via `-with-<package>=${CURRENT_INSTALLED_DIR}/(debug/)` in addition to that the configure script itself must be patched. Furthermore the configure file in the debug build must be changed so that every occourence of `-with-<package>/include` is replaced by `-with-<package>/../include` due to additional header existence checks (WTF?!? Why? You already did compiler/linker checks .....)

(and that might not be all)

Here is a list of all gdal configure options for reference:
```
Defaults for the options are specified in brackets.

Configuration:
  -h, --help              display this help and exit
      --help=short        display options specific to this package
      --help=recursive    display the short help of all the included packages
  -V, --version           display version information and exit
  -q, --quiet, --silent   do not print \`checking ...' messages
      --cache-file=FILE   cache test results in FILE [disabled]
  -C, --config-cache      alias for \`--cache-file=config.cache'
  -n, --no-create         do not create output files
      --srcdir=DIR        find the sources in DIR [configure dir or \`..']

Installation directories:
  --prefix=PREFIX         install architecture-independent files in PREFIX
                          [$ac_default_prefix]
  --exec-prefix=EPREFIX   install architecture-dependent files in EPREFIX
                          [PREFIX]

By default, \`make install' will install all the files in
\`$ac_default_prefix/bin', \`$ac_default_prefix/lib' etc.  You can specify
an installation prefix other than \`$ac_default_prefix' using \`--prefix',
for instance \`--prefix=\$HOME'.

For better control, use the options below.

Fine tuning of the installation directories:
  --bindir=DIR            user executables [EPREFIX/bin]
  --sbindir=DIR           system admin executables [EPREFIX/sbin]
  --libexecdir=DIR        program executables [EPREFIX/libexec]
  --sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
[...]

PR #17219, opened 2021-04-11
Add mingw documentation

dg0yt:
This PR adds basic documentation for using vcpkg with mingw.

The instructions weren't tested line by line from scratch. But they are based on recent work with vcpkg.

### DIFF
--- a/docs/users/mingw.md
+++ b/docs/users/mingw.md
@@ -2,6 +2,15 @@
 
 **The latest version of this documentation is available on [GitHub](https://github.com/Microsoft/vcpkg/tree/master/docs/users/mingw.md).**
 
+*MinGW is community-supported and not tested as part of vcpkg repository's CI process.*
+
+## Table of Contents
+
+ - [Mingw-w64 community triplets](#Mingw-w64-community-triplets)
+ - [Using Mingw-w64 natively on Windows](#Using-Mingw-w64-natively-on-Windows)
+   - [How to avoid mixing different installations](#How-to-avoid-mixing-different-installations)
+ - [Using Mingw-w64 to build Windows programs on other systems](#Using-Mingw-w64-to-build-Windows-programs-on-other-systems)
+
 ## Mingw-w64 community triplets
 
 Vcpkg includes
@@ -100,9 +109,9 @@ try to keep the directories of the msys subsystem (`/usr/bin`, `bin`)
 out of the `PATH` environment variable as found by vcpkg. In bash, you
 may modify the `PATH` just for a single call of vcpkg:
 
-~~~
+```bash
 PATH="${PATH/:\/usr\/bin:\/bin:/:}" ./vcpkg install libpq
-~~~
+```
 
 Alternatively, you may run vcpkg from a regular Command Prompt, after
 adding *only* the desired mingw directory (e.g. `C:\msys64\mingw64\bin`)
@@ -133,7 +142,7 @@ might be older releases which lack some useful features or bug fixes.
 An alternative independent toolchain is offered by [MXE](https://mxe.cc/).
 
 For vcpkg bootstrapping, clone the github repository and run the
-bootstrap-vcpkg.sh script:
+`bootstrap-vcpkg.sh` script:
 
 ```bash
 git clone https://github.com/microsoft/vcpkg.git
--- a/docs/users/mingw.md
+++ b/docs/users/mingw.md
@@ -41,7 +41,7 @@ must use a mingw subsystem of MSYS2, and install some packages
 | x64          | x64-mingw-dynamic, x64-mingw-static | mingw64   | mingw-w64-x86_64- |
 | x86          | x86-mingw-dynamic, x86-mingw-static | mingw32   | mingw-w64-i686-   |
 
-After the basic installation if MSYS2, you will need to install a few
+After the basic installation of MSYS2, you will need to install a few
 additional packages for software development, e.g. for x64:
 
 ```bash
--- a/docs/users/mingw.md
+++ b/docs/users/mingw.md
@@ -1,24 +1,23 @@
-# Vcpkg and MinGW
+# Vcpkg and Mingw-w64
 
 **The latest version of this documentation is available on [GitHub](https://github.com/Microsoft/vcpkg/tree/master/docs/users/mingw.md).**
 
-## MinGW community triplets
+## Mingw-w64 community triplets
 
 Vcpkg includes
-[community triplets for MinGW](https://github.com/microsoft/vcpkg/tree/master/triplets/community)
-for x64, x86, arm64 and arm. They don't depend on Visual Studio and
+[x64, x86, arm64 and arm community triplets](https://github.com/microsoft/vcpkg/tree/master/triplets/community)
+for [Mingw-w64](http://mingw-w64.org/). They don't depend on Visual Studio and
 can be used natively on Windows as well as for cross-compiling on
 other operating systems. There are two variants of each triplet,
-selecting between static and dynamic linking:
+selecting between static and dynamic linking. The actual tools
+(g++ etc.) are expected to be named with particular prefixes.
 
-- arm64-mingw-dynamic
-- arm64-mingw-static
-- arm-mingw-dynamic
-- arm-mingw-static
-- x64-mingw-dynamic
-- x64-mingw-static
-- x86-mingw-dynamic
-- x86-mingw-static
+| architecture | vcpkg community triplets                | tool name prefix     |
+|--------------|-----------------------------------------|----------------------|
+| x64          | x64-mingw-dynamic, x64-mingw-static     | x86_64-w64-mingw32-  |
+| x86          | x86-mingw-dynamic, x86-mingw-static     | i686-w64-mingw32-    |
+| arm64        | arm64-mingw-dynamic, arm64-mingw-static | aarch64-w64-mingw32- |
+| arm          | arm-mingw-dynamic, arm-mingw-static     | armv7-w64-mingw32-   |
 
 These triplets are not tested by continuous integration, so many ports
 do not build, and even existing ports may break on port updates.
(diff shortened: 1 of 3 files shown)


docs/users/mingw.md
```diff
@@ -0,0 +1,101 @@
...
+Now you can test your setup:
+
+```bash
+./vcpkg install zlib
+```
+
+## Using MinGW to build Windows programs on other systems
+
+Many Linux distributions come with MinGW toolchains which allow to
+cross-compile software on Linux to be run on Windows. You can use
```
autoantwort, line 88: Maybe add a hint what must be installed. Must I install `mingw-w64` or is `gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64` enough?

dg0yt: I extended that paragraph now. OTOH I don't want to add too much specific information. There is no point in using vcpkg with MinGW unless you either know MinGW already, or you are willing to learn from other resources.

dg0yt:
What I would like to add later is comments about the "system binaries" configuration, affecting cmake and ninja. But this really needs to be tested before published. This can be done i another PR.

docs/users/mingw.md
```diff
@@ -0,0 +1,113 @@
...
+| x86          | x86-mingw-dynamic, x86-mingw-static | mingw32   | mingw-w64-i686-   |
+
+After the basic installation if MSYS2, you will need to install a few
+additional packages for software development, e.g. for x64:
+
+```bash
+pacman -S git mingw-w64-x86_64-cmake mingw-w64-x86_64-ninja
+```
+
+The active subsystem is select by running the MSYS2 MinGW app, or
```
NancyLi1013, line 52:
```suggestion
The active subsystem is selected by running the MSYS2 MinGW app, or
```

NancyLi1013 (changes requested):

docs/users/mingw.md
dg0yt: Done.

dg0yt:
Ready, from my POV.

NancyLi1013 (approved):

docs/users/mingw.md
```diff
@@ -0,0 +1,113 @@
...
+with a special [suffix](https://repology.org/projects/?search=-mingw-w64)
+or [prefix](https://repology.org/projects/?search=mingw-w64-).
+As an example, for Debian-based distributions, you would start with
+these installation commands for the x64 targets:
+
+```
+sudo apt-get install gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64
+```
+
+Note that the pre-installed cmake might be too old for vcpkg. Also the
```
ras0219-msft, line 99:
```suggestion
Note that the pre-installed cmake might be too old for vcpkg.
```
We should auto-download a newer cmake, just the same as our normal linux instructions

dg0yt: Indeed, there is cmake and ninja in the downloads/tools folder. I will edit this PR to generally reflect the default behaviour which is downloading cmake and ninja. Dealing with the "system binaries" configuration can be done in a separate PR if desired.

dg0yt: Done, 3798b809

longnguyen2004:
Thanks for the PR! I really appreciate having proper documentation for mingw.
I think we should add a note that users should use a Mingw-w64 based toolchain, not the old MinGW based toolchain. There was a discussion about this in #14561, and an actual issue #16647. Basically put, practically everyone assumes we're using a Mingw-w64 based toolchain, so we should make that clear as well.

docs/users/mingw.md
```diff
@@ -0,0 +1,112 @@
...
+git clone https://github.com/microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.bat
+```
+
+For building packages, you need to tell vcpkg that you want to use the
+MinGW triplet. This can be done in different ways. When Visual Studio
+is not installed, you must also set the host triplet to mingw. This is
+needed to resolve host dependencies. For convenience, you can use
+environment variables to set both triplets:
```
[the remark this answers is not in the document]
dg0yt: As it is written now, it implies a mingw Terminal with bash and the mingw subsystem. No matter if this is done via starting "MSYS2 Mingw64" (for x64) from the start menu, or changing it in another MSYS2 terminal via `source shell mingw64` (for x64), it will have `/mingw/bin` in the path, before `/usr/bin`. This is enough to have the right toolchain in the path during this Terminal session.
I'm not sure if it is a good idea to add MSYS2 path to system `PATH` environment. `/usr/bin` is (mostly) the msys subsystem which is not what is desired. And adding somehing like `/mingw64/bin` would choose a particular arch where MSYS2 normally allows switching at runtime.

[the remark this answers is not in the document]
longnguyen2004: Yeah this is mostly for people using a standalone toolchain, in which case there's no MSYS2, and people who are cross-compiling, where the compiler might not be at `/usr/bin`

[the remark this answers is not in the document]
dg0yt: For the Windows section, the documentation is only for MSYS2. I can move the MSYS2 mention from the first paragraph to the section headline.

For cross-compiling, distribution packages provide prefixes binaries `x86_64-w64-mingw32-g++` in the system path.

For any other stand-alone toolchain, including the mentioned MXE, users should read the toolchain's instruction in the first place. IMO there cannot be a general instruction to change the `PATH` without the risk of breaking other things. There seems to many issues reported to vcpkg which are related to side effects of modifying `PATH`. I never had this issue with MSYS2. I never let it change the `PATH` outside the top-level bash.

dg0yt:
> I think we should add a note that users should use a Mingw-w64 based toolchain, not the old MinGW based toolchain. There was a discussion about this in #14561, and an actual issue #16647. Basically put, practically everyone assumes we're using a Mingw-w64 based toolchain, so we should make that clear as well.

I was inclined to answer again with "know your toolchain" but maybe this needs more attention indeed. I looked throught the mingw mailing lists, and it confirm the MinGW project's mood as reported in #14561, in particular when users of mingw-w64 raise questions there.

To make it really clear, I would replace MinGW with Mingw-w64 in most occurences. This is not meant to exclude MinGW if someone is willing to support that in terms of documentation, triplets and or ports. But I don't think MinGW is ready for usage with vcpkg at this stage.

In particular, I won't change the mingw.md file name. However, renaming the community triplets should be seriously considered. And last not least it is `scripts/toolchains/mingw.cmake` which selects mingw-w64 tools by `${CMAKE_SYSTEM_PROCESSOR}-w64-mingw32-` prefixes (note the w64).

longnguyen2004:
> But I don't think MinGW is ready for usage with vcpkg at this stage.

It's not about vcpkg, it's about upstream expectation. Pretty much every libraries out there expect Mingw-w64, not MinGW.
I would be happy to add a check for it in `mingw.cmake`, if there's a way to differentiate between them purely by compiler output.

dg0yt:
> > But I don't think MinGW is ready for usage with vcpkg at this stage.
> 
> It's not about vcpkg, it's about upstream expectation. Pretty much every libraries out there expect Mingw-w64, not MinGW.

IMO there is nothing wrong with upstream. It is downstream, i.e. vcpkg users, which expect the ports to magically work with MinGW when vcpkg talks about MinGW. But when there is no x86_64 in MinGW, it won't come through vcpkg.

> I would be happy to add a check for it in `mingw.cmake`, if there's a way to differentiate between them purely by compiler output.

I didn't try any MinGW toolchain in the last years, and the currently broken MinGW website doesn't help. But I don't think they use the same prefix, or report the same autoconf triplet, as mingw-w64. I assume `i386-pc-mingw32` or similar for MinGW. And for compiler output, for mingw-w64 targeting x64, I can run:
```bash
$ x86_64-w64-mingw32-gcc -dumpmachine
x86_64-w64-mingw32
```
So I would look for `-w64-mingw32`.

dg0yt:
And now I realize that vcpkg passes the MinGW autotools triplets (...-pc-mingw32) to configure. IMO it should use the ...mingw-w64 autotools triplets when it targets Mingw-w64.

longnguyen2004:
> And now I realize that vcpkg passes the MinGW autotools triplets (...-pc-mingw32) to configure. IMO it should use the ...~mingw-w64~-w64-mingw32 autotools triplets when it targets Mingw-w64.

To be honest, letting autotools decide what triplet to use is a bit unreliable, and I would really like for each triplet to specify its corresponding autotools triplet, but that would require some work on `vcpkg_configure_make`. @Neumann-A What do you think?

dg0yt:
> letting autotools decide what triplet to use is a bit unreliable

I didn't want to suggest that autotools should decide. I just want to make a distinction from vcpkg's triplets.

> I would really like for each triplet to specify its corresponding autotools triplet

In the vcpkg triplet file, at least optionally? Well, this is also what I think. It would facilitate adding new or customizing existing triplets (Android, mingw-w64 ucrt). The triplet files are in the hand of the users.

> that would require some work on vcpkg_configure_make

I got the impression that there is already a backlog of open PRs for this script, and I noticed some issues which aren't reported yet.

longnguyen2004:
> I didn't want to suggest that autotools should decide. I just want to make a distinction from vcpkg's triplets.

Oh alright, I thought you were talking about the MinGW vs. Mingw-w64 issue and how the `*-pc-mingw32` triplet is related to it.

dg0yt:
I added another section how to to avoid mixing MSYS binaries from different installations.

*Background:*
When building port `libpq` from MSYS2 bash, I ran into [strange issues](https://github.com/microsoft/vcpkg/pull/16740#issuecomment-821813346). The root cause is a mix of differents MSYS2 installations which is cause argument passing/quoting/escaping to fail. I found that this is discouraged by the MSYS2 project.

*Note: IMO the underlying problem might effect even building ports for MSVC targets.*

For the record, my intermediate goal is to build the `gdal`. At the moment I have to add 10 PRs to vcpkg HEAD, and the next one is coming.

dg0yt:
I had to extend the `PATH` recommendation to also exclude `/bin`. (My test installation were not fully consistent before this change.)

I think that it my eventually make sense to handle this `PATH` quirk in the triplet files and/or other central files. But this is out of scope for this documentation PR.

Removing the local msys2 installation from `PATH` means that even simple tools like `touch` are not available without `vcpkg_acquire_msys` / before `vcpkg_configure_make`. But it doesn't look like multiple calls to `vcpkg_acquire_msys` are a good thing: They result in multiple msys roots. In some cases, such as a call to `touch` in the non-Windows path of port gdal, this can be solved by using the portable commands of `cmake -E`.

This basically marks the point where I consider this PR done: It is tested on multiple installation of Windows. (Yes, I succeeded in building gdal with mingw on Windows. It is as broken as on Linux, #9068).

ras0219-msft (approved):
LGTM, thanks for making this awesome doc @dg0yt !

PR #17164, opened 2021-04-08
[ncurses] new port

electron271:
**Describe the pull request**

Adds ncurses

- What does your PR fix? Fixes #7110 #17167

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
Yes

PR #17226, opened 2021-04-12
[ncurses] Add new port

### LINKED ISSUE #7110: [New port Request] ncurses
I made a search about ncurse library but I can't find it in vcpkg, I think we need to add it to packages

### LINKED ISSUE #17167: [New Port Request] ncurses
Library name:

ncurses

Library description:

ncurses is a programming library providing an API that allows the programmer to write text-based user interfaces.

Source repository URL:

https://github.com/mirror/ncurses

https://invisible-island.net/ncurses/announce.html

Hoikas:
Adds the port ncurses.

- What does your PR fix?
Fixes #7110
Fixes #17167
Closes #17164

Windows (MSVC) is not supported.

Yes

### DIFF
--- a/ports/ncurses/portfile.cmake
+++ b/ports/ncurses/portfile.cmake
@@ -1,0 +1,59 @@
+vcpkg_fail_port_install(ON_TARGET "Windows" "UWP")
+
+vcpkg_download_distfile(
+    ARCHIVE_PATH
+    URLS
+        "https://invisible-mirror.net/archives/ncurses/ncurses-6.2.tar.gz"
+        "ftp://ftp.invisible-island.net/ncurses/ncurses-6.2.tar.gz"
+        "ftp://ftp.gnu.org/gnu/ncurses/ncurses-6.2.tar.gz"
+    FILENAME "ncurses-6.2.tgz"
+    SHA512 4c1333dcc30e858e8a9525d4b9aefb60000cfc727bc4a1062bace06ffc4639ad9f6e54f6bdda0e3a0e5ea14de995f96b52b3327d9ec633608792c99a1e8d840d
+)
+
+vcpkg_extract_source_archive_ex(
+    OUT_SOURCE_PATH SOURCE_PATH
+    ARCHIVE ${ARCHIVE_PATH}
+)
+
+set(OPTIONS
+    --disable-db-install
+    --enable-pc-files
+    --without-manpages
+    --without-progs
+    --without-tack
+    --without-tests
+)
+if(VCPKG_LIBRARY_LINKAGE STREQUAL dynamic)
+    list(APPEND OPTIONS
+        --with-shared
+        --with-cxx-shared
+        --without-normal
+    )
+endif()
+
+set(OPTIONS_DEBUG
+    --with-pkg-config-libdir=${CURRENT_INSTALLED_DIR}/debug/lib/pkgconfig
+    --with-debug
+)
+set(OPTIONS_RELEASE
+    --with-pkg-config-libdir=${CURRENT_INSTALLED_DIR}/lib/pkgconfig
+    --without-debug
+)
+
+vcpkg_configure_make(
+    SOURCE_PATH ${SOURCE_PATH}
+    OPTIONS ${OPTIONS}
+    OPTIONS_DEBUG ${OPTIONS_DEBUG}
+    OPTIONS_RELEASE ${OPTIONS_RELEASE}
+    NO_ADDITIONAL_PATHS
+)
+vcpkg_install_make()
+
+vcpkg_fixup_pkgconfig()
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/bin")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/bin")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+
+file(INSTALL "${SOURCE_PATH}/COPYING" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME copyright)
--- a/ports/ncurses/vcpkg.json
+++ b/ports/ncurses/vcpkg.json
@@ -1,0 +1,6 @@
+{
+  "name": "ncurses",
+  "version-string": "6.2",
+  "description": "free software emulation of curses in System V Release 4.0",
+  "supports": "!(windows | uwp)"
+}
--- a/scripts/test_ports/cmake/portfile.cmake
+++ b/scripts/test_ports/cmake/portfile.cmake
@@ -10,6 +10,10 @@ vcpkg_from_gitlab(
     HEAD_REF master
 )
 
+if(NOT VCPKG_TARGET_IS_WINDOWS AND NOT VCPKG_TARGET_IS_UWP)
+    set(BUILD_CURSES_DIALOG ON)
+endif()
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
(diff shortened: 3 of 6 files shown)


ports/ncurses/portfile.cmake
```diff
@@ -0,0 +1,43 @@
...
+    list(APPEND OPTIONS
+        --with-shared
+        --with-cxx-shared
+        --without-normal
+    )
+endif()
+
+vcpkg_configure_make(
+    SOURCE_PATH ${SOURCE_PATH}
+    OPTIONS ${OPTIONS}
```
JackBoosY, line 31: I remember there is an option named `--with-debug`.

ports/ncurses/portfile.cmake
```diff
@@ -0,0 +1,43 @@
+vcpkg_download_distfile(
```
JackBoosY, line 1:
```suggestion
vcpkg_fail_port_install(ON_TARGET "Windows")
vcpkg_download_distfile(
```

JackBoosY (changes requested):

JackBoosY (approved):

ports/ncurses/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "ncurses",
+  "version-string": "6.2",
+  "description": "free software emulation of curses in System V Release 4.0",
```
electron271, line 4: typo, replace with "Free"

electron271:
Trying to install right now..

ports/ncurses/vcpkg.json
Hoikas: This was copied and pasted directly from the project website, as is where is

ports/ncurses/portfile.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+vcpkg_extract_source_archive_ex(
+    OUT_SOURCE_PATH SOURCE_PATH
+    ARCHIVE ${ARCHIVE_PATH}
+)
+
+set(OPTIONS
+    --disable-db-install
+    --enable-pc-files
+    --with-pkg-config-libdir=${CURRENT_INSTALLED_DIR}/lib/pkgconfig
```
Neumann-A, line 21: This is only correct for release builds. For debug builds it is ${CURRENT_INSTALLED_DIR}/debug/lib/pkgconfig

ports/ncurses/portfile.cmake
```diff
@@ -0,0 +1,53 @@
...
+    --without-manpages
+    --without-progs
+    --without-tack
+    --without-tests
+)
+if(VCPKG_LIBRARY_LINKAGE STREQUAL dynamic)
+    list(APPEND OPTIONS
+        --with-shared
+        --with-cxx-shared
```
Neumann-A, line 30: this probably is not `VCPK_LIBRARY_LINKAGE` but `VCPKG_CRT_LINKAGE`

Hoikas: This is for the C++ binding. Per `./configure --help`: "--with-cxx-shared       generate C++ shared-libraries"

Neumann-A:
add testing via cmake test port in script/test_ports. The option for cmake is `BUILD_CursesDialog`

electron271:
would it be possible to make it work with windows/uwp? or do I have to stick with pdcurses

Hoikas:
> add testing via cmake test port in script/test_ports. The option for cmake is BUILD_CursesDialog

I'm sorry, but I don't understand this statement.

> would it be possible to make it work with windows/uwp? or do I have to stick with pdcurses

You will need to use MinGW for this to work on Windows. I do not have a functioning MinGW setup, so I cannot support that.

Neumann-A:
> I'm sorry, but I don't understand this statement.

To test if ncurses can be correctly used it is helpful to build an application with it. Since CMake supports building with some sort of `Curses` dependency it would be appropriate to include it in the cmake build by settting `BUILD_CursesDialog=ON`  after https://github.com/microsoft/vcpkg/blob/fe112d8301e13fde9a25dc92d94a0c4a6dc3cf19/scripts/test_ports/cmake/portfile.cmake#L29 and add a dependency on it in the manifest. 

This way it can be checked if it actually builds a useable dialog and links correctly.

electron271:
> You will need to use MinGW for this to work on Windows. I do not have a functioning MinGW setup, so I cannot support that.
well then I guess i'm going to stick to pdcurses or some other lib

Neumann-A:
@Hoikas You probably need to comment out:
https://github.com/microsoft/vcpkg/blob/fe112d8301e13fde9a25dc92d94a0c4a6dc3cf19/scripts/test_ports/cmake/portfile.cmake#L51
for a CI run so that you can get the application name. Otherwise it will simply be deleted.

Hoikas:
Will that be a permanent change, or just so we can test ccmake locally?

Neumann-A:
> Will that be a permanent change, or just so we can test ccmake locally?

temporary but since you seem to already now the application name you can add it to the copy tools section of the port. Somewhere around here: https://github.com/microsoft/vcpkg/blob/fe112d8301e13fde9a25dc92d94a0c4a6dc3cf19/scripts/test_ports/cmake/portfile.cmake#L36-L40

Do you know the deps of ncurses? Maybe I can make it somehow work for windows in #9966.

Hoikas:
Very minimal deps. There's an optional pcre2 dependency (that we're not opting into) and getopt is needed to build the programs. I initially wrote the port on Windows but got nasty build errors about things like `DWORD` not being defined. It was more than I really wanted to investigate at the time 😅

scripts/test_ports/cmake/portfile.cmake
```diff
@@ -26,6 +26,7 @@ vcpkg_configure_cmake(
         -DCMAKE_USE_SYSTEM_JSONCPP=ON
         -DCMAKE_USE_SYSTEM_LIBRHASH=OFF # not yet in VCPKG
         -DCMAKE_USE_SYSTEM_LIBUV=ON
+        -DBUILD_CursesDialog=ON
```
vicroms, line 29: Shouldn't it be `ON` only on non-Windows?

GitMensch:
Does it "just not compile" with MSVC or were there other reasons to not integrate it?
Years ago this was doomed to work because of the missing termcap model but as [ncurses 5.8 added](https://invisible-island.net/ncurses/announce-5.8.html) "added a terminal driver for Windows console" and more recent MSVC versions are not as limited as old ones I've thought that this would work - wouldn't it?

Hoikas:
ncurses does not compile with MSVC; however, I suspect you may find some success with mingw or clang-cl.

dg0yt:
Note that there is port pdcurses for windows. (However it seems to need some maintenance, in particular a port to `vcpkg_build_nmake` which might also fix issues with some triplets.)
Last not least there could be a `curses` meta-port with specific dependencies for each platform.

GitMensch:
> ncurses does not compile with MSVC; ...

Can you point to build tests  / documentation for this?

> ... however, I suspect you may find some success with mingw or clang-cl.

mingw should definitely work as this is what is used in MSYS2.
Are there compiler specific triplets?
Do you suggest compiling ncurses with mingw, then linking against the result with msvc?

> Last not least there could be a `curses` meta-port with specific dependencies for each platform.

That would be _very_ reasonable, and as long as each port installs a curses.h (and/or adds `-DHAVE_PDCURSES_H` and similar) would really help with portability of recipes - a central place for that is much better than conditionals in every package that wants to use curses to let it compile both on Win32 and on other triplets.

dg0yt:
> > ncurses does not compile with MSVC; ...
> 
> Can you point to build tests / documentation for this?

The citation you have given, complete line:

> add a terminal driver for Windows console, which supports a MinGW port to Windows.

Related:
https://invisible-island.net/ncurses/ncurses.html#download_mingw 

Additionally:

> It should port easily to any ANSI/POSIX-conforming UNIX.

> Future plans 
> ... 
> Ports to more systems, including DOS and Windows.

Briefly: Not enough POSIX with MSVC.

PR #17205, opened 2021-04-10
[vcpkg] Don't override pkg-config prefix

dg0yt:
**Describe the pull request**

- What does your PR fix? 
  This PR fixes breaking builds when pkg-config picks up .pc files from system directories: Overriding the prefix will invalidate include and lib paths. 
  Example case: gdal, picking up system OpenEXR in Azure Pipelines' Ubuntu image.

- Which triplets are supported/not supported? Have you updated the CI baseline?
  -/-

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?
 -/-

.pc file from vcpkg ports should be safe when properly using `$(pcfiledir)` and `vcpkg_fixup_pkgconfig`.
Related discussion: https://github.com/microsoft/vcpkg/pull/10402#issuecomment-817085769

Issue #17098, opened 2021-04-05
[libtasn1] build failure in osx pipeline


autoantwort:
**Host Environment**
- OS: macOS CI
- Compiler: unknown

**To Reproduce**
Run pipeline of #17073

**Failure logs** 
https://dev.azure.com/vcpkg/c1ee48cb-0df2-4ab3-8384-b1df5a79fe53/_apis/build/builds/51525/artifacts?artifactName=failure%20logs%20for%20x64-osx&api-version=6.0&%24format=zip

autoconf-x64-osx-err.log:
```
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4 -I m4-gl
autoreconf: configure.ac: tracing
autoreconf: running: glibtoolize --copy --force
autoreconf: configure.ac: not using Intltool
autoreconf: running: gtkdocize --copy
Can't exec "gtkdocize": No such file or directory at /usr/local/Cellar/autoconf/2.71/share/autoconf/Autom4te/FileUtils.pm line 293.
autoreconf: error: gtkdocize failed with exit status: 2
```

**Additional context**
It seems to be similar to #16990

autoantwort:
Seems to be the same error as #17057

NancyLi1013:
Hi @autoantwort 

Thanks for posting this issue.
Sorry, I cannot repro this on my machine.

```
vcpkg@VCPKGMM-12 vcpkg % ./vcpkg install libtasn1 --editable
Computing installation plan...
The following packages will be built and installed:
    libtasn1[core]:x64-osx -> 4.16.0
Could not locate cached archive: /Users/vcpkg/.cache/vcpkg/archives/.zip
Starting package 1/1: libtasn1:x64-osx
Building package libtasn1[core]:x64-osx...
-- Downloading https://ftp.gnu.org/gnu/libtasn1/libtasn1-4.16.0.tar.gz -> libtasn1-4.16.0.tar.gz...
-- Extracting source /Users/vcpkg/Nancy/vcpkg/downloads/libtasn1-4.16.0.tar.gz
-- Using source at /Users/vcpkg/Nancy/vcpkg/buildtrees/libtasn1/src/4.16.0-1a71df00ff
-- Configuring x64-osx-dbg
-- Configuring x64-osx-rel
-- Generating configure for x64-osx
-- Finished generating configure for x64-osx
-- Configuring x64-osx-dbg
[...]

Issue #17057, opened 2021-04-03
[libidn2] build failure on osx pipeline


autoantwort:
**Host Environment**
- OS: macOS CI
- Compiler: unknown

**To Reproduce**
Run pipeline of #17049

**Failure logs** 
https://dev.azure.com/vcpkg/c1ee48cb-0df2-4ab3-8384-b1df5a79fe53/_apis/build/builds/51476/artifacts?artifactName=failure%20logs%20for%20x64-osx&api-version=6.0&%24format=zip

autoconf-x64-osx-err.log:
```
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: running: autopoint --force
autoreconf: running: aclocal --force -I m4 -I unistring/m4
autoreconf: configure.ac: tracing
autoreconf: running: glibtoolize --copy --force
autoreconf: configure.ac: not using Intltool
autoreconf: running: gtkdocize --copy
Can't exec "gtkdocize": No such file or directory at /usr/local/Cellar/autoconf/2.71/share/autoconf/Autom4te/FileUtils.pm line 293.
autoreconf: error: gtkdocize failed with exit status: 2
```

**Additional context**
It seems to be similar to #16990

autoantwort:
Maybe you have to run https://apple.stackexchange.com/questions/108513/cant-exec-glibtoolize-no-such-file-or-directory
on the build nodes

NancyLi1013:
Hi @autoantwort 

Thanks for posting this issue.
I noticed that all CI checks have passed on #17049 now. Does this issue also exist on your machine?

On my virtual box machine, the failures like this:

```
  autoreconf: Entering directory `.'
  autoreconf: running: autopoint --force
  Can't exec "autopoint": No such file or directory at /usr/local/Cellar/autoconf/2.69/share/autoconf/Autom4te/FileUtils.pm line 345.
  autoreconf: failed to run autopoint: No such file or directory
  autoreconf: autopoint is needed because this package uses Gettext
```
I investigated this error and found the reference https://dausruddin.com/autoreconf-failed-to-run-autopoint-no-such-file-or-directory/ and http://tutorialspots.com/how-to-fix-error-autoreconf-failed-to-run-autopoint-no-such-file-or-directory-on-centos-4724.html.

Seems need to install `gettext`. But I cannot install it without permission.

So I'm not sure if it can solve the problem. If possible, you can try on your local machine.

autoantwort:
[...]

PR #17227, opened 2021-04-12
[vcpkg] Add missing autoconf dependencies to osx provisioning

dg0yt:
- What does your PR fix?
  This PR is to install dependencies which are required by autoconf when building ports such as libtasn1 or libidn2. The lack of these dependencies breaks a number of PRs (e.g. #17215, #17205, #17137, #17098, #17057).

  ~linux,~ osx

  -/-

### DIFF
--- a/scripts/azure-pipelines/linux/provision-image.sh
+++ b/scripts/azure-pipelines/linux/provision-image.sh
@@ -12,10 +12,7 @@ APT_PACKAGES="at curl unzip tar libxt-dev gperf libxaw7-dev cifs-utils \
   libxcursor-dev yasm libnuma1 libnuma-dev python-six python3-six python-yaml \
   flex libbison-dev autoconf libudev-dev libncurses5-dev libtool libxrandr-dev \
   xutils-dev dh-autoreconf autoconf-archive libgles2-mesa-dev ruby-full \
-  pkg-config meson autopoint"
-
-# Required by libgpod
-APT_PACKAGES="$APT_PACKAGES gtk-doc-tools"
+  pkg-config meson"
 
 # Additionally required by qt5-base
 APT_PACKAGES="$APT_PACKAGES libxext-dev libxfixes-dev libxrender-dev \
--- a/scripts/azure-pipelines/linux/provision-image.sh
+++ b/scripts/azure-pipelines/linux/provision-image.sh
@@ -12,7 +12,7 @@ APT_PACKAGES="at curl unzip tar libxt-dev gperf libxaw7-dev cifs-utils \
   libxcursor-dev yasm libnuma1 libnuma-dev python-six python3-six python-yaml \
   flex libbison-dev autoconf libudev-dev libncurses5-dev libtool libxrandr-dev \
   xutils-dev dh-autoreconf autoconf-archive libgles2-mesa-dev ruby-full \
-  pkg-config meson"
+  pkg-config meson autopoint"
 
 # Additionally required by qt5-base
 APT_PACKAGES="$APT_PACKAGES libxext-dev libxfixes-dev libxrender-dev \
--- a/scripts/azure-pipelines/linux/provision-image.sh
+++ b/scripts/azure-pipelines/linux/provision-image.sh
@@ -14,6 +14,9 @@ APT_PACKAGES="at curl unzip tar libxt-dev gperf libxaw7-dev cifs-utils \
   xutils-dev dh-autoreconf autoconf-archive libgles2-mesa-dev ruby-full \
   pkg-config meson autopoint"
 
+# Required by libgpod
+APT_PACKAGES="$APT_PACKAGES gtk-doc-tools"
+
 # Additionally required by qt5-base
 APT_PACKAGES="$APT_PACKAGES libxext-dev libxfixes-dev libxrender-dev \
   libxcb1-dev libx11-xcb-dev libxcb-glx0-dev libxcb-util0-dev"
--- a/scripts/azure-pipelines/osx/configuration/Vagrantfile
+++ b/scripts/azure-pipelines/osx/configuration/Vagrantfile
@@ -18,6 +18,7 @@ brew_formulas = [
   'autoconf',
   'automake',
   'bison',
+  'gettext',
   'gfortran',
   'gperf',
   'gtk-doc',
--- a/scripts/azure-pipelines/osx/configuration/Vagrantfile
+++ b/scripts/azure-pipelines/osx/configuration/Vagrantfile
@@ -20,6 +20,7 @@ brew_formulas = [
   'bison',
   'gfortran',
   'gperf',
+  'gtk-doc',
   'libtool',
   'meson',
   'mono',


JackBoosY (approved):

scripts/azure-pipelines/linux/provision-image.sh
```diff
@@ -12,7 +12,7 @@ APT_PACKAGES="at curl unzip tar libxt-dev gperf libxaw7-dev cifs-utils \
   libxcursor-dev yasm libnuma1 libnuma-dev python-six python3-six python-yaml \
   flex libbison-dev autoconf libudev-dev libncurses5-dev libtool libxrandr-dev \
   xutils-dev dh-autoreconf autoconf-archive libgles2-mesa-dev ruby-full \
-  pkg-config meson"
+  pkg-config meson autopoint"
```
Neumann-A, line 15: also add:

```
# Required by libgpod
APT_PACKAGES="$APT_PACKAGES gtk-doc-tools"
```

dg0yt: done.

dg0yt:
I added gettext also to osx, as suggest in #17057. So similar changes are done to linux and osx now

I would like to add that wile gettext/autopoint don't have significant dependencies (m4 on linux), gtk-doc-tools/gtk-doc come with a longer list on [Ubuntu](https://packages.ubuntu.com/focal/gtk-doc-tools) and [macOS](https://formulae.brew.sh/formula/gtk-doc#default). 

And for gettext, there is also a port.

strega-nil:
You may want to split out linux and osx, since osx is a lot easier to update (and I can personally do it).

dg0yt:
> You may want to split out linux and osx, since osx is a lot easier to update (and I can personally do it).

This PR is now reduced to osx (matching my initial branch naming). Linux is in #17278.

Looing forward to see the autoconf backlog being resolved.

strega-nil:
cool, thanks @dg0yt. I will try this out soon.

PhoebeHui:
@Neumann-A, I noticed that you have already added feature 'tools' which contains autopoint to gettext port by PR https://github.com/microsoft/vcpkg/pull/17278, is it possible to replace 'brew install gettext' on osx or linux?

Neumann-A:
@PhoebeHui You can try it but I don't exactly know how it interacts which the whole m4 stuff which is normally needed. Just put the corresponding binary on `PATH` and it should be found. Be aware that building `gettext[tools]` is also a bit time consuming.

dg0yt:
On Linux, the gettext port is empty at the moment:
https://github.com/microsoft/vcpkg/blob/b1e352c60e706098ac3b5421856ae012eeeb832e/ports/gettext/portfile.cmake#L1-L7

Regarding build time, it might be possible to create a separate feature for autotools. It is a script file which has its own target in make. Maybe its enough to satisfy the autoconf.

dg0yt:
> ... for autotools ...

"for autopoint"

PhoebeHui:
Verified it on macOS platform, and it works. however, the gettext doesn't work on linux, and building gettext[tools] is also a bit time consuming, we may consider to install gettext on pipeline test machines? 

@strega-nil, what do you think?

Verified:
1. ./vcpkg install 'gettext[tools]'
2. Add 'vcpkg_add_to_path(PREPEND "${CURRENT_INSTALLED_DIR}/tools/gettext/bin")' to fontconfig portfile.
3. ./vcpkg install fontconfig
4. fontconfig:x64-osx successfully installed.

Neumann-A:
`vcpkg_add_to_path(PREPEND "${CURRENT_INSTALLED_DIR}/tools/gettext/bin")`
should be 
`vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/gettext/bin")`

dg0yt:
I'm playing with the gettext port on Linux, but it is relevant for osx as well:

I was able to establish an pure "autopoint" feature, running configure in `/gettext-tools` (slow) and running install in `/gettext-tools/misc` (fast). At the end, the package is just the `autopoint` script and a compressed tarball.

Given that this "build" just consists of configuring the `autopoint.in` script template (with `@variable@` placeholders) and compressing a tarball, the slow configure step could be eliminated by using a small cmake script instead of using configure + make. The script has to fill in some paths and platform details, and install the script and tarball.

dg0yt:
> a small cmake script instead of using configure + make

Works for me, for windows/linux/osx/mingw.

Looking at the `gtkdocize`: I don't know about the build times, but `brew install gtk-doc` takes some time and disk space when it pulls in "docbook"

The autopoint solution pattern could work for gtkdocize, too: it is a script which runs during autoconf and doesn't come with direct dependencies. However there is no `gtk-doc` port yet, for adding a feature. Creating a full port with all dependencies is a much bigger task.

dg0yt:
> > a small cmake script instead of using configure + make
> 
> Works for me, for windows/linux/osx/mingw.

#17317

autoantwort:
> Looking at the gtkdocize: I don't know about the build times, but brew install gtk-doc takes some time and disk space when it pulls in "docbook"

Huh on my machine it is as fast as installing other packages. It needs 10 seconds to install the package and dependencies.

dg0yt:
> Huh on my machine it is as fast as installing other packages. It needs 10 seconds to install the package and dependencies.

I need to this in Azure Pipelines, and it takes more than one minute:
~~~
2021-04-16T07:16:19.8328510Z ==> Installing dependencies for gtk-doc: docbook, docbook-xsl, libxml2, boost and source-highlight
...
2021-04-16T07:17:24.8477650Z   /usr/local/etc/bash_completion.d
~~~

And now I see that it also pulls in `boost`. This takes some time, and generates `/usr/local/Cellar/boost/1.75.0_2: 15,058 files, 524.2MB`.

PhoebeHui:
```
Installing dependencies for gtk-doc: docbook, docbook-xsl, libxml2, boost and source-highlight
```
Yes, I concerned to install gtk-doc on osx test machines, I'm not sure if it would cause conflicts with these packages which installed by vcpkg.

autoantwort:
> Yes, I concerned to install gtk-doc on osx test machines, I'm not sure if it would cause conflicts with these packages which installed by vcpkg.

It is already installed on some machines. Otherwise the pipelines would not fail randomly based on the machine they run on.

dg0yt:
I have no clue if there is any interest in offering a lightweight `gtk-doc[gtkdocize]` feature (without the rest of gtk-doc), similar to the lightweight `gettext[autopoint]` feature. I decided to not start work on it until I see interest in such contributions.

strega-nil:
I think this is good; however, I'd prefer to roll this into PR #17376

dg0yt:
Thanks. 
I think the main decision is still whether the autoconf dependencies are a user/CI responsibility, or if vcpkg includes (at least minimimal) ports to be used by port maintainers as host dependencies. It might be helpful to add a note to `vcpkg_configure_make` documentation.

dg0yt:
Assuming that `autopoint` and `gtkdocize` are maintainer tools which are rarely actually needed for vcpkg ports, I offer one more alternative proposal to updating CI provisioning, now in #17406: Replace the calls to the missing maintainer tools by no-op calls to `true` (i.e. `/usr/bin/true`). 
- This is the fastest solution for vcpkg ports which don't really need to update to these parts of the build systems. 
- It solves the issue both for CI and for vcpkg users.
- It is entirely implemented in `vcpkg_configure_make`.
- It will still make use of the tools when (made) available.

ras0219-msft (approved):
LGTM, though this will be blocked until we reprovision OSX.

I also do very much like fast solutions :) and if we can get effectively identical results while doing less work I'm all for it.

Finally, if gettext tools can be made fast enough, building our own copies of the tools would be better than needing the system copies.

dg0yt:
Today I would rather propose to not merge this. It solves issues in CI but not for users, and better solutions start to get merged.
- The `gtkdocize` aspect of `libidn2` and `libtasn1` is resolved by merging #17419, #17420. 
  One only has to remember to add `set(ENV{GTKDOCIZE} true)` in ports which don't need to build documentation. 
- The `autopoint` aspect is solved with host dependencies either on `gettext[tools]` (slow) or on `gettext[autopoint]` (fast, from #17317)

JackBoosY:
@dg0yt So please close this issue or change it to draft.

PR #17229, opened 2021-04-12
[catch2] Add patch to lower-case cmake config dir name

### LINKED ISSUE #17228: [catch2] build failure
**Host Environment**
- OS: Ubuntu 20.04 inside VSCode dev-container running on Windows
- Compiler: clang 10

**To Reproduce**
* Use VSCode on Windows and create a Linux dev container
* Inside the mapped `/workspaces/<your-project-name>` directory clone `vcpkg`
* Run `vcpkg install catch2`

**Failure logs** 
```
Starting package 1/7: catch2:x64-linux
Building package catch2[core]:x64-linux...
Building package catch2[core]:x64-linux... done
Installing package catch2[core]:x64-linux...
failed: /workspaces/project/build/foo/x64-linux/share/catch2: File exists
failed: /workspaces/project/build/foo/x64-linux/share/catch2/Catch.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/ParseAndAddCatchTests.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/CatchAddTests.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/Catch2Config.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/Catch2ConfigVersion.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/Catch2Targets.cmake: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/copyright: No such file or directory
failed: /workspaces/project/build/foo/x64-linux/share/catch2/vcpkg_abi_info.txt: No such file or directory
Installing package catch2[core]:x64-linux... done
```

wildmichael:
Otherwise on Linux there will be both share/Catch2 and
share/catch2 and when developing on Windows inside a Linux
dev-container, installation in the mapped file system fails in
vcpkg_fixup_targets() because vcpkg is trying to make the directory
share/cmake/catch2 but share/cmake/Catch2 already exists.

**Describe the pull request**

- What does your PR fix? Fixes #17228

- Which triplets are supported/not supported? Have you updated the CI baseline? N/A

- Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)? to the best of my knowledge...

### DIFF
--- a/ports/catch2/portfile.cmake
+++ b/ports/catch2/portfile.cmake
@@ -4,6 +4,8 @@ vcpkg_from_github(
     REF fd9f5ac661f87335ecd70d39849c1d3a90f1c64d # v2.13.1
     SHA512 4fafd06006034cc02dddd22c381b5817549834dae0aff29ed598edd21a3c67f8ac61a77f51b06f3c59baa96a114ecb19c6df09126215bfc00bef94f8f77b810d
     HEAD_REF master
+    PATCHES
+        lowercase-cmake-config-dir.patch
 )
 
 vcpkg_configure_cmake(
@@ -16,7 +18,7 @@ vcpkg_configure_cmake(
 
 vcpkg_install_cmake()
 
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/Catch2)
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/catch2)
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug ${CURRENT_PACKAGES_DIR}/lib)
 
--- a/ports/catch2/CONTROL
+++ b/ports/catch2/CONTROL
@@ -1,4 +1,5 @@
 Source: catch2
 Version: 2.13.1
+Port-Version: 1
 Description: A modern, header-only test framework for unit testing.
 Homepage: https://github.com/catchorg/Catch2
--- a/ports/catch2/lowercase-cmake-config-dir.patch
+++ b/ports/catch2/lowercase-cmake-config-dir.patch
@@ -1,0 +1,22 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index b515105..c69c96e 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -108,7 +108,7 @@ add_library(Catch2::Catch2 ALIAS Catch2)
+ # see https://github.com/catchorg/Catch2/issues/1373
+ if (NOT_SUBPROJECT)
+     include(CMakePackageConfigHelpers)
+-    set(CATCH_CMAKE_CONFIG_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/Catch2")
++    set(CATCH_CMAKE_CONFIG_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/catch2")
+ 
+     configure_package_config_file(
+         ${CMAKE_CURRENT_LIST_DIR}/CMake/Catch2Config.cmake.in
+@@ -199,7 +199,7 @@ if (NOT_SUBPROJECT)
+         "contrib/gdbinit"
+         "contrib/lldbinit"
+       DESTINATION
+-        ${CMAKE_INSTALL_DATAROOTDIR}/Catch2
++        ${CMAKE_INSTALL_DATAROOTDIR}/catch2
+     )
+     endif()
+ 
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -1110,7 +1110,7 @@
     },
     "catch2": {
       "baseline": "2.13.1",
-      "port-version": 0
+      "port-version": 1
     },
     "cccapstone": {
       "baseline": "9b4128ee1153e78288a1b5433e2c06a0d47a4c4e-1",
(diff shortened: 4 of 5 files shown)


ports/catch2/portfile.cmake
```diff
@@ -16,7 +18,7 @@ vcpkg_configure_cmake(
 
 vcpkg_install_cmake()
 
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/Catch2)
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/catch2)
```
NancyLi1013, line 21: Currently, seems that the config and targets cmake files are specified to different directories. On Windows and osx, those files are listed in _share/Catch2_ directory, while on Linux, they are listed in  _share/catch2_ directory.
So we might consider to update as follows:
```
if(VCPKG_TARGTE_IS_WINDOWS OR VCPKG_TARGET_IS_OSX)
    vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/Catch2)
elseif(VCPKG_TARGET_IS_LINUX)
    vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/Catch2 TARGET_PATH share/catch2)
endif()
```
Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_fixup_cmake_targets.md#target_path

So could you please if this can work fine for you?

NancyLi1013 (changes requested):

ports/catch2/portfile.cmake
wildmichael: I think the different listings on Windows and OSX are only due to their case-insensitive file systems. I tried using `TARGET_PATH`, it didn't fix anything. The root problem is, if I understand things correctly, is that `vcpkg_fixup_cmake_targets` tries to use move things to a directory named like the port, i.e. `share/catch2`. However, `share/Catch2` already exists, and this trips things. On Linux you get the two directories (the fixup function seems to leave some files behind, namely `lldbinit` and `gdbinit`) because the they can coexist. On pure Windows and Mac you only get one directory, because it is assumed that path names are case-insensitive. It breaks down when CMake (or the fixup) script assumes the file system to be case sensitive (after all, it runs in Linux), but the underlying file system in fact isn't.

NancyLi1013: Thanks for your further investigation about this.

ports/catch2/portfile.cmake
```diff
@@ -4,6 +4,8 @@ vcpkg_from_github(
     REF fd9f5ac661f87335ecd70d39849c1d3a90f1c64d # v2.13.1
     SHA512 4fafd06006034cc02dddd22c381b5817549834dae0aff29ed598edd21a3c67f8ac61a77f51b06f3c59baa96a114ecb19c6df09126215bfc00bef94f8f77b810d
     HEAD_REF master
+    PATCHES
+      lowercase-cmake-config-dir.patch
```
NancyLi1013, line 8:
```suggestion
    PATCHES
        lowercase-cmake-config-dir.patch
```

NancyLi1013 (changes requested):

Neumann-A:
Remove patch use TARGET_PATH in `vcpkg_fixup_cmake_targets` with correct case instead.

wildmichael:
@Neumann-A I tried this, it fails in this particular case.

wildmichael:
Ok, I found the real culprit. The problem is this:

Catch2 installs the CMake configuration files to `lib/cmake/Catch2/` and some debugger initialization scripts to `share/Catch2`.

`vcpkg_fixup_cmake_targets()` however, tries to move the files from `lib/cmake/Catch2/` to `share/catch2/`.  And that's where things go wrong, because that directory already exists, but starting with an upper-case `C`.

The fix is quite simple:

```diff
diff --git a/ports/catch2/portfile.cmake b/ports/catch2/portfile.cmake
index 72365eb46..dc273552a 100644
--- a/ports/catch2/portfile.cmake
+++ b/ports/catch2/portfile.cmake
@@ -16,6 +16,9 @@ vcpkg_configure_cmake(
 
 vcpkg_install_cmake()
 
+file(RENAME ${CURRENT_PACKAGES_DIR}/share/Catch2 ${CURRENT_PACKAGES_DIR}/share/catch2)
+file(RENAME ${CURRENT_PACKAGES_DIR}/debug/share/Catch2 ${CURRENT_PACKAGES_DIR}/debug/share/catch2)
+
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/Catch2)
 
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug ${CURRENT_PACKAGES_DIR}/lib)
```

I like this much better than the patch file. What do you think @Neumann-A and @NancyLi1013?

Neumann-A:
needs extra `"` around the paths

wildmichael:
It doesn't have it in the other places where it uses `${CURRENT_PACKAGES_DIR}`. Should I fix those too?

wildmichael:
I opened a new separate PR #17271 with the above described fix plus the path quoting. If you agree, we can reject this PR here.

NancyLi1013:
This PR may be closed since you have opened another PR #17271 to fix the same issue. What do you think? @wildmichael

wildmichael:
@NancyLi1013, yes, I agree.

PR #17235, opened 2021-04-12
[fluidsynth] update portfile features.

Neumann-A:
Just to make sure it does not pick up any optional dependencies.


ports/fluidsynth/vcpkg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "fluidsynth",
+  "version": "2.1.4",
+  "port-version": 1,
+  "description": "FluidSynth reads and handles MIDI events from the MIDI input device. It is the software analogue of a MIDI synthesizer. FluidSynth can also play midifiles using a Soundfont.",
+  "homepage": "https://github.com/FluidSynth/fluidsynth",
+  "dependencies": [
+    "glib"
+  ]
```
autoantwort, line 9: Did you not have to list the features here that you use in the portfile ?

Neumann-A: For #13100 I need pkgconfig enabled to correctly link glib. But fluidsynth has a lot of optional dependencies which are picked up by that. So there is two options here. I simply disable them all in the portfile or I use vcpkg_check_features without defining the features. In the latter case somebody who actually cares about the features can simply add them to the vcpkg.json and test them without additionally touching the portfile.

PhoebeHui: For me, I think you can simply disable them all in portfile, and when some features required from user, we can add that back as a feature.

@strega-nil, what do you think?

strega-nil: I agree, personally

PhoebeHui (approved):

strega-nil:
Thanks @Neumann-A :)

PR #17280, opened 2021-04-14
[fontconfig] Don't require uuid port for mingw

dg0yt:
- #### What does your PR fix?  
  This PR fixes a failing dependency of fontconfig on mingw: The libuuid port is `linux|osx`. On mingw as on a regular windows, there is a static uuid system library.

  mingw, -/-.

  Yes

  Yes

(There is another fontconfig mingw issue due to the pthread dependency but I would like to resolve this there.)

### DIFF
--- a/ports/fontconfig/CONTROL
+++ b/ports/fontconfig/CONTROL
@@ -1,6 +1,6 @@
 Source: fontconfig
 Version: 2.13.1
-Port-Version: 5
+Port-Version: 6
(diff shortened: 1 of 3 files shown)


JonLiu1993:
@dg0yt ,The failures caused by popsift will be fixed by #17277.

strega-nil-ms:
It's so weird to me that mingw =/> windows... thanks for the PR @dg0yt :)

Issue #17072, opened 2021-04-04
[wxwidgets] error "No Target! You should use wx-config program for compilation flags!"


SamuelMarks:
**Is your feature request related to a problem? Please describe.**

### `CMakeLists.txt`
```cmake
cmake_minimum_required(VERSION 3.0)
cmake_policy(SET CMP0048 NEW)
project(pp VERSION 0.0.0 LANGUAGES CXX)
set(CMAKE_CXX_STANDARD 14)


file(DOWNLOAD https://raw.githubusercontent.com/wxWidgets/wxWidgets/c034a78/samples/widgets/button.cpp src/main.cpp
     EXPECTED_HASH SHA256=c94699427f8b4a90c7ae3e075210bc44ae80098b5303a564f11d3991c620328f)
file(DOWNLOAD https://raw.githubusercontent.com/wxWidgets/wxWidgets/WX_3_0_BRANCH/samples/widgets/icons/button.xpm icons/button.xpm
     EXPECTED_HASH SHA256=c7f7bf9a4b678376a20af95adc213a467450a71040710dacc168fa6885a1ad0f)

set(src ${CMAKE_CURRENT_BINARY_DIR}/src/main.cpp)
find_package(wxWidgets REQUIRED)
add_executable(${PROJECT_NAME} ${src})
target_include_directories(${PROJECT_NAME} PRIVATE ${wxWidgets_INCLUDE_DIRS})
target_link_libraries(${PROJECT_NAME} PRIVATE ${wxWidgets_LIBRARIES})
````

### `shell`
```sh
$ [path]/vcpkg install wxwidgets
$ mkdir build && cd $_
$ cmake -DCMAKE_TOOLCHAIN_FILE=[root]/vcpkg/scripts/buildsystems/vcpkg.cmake ..
$ cmake --build .
[ 50%] Building CXX object CMakeFiles/pp.dir/src/main.cpp.o
In file included from /tmp/a/build/src/main.cpp:20:
In file included from vcpkg/packages/wxwidgets_x64-osx/include/wx-3.1/wx/wxprec.h:12:
vcpkg/packages/wxwidgets_x64-osx/include/wx-3.1/wx/defs.h:62:13: error: "No Target! You should use wx-config program for compilation flags!"
#           error "No Target! You should use wx-config program for compilation flags!"
            ^
In file included from /tmp/a/build/src/main.cpp:28:
[...]

PR #17314, opened 2021-04-16
[wxWidgets] update to 3.1.5

### LINKED ISSUE #17313: [wxWidgets] update to 3.1.5
Library name: `wxwidgets`

New version number: `3.1.5`

https://github.com/wxWidgets/wxWidgets/releases/tag/v3.1.5

https://www.wxwidgets.org/news/2021/04/wxwidgets-3.1.5-released/

electron271:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #17313

  UWP is not supported

  `I think so..`

 No

### DIFF
--- a/ports/wxwidgets/vcpkg.json
+++ b/ports/wxwidgets/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "wxwidgets",
-  "version-semver": "3.1.4",
-  "port-version": 7,
+  "version-semver": "3.1.5",
+  "port-version": 8,
   "description": "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications.",
   "homepage": "https://github.com/wxWidgets/wxWidgets",
   "supports": "!uwp",
--- a/ports/wxwidgets/vcpkg.json
+++ b/ports/wxwidgets/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "wxwidgets",
   "version-semver": "3.1.5",
-  "port-version": 8,
+  "port-version": 0,
   "description": "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications.",
   "homepage": "https://github.com/wxWidgets/wxWidgets",
   "supports": "!uwp",
--- a/versions/w-/wxwidgets.json
+++ b/versions/w-/wxwidgets.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "b284aa561d3ec113a892e6fb37e4b4333bf60517",
+      "version-semver": "3.1.5",
+      "port-version": 8
+    },
+    {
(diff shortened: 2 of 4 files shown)


electron271:
how exactly to update the patch?

SamuelMarks:
@electron271 You might also want to look at #17072 and #17111

JonLiu1993:
@electron271 ,usually update the port, there are the following steps:
1. Convert the control file to the vcpkg.json file, and modify the version inside to the target version. If the version is a special combination method, you need to use version, version-string, version-semver or version-date respectively. For details, please refer to [here](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#versioning). If there is port-version before update, it needs to be set to 0 or deleted.
2. Modify the ref and hash value in portfile.cmake.
3. If you need to update the patch in the port, you need to investigate the port history update record and combine the target version in the soure to remake the patch.

- If the previous patch has been successfully added to the source of the target version, you can delete the patch.
- If the latest version of the source does not contain the modifications in the patch and the patch cannot be used normally, you need to remake the patch.
- If the latest version of the source does not contain the patch modification content, the patch can be used directly, just keep the patch intact and use it directly.

The above is the most basic way to update ports. Of course, this is the only way to update all ports. We have more small details and gradual update strategies that need to be paid attention to at any time. Hope these can help you, thanks for your contribution

ports/wxwidgets/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "wxwidgets",
-  "version-semver": "3.1.4",
-  "port-version": 7,
+  "version-semver": "3.1.5",
+  "port-version": 8,
```
JonLiu1993, line 4: @electron271 ,please delete port-version or set port-version to 0

versions/w-/wxwidgets.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "b284aa561d3ec113a892e6fb37e4b4333bf60517",
+      "version-semver": "3.1.5",
+      "port-version": 8
```
JonLiu1993, line 6: please delete port-version or set port-version to 0

JonLiu1993:
@electron271 ,I found that the current problem of CI is due to the incorrect use of patch. You can try to remake the patch.
````
Applying patch fix-stl-build-vs2019-16.6.patch
CMake Error at scripts/cmake/z_vcpkg_apply_patches.cmake:57 (message):
  Applying patch failed: Checking patch include/wx/hashmap.h...

````

electron271:
how would I remake the patch?

JonLiu1993:
> how would I remake the patch?

@electron271 ,You can follow the steps below

Direct   to the libs path src: such as :C:\fixDate\vcpkg\buildtrees\date\src\date-2.4
--
Git init .
Git add .
Git commit -m  "test amp"
Modify source
git   diff  | out-file -enc ascii fix-.patch

Issue #17237, opened 2021-04-12
[ffmpeg] support community builds that disable unneeded codecs/filters/(de)muxers/devices/...


mcmtroffaes:
****

**Is your feature request related to a problem? Please describe.**
It is very common for users to wish to do custom FFmpeg builds that disable unneeded features, to save resources such as build times or file sizes. FFmpeg supports hundreds of configure options to enable or disable specific filters, codecs, muxers, demuxers, devices, ... It's not feasible for vcpkg to expose each individual feature of FFmpeg. It would be great if vcpkg were to allow community builds to disable unneeded features. The question is whether this is feasible, and if so, how best to achieve this.

**Proposed solution**
One solution would be to adding a text file to the FFmpeg port folder where users can add their own configure options. These options would then be read and passed to the configure script by the FFmpeg port file.

**Describe alternatives you've considered**
I'm not sure. I'm posting this here to elicit some feedback as to whether the proposed solution is a sensible approach, or whether some other/better solution for this already exists in vcpkg.

**Additional context**
N/A

Neumann-A:
Expose all features in the FFMPEG port. Define common features sets as separate features. Let people not covered by this case use manifest mode and explicitly list the features? I don't see a problem in exposing a lot of feature flags as long as they are additive.

mcmtroffaes:
Yes, it would be great to have a single feature for each individual codec, filter, (de)muxer, etc. but wouldn't that be too bloated? Can vcpkg handle several hundreds of features?

Neumann-A:
> Yes, it would be great to have a single feature for each individual codec, filter, (de)muxer, etc. but wouldn't that be too bloated? 

Only export features which can actually be toggled and make sense for a user to toggle. E.g. qt(6)base has >400 possible features but some of them don't make sense, e.g. `FEATURE_xml` can be toggled but the build will simply fail because the targets behind that feature are used elsewhere in the buildsystem.

> Can vcpkg handle several hundreds of features?

Why shouldn't it be able to handle it?

mcmtroffaes:
> Only export features which can actually be toggled and make sense for a user to toggle. E.g. qt(6)base has >400 possible features [...]

It's encouraging to know that vcpkg can handle packages with over 400 individual features.

> > Can vcpkg handle several hundreds of features?
> 
> Why shouldn't it be able to handle it?

[...]

PR #17323, opened 2021-04-16
[ffmpeg] expose individual encoder/decoder groups as features

mcmtroffaes:
**Describe the pull request**

- #### What does your PR fix?  
  See #17237. This PR adds support for selecting groups of encoders/decoders, enabling ffmpeg builds that do not have all encoders/decoders enabled: audio-only builds, video-only builds, encoding-only builds, or anything else as long as the feature set allows it. Common codecs have been exposed as individual features, but more can be easily added. Because there are so many codecs, the list of codecs have been parsed from the ffmpeg source tree. The parsing source code (written in Python) is also included with the PR. This script can be rerun on ffmpeg version updates (or, the code could also be manually maintained, it's not very complicated, but having a script seems convenient to ensure every possible codec is caught from the ffmpeg source). The script can also help adding new codec groups as features.

  I've manually tested a few features and it works but by no means have I fully tested this yet. I'm mainly aiming to get some feedback as to whether this approach is sensible, whether anything can be improved, and perhaps whether or not this level of detail for ffmpeg is desired for vcpkg. If this gets a go-ahead, I'll do a much more in-depth testing run.

  No changes.

  Yes, I've done my best to do so.

  Yes.


mcmtroffaes:
Rebased and conflicts resolved.

mcmtroffaes:
I've been thinking how to remove the need for the external python script. If this can be rewritten in cmake, the code parsing could in principle happen directly from the portfile, removing the need for an external script to identify the appropriate codec names for each feature. I would need to have a think however on how to best represent each feature in cmake, in a way that the portfile remains easily maintainable.

In the meantime, I'd appreciate some feedback on this.

autoantwort:
> In the meantime, I'd appreciate some feedback on this.

I have no idea about ffmpeg and this features, but I thing in general when this solution will be followed up it should generate a vcpkg.json file instead of a CONTROL file to be future proof.

mcmtroffaes:
Thanks @autoantwort that's a fair comment, and it should be easy to adapt. The conversion to json for ffmpeg should probably go into a separate PR. I think I have something that also avoids having to run anything for generating the portfile part by implementing the source code scanning in cmake, unless folks want this to go in to wait for improvements of implementation later.

@cenit @Sibras Any thoughts from your end on this as ffmpeg port specialists (both conceptual and/or on the implementation)?

cenit:
@autoantwort is right. You cannot modify CONTROL files adding features anymore per vcpkg rules ("all new features are added only to vcpkg.json manifest files", https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/control-files.md), so this PR should follow that and already move ffmpeg to manifest.
It's easy, it's just a matter of `./vcpkg format-manifest ports/<portname>/CONTROL
`

Sibras:
This kind of feature is a can of worms which is why I havnt commented earlier as what youve done is admirable but it requires alot of work to get completed and maintain. 

From my own experience, scraping the ffmpeg configure script is not trivial. A quick check showed that ffmpeg currently is exposing 2,379 different configurable decoders/encoders/demuxers/muxers/filters/devices/hwaccels/parsers/bsfs/protocols/indevs/outdevs etc. At the moment your only doing encoders/decoders but if you were to continue on this path to all other features then it gets complex fast.

You then have to work out which external dependency libs map to which configurable option which currently you appear to be hard coding (which makes maintenance an issue as they do change often enough to be a pain). For reference you can scan 'configure' for ${config_option}_deps/${config_option}_deps_any/${config_option}_select/${config_option}_suggest for the dependencies (for just external libs '_deps' or '_deps_any' has that information). But as several config options are dependent on other options (such as decoders depending on bsf's or filters depending on encoders) then _select and _suggest need to be checked to detect the dependency trees between them. Some external dependencies are also used by different options (such as decoders and demuxers) so its not 1:1.

There are also some configure options that are actually just aliases for others and some options that when enabled will internally enable/disable others (such as nvdec/cuvid or sdl/sdl2 etc.).

For reference as your currently hardcoding the licensing dependencies, you can scrape them from the configure file by searching for the lists EXTERNAL_LIBRARY_GPL_LIST/EXTERNAL_LIBRARY_NONFREE_LIST etc.

Hardcoding the decoders/encoders into groups like what you are currently doing simplifies the options exposed to the user (which is not a bad thing) but then a discussion probably needs to be had about whether if your going down this route then whether all options should just be exposed anyway. Hardcoding the groups also seems like a long term maintenance burden as there are a lot of odd components that dont easily fit into the existing groups and new ones pop in quite frequently.

You then also have the problem that downstream ports that use ffmpeg require specific components to be enabled so keeping them updated to detect when ffmpeg hasnt been built with those specific components could be an significant issue.

For starters ffmpeg currently has options for bulk disabling decoders/demuxers/devices etc. that can be easily exposed in the port without requiring any scripting so thats alot easier to add (11 options is alot less than 2379). However unfortunately it doesnt offer the granularity to specify audio/video decoders etc.

mcmtroffaes:
Thanks for your insightful comments, everyone. Happy to switch to a json format for this PR.

I agree that trying to reproduce and maintain the entire dependency structure from the configure script would be a horrible burden. External dependencies that reside in vcpkg do need to be tracked of course, to the best of our ability, but these are fairly limited, and are quite easy to detect from the configure script. For other dependencies, I think it's enough for vcpkg to print the configure warning messages to the user. A hypothetical example:

```bash
$ ./configure --disable-avcodec --disable-filters --enable-filter=afftdn
[...]
WARNING: Disabled afftdn_filter because not all dependencies are satisfied: avcodec
```

Even if there are external dependencies that are not yet supported, the configure script will do the right thing for us:

```bash
$ ./configure --disable-encoders --enable-encoder=h264_amf
[...]
WARNING: Disabled h264_amf_encoder because not all dependencies are satisfied: amf
```

The key thing here is that _these warnings don't cause the build to fail_. The only consequence is that the requested feature will not be available due to a missing dependency.

Of course, these dependencies could be manually maintained, however maybe we don't have to. I'd propose to simply make a clear statement that it is up to the individual user to adjust the dependencies of their build to enable the features they want (e.g. by adding a statement along with any warning messages captured from the configure output). This way, it does not become a maintenance burden, whilst we can still support highly configurable builds.

NancyLi1013:
@mcmtroffaes Is work still being for this PR? Seems there is no any progress for a long time.

mcmtroffaes:
Thanks for checking up on this. Yes, I think this can be closed for now. Should the corresponding issue also be closed? I'm not sure if anyone is waiting for it. I might do a more modest implementation of it at some point (i.e. enabling/disabling all encoders or decoders, since that's just a simple configure switch).

NancyLi1013:
Thanks for your instant response and also for your contribution. The corresponding issue can be kept open, since it is not solved now. You can ping us to reopen this PR if you still want to continue working on this PR in the future.

PR #17332, opened 2021-04-17
[aom] Add new port

HappySeaFox:
**Describe the pull request**

Add AV1 codec library

- #### What does your PR fix?  
\-

Triplets: Everything except UWP and ARM.

CI baseline is not updated.

Yes.

Yes.

### DIFF
--- a/ports/aom/aom-install.diff
+++ b/ports/aom/aom-install.diff
@@ -1,0 +1,72 @@
+diff -pruN aom-3.0.0.o/cmake/aomConfig.cmake.in aom-3.0.0/cmake/aomConfig.cmake.in
+--- aom-3.0.0.o/cmake/aomConfig.cmake.in	1970-01-01 03:00:00.000000000 +0300
++++ aom-3.0.0/cmake/aomConfig.cmake.in	2021-04-16 16:06:19.560725100 +0300
+@@ -0,0 +1,2 @@
++@PACKAGE_INIT@
++include(${CMAKE_CURRENT_LIST_DIR}/aomTargets.cmake)
+diff -pruN aom-3.0.0.o/CMakeLists.txt aom-3.0.0/CMakeLists.txt
+--- aom-3.0.0.o/CMakeLists.txt	2021-04-15 20:05:52.695181200 +0300
++++ aom-3.0.0/CMakeLists.txt	2021-04-16 16:08:54.160941600 +0300
+@@ -49,6 +49,9 @@ endif()
+ set(SO_VERSION 3)
+ set(SO_FILE_VERSION 3.0.0)
+ 
++include(GNUInstallDirs)
++include(CMakePackageConfigHelpers)
++
+ include("${AOM_ROOT}/build/cmake/aom_configure.cmake")
+ include("${AOM_ROOT}/aom_dsp/aom_dsp.cmake")
+ include("${AOM_ROOT}/aom_mem/aom_mem.cmake")
+@@ -273,6 +276,52 @@ if(BUILD_SHARED_LIBS)
+   endif()
+ endif()
+ 
++set(PUBLIC_HEADERS
++    aom/aom.h
++    aom/aom_codec.h
++    aom/aom_decoder.h
++    aom/aom_encoder.h
++    aom/aom_frame_buffer.h
++    aom/aom_image.h
++    aom/aom_integer.h
++    aom/aomcx.h
++    aom/aomdx.h
++)
++
++set_target_properties(aom PROPERTIES
++                          PUBLIC_HEADER "${PUBLIC_HEADERS}")
++
++
++target_include_directories(aom
++                            PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
++                                   $<INSTALL_INTERFACE:include>)
++
++install(TARGETS aom
++        EXPORT aomTargets
++        ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
++        LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
++        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
++        PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/aom")
++
++install(EXPORT aomTargets
++        FILE aomTargets.cmake
++        NAMESPACE AOM::
++        DESTINATION lib/cmake/aom)
++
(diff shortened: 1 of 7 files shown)


ports/aom/aom-rename-static.diff
```diff
@@ -0,0 +1,12 @@
+diff -pruN aom-3.0.0.o/CMakeLists.txt aom-3.0.0/CMakeLists.txt
+--- aom-3.0.0.o/CMakeLists.txt	2021-04-15 20:05:52.695181200 +0300
++++ aom-3.0.0/CMakeLists.txt	2021-04-15 22:34:16.147522600 +0300
+@@ -249,7 +249,7 @@ endif()
+ add_library(aom ${AOM_SOURCES} $<TARGET_OBJECTS:aom_rtcd>)
+ if(BUILD_SHARED_LIBS)
+   add_library(aom_static STATIC ${AOM_SOURCES} $<TARGET_OBJECTS:aom_rtcd>)
+-  set_target_properties(aom_static PROPERTIES OUTPUT_NAME aom)
++  set_target_properties(aom_static PROPERTIES OUTPUT_NAME aom_static)
```
NancyLi1013, line 9: Why do you rename the output name?

happy-fox-shark: It conflicts with the shared target name. Ninja fails to build it.

ports/aom/portfile.cmake
```diff
@@ -0,0 +1,47 @@
...
+vcpkg_from_git(
+    OUT_SOURCE_PATH SOURCE_PATH
+    URL "https://aomedia.googlesource.com/aom"
+    REF 307ce06ed82d93885ee8ed53e152c9268ac0d98d
+    PATCHES
+        aom-rename-static.diff
+        aom-install.diff
+)
+
+vcpkg_configure_cmake(
```
NancyLi1013, line 22: Would you like to update the deprecated function as the new one? Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_configure_cmake.md

We prefer to use the suggested function for new ports.

ports/aom/portfile.cmake
```diff
@@ -0,0 +1,47 @@
...
+    PREFER_NINJA
+    OPTIONS
+        -DENABLE_DOCS=OFF
+        -DENABLE_EXAMPLES=OFF
+        -DENABLE_TESTDATA=OFF
+        -DENABLE_TESTS=OFF
+        -DENABLE_TOOLS=OFF
+)
+
+vcpkg_install_cmake()
```
NancyLi1013, line 33: The same as above, use `vcpkg_cmake_install()` here.

ports/aom/portfile.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+vcpkg_copy_pdbs()
+
+# Move cmake configs
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/aom)
```
NancyLi1013, line 38: The same as above, use `vcpkg_cmake_config_fixup()`.

ports/aom/portfile.cmake
happy-fox-shark: Sure, I'll update it.

NancyLi1013 (changes requested):
Please also add dependency list if we use the suggested functions.

```
"dependencies": [
    {
      "name": "vcpkg-cmake",
      "host": true
    },
    {
      "name": "vcpkg-cmake-config",
      "host": true
    }
]
```

ports/aom/aom-rename-static.diff
```diff
@@ -0,0 +1,12 @@
+diff -pruN aom-3.0.0.o/CMakeLists.txt aom-3.0.0/CMakeLists.txt
+--- aom-3.0.0.o/CMakeLists.txt	2021-04-15 20:05:52.695181200 +0300
++++ aom-3.0.0/CMakeLists.txt	2021-04-15 22:34:16.147522600 +0300
+@@ -249,7 +249,7 @@ endif()
+ add_library(aom ${AOM_SOURCES} $<TARGET_OBJECTS:aom_rtcd>)
+ if(BUILD_SHARED_LIBS)
+   add_library(aom_static STATIC ${AOM_SOURCES} $<TARGET_OBJECTS:aom_rtcd>)
+-  set_target_properties(aom_static PROPERTIES OUTPUT_NAME aom)
++  set_target_properties(aom_static PROPERTIES OUTPUT_NAME aom_static)
+
```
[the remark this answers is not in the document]
happy-fox-shark: >If we are building shared libs, then we should probably not even bother with building the static one

Totally agree. But this is upstream who builds a separate static lib when `BUILD_SHARED_LIBS=ON`. I decided not to remove this behavior.

Having a separate static lib `aom_static` is not a big deal for us as it 's simply not installed by the `install` rules added above. The upstream build system builds it, and our rules don't install it.

ports/aom/aom-install.diff
```diff
@@ -0,0 +1,72 @@
...
++                              NO_SET_AND_CHECK_MACRO
++                              NO_CHECK_REQUIRED_COMPONENTS_MACRO)
++
++write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/aomConfigVersion.cmake
++                                 VERSION ${SO_FILE_VERSION}
++                                 COMPATIBILITY SameMajorVersion)
++
++install(FILES ${CMAKE_CURRENT_BINARY_DIR}/aomConfig.cmake
++              ${CMAKE_CURRENT_BINARY_DIR}/aomConfigVersion.cmake
++        DESTINATION lib/cmake/aom)
```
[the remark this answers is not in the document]
happy-fox-shark: I totally agree. I filed a feature request https://bugs.chromium.org/p/aomedia/issues/detail?id=3029 . Anyways, it will not be available in the current stable release. I would vote for temporarily accepting this patch and remove it when AOM has native installation/configs integration in upstream.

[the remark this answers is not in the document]
happy-fox-shark: Upstream quickly responded and would probably prioritize the feature request. Let's see how it goes on their side. Meanwhile, I'll convert the pr into draft. If they add cmake configs quickly, I'll update it with the new version.

happy-sea-fox:
@Hoikas It looks like upstream delayed and de-prioritized the task to add CMake installation rules. Can we add CMake installation rules in the port as I did earlier with a patch? If it's not recommended, then I'm afraid we can close this PR.

Hoikas:
That sounds good to me, I just think the namespace needs to be changed to `unofficial` to match the rest of the targets added in vcpkg. :smile:

PR #17338, opened 2021-04-18
[aws-sdk-cpp] [openssl] Fix iOS build

ahmedyarub:
Don't pass the the compiler to OpenSSL when building for iOS. That would end generating a MacOS library which would cause linking errors when the libraries are used. The OpenSSL configuration scripts know how to detect the target OS and compile using the correct command, example: `xcrun -sdk iphonesimulator cc`. Note that this change should fix building for all iOS targets, including both simulators and actual devices.
Add try_compile() parameters required to test the features of curl when cross-compiling AWS SDK for C++.

- #### What does your PR fix?  
  Fixes building OpenSSL and AWS SDK for C++ for iOS.

  All iOS triplets

  Yes

  Yes


versions/a-/aws-sdk-cpp.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "2795f006adfb732d610a140ef9eb036f053a5997",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "9c60baa2cff0bbfc024a8241b96f6685f4e083d3",
```

versions/o-/openssl.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "2797d49ed1aa47fd6f37965c3db7dc48ca64d1f5",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "3388a88872808e9926c9d94a702bca7e9f2736f9",
```

JackBoosY (changes requested):

JackBoosY:
I think you should wait for my PR https://github.com/microsoft/vcpkg/pull/17341 merge.

JackBoosY (approved):

versions/a-/aws-sdk-cpp.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "422c3181effad9596e2a70905d357e39ec33d5cc",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "c23d66191b235f6f388187395b62f38b8e6e35f5",
```

versions/o-/openssl.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "f64c492c931c3cb248575a39ae92645392564ad3",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "01fede48f331d05f2feb591061d9e5c1bc076fb0",
```

ports/openssl/unix/CMakeLists.txt
```diff
@@ -143,37 +146,50 @@ else()
...
-        ${PLATFORM}
-        "--prefix=${CMAKE_INSTALL_PREFIX}"
-        "--openssldir=/etc/ssl"
-        ${CFLAGS}
-    COMMAND "${CMAKE_COMMAND}" "-DDIR=${BUILDDIR}" -P "${CMAKE_CURRENT_LIST_DIR}/remove-deps.cmake"
-    VERBATIM
-    WORKING_DIRECTORY "${BUILDDIR}"
-)
+if(IOS)
+    add_custom_command(
```
ras0219-msft, line 150: This call should be combined with the main `add_custom_command()` with only the needed parts changed.

For example, `DISABLES` could be set in an if/else chain.

ras0219-msft (changes requested):
Thanks for the PR!

Generally, this smells like we simply aren't correctly propagating flags to the build -- we're missing some flags from the big list on lines 50-77 or so.

This file comes from a time before we had https://github.com/microsoft/vcpkg/blob/master/scripts/cmake/vcpkg_internal_get_cmake_vars.cmake, which is the new way of approaching this problem. See https://github.com/microsoft/vcpkg/blob/master/scripts/get_cmake_vars/CMakeLists.txt for the "modern" version.

I'm not opposed to this stop-gap for now as long as it's better integrated with the existing behavior, but the long term goal will be to move to `vcpkg_internal_get_cmake_vars()`.

ahmedyarub:
> Thanks for the PR!
> 
> Generally, this smells like we simply aren't correctly propagating flags to the build -- we're missing some flags from the big list on lines 50-77 or so.
> 
> This file comes from a time before we had https://github.com/microsoft/vcpkg/blob/master/scripts/cmake/vcpkg_internal_get_cmake_vars.cmake, which is the new way of approaching this problem. See https://github.com/microsoft/vcpkg/blob/master/scripts/get_cmake_vars/CMakeLists.txt for the "modern" version.
> 
> I'm not opposed to this stop-gap for now as long as it's better integrated with the existing behavior, but the long term goal will be to move to `vcpkg_internal_get_cmake_vars()`.

Appreciate the feedback @ras0219-msft . Please give me a couple of days to redo this PR and the Emscripten/WASM32. I'll make the suggested modernization and test on Windows, Ubuntu 20.04 (WSL), MacOS Big Sur, latest Android NDK, iOS and Emscripten SDK on WSL.

ahmedyarub:
> Thanks for the PR!
> 
> Generally, this smells like we simply aren't correctly propagating flags to the build -- we're missing some flags from the big list on lines 50-77 or so.
> 
> This file comes from a time before we had https://github.com/microsoft/vcpkg/blob/master/scripts/cmake/vcpkg_internal_get_cmake_vars.cmake, which is the new way of approaching this problem. See https://github.com/microsoft/vcpkg/blob/master/scripts/get_cmake_vars/CMakeLists.txt for the "modern" version.
> 
> I'm not opposed to this stop-gap for now as long as it's better integrated with the existing behavior, but the long term goal will be to move to `vcpkg_internal_get_cmake_vars()`.

I thought that this method was not supposed to be used by us port devs, right? I tried using vcpkg_configure_make() but it is not compatible at all with OpenSSL and I had to close the PR.

JackBoosY:
Depends on https://github.com/microsoft/vcpkg/pull/17983.

JackBoosY (approved):

strega-nil-ms:
@ahmedyarub could you rebase onto latest?

ahmedalnuaimi:
> @ahmedyarub could you rebase onto latest?

Done.

ahmedalnuaimi:
Please hold on merging I need to make more tests since the rebase...

ahmedyarub:
> @ahmedyarub could you rebase onto latest?

Please hold on merging I need to make more tests since the rebase...

ahmedalnuaimi:
OK tested both on iOS and Emscripten on Ubuntu. Fixed a couple of bugs and not it should be good to go.

ahmedyarub:
OK tested both on iOS and Emscripten on Ubuntu. Fixed a couple of bugs and not it should be good to go.

strega-nil-ms (approved):

strega-nil-ms:
Thanks @ahmedyarub :)

PR #17363, opened 2021-04-19
Add arm64-osx-dynamic triplet

ghost:
**Describe the pull request**

- #### What does your PR fix?  
Adds a new triplet for dynamic libraries for arm64 on osx. I copied the existing `arm64-osx` triplet and simply changed `VCPKG_LIBRARY_LINKAGE` to `dynamic`. I tested with `./vcpkg install sdl2:arm64-osx-dynamic` and it successfully built a `.dylib` instead of a `.a`.

N/A, this is just adding a new community triplet.

Yes

N/A, not a change to any ports


JackBoosY (approved):

andrewchoi5 (approved):

BillyONeal:
I'm marking this requires: discussion because to my understanding this could be bad for the same reason we don't provide dynamic linux triplets, even under community; because we have no means of arranging for the loader to ever find the right libs. (getting into community means we have to not know it won't work)

I do observe there already is an x64-osx-dynamic. Either:

1. osx is different than linux enough in how it finds dynamic libs that both triplets are fine, or
2. osx is exactly like linux and both this port should be rejected and x64-osx-dynamic should be deleted.

Related: https://github.com/microsoft/vcpkg/issues/15006 https://github.com/microsoft/vcpkg/issues/16332

nickgravelyn:
Can't say I have deep knowledge in dynamic loading nor am I heavily invested either way for vcpkg; I only submitted this PR while testing vcpkg for a project and noticing there was no dynamic arm64 triplet.

For my purposes, the games I build on macOS use dylibs for things like SDL2 which use `@rpath` for their ID. When I build my app against those libs, I use Xcode's feature to embed and sign the dylibs so that my app bundle includes the dynamic libraries in the correct location in the bundle, with my executable configuring the `rpath` such that the dylibs load correctly. 

I personally don't ever build and distribute non-bundled apps on macOS so that distribution channel might have more issues here, but I believe `rpath` can still be used there.

BillyONeal:
I see, so the bundling process fixes up rpath to be all relative within the bundle correctly? That is something meaningfully different than linux which argues to merge this.

BillyONeal:
After a conversation with @strega-nil we're going to merge this.

Thanks for your contribution!

PR #17367, opened 2021-04-19
[spirv-cross] small cleanup

fran6co:
This is part of #12199

PR #17361, opened 2021-04-19
[robin-map] add target

fran6co:
This is part of #12199

add a target for robin-map

PR #17369, opened 2021-04-19
[filament] new port

fran6co:
This is part of #12199

Depends on:

https://github.com/microsoft/vcpkg/pull/17367
https://github.com/microsoft/vcpkg/pull/17361
https://github.com/microsoft/vcpkg/pull/17360


fran6co:
It only builds with clang in linux

cenit:
thank you for resurrecting my old PR (it was huge, i agree, and that it was also its very weak point). i never had time to split it into smaller more manageable PRs

in the end filament and open3d were buildable on mac, win and linux (clang). Is it working for you now only with linux(clang)?

fran6co:
@cenit I'm trying to get it to work on linux, I didn't manage to get open3d compiling yet but I'll be pushing for more PRs as I move forward with it. It's very useful having your initial PR as a base

cenit:
i'm glad for that. I was hoping on someone like you to resurrect it :)

if you need any help, please let me know

ports/filament/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/spirv-tools")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/stb")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/tinyexr")
+
+if(VCPKG_TARGET_IS_WINDOWS)
+    vcpkg_add_to_path("${CURRENT_INSTALLED_DIR}/debug/bin")  #so that tools that are self-produced during build process, which might depend on vcpkg-provided shared libraries, do not silently fail due to missing .dlls
+    vcpkg_add_to_path("${CURRENT_INSTALLED_DIR}/bin")
+endif()
+
+vcpkg_find_acquire_program(CLANG)
```
NancyLi1013, line 38: Does this require on all platforms?

fran6co: no, only linux... I have been testing this only on linux so far so I forgot to make it conditional. I'll change it

NancyLi1013 (changes requested):

ports/filament/portfile.cmake
NancyLi1013: If no `CLANG `on Linux, can it build fine?

ports/filament/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/libassimp")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/libgtest")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/libpng")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/libsdl2")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/libz")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/robin-map")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/spirv-cross")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/spirv-tools")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/stb")
+file(REMOVE_RECURSE "${SOURCE_PATH}/third_party/tinyexr")
```
NancyLi1013, line 31: Can you make a patch to comment these libs out?

ports/filament/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+
+if(VCPKG_TARGET_IS_WINDOWS)
+    vcpkg_add_to_path("${CURRENT_INSTALLED_DIR}/debug/bin")  #so that tools that are self-produced during build process, which might depend on vcpkg-provided shared libraries, do not silently fail due to missing .dlls
+    vcpkg_add_to_path("${CURRENT_INSTALLED_DIR}/bin")
+endif()
+
+vcpkg_find_acquire_program(CLANG)
+
+vcpkg_configure_cmake(
```
NancyLi1013, line 40: This function has been deprecated now. Can you update it as the new one? We prefer to use new functions for new ports.

ports/filament/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+        -DFILAMENT_ENABLE_JAVA=OFF
+        -DFILAMENT_USE_EXTERNAL_GLES3=OFF
+        -DFILAMENT_USE_SWIFTSHADER=OFF
+        -DFILAMENT_GENERATE_JS_DOCS=OFF
+        -DFILAMENT_ENABLE_LTO=OFF
+        -DFILAMENT_SKIP_SAMPLES=OFF
+        -DCMAKE_DEBUG_POSTFIX=d
+)
+
+vcpkg_install_cmake()
```
NancyLi1013, line 55: The same as above. Please do not use a deprecated function in a new port.

ports/filament/vcpkg.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "filament",
+  "version-string": "1.9.20",
```
NancyLi1013, line 3:
```suggestion
  "version-semver": "1.9.20",
```
This field should be more proper. Please see https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.reference.md

ports/filament/vcpkg.json
```diff
@@ -0,0 +1,22 @@
...
+    "gtest",
+    "imgui",
+    "libpng",
+    "robin-map",
+    "sdl2",
+    "spirv-cross",
+    "spirv-tools",
+    "stb",
+    "tinyexr",
+    "zlib"
```
NancyLi1013, line 20: We should add vcpkg-cmake to the dependency list like this if we use new functions:
```
    {
      "name": "vcpkg-cmake",
      "host": true
    }
```

NancyLi1013 (changes requested):

ports/filament/portfile.cmake
fran6co: No, filament doesn't build on Linux without clang -> https://github.com/google/filament/issues/2580

ports/filament/portfile.cmake
```diff
@@ -0,0 +1,78 @@
...
+    OPTIONS
+        -DCMAKE_CXX_COMPILER=${CLANG}++ -DCMAKE_C_COMPILER=${CLANG}
+        -DUSE_STATIC_CRT=${USE_STATIC_CRT}
+        -DFILAMENT_ENABLE_JAVA=OFF
+        -DFILAMENT_USE_EXTERNAL_GLES3=OFF
+        -DFILAMENT_USE_SWIFTSHADER=OFF
+        -DFILAMENT_GENERATE_JS_DOCS=OFF
+        -DFILAMENT_ENABLE_LTO=OFF
+        -DFILAMENT_SKIP_SAMPLES=OFF
+        -DCMAKE_DEBUG_POSTFIX=d
```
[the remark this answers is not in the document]
cenit: i wrote also the Findfilament.cmake module. And in order to have a working cmake module (note, not a config file, because the filament cmake does not export targets!), you need to distinguish debug from release builds, and that's the reason for the 'd' suffix

ports/filament/portfile.cmake
cenit: ? isn't it a standard procedure to remove external 3rd party libs from repo, so that an incomplete patch fails and the port is totally unable to use its own 3rd party libs instead of vcpkg-provided ones?

NancyLi1013: So some of 3rd party libs from repo are used, some are not, right? If so, the current way seems reasonable.

cenit: I don't remember exactly, maybe you're right. Of course the idea would be to try to use all libraries externally, adding to vcpkg those that are missing. In the meantime it might be a good approach. I still remember that for that PR I had to completely rewrite assimp port, because in vcpkg was using *all* internal 3rd party libraries even if all of them were available also in vcpkg, creating a mess for downstream projects!

cenit: (those fixes related to assimp were already cherry-picked by @ras0219-msft and are already merged in master)

NancyLi1013:
Can you resolve the conflicts in this PR @fran6co?

NancyLi1013:
The failures will be fixed by PR #19219.
```
CMake Error at D:/installed/x64-windows/share/SPIRV-Tools/SPIRV-ToolsTarget.cmake:81 (message):
  The imported target "SPIRV-Tools-shared" references the file

     "D:/installed/x64-windows/debug/bin/SPIRV-Tools-shared.dll"

     "D:/installed/x64-windows/share/SPIRV-Tools/SPIRV-ToolsTarget.cmake"

  D:/installed/x64-windows/share/SPIRV-Tools/SPIRV-ToolsConfig.cmake:1 (include)
  C:/a/1/s/scripts/buildsystems/vcpkg.cmake:782 (_find_package)
  CMakeLists.txt:59 (find_package)
```

JonLiu1993:
@fran6co ,Is this PR being worked on?

fran6co:
> @fran6co ,Is this PR being worked on?

no, I don't have the time to work on this

JonLiu1993:
Temporarily close this issue, please feel free reopen it if you have time to deal with it

PR #17389, opened 2021-04-20
[libxml2] Pass version to rc via configured file

dg0yt:
- #### What does your PR fix?  
  This change fixes a build failure on mingw/Windows from issues with quoting/escaping the dotted version string when passing it to the resource compiler.

  all, no

  yes

  yes


JonLiu1993 (approved):

ras0219-msft:
LGTM, thanks for the fix!

JackBoosY:
Does the upstream know this bug or fixed it?

dg0yt:
There is no upstream bug. The `CMakeLists.txt` comes from vcpkg. That's why the include was missing.

dg0yt:
FTR, upstream creates `rcVersion.h` here: 
https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/win32/configure.js#L283-294

PR #17404, opened 2021-04-21
[forge] Fix usage

### LINKED ISSUE #17396: [forge:x64-windows] Compiles ok, but Forge_FOUND==false after find_package
vcpkg version  2021-01-13-d67989bce1043b98092ac45996a8230a059a2d7e
forge version 1.0.5
OS Windows 10

- `.\vcpkg.exe install forge:x64-windows` successful
- find_package(Forge CONFIG REQUIRED),    target_link_libraries(main PRIVATE Forge::forge) provided
- VisualStudio->Create New CMake project
- Insert `find_package(Forge CONFIG REQUIRED)` in generated CMakeLists.txt
- Error: 
[CMake] CMake Error at C:/Dev/vcpkg/scripts/buildsystems/vcpkg.cmake:861 (_find_package):
1> [CMake]   Found package configuration file:
1> [CMake] 
1> [CMake]     C:/Dev/vcpkg/installed/x64-windows/share/forge/ForgeConfig.cmake
1> [CMake] 
1> [CMake]   but it set Forge_FOUND to FALSE so package "Forge" is considered to be NOT
1> [CMake]   FOUND.
1> [CMake] Call Stack (most recent call first):
1> [CMake]   forgetest/CMakeLists.txt:6 (find_package)
1> [CMake] -- Configuring incomplete, errors occurred!

[CMakeOutput.log](https://github.com/microsoft/vcpkg/files/6345678/CMakeOutput.log)

JackBoosY:
Fix ForgeConfig.cmake, suitable for vcpkg.

Fixes #17396.

### DIFF
--- a/ports/forge/fix-usage.patch
+++ b/ports/forge/fix-usage.patch
@@ -1,0 +1,29 @@
+diff --git a/CMakeModules/ForgeConfig.cmake.in b/CMakeModules/ForgeConfig.cmake.in
+index e4c7737..6429069 100644
+--- a/CMakeModules/ForgeConfig.cmake.in
++++ b/CMakeModules/ForgeConfig.cmake.in
+@@ -52,16 +52,18 @@ set(Forge_FOUND OFF)
+ 
+ if (NOT TARGET Forge::forge AND
+         NOT TARGET forge AND
+-        EXISTS @PACKAGE_CMAKE_DIR@/ForgeTargets.cmake)
+-    include(@PACKAGE_CMAKE_DIR@/ForgeTargets.cmake)
++        EXISTS ${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
++    include(${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
+ endif ()
+ 
+ if (TARGET Forge::forge)
+-    get_property(config TARGET Forge::forge PROPERTY IMPORTED_CONFIGURATIONS)
+-    if(NOT config)
+-      set(config "NOCONFIG")
++    get_property(configs TARGET Forge::forge PROPERTY IMPORTED_CONFIGURATIONS)
++    if(NOT configs)
++      set(configs "NOCONFIG")
+     endif()
+-    get_property(loc TARGET Forge::forge PROPERTY IMPORTED_LOCATION_${config})
++    foreach(config ${configs})
++        get_property(loc TARGET Forge::forge PROPERTY IMPORTED_LOCATION_${config})
++    endforeach()
+ endif ()
+ 
+ if ((TARGET Forge::forge AND EXISTS ${loc})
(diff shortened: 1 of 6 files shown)


ports/forge/fix-usage.patch
```diff
@@ -0,0 +1,29 @@
...
+-        EXISTS @PACKAGE_CMAKE_DIR@/ForgeTargets.cmake)
+-    include(@PACKAGE_CMAKE_DIR@/ForgeTargets.cmake)
++        EXISTS ${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
++    include(${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
+ endif ()
+ 
+ if (TARGET Forge::forge)
+-    get_property(config TARGET Forge::forge PROPERTY IMPORTED_CONFIGURATIONS)
+-    if(NOT config)
+-      set(config "NOCONFIG")
```
JackBoosY, line 18: Here, since vcpkg moved `debug` and `release` configure file to the same directory, `config` will contain two fields `DEBUG` / `RELEASE`. So we should treat it as a array, not one value.

PhoebeHui (approved):

ports/forge/fix-usage.patch
JackBoosY: Full code in ForgeConfig.cmake:
```cmake
set(Forge_FOUND OFF)
if (NOT TARGET Forge::forge AND
        NOT TARGET forge AND
        EXISTS ${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
    include(${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
endif ()

if (TARGET Forge::forge)
    get_property(configs TARGET Forge::forge PROPERTY IMPORTED_CONFIGURATIONS)
    message("configs: ${configs}")
    if(NOT configs)
      set(configs "NOCONFIG")
    endif()
    foreach(config ${configs})
        get_property(loc TARGET Forge::forge PROPERTY IMPORTED_LOCATION_${config})
    endforeach()
endif ()

if ((TARGET Forge::forge AND EXISTS ${loc})
    OR
    TARGET forge)
    set(Forge_FOUND ON)
    if (TARGET forge AND NOT TARGET Forge::forge)
        add_library(Forge::forge ALIAS forge)
    endif ()
    set(Forge_LIBRARIES Forge::forge)
else ()
    set(Forge_FOUND OFF)
endif ()
```

PhoebeHui: I think we can remove 'AND EXISTS ${loc}' from 'if ((TARGET Forge::forge AND EXISTS ${loc})', I'm don't think it's necessary to detect it.

ras0219-msft: Why can we not replace this entire file with:
```
include(${CMAKE_CURRENT_LIST_DIR}/ForgeTargets.cmake)
set(Forge_LIBRARIES Forge::forge)
set(Forge_FOUND ON)
```
?

JackBoosY: @ras0219-msft I think we should try not to change the upstream source code, what do you think about this?

ras0219-msft: Yeah, you're right. This is better since it's still a reasonably short patchfile.

PR #17419, opened 2021-04-21
[Libidn2] disable documentation

xandox:
Documentation generates by gtk tool which not must exists. For PR #15605 

same,  No

 yes

  Yes


dg0yt:
Same as for https://github.com/microsoft/vcpkg/pull/17420#issuecomment-824200036

xandox:
`gt-html` is on by default according to `configure --help`

ports/libidn2/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "libidn2",
   "version-string": "2.3.0",
```
NancyLi1013, line 3:
```suggestion
  "version": "2.3.0",
```
We prefer to use more proper field here. Please see https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md.

NancyLi1013 (changes requested):

dg0yt:
osx: #17057 reproduced.

dg0yt:
The autoreconf patch will probaly fix the build, but it is tightly coupled to the current source files.

In the meantime, I could verify that my proposal from https://github.com/microsoft/vcpkg/pull/17420#issuecomment-824200036 works. So alternative to patching, it can be solved just by configuration if you take our pieces together:
~~~
set(ENV{GTKDOCIZE} true)
...
   --disable-gtk-doc...
~~~

xandox:
I do not known what is wrong with `getdns`. It's perfectly builds on my host.

xandox:
@dg0yt thank you. Your suggestion better.

dg0yt:
>  do not known what is wrong with `getdns`.

I don't think it is related to the doc changes here. This PR just triggers the rebuild.

> @dg0yt thank you. Your suggestion better.

Thanks for integrating my suggestion.

ras0219-msft (approved):
LGTM, thanks for the improvement!

NancyLi1013 (approved):

PR #17435, opened 2021-04-22
[catch2] updated catch2 to version 2.13.6

georg-emg:
I updated the version in the catch2 port to 2.13.6. I also replaced the old CONTROL file with a proper manifest.


ports/catch2/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "catch2",
+  "version-string": "2.13.6",
```
NancyLi1013, line 3: This filed might be `version-semver` here. Please refer to https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md.

versions/c-/catch2.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "0bbe381f5f7a5df720a1bdaca43da2ec942004c1",
+      "version-string": "2.13.6",
```
NancyLi1013, line 5: The same as above.

NancyLi1013 (changes requested):
It seems that you forgot to update portfile.cmake file.

georg-emg:
> It seems that you forgot to update portfile.cmake file.

Only took me two tries to fix! (I hope)

ports/catch2/vcpkg.json
NancyLi1013: Seems you missed this point.

NancyLi1013 (changes requested):

NancyLi1013 (approved):

PR #17466, opened 2021-04-23
[jsoncons] Update to version 0.163.3

### LINKED ISSUE #317: jmespath issue with reusing compiled expressions
First to all thank you providing an incredible library! 

We use `JMESPath` implementation quite extensively but recently we had a big higher throughput requirements which forced use to reuse compiled expression. The usual workaround for this bug (which has been around for a while) was to compile expression every time before usage. `jsoncons::jmespath::search` takes around 3x longer than `evaluate` on expression

**Describe the bug**

I did a bit more digging, please have a look at the snippet 

```c++

#include <jsoncons/json.hpp>
#include <jsoncons_ext/jmespath/jmespath.hpp>

int main(int argc, char *argv[]) {
  using namespace jsoncons; // for convenience

  std::string jtext = R"(
            {
              "people": [
                {
                  "age": 20,
                  "other": "foo",
                  "name": "Bob"
                },
                {
                  "age": 25,
                  "other": "bar",
                  "name": "Fred"
                },
                {
                  "age": 30,
                  "other": "baz",
                  "name": "George"
                }
              ]
            }
        )";

  auto expr = jmespath::jmespath_expression<json>::compile("sum(people[].age)");

  json doc = json::parse(jtext);

  json result = expr.evaluate(doc);
  std::cout << pretty_print(result) << "\n\n";
}
```

It will result in exception `Function called with wrong number of arguments` or just outright `SIGSEGV`
but this problem will never occur when using `jsoncons::jmespath::search`. This seem not to happen with every expression. We seem to have issues with functions like `max` or `sum` i.e `max([].my_object.sub_field)`

**But now** if we re-arrange to first `::parse` and then `::compile` everything is OK and I get result `75.0`

```c++
json doc = json::parse(jtext);
auto expr = jmespath::jmespath_expression<json>::compile("sum(people[].age)");
```

It looks to me like `::parse` breaks a global state that `compile` introduces

**What compiler, architecture, and operating system?**

- Compiler: gcc 9.2 / gcc 10.2
- Architecture (e.g. x86, x64) x64
- Operating system: linux

**What jsoncons library version?**
 release v0.162.2, however this problem has been reoccurring for all previous versions

danielaparker:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes danielaparker/jsoncons#317

  All triplets are supported. Baseline as been updated.

  Yes

  Yes


ras0219-msft:
Thanks for the great library!

Issue #17427, opened 2021-04-21
How do I use the yara port?


VelocityRa:
Normally when installing a port, doing simply:
```cmake
find_package(myport CONFIG REQUIRED)
target_link_libraries(myproj PRIVATE myport)
```
works.
But the [yara port](https://github.com/microsoft/vcpkg/tree/master/ports/yara) does not seem to install a config file, so the `find_package` fails. I also tried just the `target_link_libraries` with both `yara` and `libyara` (the installed `.lib` is called this) to no avail.

Whatever the method is, it would be useful for instruction to be printed when you install the port, like some ports do. Especially since it seems that you need a non-'standard' way to install it (or I am missing something glaringly obvious).

NancyLi1013:
Hi @VelocityRa 

Thanks for posting this issue.

As you said above, there are no config cmake file and FindYara.cmake for `yara`. So you cannot use `find_package()` to find `yara`.

For most ports that don't provide these files(xxxConfig.cmake or Findxxx.cmake), we use the ports like this:

```
find_library()
find_path()
```

VelocityRa:
Thanks @NancyLi1013!

For anyone coming across this, I ended up using:
```cmake
# Dependency: jansson (needed by yara)
find_package(jansson CONFIG REQUIRED)
target_link_libraries(myproj PRIVATE jansson::jansson)

# Dependency: yara
find_library(YARA_LIBRARY libyara)
target_link_libraries(myproj PRIVATE ${YARA_LIBRARY})
```

I tried to commit these as a helpful message when installing the yara port to save people some time but I actually can't find example code of existing ports that do this 🤔 I'm pretty sure I remember some doing it, was it removed?

Anyway, feel free to close this.

NancyLi1013:
Thanks for your help to do this work.
Do you mean the usage of ports? You want to add the usage for yara, but you cannot find example code of others ports that how to add usage. Please help me confirm if i understand this correctly.

VelocityRa:
Correct, I wanted to make vcpkg print usage instructions for yara, specifically the code i posted above (or similar).

NancyLi1013:
Maybe you can search `usage ` in _ports/_ folder, you will find many usage files listed in `ports/PortName`.

You can refer some of them. Such as https://github.com/microsoft/vcpkg/blob/master/ports/libqrencode/usage
https://github.com/microsoft/vcpkg/blob/master/ports/discord-game-sdk/usage
[...]

PR #17473, opened 2021-04-23
[yara] Usage instructions

VelocityRa:
Related: #17427


VelocityRa:
```cmake
CMake Error at ports/yara/portfile.cmake:27 (file):
  file COPY cannot find "C:/a/1/s/ports/yara/usage": File exists.
```
Huh?

Hoikas:
Rename `usage.txt` to `usage`.

VelocityRa:
Oh yea my editor added that.

JackBoosY:
No, I think the correct way is export the unofficial target instead of this.
The way which you provided is only find the release library.

VelocityRa:
@JackBoosY Closed it because I don't know how to do that. I can work on it if you point me to the right cmake commands to use.

VelocityRa:
If it's so hard to actually use, I don't really understand why the port was merged in the first place in such a state.

VelocityRa:
Current PR is also wrong because it depends on openssl too not just openssl. But I assume doing it the 'correct' way you said would fix this.

JackBoosY:
@VelocityRa I provide you 2 ways:
1. export the unofficial target, example: https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/libmariadb/export-cmake-targets.patch
2. Configure the `unofficial-yara-config.cmake.in`, example: https://github.com/microsoft/vcpkg/blob/master/ports/pcre/unofficial-pcre-config.cmake

I perfer the first one.

PR #17526, opened 2021-04-27
[vcpkg] Add triplet option VCPKG_NO_COMPILER_TRACKING

### LINKED ISSUE #16631: [vcpkg] Possibility to control the compiler hash for binary caching

klalumiere:
**Describe the pull request**

- #### What does your PR fix?
  Fixes #16631

Implements @ras0219-msft suggestion in the pull request #16632

>Alternatively, we could expose (effectively) --feature-flags=-compilertracking as a triplet setting, which would omit the detect compiler step entirely and rely on the hash of the triplet file + toolchain to distinguish package ABI. In this case, the user could add a comment to their triplet file containing any information they want to distinguish (such as compiler version). I think this is a more promising approach because it has the least interface churn.

~The doc is not written yet: I'll wait for the feedback before writing it.~ :slightly_smiling_face: 
The doc is written. :slightly_smiling_face: 

  all, No

  Yes


klalumiere:
:thinking: I can't think of a way if that the CI failures are related to the changes since everything is inside a `if` that depends on a variable `VCPKG_NO_COMPILER_TRACKING` that wasn't use yet.

JackBoosY:
@klalumiere  I think it's not related to this PR.

JackBoosY:
Depends on https://github.com/microsoft/vcpkg/pull/17528.

JackBoosY (approved):

strega-nil-ms:
This needs to be a change in the vcpkg tool; we need for `VCPKG_NO_COMPILER_TRACKING` to result in `detect_compiler` not even being called.

klalumiere:
Hey Nicole!

Thanks for the feedback!

> This needs to be a change in the vcpkg tool; we need for VCPKG_NO_COMPILER_TRACKING to result in detect_compiler not even being called.

This was my first thought actually. I agree that with the current implementation, some people could have problems with the compiler detection when `VCPKG_NO_COMPILER_TRACKING` is set to true, which would be both confusing and frustrating.

My initial idea was to add something like `&& ! triplet.has_no_compiler_tracking()` in [vcpkg-tool/src/vcpkg/build.cpp:526](https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/build.cpp#L526). Or maybe initialize `m_compiler_tracking` by considering the hypothetical `triplet.has_no_compiler_tracking()` directly. The problem I encountered is that I was unable to found anywhere where we read or parse the triplet file *content*. In particular, the [triplet API](https://github.com/microsoft/vcpkg-tool/blob/main/include/vcpkg/triplet.h) seems to only deal with triplets name.

So my questions:

If there is any such C++ API in `vcpkg-tool` that allows to read/parse the triplet content, could you point me to it please? Otherwise, I guess we would need to also add such an API as a prerequisite to the feature `VCPKG_NO_COMPILER_TRACKING`, right? Is there any other way that I don't see? :thinking:

Thanks!

JackBoosY:
Ping @strega-nil @strega-nil-ms for response.

autoantwort:
Maybe you can use `CMakeVarProvider::load_generic_triplet_vars(Triplet triplet)` and `CMakeVarProvider::get_generic_triplet_vars(Triplet triplet)`

strega-nil:
ah, shoot, sorry I missed this; if I don't see something, please ping me on discord. I have way too many things on my plate haha!

I think @ras0219 would understand better... poking him

ras0219-msft:
https://github.com/microsoft/vcpkg/blob/master/scripts/vcpkg_get_tags.cmake is the file that gets run to extract values from the triplet into C++.

Specifically, you'll need to change the "generic" phase, which is used to determine the common behaviors across all ports: https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/cmakevars.cpp#L224. You'll probably want to change https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/build.cpp#L548 to consider this new flag, which should be smuggled through the `abi_info.pre_build_info`

klalumiere:
**autoantwort**: Exactly what I was searching for, thanks! :slightly_smiling_face: 

@strega-nil : No problems, I had switched focus to other projects :slightly_smiling_face:. I considered at some point trying to poke somebody else and/or to go talk on somewhere interactive like the C++ Slack. About Discord, which server are you talking about? Is there a public vcpkg server?

ras0219-msft: Great, thanks! I'll try to work on this soon. I'll leave the current pull request open for the documentation.

autoantwort:
There is a vcpkg channel on [this](https://discord.com/invite/WjvWcZTTuN) discord server which is quite active.

klalumiere:
All right! I think that the two pull requests (this one and microsoft/vcpkg-tool#110) are ready for review! :smile:

JackBoosY:
Ping @ras0219 @ras0219-msft for review again.

strega-nil-ms:
Waiting for microsoft/vcpkg-tool#110 to be merged :)

strega-nil-ms (approved):

PR #17549, opened 2021-04-28
[doc] Update manifest to reference versioning spec

### LINKED ISSUE #17548: Update docs manifest
**Describe the bug**
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/manifest-files.md#version-fields

it only says that version-string exists. however, version, version-string and version-semver exist (to my knowledge). 

It would be good if someone with deeper knowlege would add the new version types.

mathisloge:
Currently only `version-string` is included in the manifest-files.md. However there are more variants of this field. This PR adds a cross link to the versioning specification. 

fixes #17548

### DIFF
--- a/docs/maintainers/manifest-files.md
+++ b/docs/maintainers/manifest-files.md
@@ -79,7 +79,7 @@ Manifest property | Versioning scheme
 `version-string`  | For arbitrary strings
 
 
-See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions, for more details.
+See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions for more details.
 
 Additionally, `"port-version"` is used to differentiate between port changes that don't change the underlying library version.
 
--- a/docs/maintainers/manifest-files.md
+++ b/docs/maintainers/manifest-files.md
@@ -74,6 +74,7 @@ however, more version kinds will be added later. Additionally,
 `"port-version"` is used to differentiate between port changes that don't change the underlying library version.
 
 #### `"version-string"`
+See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions , for further cases (semantic versioning etc.)
 
 This field is an ascii string, and may contain alphanumeric characters, `.`, `_`, or `-`. No attempt at ordering versions is made; all versions are treated as byte strings and are only evaluated for equality.
 
--- a/docs/maintainers/manifest-files.md
+++ b/docs/maintainers/manifest-files.md
@@ -69,12 +69,13 @@ numbers, and hyphens, and it must not begin nor end with a hyphen.
 
 ### Version fields
 
+See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions, for further cases (semantic versioning etc.)
+
 The library version. There is currently only one kind of version, a `"version-string"` -
 however, more version kinds will be added later. Additionally,
 `"port-version"` is used to differentiate between port changes that don't change the underlying library version.
 
 #### `"version-string"`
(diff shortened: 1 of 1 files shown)


docs/maintainers/manifest-files.md
```diff
@@ -74,6 +74,7 @@ however, more version kinds will be added later. Additionally,
 `"port-version"` is used to differentiate between port changes that don't change the underlying library version.
 
 #### `"version-string"`
+See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions , for further cases (semantic versioning etc.)
```
NancyLi1013, line 77:
```suggestion
See https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions, for further cases (semantic versioning etc.)
```
It might be better to add these specifications in `Version fields`. We can also update the description for` Version fields`. Since we have supported other kinds of version now.

NancyLi1013 (changes requested):

mathisloge:
@NancyLi1013 maybe we should remove the whole `#### "version-string"` section and just add the link to the specification? 
I don't think it's really maintainable or useful to have the information duplicated.

NancyLi1013:
> @NancyLi1013 maybe we should remove the whole `#### "version-string"` section and just add the link to the specification?
> I don't think it's really maintainable or useful to have the information duplicated.

I agree with your suggestions. This section seems outdated. But since it is about version naming, we cannot remove it directly. We can rewrite this section. What do you think?

mathisloge:
I was wondering if the changes introduced with https://github.com/microsoft/vcpkg/pull/17549/commits/66619e0d42093e8bf077d324c193f7bbe9411929  are sufficient. Since all details are available in the specification (with examples) and the needed information, which fields exists are given in the table added with the mentioned commit

docs/maintainers/manifest-files.md
```diff
@@ -69,9 +69,19 @@ numbers, and hyphens, and it must not begin nor end with a hyphen.
...
+Currently there are different fields for special versioning. Namely:
+
+Manifest property | Versioning scheme
+------------------|------------------------------------
+`version`         | For dot-separated numeric versions
+`version-semver`  | For SemVer compliant versions
+`version-date`    | For dates in the format YYYY-MM-DD
+`version-string`  | For arbitrary strings
+
+
```
NancyLi1013, line 81:
```suggestion
```

NancyLi1013 (changes requested):

mathisloge:
@NancyLi1013  the "thumbs up" is the indication that we will drop `#### "version-string"` in favor of https://github.com/microsoft/vcpkg/pull/17549/commits/66619e0d42093e8bf077d324c193f7bbe9411929 ? 😄

NancyLi1013:
> @NancyLi1013 the "thumbs up" is the indication that we will drop `#### "version-string"` in favor of [66619e0](https://github.com/microsoft/vcpkg/commit/66619e0d42093e8bf077d324c193f7bbe9411929) ? 😄

Currently, it might be not. I think we should rewrite this section. But I have no idea how to make it more reasonable. We have supported several kinds of version, only listing `version-string` here seems not good. The current changes look good to me. `#### "version-string"` section should be discussed.

Do you have any suggestion?

ras0219-msft (approved):
Further tweaks and consolidation is needed, but this is certainly an incremental improvement.

mathisloge:
@NancyLi1013 From my point of view, the table with the possible fields would be sufficient. Since, in my opinion, all the information we would write here would be duplicated with the specification. 
In my case, as someone who only updates or creates ports, this would be perfectly sufficient for me, since I know that there are other fields than version-string and I know where to look for examples and detailed information (in the linked spec.)

Maybe first merge these changes and discuss further changes in discussions or in a seperate PR?

ras0219-msft:
Thanks for the PR!

PR #17563, opened 2021-04-28
[winsparkle] inital port

### LINKED ISSUE #16607: [New Port Request] winsparkle
Library name:winsparkle

Library description: WinSparkle is a plug-and-forget software update library for Windows applications. It is heavily inspired by the Sparkle framework for OS X written by Andy Matuschak and others, to the point of sharing the same updates format (appcasts) and having very similar user interface.

Source repository URL:https://github.com/vslavik/winsparkle

Project homepage (if different from the source repository):https://winsparkle.org/

daschuer:
This PR adds winsparkle to the ports 

Fixes  #16607

- 
winsparkle:arm-uwp=fail
winsparkle:x64-linux=fail
winsparkle:x64-osx=fail
winsparkle:x64-uwp=fail

Yes

Yes

### DIFF
--- a/ports/winsparkle/portfile.cmake
+++ b/ports/winsparkle/portfile.cmake
@@ -1,5 +1,3 @@
-vcpkg_fail_port_install(MESSAGE "winsparkle only supports Windows platforms" ON_TARGET "osx" "linux" "uwp")
-
 vcpkg_download_distfile(ARCHIVE
     URLS "https://github.com/vslavik/winsparkle/releases/download/v0.7.0/WinSparkle-0.7.0-src.zip"
     FILENAME "winsparkle-070.zip"
--- a/ports/winsparkle/portfile.cmake
+++ b/ports/winsparkle/portfile.cmake
@@ -1,0 +1,44 @@
+vcpkg_fail_port_install(MESSAGE "winsparkle only supports Windows platforms" ON_TARGET "osx" "linux" "uwp")
+
+vcpkg_download_distfile(ARCHIVE
+    URLS "https://github.com/vslavik/winsparkle/releases/download/v0.7.0/WinSparkle-0.7.0-src.zip"
+    FILENAME "winsparkle-070.zip"
+    SHA512 58991c821b31bbc2e7af342af6cfce2fc095841200a2a2359d3fd784d804e1bfddcb640ba8427614732e43bc1608f203e2f03274d8f942e66ad40d5ac2554891
+)
+
+vcpkg_extract_source_archive_ex(
+    OUT_SOURCE_PATH SOURCE_PATH
+    ARCHIVE ${ARCHIVE}
+)
+
+# build and install 
+vcpkg_install_msbuild(
+    SOURCE_PATH "${SOURCE_PATH}"
+    PROJECT_SUBPATH "WinSparkle-2017.sln"
+    PLATFORM ${BUILD_ARCH}
+    TARGET_PLATFORM_VERSION ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}
+    PLATFORM_TOOLSET ${VCPKG_PLATFORM_TOOLSET}
+    OPTIONS /t:restore /p:RestorePackagesConfig=true
+    USE_VCPKG_INTEGRATION
+)
+
+# These libraries are useless, so remove.
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libcharset-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libcharset-1.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libgcc_s_dw2-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libgcc_s_dw2-1.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libgettextlib-0-20-2.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libgettextlib-0-20-2.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libgettextpo-0.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libgettextpo-0.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libgettextsrc-0-20-2.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libgettextsrc-0-20-2.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libgomp-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libgomp-1.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libiconv-2.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libiconv-2.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libintl-8.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libintl-8.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libstdc++-6.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libstdc++-6.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libtextstyle-0.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libtextstyle-0.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libwinpthread-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libwinpthread-1.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libcharset-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libcharset-1.dll)
+
+# Install includes
+file(INSTALL ${SOURCE_PATH}/include/winsparkle.h DESTINATION ${CURRENT_PACKAGES_DIR}/include/winsparkle.h)
+file(INSTALL ${SOURCE_PATH}/include/winsparkle-version.h DESTINATION ${CURRENT_PACKAGES_DIR}/include/winsparkle-version.h)
+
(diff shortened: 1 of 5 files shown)


NancyLi1013:
Hi @daschuer 

Could you please sign CLA first?

ports/winsparkle/portfile.cmake
```diff
@@ -0,0 +1,44 @@
+vcpkg_fail_port_install(MESSAGE "winsparkle only supports Windows platforms" ON_TARGET "osx" "linux" "uwp")
```
NancyLi1013, line 1:
```suggestion
vcpkg_fail_port_install(ON_TARGET "osx" "linux" "uwp")
```

ports/winsparkle/portfile.cmake
```diff
@@ -0,0 +1,44 @@
...
+vcpkg_download_distfile(ARCHIVE
+    URLS "https://github.com/vslavik/winsparkle/releases/download/v0.7.0/WinSparkle-0.7.0-src.zip"
+    FILENAME "winsparkle-070.zip"
+    SHA512 58991c821b31bbc2e7af342af6cfce2fc095841200a2a2359d3fd784d804e1bfddcb640ba8427614732e43bc1608f203e2f03274d8f942e66ad40d5ac2554891
+)
+
+vcpkg_extract_source_archive_ex(
+    OUT_SOURCE_PATH SOURCE_PATH
+    ARCHIVE ${ARCHIVE}
+)
```
NancyLi1013, line 12: Why do you not use `vcpkg_from_github()` to download source codes?

ports/winsparkle/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "winsparkle",
+  "version-string": "0.7.0",
```
NancyLi1013, line 3:
```suggestion
  "version": "0.7.0",
```
version seems more proper here. Please see https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md

scripts/ci.baseline.txt
```diff
@@ -1704,6 +1704,10 @@ winreg:x64-linux=fail
 winreg:x64-osx=fail
 winsock2:x64-linux=fail
 winsock2:x64-osx=fail
+winsparkle:arm-uwp=fail
+winsparkle:x64-linux=fail
+winsparkle:x64-osx=fail
+winsparkle:x64-uwp=fail
```
NancyLi1013, line 1710:
```suggestion
```
We can add `supports: windows`field in vcpkg.json.

ports/winsparkle/portfile.cmake
```diff
@@ -0,0 +1,44 @@
...
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libtextstyle-0.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libtextstyle-0.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libwinpthread-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libwinpthread-1.dll)
+file(REMOVE ${CURRENT_PACKAGES_DIR}/bin/libcharset-1.dll ${CURRENT_PACKAGES_DIR}/debug/bin/libcharset-1.dll)
+
+# Install includes
+file(INSTALL ${SOURCE_PATH}/include/winsparkle.h DESTINATION ${CURRENT_PACKAGES_DIR}/include/winsparkle.h)
+file(INSTALL ${SOURCE_PATH}/include/winsparkle-version.h DESTINATION ${CURRENT_PACKAGES_DIR}/include/winsparkle-version.h)
+
+# Handle copyright
+file(INSTALL ${SOURCE_PATH}/COPYING DESTINATION ${CURRENT_PACKAGES_DIR}/share/winsparkle RENAME copyright)
```
NancyLi1013, line 44:
```suggestion
file(INSTALL ${SOURCE_PATH}/COPYING DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)

```

ports/winsparkle/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "winsparkle",
+  "version-string": "0.7.0",
+  "homepage": "https://winsparkle.org",
+  "description": "WinSparkle is an easy-to-use software update library for Windows developers."
```
NancyLi1013, line 5:
```suggestion
  "description": "WinSparkle is an easy-to-use software update library for Windows developers.",
  "supports: "windows"
```

NancyLi1013 (changes requested):

NancyLi1013:
Please run `vcpkg format-manifest ./ports/winsparkle/vcpkg.json` to format the json file.

ports/winsparkle/portfile.cmake
daschuer: Because the repo contains submodules and the package I have picked here has the submodules included.
I have tried to use  vcpkg_from_github calls the compose the same on the fly, but I failed.

NancyLi1013:
Hi @daschuer 

daschuer:
I consider this as done so far. 
It is missing the evidence that it is working in a productive environment, because we have not yet integrated it into Mixxx. I am afraid this will take us some time because we are currently fighting other issues.

Cheney-W:
@daschuer   Is work still being done for this PR?

daschuer:
The status is unchanged since my last post.
Do you consider to use this port?

Cheney-W:
Thanks for your reply!
No, I don't use this port, I'm just asking about the current status of this PR based on the PR review process.
Also, If you don't have time to do this PR at the moment, I think we can close it and reopen it when you have time to do it. What do you think?

daschuer:
My idea was to leave it open, that it is visible.
Without it, there is a risk that someone else's repeat my work. 
Alternative I can polish it for passing the CI and merge without a using project. If one discovers that it is not working in his environment, we have at least a good start.

Cheney-W:
LGTM.

ports/winsparkle/portfile.cmake
```diff
@@ -0,0 +1,44 @@
+vcpkg_fail_port_install(ON_TARGET "osx" "linux" "uwp")
```
Cheney-W, line 1: This sentence could be remove because the vcpkg.json has include the "supports" item.

Cheney-W:
I don't think the [commit: Debug vcpkg_copy_pdbs](https://github.com/microsoft/vcpkg/pull/17563/commits/8ab0b529b66863979c669d808a6ed2c3b6c6e572) should be added.

strega-nil-ms:
We need to patch out the usage of submodules here; building openssl and everything instead of using openssl from vcpkg is not great. Can you take a look at this @daschuer? You should be able to patch out all the `3rdparty` references, and then pass `USE_VCPKG_INTEGRATION` to `vcpkg_build_msbuild`.

daschuer:
Unfortunately I don't know how to make the changes in the msbuild project. There is also a broken cmake build system with a pending PR https://github.com/vslavik/winsparkle/pull/230
Once this is merged a cmake based solution would be easy for me. 
However I am unsure if it a good idea to do this without upstream support. This will put the maintenance burden on us to maintain and extra build system which is not tested by upstream. 

I have filed an issue to here @vslavik opinion:
https://github.com/vslavik/winsparkle/issues/238

Cheney-W:
@daschuer 
I'll get to it, but it may take some time, in the meantime, can we convert this PR to draft?

daschuer:
Yes sure, thank you.

vslavik:
> However I am unsure if it a good idea to do this without upstream support. This will put the maintenance burden on us to maintain and extra build system which is not tested by upstream.

…as well as additional burden on *upstream* who'd now have to deal with issues caused by the non-standard build…

It is an explicit design goal for WinSparkle to be a single self-contained DLL with no external dependencies (to the point that it even links to static CRT!). This matters for e.g. in-app delta updates or re-launching the app after update (which, OK, aren't implemented yet) when WinSparkle code requires special handling to keep running. As long as that means just a single WinSparkle.dll and nothing else, it is manageable.

As upstream, I'd really appreciate it if something as high-profile as Vcpkg didn't provide non-standard builds.

daschuer:
OK, that means vcpkg should produce always a dynamic linked DLL, even If the whole application is build static and this DLL shall link to the static version of the depended libraries, right? 

I am afraid afraid vcpkg is not prepared for that right now. There is no way to identify the matching dynamic triplet in a static build and wise versa. There is also no general solution in sight, so there is no point of hacking around the upstream build system AND the vcpkg dependency tree.  

For me the private build of dependency is a good solution and the current state is ready to be merged.

dg0yt:
> OK, that means vcpkg should produce always a dynamic linked DLL, even If the whole application is build static and this DLL shall link to the static version of the depended libraries, right?
> 
> I am afraid afraid vcpkg is not prepared for that right now.

IIUC this is done by
~~~
~~~

daschuer:

Does not work, because all other packages will be created with the same triplet. But we want WinSparkles to be dll, statically linked to it's dependencies.

dg0yt:
> Does not work, because all other packages will be created with the same triplet.

Pardon? It will create a dll in a triplet where all dependencies only exist as static libs.

daschuer:
Yes, but only if all other libraries build are static.
The requirement here is to always link the depending libraries statically and use CRT. 

So the issue is not about the port itself, it is the question how to link the dependencies statically. The current solution is to build them together with the port. 

The alternative would be a kind of package dependency that includes a triplet, depending on the selected main triplet and a we need to hack around the upstream build environment ant pull the related maintenance to vcpkg.

strega-nil-ms:
So given the specific case here, where the DLL is completely encapsulated - there are no libraries that both the DLL and the consumer depend on that can step on each other's toes, and the author also told us not to do anything else besides this, contrary to our normal de-vendoring policies - it seems to me that there are no benefits to building this library, rather than downloading the pre-built binary. There aren't any ABI concerns that building the source would solve, in this case.

@daschuer, would you mind making this change? (downloading WinSparkle-0.7.0.zip instead of WinSparkle-0.7.0-src.zip, and then just manually copying files over instead of building).

Cheney-W:
Can this pr be closed because https://github.com/microsoft/vcpkg/pull/23194 has been merged?

daschuer:
Yes, I have kept this open for reference only.

PR #17565, opened 2021-04-29
[v8] Add support for clang builds.

### LINKED ISSUE #17536: [v8] build failure
**Host Environment**
- OS: Linux
- Compiler: Clang 12

**To Reproduce**
`./vcpkg install v8`

**Failure logs** 
/bin/sh: 1: g++: not found

**Additional context**
v8 fails to build if g++ isn't installed. Since vcpkg ports shouldn't assume a particular compiler, this is clearly a port bug.

Kwizatz:
Environment Variables must be set accordingly:

CC=clang
CXX=clang
AR=llvm-ar
NM=llvm-nm

Fixes #17536

### DIFF
--- a/ports/v8/portfile.cmake
+++ b/ports/v8/portfile.cmake
@@ -1,5 +1,5 @@
 
-set(pkgver "9.0.257.17")
+set(pkgver "9.1.269.39")
 
 set(ENV{DEPOT_TOOLS_WIN_TOOLCHAIN} 0)
 
@@ -71,7 +71,7 @@ endfunction()
 vcpkg_from_git(
     OUT_SOURCE_PATH SOURCE_PATH
     URL https://chromium.googlesource.com/v8/v8.git
-    REF 462fc27a2892702a4d42ffd647789c58ffcee747
+    REF 7d3d62c91f69a702e5aa54c6b4dbbaa883683717
     PATCHES ${CURRENT_PORT_DIR}/v8.patch
 )
 
@@ -79,7 +79,7 @@ message(STATUS "Fetching submodules")
 v8_fetch(
         DESTINATION build
         URL https://chromium.googlesource.com/chromium/src/build.git
-        REF acacc4cc0668cb4dc7f44a3f4430635f438d7478 
+        REF fd86d60f33cbc794537c4da2ef7e298d7f81138e 
         SOURCE ${SOURCE_PATH}
         PATCHES ${CURRENT_PORT_DIR}/build.patch)
 v8_fetch(
@@ -107,7 +107,12 @@ v8_fetch(
         URL https://chromium.googlesource.com/chromium/src/third_party/markupsafe.git
         REF 8f45f5cfa0009d2a70589bcda0349b8cb2b72783
         SOURCE ${SOURCE_PATH})
-
+v8_fetch(
+          DESTINATION tools/clang
+          URL https://chromium.googlesource.com/chromium/src/tools/clang.git
+          REF ba668f13d135f1d01faf9b03c9a05f5877ac3f51
+          SOURCE ${SOURCE_PATH})
+  
 vcpkg_execute_required_process(
         COMMAND ${PYTHON2} build/util/lastchange.py -o build/util/LASTCHANGE
         WORKING_DIRECTORY ${SOURCE_PATH}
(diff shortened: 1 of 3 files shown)


JackBoosY (changes requested):

JackBoosY (approved):

ports/v8/portfile.cmake
```diff
@@ -68,6 +68,13 @@ function(v8_fetch)
   endforeach()
 endfunction()
 
+set(is_clang "false")
+if($ENV{CXX} MATCHES "clang")
+    message(STATUS "Detected clang compiler at CXX")
+    set(is_clang "true")
+    set(clang_vars "clang_use_chrome_plugins=false")
+endif()
```
strega-nil-ms, line 76: This is almost certainly not the correct way to check for clang. I'm not sure how to solve it, but it's likely that we'd need to use `vcpkg_internal_get_cmake_vars`. The question here is why `is_clang` is needed for v8 here; what does it do?

strega-nil-ms (changes requested):
We need to figure out a better way to do this.

ports/v8/portfile.cmake
Kwizatz: > 
> 
> This is almost certainly not the correct way to check for clang. I'm not sure how to solve it, but it's likely that we'd need to use `vcpkg_internal_get_cmake_vars`. The question here is why `is_clang` is needed for v8 here; what does it do?

is_clang determines whether GN would use a clang toolchain vs a native toolchain, it is a chromium construct, it is not necessarily "the correct" or "the cannon" GN way to do this, but it is how chromium and v8 does it.

Just like CMake, GN detects the compiler from ENV variables CXX, CC and so on, CMAKE_CXX_COMPILER should provide the value of CXX for example, but at the portfile, this variable is not set. vcpkg_internal_get_cmake_vars can provide this variable, but it is the only variable we need, and it is right there, not in a generated file from a different build system (cmake vs gn).

Kwizatz:
> 
> 
> We need to figure out a better way to do this.

Agreed

ports/v8/portfile.cmake
```diff
@@ -146,10 +162,9 @@ else()
 endif()
 
 message(STATUS "Generating v8 build files. Please wait...")
-
 vcpkg_configure_gn(
     SOURCE_PATH ${SOURCE_PATH}
-    OPTIONS "is_component_build=${is_component_build} target_cpu=\"${VCPKG_TARGET_ARCHITECTURE}\" v8_monolithic=${v8_monolithic} v8_use_external_startup_data=${v8_use_external_startup_data} use_sysroot=false is_clang=false use_custom_libcxx=false v8_enable_verify_heap=false icu_use_data_file=false" 
+    OPTIONS "is_component_build=${is_component_build} target_cpu=\"${VCPKG_TARGET_ARCHITECTURE}\" v8_monolithic=${v8_monolithic} v8_use_external_startup_data=${v8_use_external_startup_data} use_sysroot=false is_clang=${is_clang} use_custom_libcxx=false v8_enable_verify_heap=false icu_use_data_file=false ${custom_toolchain} ${clang_vars}"
```
Kwizatz, line 167: @strega-nil-ms This is where is_clang is used to tell GN to use the clang toolchain vs the msvc toolchain.
I agree with you that the method to determine the toolchain method can be improved, but we do need this variable, other things being equal, like with other packages, cmake/gn would just check CXX and their kind and take proper acction, but we do need to pass this variable here, its value determines more than just what compiler binary to run.

strega-nil-ms: so I don't understand; why does gn need to be told to use the clang toolchain or the msvc toolchain? shouldn't they take it from `vcpkg_internal_get_cmake_vars`?

Kwizatz: Is it? chromium needs to support many platforms and configurations, like I said, the variable is part of the chromium build repo, we could not pass it for clang builds, but we'll have to pass it for msvc builds, so it is still needed.

I am not sure how could vcpkg_configure_gn be changed/fixed to handle this, it would be like assuming all CMake projects always had an OPTION(is_clang ... OFF) directive.

dan-shaw:
@Kwizatz As a temporary fix, I think we should use `vcpkg_internal_get_cmake_vars` as @strega-nil-ms suggested. Since this is an internal script and not yet stable, we likely will need to use a snapshot of the current script.

Kwizatz:
@dan-shaw do you mean vcpkg_internal_get_GN_vars? v8 uses GN.

It may be possible to just detect the environment variables from the gn script and not have to pass is_clang explicitly, but I can already see the argument against "adding features via patches" being made.

Anyway, I am done for now, if anyone is interested on taking over and making the changes feel free to do so, I'll come back to this when its time for a version update, as it is, there's a bigger issue regarding mixing C++ style attributes ([[nodiscard]] specifically) and gcc attributes ((__attribute__(visibility=default)) specifically) on Linux.

PR #17579, opened 2021-04-29
[python3] Disable registry access to determine WinSDK

ras0219-msft:
Currently, `python3` uses code like the following to get the Windows SDK Version:
```xml
  <PropertyGroup Condition="$(DefaultWindowsSDKVersion) == ''">
    <!--
    Attempt to select the latest installed WinSDK. If we don't find any, then we will
    let the MSBuild targets determine which one it wants to use (typically the earliest
    possible version). Since we limit WINVER to Windows 7 anyway, it doesn't really
    matter which WinSDK version we use.
    -->
    <_RegistryVersion>$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\Windows\v10.0@ProductVersion)</_RegistryVersion>
    <_RegistryVersion Condition="$(_RegistryVersion) == ''">$(Registry:HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Microsoft SDKs\Windows\v10.0@ProductVersion)</_RegistryVersion>
    <!-- Sometimes the version in the registry has to .0 suffix, and sometimes it doesn't. Check and add it -->
    <_RegistryVersion Condition="$(_RegistryVersion) != '' and !$(_RegistryVersion.EndsWith('.0'))">$(_RegistryVersion).0</_RegistryVersion>

    <!-- The minimum allowed SDK version to use for building -->
    <DefaultWindowsSDKVersion>10.0.10586.0</DefaultWindowsSDKVersion>
    <DefaultWindowsSDKVersion Condition="$([System.Version]::Parse($(_RegistryVersion))) > $([System.Version]::Parse($(DefaultWindowsSDKVersion)))">$(_RegistryVersion)</DefaultWindowsSDKVersion>
  </PropertyGroup>
```

This is bad for a variety of reasons, but specifically we want to avoid registry access.

### DIFF
--- a/ports/python3/portfile.cmake
+++ b/ports/python3/portfile.cmake
@@ -88,6 +88,11 @@ if(VCPKG_TARGET_IS_WINDOWS OR VCPKG_TARGET_IS_UWP)
             "/p:ForceImportBeforeCppTargets=${SOURCE_PATH}/PCbuild/python_vcpkg.props"
         )
     endif()
+    string(REPLACE "\\" "" WindowsSDKVersion "$ENV{WindowsSDKVersion}")
+    list(APPEND OPTIONS
+        "/p:WindowsTargetPlatformVersion=${WindowsSDKVersion}"
+        "/p:DefaultWindowsSDKVersion=${WindowsSDKVersion}"
+    )
     if(VCPKG_TARGET_IS_UWP)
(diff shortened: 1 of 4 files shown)


BillyONeal (approved):

aekewu1111 (commented):
The

Issue #17607, opened 2021-05-01
Package installation layout for tools


dg0yt:
I fail to find documentation (or a specification) for the package installation layout.

In particular, I would like to know where to properly place tools like `curl-config` or `geos-config`, i.e. executable scripts which provide information about how to use the corresponding libraries. 
Software which uses theses script (e.g. gdal) expects them to be found by name via the `PATH` environment variable or to be provided explicit by parameters (e.g. gdal's `--with-geos`)

This is what I found in vcpkg:
- ~The release variant of these scripts (`curl-config`, `geos-config`) is currently installed to 
  `share/${PORT}`~
  ~The debug variant (which could provide the d-suffixed library names) is dropped,~
  Fixed.
- The following locations are automatically added to the `PATH` environment variable by `vcpkg.cmake`, ~even for cross-builds (Android, iOS)~(since #23322:) for the host triplet: 
  `tools` 
  `tools/*` 
  ~(but no extra subdirectories)~ and all subdirectories (since #23322) 
- For all declared dependencies, the following locations are automatically added to the `CMAKE_PROGRAM_PATH` cmake variable:
  `tools\<DEPENDENCY>`
  This variable is used by `cmake_find_program` which is sometimes called with `PATH_SUFFIXES bin` (`FindMPI.cmake` #18151), sometimes without suffixes (`FindGettext.cmake` #13518).
- sqlite3 executable is installed in
  `tools`
- vcpkg_copy_tools defaults to dealing with
   `tools/${PORT}`
- vcpkg_configure_make installs tools to
  `tools/${PORT}/bin` 
  `tools/${PORT}/sbin` 
  `tools/${PORT}/debug/bin`
  `tools/${PORT}/debug/sbin`
- vcpkg_configure_meson has undocumented options ADDITIONAL_NATIVE_BINARIES, ADDITIONAL_CROSS_BINARIES which are used by some ports like
  `tools/${PORT}/<COMPONENT>`
- vcpkg_cmake_config_fixup adjusts the path of tools references in cmake config files to
  `tools/${PORT}`
- dbus installs debug tools to  
  `debug/tools/${PORT}`.
- Some tools which shall not automatically be in the (CMAKE_PROGRAM_)PATH are installed to  
  `manual-tools/${PORT}`.
  The Linuxish standard for this would be  
  `libexec`  
  and it is used, too.

My questions:
1. When to use `tools` and when to use another prefix?
2. Where to place debug variants, if needed?
[...]

Issue #16602, opened 2021-03-08
x64-dynamic-linux build problem for geos (Geometry Engine Open Source)


SergeiDelov:
When attempted to build Linux packages as dynamic libraries (see https://github.com/microsoft/vcpkg/blob/master/docs/examples/overlay-triplets-linux-dynamic.md) got the following log:
````
Building package geos[core]:x64-dynamic-linux...
-- [OVERLAY] Loading triplet configuration from: /mnt/c/vcpkg/custom-triplets/x64-dynamic-linux.cmake
-- Downloading http://download.osgeo.org/geos/geos-3.9.0.tar.bz2 -> geos-3.9.0.tar.bz2...
-- Extracting source /mnt/c/vcpkg/downloads/geos-3.9.0.tar.bz2
-- Applying patch dont-build-docs.patch
-- Applying patch dont-build-astyle.patch
-- Using source at /mnt/c/vcpkg/buildtrees/geos/src/3.9.0-27800c28d2.clean
-- Configuring x64-dynamic-linux-dbg
-- Configuring x64-dynamic-linux-rel
-- Building x64-dynamic-linux-dbg
-- Building x64-dynamic-linux-rel
-- Performing post-build validation
There should be no empty directories in /mnt/c/vcpkg/packages/geos_x64-dynamic-linux
The following empty directories were found:

    /mnt/c/vcpkg/packages/geos_x64-dynamic-linux/bin
    /mnt/c/vcpkg/packages/geos_x64-dynamic-linux/debug/bin

If a directory should be populated but is not, this might indicate an error in the portfile.
If the directories are not needed and their creation cannot be disabled, use something like this in the portfile to remove them:

    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/a/dir" "${CURRENT_PACKAGES_DIR}/some/other/dir")


Found 1 error(s). Please correct the portfile:
    /mnt/c/vcpkg/ports/geos/portfile.cmake
-- Performing post-build validation done
Error: Building package geos:x64-dynamic-linux failed with: POST_BUILD_CHECKS_FAILED
Please ensure you're using the latest portfiles with `./vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: geos:x64-dynamic-linux
  Vcpkg version: 2021-01-13-unknownhash
````
Both 
````
/mnt/c:/vcpkg/buildtrees/geos/x64-dynamic-linux-dbg/lib
```` 
and
[...]

Issue #11681, opened 2020-05-31
[geos] build failure dynamic linkage


ASxa86:
**Host Environment**
- OS: Ubuntu 20.04
- Compiler: gcc v9.3.0

**To Reproduce**
`./vcpkg install geos:x64-linux-shared`

**Failure logs** 
There should be no empty directories in /home/aaron/dev/vcpkg/packages/geos_x64-linux-shared
The following empty directories were found:

    /home/aaron/dev/vcpkg/packages/geos_x64-linux-shared/bin
    /home/aaron/dev/vcpkg/packages/geos_x64-linux-shared/debug/bin

If a directory should be populated but is not, this might indicate an error in the portfile.
If the directories are not needed and their creation cannot be disabled, use something like this in the portfile to remove them:

    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/a/dir" "${CURRENT_PACKAGES_DIR}/some/other/dir")

Found 1 error(s). Please correct the portfile:
    /home/aaron/dev/vcpkg/ports/geos/portfile.cmake
-- Performing post-build validation done
Error: Building package geos:x64-linux-shared failed with: POST_BUILD_CHECKS_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: geos:x64-linux-shared
  Vcpkg version: 2020.02.04-unknownhash

Additionally, attach any relevant sections from the log files above.

**Additional context**
Fix by removing the linkage check in to portfile.cmake prior to removing the bin directories.

JackBoosY:
Solved by #17616.

PR #17616, opened 2021-05-01
[geos] Relocatable geos-config; dynamic builds on Unix

dg0yt:
- #### What does your PR fix?  
  This PR
  - makes `geos-config` relocatable, and adds a `geos-config-debug` for the debug build,
    as a step towards solving the gdal port issues (#9068, #8943)
  - fixes dynamic builds on Unix systems (#16602, #11681)

  unchanged, no

  yes

  yes

I still need to test the latest state with my gdal branch, but it is ready for vcpkg CI. The question in #17607 are related, too.


JackBoosY (approved):

ports/geos/portfile.cmake
```diff
@@ -34,19 +34,28 @@ vcpkg_configure_cmake(
...
+    file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/share/${PORT})
+    file(RENAME ${CURRENT_PACKAGES_DIR}/bin/geos-config ${CURRENT_PACKAGES_DIR}/share/${PORT}/geos-config)
+endif()
+if(EXISTS ${CURRENT_PACKAGES_DIR}/debug/bin/geos-config)
+    file(READ "${CURRENT_PACKAGES_DIR}/debug/bin/geos-config" GEOS_CONFIG)
+    string(REGEX REPLACE "(\nprefix=)[^\n]*" [[\1$(CDPATH= cd -- "$(dirname -- "$0")"/../.. && pwd -P)]] GEOS_CONFIG "${GEOS_CONFIG}")
+    string(REGEX REPLACE "(-lgeos(_c)?)d?( |\n)" "\\1d\\3" GEOS_CONFIG "${GEOS_CONFIG}")
+    file(WRITE "${CURRENT_PACKAGES_DIR}/debug/bin/geos-config" "${GEOS_CONFIG}")
+    file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/share/${PORT})
+    file(RENAME ${CURRENT_PACKAGES_DIR}/debug/bin/geos-config ${CURRENT_PACKAGES_DIR}/share/${PORT}/geos-config-debug)
```
ras0219-msft, line 50: Can you give an example of how this would be consumed? My understanding is that `xyz-config` scripts like this are usually referred to like `"${XYZ_ROOT}/xyz-config"` and so having `xyz-config-debug` won't be directly usable. Instead, we should provide `debug/share/geos/geos-config` and set `GEOS_ROOT` to `debug/share/geos` when in debug mode.

dg0yt: gdal `configure` says:
~~~
  --with-geos=ARG         Include GEOS support (ARG=yes, no or geos-config
                          path)
~~~
and similar for curl, kea, sfcgal.

So this is to be consumed in the gdal portfile by:
~~~
    list(APPEND OPTIONS_RELEASE --with-geos=${CURRENT_INSTALLED_DIR}/share/geos/geos-config)
    list(APPEND OPTIONS_DEBUG --with-geos=${CURRENT_INSTALLED_DIR}/share/geos/geos-config-debug)
~~~
I tested this locally, but this just one out of a pile of pending changes.

dg0yt: >  provide `debug/share/geos/geos-config` and set set GEOS_ROOT to debug/share/geos

If I adopt this: Outside vcpkg, the `<package>_ROOT` variable generally (e.g. in CMake find modules) means the installation *prefix*, i.e. `${CURRENT_INSTALLED_DIR}`. 
As a consumer in a port file, I would still have to assign the roots to per-build-type parameters because the build types are handled deeper down in the cmake call stack.

dg0yt: > provide `debug/share/geos/geos-config`

I tried that when preparing the PR, and I verified today: It leads into a build error:
~~~
/debug/share should not exist. Please reorganize any important files, then use
~~~

JackBoosY: Which file is in _${CURRENT_PACKAGES_DIR}/debug/share_?

dg0yt: > Which file is in _${CURRENT_PACKAGES_DIR}/debug/share_?

None with the current PR. @ras0219-msft proposed to have `debug/share/geos/geos-config`, and I reported what would be the result.

dg0yt: The alternative to fixing geos-config, curl-config etc. is forcing the consumers to use pkg-config where possible, by removing these files. Consumers also need to know whether to ask the scripts for static or standard flags, and this is solved for `PKG_CONFIG`.
E.g. port gdal will need a number of patches anyway, e.g. to use pkg-config for tiff.

JackBoosY:
Ping @ras0219-msft for review this PR again.

dg0yt:
More side notes: 
- This port sets the debug postfix ('d'). This might be contrary to the current maintainer guidelines. `git grep geosd` finds a small number of ports which depend on this.
- pc files do not carry the debug postfix.
- pc files are not passed to fixup. (Prefix pointing to packages.)
- Static linking dependencies are incomplete in the pc files.

dg0yt:
Will add pc file handling soon.

dg0yt:
New changes:
- pc files rectified.
- geos-config moved to tools/bin and tools/debug/bin.
- Ported to vcpkg_configure_....
- Building of benchmarks disabled.
- Usage instructions added.

ports/geos/portfile.cmake
```diff
@@ -23,30 +25,45 @@ else()
...
+endif()
+if(EXISTS "${CURRENT_PACKAGES_DIR}/bin/geos-config")
+    file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin")
+    file(RENAME "${CURRENT_PACKAGES_DIR}/bin/geos-config" "${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin/geos-config")
+endif()
+if(EXISTS "${CURRENT_PACKAGES_DIR}/debug/bin/geos-config")
+    file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/tools/${PORT}/debug/bin")
+    file(RENAME "${CURRENT_PACKAGES_DIR}/debug/bin/geos-config" "${CURRENT_PACKAGES_DIR}/tools/${PORT}/debug/bin/geos-config")
+    geos_add_debug_postfix("${CURRENT_PACKAGES_DIR}/tools/${PORT}/debug/bin/geos-config")
 endif()
```
JackBoosY, line 57: Can we use `vcpkg_copy_tools` instead?

JackBoosY (changes requested):

ports/geos/portfile.cmake
dg0yt: > Can we use `vcpkg_copy_tools` instead?

No, no, and no.
- `vcpkg_copy_tools` doesn't handle tools without `VCPKG_TARGET_EXECUTABLE_SUFFIX`.
- `vcpkg_copy_tools` doesn't handle tools for the debug variant.
- For release+debug, the filesystem layout practice of `vcpkg_configure_make` is different from `vcpkg_copy_tools`.

File system layout for tools: That's why I wrote #17607, 23 days ago. Not to forget #13649. This is a blocker for quality of contributions, and also for vcpkg adoption.

JackBoosY (approved):

strega-nil-ms:
Cool, LGTM! Thanks @dg0yt :)

PR #17617, opened 2021-05-01
[rtaudio] add ASIO backend as a feature

dyfer:
**Describe the pull request**

- #### What does your PR fix?  
  My PR adds ASIO support as a feature to the RtAudio library. 

  No change (new feature is available on windows only though)

  Yes

  Yes

### DIFF
--- a/ports/rtaudio/portfile.cmake
+++ b/ports/rtaudio/portfile.cmake
@@ -6,6 +6,7 @@ vcpkg_from_github(
     REF 34a3752e0c8249dc1780d196cd24e745425f0c77
     SHA512 00fea107f409f6dc43154aaf69aeffa1a3385031778b5f7d1ae1cc8337ed4ab92a7917cc9eade848dedd746016b6eff6234088619cb8d6a9a3f26a63efde493e
     HEAD_REF master
+    PATCHES remove-unused-variable.patch
 )
 
 if(VCPKG_CRT_LINKAGE STREQUAL static)
@@ -14,10 +15,15 @@ else()
     set(RTAUDIO_STATIC_MSVCRT OFF)
 endif()
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        asio  RTAUDIO_API_ASIO
+)
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
-    OPTIONS -DRTAUDIO_STATIC_MSVCRT=${RTAUDIO_STATIC_MSVCRT}
+    OPTIONS -DRTAUDIO_STATIC_MSVCRT=${RTAUDIO_STATIC_MSVCRT} ${FEATURE_OPTIONS}
 )
 
 vcpkg_install_cmake()
--- a/ports/rtaudio/vcpkg.json
+++ b/ports/rtaudio/vcpkg.json
@@ -1,7 +1,12 @@
 {
   "name": "rtaudio",
-  "version-date": "2021-01-25",
+  "version-date": "2021-04-30",
   "description": "A set of C++ classes that provide a common API for realtime audio input/output across Linux (native ALSA, JACK, PulseAudio and OSS), Macintosh OS X (CoreAudio and JACK), and Windows (DirectSound, ASIO and WASAPI) operating systems.",
   "homepage": "https://github.com/thestk/rtaudio",
-  "supports": "!uwp"
+  "supports": "!uwp",
+  "features": {
+    "asio": {
+      "description": "Build with ASIO backend"
+    }
+  }
 }
--- a/ports/rtaudio/remove-unused-variable.patch
+++ b/ports/rtaudio/remove-unused-variable.patch
@@ -1,0 +1,20 @@
+diff --git a/include/asiolist.cpp b/include/asiolist.cpp
+index e4c73c2..333a662 100644
+--- a/include/asiolist.cpp
++++ b/include/asiolist.cpp
+@@ -166,7 +166,6 @@ AsioDriverList::AsioDriverList ()
+ 	LPASIODRVSTRUCT	pdl;
+ 	LONG 			cr;
+ 	DWORD			index = 0;
+-	BOOL			fin = FALSE;
+ 
+ 	numdrv		= 0;
+ 	lpdrvlist	= 0;
+@@ -184,7 +183,6 @@ AsioDriverList::AsioDriverList ()
+ #endif
+ 			lpdrvlist = newDrvStruct (hkEnum,keyname,0,lpdrvlist);
+ 		}
+-		else fin = TRUE;
+ 	}
+ 	if (hkEnum) RegCloseKey(hkEnum);
+ 
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -5381,7 +5381,7 @@
       "port-version": 1
     },
     "rtaudio": {
-      "baseline": "2021-01-25",
+      "baseline": "2021-04-30",
       "port-version": 0
     },
     "rtlsdr": {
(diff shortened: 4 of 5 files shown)


ports/rtaudio/vcpkg.json
```diff
@@ -1,7 +1,13 @@
...
   "name": "rtaudio",
-  "version-date": "2021-01-25",
+  "version-date": "2021-04-30",
   "description": "A set of C++ classes that provide a common API for realtime audio input/output across Linux (native ALSA, JACK, PulseAudio and OSS), Macintosh OS X (CoreAudio and JACK), and Windows (DirectSound, ASIO and WASAPI) operating systems.",
   "homepage": "https://github.com/thestk/rtaudio",
-  "supports": "!uwp"
+  "supports": "!uwp",
+  "features": {
+    "asio": {
+      "description": "Build with ASIO backend"
```
autoantwort, line 9:
```suggestion
      "description": "Build with ASIO backend",
```
You can also use `./vcpkg format-manifest --all` to format

dyfer:
The builds fail with 
`Command failed: ninja -v`
https://dev.azure.com/vcpkg/public/_build/results?buildId=53034&view=logs&j=821e2c31-c252-5a5c-eeb9-89df78f2f610&t=34ac2787-4652-59a0-fda0-8b82a81de66d&l=1760
which doesn't seem to be related to my changes. Locally I have experiences this as well (before my change) and worked around it by adding `ninja` to my `PATH`, though I don't use VCPKG often and I'm not sure what would be the proper way to address this.

autoantwort:
You can download the log files if you click on the jobs root node -> artifacts -> tree dots on the right (new page) -> download. The content is:
```
-- The CXX compiler identification is MSVC 19.28.29914.0
-- Detecting CXX compiler ABI info - failed
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.28.29910/bin/Hostx64/x86/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.28.29910/bin/Hostx64/x86/cl.exe - broken
CMake Error at D:/downloads/tools/cmake-3.20.1-windows/cmake-3.20.1-windows-i386/share/cmake-3.20/Modules/CMakeTestCXXCompiler.cmake:59 (message):
  The C++ compiler

    "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.28.29910/bin/Hostx64/x86/cl.exe"

    Change Dir: D:/buildtrees/rtaudio/x86-windows-dbg/CMakeFiles/CMakeTmp
    
    Run Build Command(s):D:/downloads/tools/ninja/1.10.1-windows/ninja.exe cmTC_d9e12 && [1/2] Building CXX object CMakeFiles\cmTC_d9e12.dir\testCXXCompiler.cxx.obj
    FAILED: CMakeFiles/cmTC_d9e12.dir/testCXXCompiler.cxx.obj 
    C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1428~1.299\bin\Hostx64\x86\cl.exe  /nologo /TP   -Wno-unused  /D_DEBUG /MDd /Z7 /Ob0 /Od /RTC1 /showIncludes /FoCMakeFiles\cmTC_d9e12.dir\testCXXCompiler.cxx.obj /FdCMakeFiles\cmTC_d9e12.dir\ /FS -c testCXXCompiler.cxx
    cl : Command line error D8021 : invalid numeric argument '/Wno-unused'

  CMake will not be able to correctly generate this project.
  CMakeLists.txt:5 (project)
```

The important line: `cl : Command line error D8021 : invalid numeric argument '/Wno-unused'`

ports/rtaudio/portfile.cmake
```diff
@@ -14,10 +14,16 @@ else()
...
+    FEATURES
+        asio  RTAUDIO_API_ASIO
+)
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
-    OPTIONS -DRTAUDIO_STATIC_MSVCRT=${RTAUDIO_STATIC_MSVCRT}
+    OPTIONS -DRTAUDIO_STATIC_MSVCRT=${RTAUDIO_STATIC_MSVCRT} ${FEATURE_OPTIONS}
+    OPTIONS_DEBUG -DCMAKE_CXX_FLAGS="-Wno-unused"
```
autoantwort, line 26: Is this line necessary?

dyfer: I think so... The debug build with MinGW was failing because there's an unused variable in [asiolist.cpp](https://github.com/thestk/rtaudio/blob/master/include/asiolist.cpp#L169)

dyfer:
> The important line: cl : Command line error D8021 : invalid numeric argument '/Wno-unused'

Ah I see, thanks for that.

The problem is, that when built with MinGW, `-Werror` option is set in the library's CMakeList (I think only for the debug build) and the build fails. Is there a way to set `Wno-unused` only for non-msvc build?

autoantwort:
> Is there a way to set `-Wno-unused` only for non-msvc build?

There is `VCPKG_TARGET_IS_MINGW`

ports/rtaudio/portfile.cmake
autoantwort: But the variable is used [here](https://github.com/thestk/rtaudio/blob/master/include/asiolist.cpp#L187)?

autoantwort: Ah, but the variable is only written to but never read. So unused.

dyfer: Yes, it seems so. In any case, compilation with MinGW pointed to that line with the unused variable error. I'm working now on change to define that CXX flag only for MinGW.

autoantwort: Another option would be to add a patch that simply removes the variable

ports/rtaudio/portfile.cmake
```diff
@@ -14,10 +14,20 @@ else()
...
 endif()
 
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        asio  RTAUDIO_API_ASIO
+)
+
+if(VCPKG_TARGET_IS_MINGW)
+    set(VCPKG_CXX_FLAGS_DEBUG "-Wno-unused")
+    set(VCPKG_C_FLAGS_DEBUG "-Wno-unused")
```
ras0219-msft, line 24: We should instead patch out the `-Werror` in the underlying build, which fixes the true problem. Otherwise, compiler updates that introduce new warnings will result in new build failures.

dyfer: All right. As there's a [pending PR](https://github.com/thestk/rtaudio/pull/288) that would fix this in the source, I'll wait a few days in case that happens, and if not, I'll add a patch to this port, does that sound okay?

dyfer: I have updated this PR to include the patch, as the changes have not been merged upstream.

dyfer: Ugh, maybe I misunderstood. I added the patch to remove the unused variable... was I supposed to remove `-Werror` instead?

PhoebeHui: @dyfer, I think you don't need to remove -Werror since the real issue was fixed by the patch.

PhoebeHui (approved):
LGTM, all features test passed with x64-windows.

autoantwort:
It is now fixed upstream

dyfer:
@PhoebeHui should I update this PR with the upstream fix?

PhoebeHui:
@dyfer, you can create a new PR for this.

PR #17629, opened 2021-05-02
[msmpi] Update the download hash

### LINKED ISSUE #17622: [msmpi] build failure
**Host Environment**
- OS: [Windows 10 Pro Version 20H2 build 19042.928 with Windows Feature Experience Pack 120.2212.551.0
- Compiler: Microsoft Visual Studio Community 2019 Version 16.9.4

**To Reproduce**
`./vcpkg install msmpi --triplet x64-windows`

This fails with the following error.

```

          File path: [ D:/vcpkg/downloads/temp/msmpisdk-10.1.12498.msi ]
      Expected hash: [ 330fad53c1979dfae786abed228d82c82207f0d61148e2efc1f37c8931838b806eb50554e7f006a56962f748a21f7017c540bd70444b7a93a72c313a64e9254c ]
        Actual hash: [ b0087dfd21423bf87b94b17d7cb03576838585371bbf8b03cca95c3ad73670108c7bc6517b0de852ef595072cc4143be2011636e7242bcb080394d94294848a7 ]

  environment variables to
```

I **do not** have a proxy configured!

JackBoosY:
Fixes #17622, seems the upstream was updated this package.

Note: already checked `msmpisetup-${MSMPI_VERSION}.exe` hash, don't need to update.


ports/msmpi/portfile.cmake
```diff
@@ -6,7 +6,7 @@ set(SOURCE_PATH ${CURRENT_BUILDTREES_DIR}/src/msmpi-${MSMPI_VERSION})
 vcpkg_download_distfile(SDK_ARCHIVE
     URLS "https://download.microsoft.com/download/a/5/2/a5207ca5-1203-491a-8fb8-906fd68ae623/msmpisdk.msi"
     FILENAME "msmpisdk-${MSMPI_VERSION}.msi"
```
ras0219-msft, line 8:
```suggestion
    FILENAME "msmpisdk-${MSMPI_VERSION}-b0087dfd.msi"
```
Whenever changing the SHA of a download, we need to also change the filename.

JackBoosY: @ras0219-msft got it! Thanks for notice.

versions/m-/msmpi.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "51eaeec27ead5e4291172b6be205b731f966da9f",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "d333159a4463603aade50bfa48dbb4bd5b33946e",
```

PhoebeHui (approved):

strega-nil-ms:
Thanks @JackBoosY!

PR #17641, opened 2021-05-03
[DCMTK] Update to 3.6.6

BigSeb:
**Describe the pull request**

- #### What does your PR fix?  
  Update DCMTK to 3.6.6

No change of triplets have been made, it should be the same as the previous version. It was personally tested with x86-windows and x64-windows.

  Yes

  Yes


ports/dcmtk/dcmtk.patch
```diff
@@ -1,16 +1,16 @@
-diff --git a/CMake/dcmtkPrepare.cmake b/CMake/dcmtkPrepare.cmake
-index e9f77b5..4f7ebe3 100644
```
ras0219-msft, line 2: Is the change to this patch file needed? It looks like you've just uppercased the resulting `set` directive.

BigSeb: Line numbers where the patch must be applied have been changed. I have generated a new patch using git in order to generate this new patch version. I did not change the original case of the CMake code (but the creator of the patch did change the case). I agree that case of the command is not important for the patch, only the context in which the patch must be applied need to change (line numbers).

PhoebeHui (approved):
LGTM, thanks for the PR!

strega-nil-ms:
Cool, thanks @BigSeb :)

PR #17643, opened 2021-05-03
[flashlight-cpu] Update port

jacobkahn:
- #### What does your PR fix?  

Updates the port to include the fix to https://github.com/flashlight/flashlight/issues/574.

Correctly moves CMake assets in `[installed]/share/flashlight-cpu` into `[installed]/share/flashlight`

No change to supported triplets

Yes

Yes


ports/flashlight-cpu/usage
```diff
@@ -0,0 +1,11 @@
+The package flashlight-cpu provides CMake targets:
```
JackBoosY, line 1: If this port exports its target, this usage information should be displayed automatically.
It is not clear why it needs to be added.

jacobkahn: The issue is that the name of the port is flashlight-cpu, whereas the CMake config file that's generated is called flashlightConfig.cmake.

The auto-generated usage message from vcpkg says that the user should use `find_package(flashlight-cpu ...)` - which is wrong - instead of `find_package(flashlight ...)`. See the issue in the original PR description for an example of that confusion.

JackBoosY: So I think we should add `vcpkg_fixup_cmake_targets(CONFIG_PATH share/flashlight-cpu TARGET_PATH share/flashlight)` and remove the usage.

jacobkahn: Good point. Let me fix this up and test it.

jacobkahn: @JackBoosY fixed! Ready for review

JackBoosY (changes requested):

versions/f-/flashlight-cpu.json
```diff
@@ -1,7 +1,12 @@
 {
   "versions": [
     {
-      "git-tree": "63e05566899f6a034fb93cf6556cf3d5ca387f61",
+      "git-tree": "9de5ce0a5763ebc2b46378a46172b8cda7ac95c5",
+      "version": "0.3",
+      "port-version": 1
+    },
+    {
+      "git-tree": "8a2fdb6cf5423b0dd078ccaebbddace28ac896e8",
```
JackBoosY, line 9:
```suggestion
      "git-tree": "63e05566899f6a034fb93cf6556cf3d5ca387f61",
```

JackBoosY (changes requested):

JackBoosY (approved):

PR #17645, opened 2021-05-03
[flashlight-cuda] Update port

jacobkahn:
Same as https://github.com/microsoft/vcpkg/pull/17643, but for `flashlight-cuda`.

- #### What does your PR fix?  

Updates the port to include the fix to https://github.com/flashlight/flashlight/issues/574.

Correctly moves CMake assets in `[installed]/share/flashlight-cuda` into `[installed]/share/flashlight`

No change to supported triplets

Yes

Yes


ports/flashlight-cuda/usage
```diff
@@ -0,0 +1,11 @@
+The package flashlight-cuda provides CMake targets:
```
ras0219-msft, line 1: Do our built-in automatic heuristics not work here?

Also, this is missing a call to `vcpkg_cmake_config_fixup()`: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.md

```cmake
vcpkg_cmake_config_fixup()
```

jacobkahn: Good point. Let me give that a try and see if this can mean we can remove the usage.

jacobkahn: @ras0219 fixed using `vcpkg_fixup_cmake_targets(...)`. Ready for review

JonLiu1993 (approved):

versions/f-/flashlight-cuda.json
```diff
@@ -1,7 +1,12 @@
 {
   "versions": [
     {
-      "git-tree": "bd616669b5eb7413794d727b9909309ae686cd04",
+      "git-tree": "9147d15ccbb61b2de168a8cc78527341edf1540d",
+      "version": "0.3",
+      "port-version": 1
+    },
+    {
+      "git-tree": "56fb4895272b9e0d75b474e4b6caf941733fc495",
```
dan-shaw, line 9: The `git-tree` reference of the previous version should be the same.

PhoebeHui (approved):
Thanks for the PR!

Issue #17650, opened 2021-05-04
[libgnutls] Static use depends on -ldl, not carried in pc file


dg0yt:
**Describe the bug**
While trying to update port libmicrohttpd, I stumbled over unresolved symbols when using static build of port libgnutls in config.log:

~~~
configure:24362: checking whether GnuTLS is usable
configure:24385: gcc -o conftest -I/home/kpa/Projekte/installed/x64-linux/debug/lib/pkgconfig/../../../include -fPIC -g -fno-strict-aliasing  -L/home/kpa/Projekte/installed/x64-linux/debug/lib  conftest.c  -L/home/kpa/Projekte/installed/x64-linux/debug/lib/pkgconfig/../../lib -lgnutls -pthread -ltspi -lgmp -latomic -lhogweed -lgmp -lnettle -ltasn1 -lidn2 >&5
/home/kpa/Projekte/installed/x64-linux/debug/lib/libgnutls.a(tpm.o): In function `check_init':
/home/kpa/Projekte/buildtrees/libgnutls/x64-linux-dbg/lib/.././../src/3.6.15-4050d06d46.clean/lib/tpm.c:123: undefined reference to `dlopen'
/home/kpa/Projekte/buildtrees/libgnutls/x64-linux-dbg/lib/.././../src/3.6.15-4050d06d46.clean/lib/tpm.c:129: undefined reference to `dlsym'
/home/kpa/Projekte/buildtrees/libgnutls/x64-linux-dbg/lib/.././../src/3.6.15-4050d06d46.clean/lib/tpm.c:129: undefined reference to `dlclose'
...
collect2: error: ld returned 1 exit status
...
configure:24397: result: no
configure:24407: WARNING: pkg-config reports that GnuTLS is present, but GnuTLS can't be used
~~~

I'm not sure if this is desired for the static build, or if it is just a bug in the `.pc` files.

**Environment**
- OS: linux
- Triplet: default (x64-linux)

**To Reproduce**
1. Add libgnutls dependency to libmicrohttpd
2. Add `--with-https` to OPTIONS for vcpkg_configure_make in libmicrohttpd portfile.
3. `./vcpkg install libmicrohttpd`
4. Inspect config.log

**Expected behavior**
No undefined references when consuming libgnutls via .pc files in a static build

wrobelda:
I just tried the same under Ubuntu 18.04.5 LTS and getting:

```
configure:24362: checking whether GnuTLS is usable
[...]

PR #17653, opened 2021-05-04
[vcpkg baseline][libmicrohttpd] Control dependencies, doc, examples, test

### LINKED ISSUE #17636: [libmicrohttpd] Port doesn't declare its dependencies
The configuration of libmicrohttpd tests for dependencies, but the port doesn't declare any dependencies.

- curl
- gettext
- libgnutls

On macOS, it could pick up files provided either from brew or from a port if it is installed. The mixing of brew and overridden prefix creates volatile configurations.

Output from `configure` for macOS in CI:
~~~
configure: GNU libmicrohttpd 0.9.63 Configuration Summary:
  Target directory:  /Users/vagrant/Data/installed/x64-osx
  Cross-compiling:   no
  Operating System:  darwin20.3.0
  Shutdown of listening socket trigger select: no
  Inter-thread comm: pipe
  poll support:      yes
  epoll support:     no
  sendfile used:     yes, Darwin-style
  HTTPS support:     yes (using libgnutls)
  Threading lib:     posix
  Use thread names:  yes
  Use debug asserts: no
  Messages:          yes
  Gettext:           yes
  Basic auth.:       yes
  Digest auth.:      yes
  HTTP "Upgrade":    yes
  Postproc:          yes
  Build docs:        yes
  Build examples:    yes
  Test with libcurl: no, many unit tests will not run
~~~

(From the failed osx build in #17205. Even libcurl is found in some other contexts.)

_Originally posted by @dg0yt in https://github.com/microsoft/vcpkg/issues/17205#issuecomment-831057544_

dg0yt:
- #### What does your PR fix?  
  - Fixes #17636 by explicitly
    - disabling https (was inactive on Windows; was used randomly when libgnutls happened to be found in the system; has issues when linking the static build of port libgnutls, #17650)
    - disabling curl (used for tests)
    - declaring gettext dependency (if not windows)
  - Disables building of doc and examples (common practice in vcpkg)
  - Adds mingw support

  unchanged, no 
  Note that CI baseline contains: `libmicrohttpd:x64-windows-static-md=fail`

  yes

  yes

CC @NancyLi1013

### DIFF
--- a/ports/libmicrohttpd/portfile.cmake
+++ b/ports/libmicrohttpd/portfile.cmake
@@ -38,6 +38,11 @@ else()
     endif()
     vcpkg_configure_make(
         SOURCE_PATH "${SOURCE_PATH}"
+        OPTIONS
+            --disable-doc
+            --disable-examples
+            --disable-curl
+            --disable-https
     )
 
     vcpkg_install_make()
--- a/ports/libmicrohttpd/portfile.cmake
+++ b/ports/libmicrohttpd/portfile.cmake
@@ -11,7 +11,7 @@ vcpkg_download_distfile(ARCHIVE
 )
 
 vcpkg_extract_source_archive_ex(
-    ARCHIVE ${ARCHIVE}
+    ARCHIVE "${ARCHIVE}"
     OUT_SOURCE_PATH SOURCE_PATH
     PATCHES fix-msvc-project.patch
 )
@@ -30,8 +30,8 @@ if (VCPKG_TARGET_IS_WINDOWS)
         DEBUG_CONFIGURATION "Debug-${CFG_SUFFIX}"
     )
     
-    file(GLOB MICROHTTPD_HEADERS ${SOURCE_PATH}/src/include/microhttpd*.h)
-    file(COPY ${MICROHTTPD_HEADERS} DESTINATION ${CURRENT_PACKAGES_DIR}/include)
+    file(GLOB MICROHTTPD_HEADERS "${SOURCE_PATH}/src/include/microhttpd*.h")
+    file(COPY ${MICROHTTPD_HEADERS} DESTINATION "${CURRENT_PACKAGES_DIR}/include")
 else()
     if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "static")
         set(ENV{LIBS} "$ENV{LIBS} -framework Foundation -framework AppKit") # TODO: Get this from the extracted cmake vars somehow
@@ -42,7 +42,7 @@ else()
 
     vcpkg_install_make()
     
-    file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
 endif()
 
-file(INSTALL ${SOURCE_PATH}/COPYING DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+file(INSTALL "${SOURCE_PATH}/COPYING" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME copyright)
--- a/ports/libmicrohttpd/portfile.cmake
+++ b/ports/libmicrohttpd/portfile.cmake
@@ -16,7 +16,7 @@ vcpkg_extract_source_archive_ex(
     PATCHES fix-msvc-project.patch
 )
 
-if (VCPKG_TARGET_IS_WINDOWS)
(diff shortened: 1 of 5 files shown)


dg0yt:
PS: A https feature might be added later, even for Windows (https://stackoverflow.com/questions/39928579/libmicrohttpd-and-ssl). Note that this feature limits licensing to "LGPL". Without https, licensing is "LGPL or eCos".

dg0yt:
Converted `version-string` to `version` now.

NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your PR @dg0yt.

ports/libmicrohttpd/portfile.cmake
```diff
@@ -30,19 +30,24 @@ if (VCPKG_TARGET_IS_WINDOWS)
...
     if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "static")
         set(ENV{LIBS} "$ENV{LIBS} -framework Foundation -framework AppKit") # TODO: Get this from the extracted cmake vars somehow
     endif()
     vcpkg_configure_make(
         SOURCE_PATH "${SOURCE_PATH}"
+        OPTIONS
+            --disable-doc
+            --disable-examples
+            --disable-curl
+            --disable-https
```
JackBoosY, line 45: I think this is not enough, we should ban gnutls as much as possible:
```
Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-pic[=PKGS]       try to use only PIC/non-PIC objects [default=use
                          both]
  --with-aix-soname=aix|svr4|both
                          shared library versioning (aka "SONAME") variant to
                          provide on AIX, [default=aix].
  --with-gnu-ld           assume the C compiler uses GNU ld [default=no]
  --with-sysroot[=DIR]    Search for dependent libraries within DIR (or the
                          compiler's sysroot if not specified).
  --with-gnu-ld           assume the C compiler uses GNU ld [default=no]
  --with-libiconv-prefix[=DIR]  search for libiconv in DIR/include and DIR/lib
  --without-libiconv-prefix     don't search for libiconv in includedir and libdir
  --with-libintl-prefix[=DIR]  search for libintl in DIR/include and DIR/lib
  --without-libintl-prefix     don't search for libintl in includedir and libdir
  --with-threads=LIB      choose threading library (posix, w32, auto, none)
                          [auto]
  --with-libcurl=PREFIX   look for the curl library in PREFIX/lib and headers
                          in PREFIX/include
  --with-gnutls[=PFX]     use GnuTLS for HTTPS support, optional PFX overrides
                          pkg-config data for GnuTLS headers (PFX/include) and
                          libs (PFX/lib)
  --with-libgcrypt-prefix=PFX
                          prefix where LIBGCRYPT is installed (optional)
```

dg0yt: Thanks for reviewing again. 

Upstreams webpage says "for TLS (HTTPS) support it require libgnutls", and "If you disable HTTPS/TLS support, you can also choose the second license, the eCos License."

I assume that `--disable-https` is enough to disable gnutls, because it is the *feature* which would make use of gnutls, and the licensing change is documented in output from `configure`.
The gnutls parameters come from the *pkgconfig* macros but I don't expect them to activate additional features.
Anyway I could add `--without-gnutls` on top of the changes if you think this is mandatory to be safe.

At least I think that no other PR should pass review without adding `--disable-doc --disable-examples`.

dg0yt: Verified: disabling https (for identical feature set on all targets) is disabling libgnutls lookup:
https://git.gnunet.org/libmicrohttpd.git/tree/configure.ac#n1759

JackBoosY: Yes, I know this, but if there is no clear statement, if someone at the vcpkg level is going to add gnutls support in the future, there will be a conflict. 
I think you need to add this parameter to the comment at least in this line.

dg0yt: Added `--with-gnutls=no`.

JackBoosY (approved):

JackBoosY:
Ping @dan-shaw @strega-nil-ms  for review and merge this PR first.

dg0yt:
FTR, I also noticed a missing `vcpkg_fixup_pkgconfig`. The prefix is the installed dir, absolutely. 
However, I cannot add the missing call before #17898 is merged.

JackBoosY:
@dg0yt It doesn't matter, we can add it in other PRs

dg0yt:
Reminder: The microhttpd build issue affects all changes to `vcpkg_configure_make` on osx. And these CI runs are expensive.

PR #17672, opened 2021-05-05
[docs registries] Add reference documentation

strega-nil:
This PR adds reference documentation for both users of registries as well as creators of registries. Please review!

### DIFF
--- a/docs/maintainers/registries.md
+++ b/docs/maintainers/registries.md
@@ -10,21 +10,64 @@ have created, please read [this documentation](../users/registries.md).
 
 - [Creating Registries](#creating-registries)
   - [Table of Contents](#table-of-contents)
-  - [Layout of Registries](#layout-of-registries)
-    - [Version Objects](#version-objects)
-      - [Version Objects: location field](#version-objects-location-field)
-      - [Version Objects: version field](#version-objects-version-field)
-  - [Adding Versions to a Registry](#adding-versions-to-a-registry)
+  - [Overview](#overview)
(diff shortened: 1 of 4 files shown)


docs/maintainers/registries.md
```diff
@@ -0,0 +1,350 @@
...
+
+Once you've finished your testing, you'll need to make sure that the directory
+as it is is under git's purview. You'll do this by creating a temporary commit:
+
+```pwsh
+> git add ports/kitten
+> git commit -m 'temporary commit'
+```
+
+Then, get the git tree ID of the directory:
```
dg0yt, line 137: Isn't this what is automated via `vcpkg x-add-version`?

strega-nil-ms: It is, but only in the builtin registry (it doesn't work for the not-builtin registry)

docs/users/registries.md
```diff
@@ -0,0 +1,127 @@
...
+`vcpkg-configuration.json`, one can tell exactly from which registry a
+package definition will be fetched from.
+
+The name resolution algorithm is as follows:
+
+- If there is a package registry that claims the package name,
+  use that registry; otherwise
+- If there is a default registry defined, use that registry; otherwise
+- If the default registry is set to `null`, error out; otherwise
+- use the built-in registry.
```
AugP, line 127: May be worth adding  a heading after this on "Versioning & Registries" and just call out that registries works with the versioning feature the same way as the public catalog, and link to the versioning docs if people want to learn more about versioning.

PR #17696, opened 2021-05-06
[cfitsio] Fix pc files

dg0yt:
- #### What does your PR fix?  
  - Fixes missing version, -l flags and dependencies in pc files.
  - ~Add 'd' suffix to the debug library.~

  unchanged, CI baseline contains:
  ~~~
  cfitsio:arm-uwp=fail
  cfitsio:x64-uwp=fail
  ~~~

  yes

  yes

Tested usage with pkg-config in WIP gdal branch for x64-linux.


dg0yt:
PS: I have no idea if the Windows-only pthreads feature would need a "Requires" item in the pc file, too. But at least nobody missed it so far.

ports/cfitsio/portfile.cmake
```diff
@@ -34,8 +39,15 @@ endif()
...
     PREFER_NINJA
-    OPTIONS ${FEATURE_OPTIONS}
+    OPTIONS
+        ${FEATURE_OPTIONS}
         -DUSE_PTHREADS=${WITH_PTHREADS}
+        "-DPKG_CONFIG_REQUIRES=${PKG_CONFIG_REQUIRES}"
+    OPTIONS_RELEASE
+        -DPKG_CONFIG_LIBS=-lcfitsio
+    OPTIONS_DEBUG
+        -DCMAKE_DEBUG_POSTFIX=d
```
PhoebeHui, line 49: If the upstream library has 'd' in debug? if not, it would break our policy, 
see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-rename-binaries-outside-the-names-given-by-upstream.

PhoebeHui (changes requested):
@dg0yt, thanks for the PR!

ports/cfitsio/portfile.cmake
dg0yt: I see. Fixed.

PhoebeHui (approved):

dg0yt:
I rechecked this PR for mingw and feature configuration, but I don't want to modify this PR. Everything left as reviewed. Thanks.

dg0yt:
I moved the dependencies to `Requires.private` instead of `Requires`. This for linking statically.

PR #17701, opened 2021-05-06
[embree3] Fix Embree 3 static compilation patch

chrismile:
**Describe the pull request**

- #### What does your PR fix?  
  ports/embree3/fix-static-usage.patch misses checks for AVX512KNL and AVX512SKX. Without these checks, finding the package in CMake will fail on systems enabling one of these two configurations. The pull request adds the missing checks.

  all, no

  Yes.

  Yes.

### DIFF
--- a/ports/embree3/CONTROL
+++ b/ports/embree3/CONTROL
@@ -1,6 +1,6 @@
 Source: embree3
 Version: 3.11.0
-Port-Version: 1
+Port-Version: 2
 Homepage: https://github.com/embree/embree
 Description: High Performance Ray Tracing Kernels.
 Build-Depends: tbb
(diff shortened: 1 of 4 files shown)


NancyLi1013 (approved):

NancyLi1013:
Hi @chrismile 

Thanks for your fixing. Have you tested the feature checks added in this PR?

chrismile:
Yes, I have tested the feature checks. Once using a manually patched version of the .cmake files, and once using 'vcpkg install embree3' after adding the commit in this PR. I can confirm that a program linking to Embree with AVX512SKX compiles and runs successfully. I can't directly test whether AVX512KNL works, as I do not have access to a machine supporting it. But if this PR successfully fixes the problems with AVX512SKX, then there should be no reason why AVX512KNL shouldn't work.

NancyLi1013 (approved):

PR #17716, opened 2021-05-07
[avcpp] Add new port

barcharcraz:
Adds the AvCpp library, it's static only because they don't export their dynamic libraries.

### DIFF
--- a/ports/avcpp/0001-remove-problematic-compound-literal.patch
+++ b/ports/avcpp/0001-remove-problematic-compound-literal.patch
@@ -1,0 +1,25 @@
+From 8790ff3520bf2648ff2b5b436686878ec8d37ff3 Mon Sep 17 00:00:00 2001
+From: Charles Barto <barto.charlie@gmail.com>
+Date: Thu, 6 May 2021 17:47:23 -0700
+Subject: [PATCH] remove problematic compound literal
+
+---
+ src/avutils.h | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/avutils.h b/src/avutils.h
+index 29c4ae0..3bfaf82 100644
+--- a/src/avutils.h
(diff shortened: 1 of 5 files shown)


barcharcraz:
They don't export any symbols from their dynamic libraries that is.

ports/avcpp/0001-remove-problematic-compound-literal.patch
```diff
@@ -0,0 +1,25 @@
+From 8790ff3520bf2648ff2b5b436686878ec8d37ff3 Mon Sep 17 00:00:00 2001
+From: Charles Barto <barto.charlie@gmail.com>
+Date: Thu, 6 May 2021 17:47:23 -0700
+Subject: [PATCH] remove problematic compound literal
+
+---
+ src/avutils.h | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
```
PhoebeHui, line 8: Can we remove the above lines so that the patch file is minimized? Could you also create an issue or PR about the patch to Upstream as well?

PhoebeHui (changes requested):
@barcharcraz, thanks for your contribution!

ports/avcpp/0001-remove-problematic-compound-literal.patch
PhoebeHui, line 8: Can we remove the above lines so that the patch file is minimized? Could you also create an issue or PR about the patch to Upstream as well?

PhoebeHui:
@barcharcraz, could you format the vcpkg.json file? you can simply apply the following patch:
[format.diff.zip](https://github.com/microsoft/vcpkg/files/6444934/format.diff.zip)

barcharcraz:
applied comments.

PhoebeHui (approved):
Thanks for your updates!

PR #17727, opened 2021-05-07
[docs] Replace 'git-tree' in fs registry docs with 'path'

ras0219-msft:
@strega-nil-ms @strega-nil


strega-nil-ms:
Thanks robert, sorry about this :)

PR #17706, opened 2021-05-06
[vcpkg] Update cmake to 3.20.2

Neumann-A:
[x] Getting hashes from CI errors

Fixes: 
Qt autogen hangs in CI

closes: #17612

PR #17747, opened 2021-05-08
[libxml2] Improve pc file

dg0yt:
- #### What does your PR fix?  
  The pc file did not list any extra libraries needed for static linking. This PR adds them, e.g. for x64-linux-dbg:  
  `Libs.private:  -pthread    -lm  -L/vcpkg/installed/x64-linux/debug/lib/pkgconfig/../../lib -llzmad -pthread -lz`
  This ~helps with~ fixes the libxml2 test in gdal `configure`.

  In addition, the PR also shortcuts a copying step introduced in my previous PR.

  all, no.

  yes

  yes

- ### Remarks

  - Similar configuration is done in upstream HEAD's CMakeLists.txt. CC @JackBoosY 
  - Some requirements are collected via `x_vcpkg_pkgconfig_get_modules`, in order to easily to resolve recursively and with debug postfixes as needed. CC @Neumann-A

### DIFF
--- a/ports/libxml2/CMakeLists.txt
+++ b/ports/libxml2/CMakeLists.txt
@@ -166,6 +166,13 @@ install(TARGETS libxml2
 )
 
 # pkgconfig
+if(NOT Iconv_IS_BUILT_IN)
+    set(ICONV_LIBS "-liconv")
+endif()
+if(UNIX)
+    set(M_LIBS "-lm")
+    set(THREAD_LIBS "-pthread")
+endif()
 set(prefix ${CMAKE_INSTALL_PREFIX})
 set(exec_prefix "\${prefix}")
 set(libdir ${prefix}/${CMAKE_INSTALL_LIBDIR})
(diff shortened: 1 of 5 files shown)


dg0yt:
Probably there is also a dependency on CMAKE_DL_LIBS, given that HAVE_DLOPEN is defined for linux and osx. Not sure if this desired for static builds at all.

dg0yt:
x64_windows_static_md: paraview: `FAILED: Qt/Components/pqComponents_autogen/timestamp` - seems unrelated.

Neumann-A:
> x64_windows_static_md: paraview: FAILED: Qt/Components/pqComponents_autogen/timestamp - seems unrelated.
@dg0yt: Fixed by #17706. Is a CMake autogen regression

JackBoosY (approved):

dg0yt:
Thanks for reviewing.

After submitting the PR, I started wondering whether this usage of `x_vcpkg_pkgconfig_get_modules` isn't a anti-pattern: 
The result is only used to insert `Libs` into the pc file. This result can probably also be achieved just by adding corresponding `Requires:` (or `Requires.private:` ?) modules to the the pc file template. This would avoid the expensive extra acquisition and run of pkg-config when building the port.

dg0yt:
> This result can probably also be achieved just by adding corresponding `Requires:` (or `Requires.private:` ?) modules to the the pc file template.

Confirmed. I will update the PR unless someone objects.

dg0yt:
> adding corresponding `Requires:` (or `Requires.private:` ?) modules to the the pc file template.

Done: `Requries.private`.

dg0yt:
#17945 will also solve the pc file problem. If it is ready for merge soon, this PR becomes obsolete.

JackBoosY (approved):

JackBoosY:
I prefer to merge this PR first because my PR needs to update cmake to the latest.

PR #17769, opened 2021-05-10
[tre, libmagic] Windows + mingw support

### LINKED ISSUE #11832: [libmagic] Clean build failure on Windows
**Host Environment**
- OS: Windows 10 64-bit
- Compiler: not applicable

**To Reproduce**
```
git clean -fXd
.\bootstrap-vcpkg.bat
.\vcpkg.exe install libmagic
```

**Failure logs** 
```
CMake Error at scripts/cmake/vcpkg_execute_required_process.cmake:72 (message):
    Command failed: D:/Projects/GitHub/vcpkg/downloads/tools/msys2/msys64/usr/bin/bash.exe --noprofile --norc --debug -c "autoreconf -vfi"
    Working Directory: D:/Projects/GitHub/vcpkg/buildtrees/libmagic/src/9852e25398-4363addce8/
    Error code: 127
      D:\Projects\GitHub\vcpkg\buildtrees\libmagic\autoconf-x64-windows-err.log

  scripts/cmake/vcpkg_configure_make.cmake:349 (vcpkg_execute_required_process)
  ports/libmagic/portfile.cmake:9 (vcpkg_configure_make)
  scripts/ports.cmake:76 (include)
```
[autoconf-x64-windows-err.log](https://github.com/microsoft/vcpkg/files/4744721/autoconf-x64-windows-err.log)

**Additional context**
I believe it's a defect in `vcpkg_configure_make`, since the necessity of `autoconf` gets decided too late for it to be acquired.

Here's where decision to use `autoconf` is made:
https://github.com/microsoft/vcpkg/blob/27c6c9420277c533cc4333bdd880f9862012cb72/scripts/cmake/vcpkg_configure_make.cmake#L322-L332
And here's where it gets acquired:
https://github.com/microsoft/vcpkg/blob/27c6c9420277c533cc4333bdd880f9862012cb72/scripts/cmake/vcpkg_configure_make.cmake#L154-L161

@Neumann-A, could you, please, take a look?

longnguyen2004:
After a year...

**Describe the pull request**

- #### What does your PR fix?  
  Fixes #11832 
  Also fixes a small bug in `vcpkg_configure_make` where `uuid` isn't removed on mingw triplets.
  Windows + mingw triplets, Yes

  Yes

  Yes


scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -561,10 +561,10 @@ function(vcpkg_configure_make)
     if(_VCPKG_TRANSFORM_LIBS)
         set(_lprefix "-l")
         list(TRANSFORM ALL_LIBS_LIST REPLACE "(.dll.lib|.lib|.a|.so)$" "")
+        list(TRANSFORM ALL_LIBS_LIST REPLACE "^(${_lprefix})" "")
         if(VCPKG_TARGET_IS_WINDOWS)
             list(REMOVE_ITEM ALL_LIBS_LIST "uuid")
         endif()
-        list(TRANSFORM ALL_LIBS_LIST REPLACE "^(${_lprefix})" "")
```
dg0yt, line 567: IIRC this is what I initially proposed in #17137, but it didn't seem to find acceptance.

longnguyen2004: So now we have 2 different approaches to the same problem, let's see which one gets accepted first.
Since on MSVC triplets, `uuid.lib` gets stripped entirely, I'm doing the same with mingw.

longnguyen2004:
I can't seem to get the diff artifact, can someone help me?
https://dev.azure.com/vcpkg/public/_build/results?buildId=53405&view=artifacts&pathAsName=false&type=publishedArtifacts

longnguyen2004:
Also, now that we have host dependencies, theoretically `libmagic` can be cross compiled as well, but that's for another day.

ports/libmagic/0001-Use-libtre.patch
```diff
@@ -0,0 +1,53 @@
...
+diff --git a/configure.ac b/configure.ac
+index 02eac8f..82d76d3 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -185,7 +185,7 @@ if test "$enable_libseccomp" != "no"; then
+     AC_CHECK_LIB(seccomp, seccomp_init)
+ fi
+ if test "$MINGW" = 1; then
+-  AC_CHECK_LIB(gnurx,regexec,,AC_MSG_ERROR([libgnurx is required to build file(1) with MinGW]))
++  AC_CHECK_LIB(tre,tre_regexec,,AC_MSG_ERROR([libtre is required to build file(1) with MinGW]))
```
Neumann-A, line 21: `AC_SEARCH_LIBS` instead?

ports/libmagic/0001-Use-libtre.patch
```diff
@@ -0,0 +1,53 @@
...
+diff --git a/src/file.h b/src/file.h
+index c548e97..299ac0c 100644
+--- a/src/file.h
++++ b/src/file.h
+@@ -79,7 +79,7 @@
+ #include <stdio.h>	/* Include that here, to make sure __P gets defined */
+ #include <errno.h>
+ #include <fcntl.h>	/* For open and flags */
+-#include <regex.h>
++#include <tre/regex.h>
```
Neumann-A, line 47: pass `-I"<vcpkginclude>/tre"` instead of patching

ports/libmagic/0002-Change-zlib-lib-name-to-match-CMake-output.patch
```diff
@@ -0,0 +1,39 @@
...
+diff --git a/configure.ac b/configure.ac
+index 82d76d3..5923900 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -173,7 +173,7 @@ AC_REPLACE_FUNCS(getopt_long asprintf vasprintf strlcpy strlcat getline ctime_r
+ 
+ dnl Checks for libraries
+ if test "$enable_zlib" != "no"; then
+-  AC_CHECK_LIB(z, gzopen)
++  AC_CHECK_LIB(zlib, gzopen)
```
Neumann-A, line 19: `AC_SEARCH_LIBS` ! due to debug suffix on windows. Search for `z zlib zlibd`

longnguyen2004: We don't have to worry about which one gets found first, since only one will be available in each build.

Neumann-A: as longnguyen2004 said. This is not cmake where we have to deal with both prefixes/libdirs being visible.

dg0yt: For z, fair enough. But not in general. This is part of the "gdal completely broken on Linux" story. On Linux, there is always a system path, in addition to the debug path. And generally there is no control about the user environment (manifest mode, registries, overlay ports).

longnguyen2004: @dg0yt We can probably uses `-nostdinc`/`-nostdlib` and manually add back in the necessary include and library paths, but that gets tedious fast.

dg0yt:
> `file` requires a native `file` of the same version to compile the magic file, so we have to add `file` itself as a host dependency (is it possible?)

Maybe a host dependency of a non-core default feature on a non-default feature?

No way to pass the tool path explicitly instead of modifying `PATH`?

ports/libmagic/portfile.cmake
```diff
@@ -34,7 +34,9 @@ vcpkg_configure_make(
     SOURCE_PATH ${SOURCE_PATH}
 )
 
-if(VCPKG_TARGET_IS_WINDOWS AND VCPKG_LIBRARY_LINKAGE STREQUAL dynamic)
+if(NOT HOST_TRIPLET STREQUAL TARGET_TRIPLET)
```
JackBoosY, line 37:
```suggestion
if(VCPKG_CROSSCOMPILING)
```

JackBoosY (changes requested):

longnguyen2004:
@Neumann-A Can you check `x86-windows` and see if there's a workaround?
The build system detects the correct `file`, but `file --version` returns `D:\installed\x64-windows\tools\libmagic\bin\file.exe-5.40` for some reason, while the build expect `file-x.xx`, thus it fails to detect the version.

ports/libmagic/portfile.cmake
longnguyen2004: I'll merge this later, for now I need to find out what's going on with the build process.

Neumann-A:
> This line only checks for '/' in argv[0], which is not the case on Windows (?)

Yeah probably need to fix that + adding ${EXEEXT} in the Makefile doing the check. 
Also install the magic.mgc file into `tools/libmagic/share/misc` so that the tool automatically picks it up

longnguyen2004:
Eh...`git rebase -i` messes everything up :)

strega-nil-ms (approved):
This looks good to me, just needs a minor change

ports/libmagic/portfile.cmake
```diff
@@ -1,22 +1,66 @@
...
     SOURCE_PATH ${SOURCE_PATH}
 )
-vcpkg_install_make()
-vcpkg_copy_pdbs()
+
+if(VCPKG_CROSSCOMPILING)
+    vcpkg_add_to_path(PREPEND ${CURRENT_HOST_INSTALLED_DIR}/tools/libmagic/bin)
+elseif(VCPKG_TARGET_IS_WINDOWS AND VCPKG_LIBRARY_LINKAGE STREQUAL dynamic)
+    set(EXTRA_ARGS "ADD_BIN_TO_PATH")
+endif()
```
Neumann-A, line 43: Why would I want to run against debug tools in a native build? 
Wouldn't `vcpkg_add_to_path(PREPEND ${CURRENT_HOST_INSTALLED_DIR}/tools/libmagic/bin)` be enough?

ports/libmagic/portfile.cmake
```diff
@@ -1,22 +1,66 @@
...
+
+vcpkg_install_make(${EXTRA_ARGS})
+vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin)
+vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/tools/${PORT}/debug/bin)
```
Neumann-A, line 47: Is there a reason to keep the debug tools?

ports/libmagic/portfile.cmake
strega-nil-ms: no, because the tools don't exist in `${CURRENT_HOST_INSTALLED_DIR}` yet.

ports/libmagic/portfile.cmake
longnguyen2004: Debug-only builds, for instance.

ports/libmagic/portfile.cmake
longnguyen2004: In native builds, after building `file.exe`, it'll use that to compile the magic file. So we need `ADD_BIN_TO_PATH` in order to let it find the correct DLL.

ports/libmagic/portfile.cmake
dg0yt: Debug-only builds are generally broken for many ports. But did you try this with a release-only build?

longnguyen2004: Not yet, can you test for me? My computer is currently busy.

dg0yt: > Not yet, can you test for me? My computer is currently busy.

I think it is okay, the function uses globs which should produce nothing in that case.

dg0yt:
On mingw, the build failed:
~~~
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ./.libs/libmagic.a(compress.o):compress.c:(.text+0xae): undefined reference to `ioctl'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ./.libs/libmagic.a(compress.o):compress.c:(.text+0x108): undefined reference to `__imp_select'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ./.libs/libmagic.a(compress.o):compress.c:(.text+0x12e): undefined reference to `ioctl'
collect2.exe: error: ld returned 1 exit status
~~~

longnguyen2004:
@dg0yt Try applying this MSYS2 patch https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-file/0001-file-5.38-mingw-build.patch

longnguyen2004:
@dg0yt Is the problem solved?

dg0yt:
> @dg0yt Is the problem solved?

Yes, mingw build succeeds now. :+1:

longnguyen2004:
Now that #17137 is merged, I'll take that fix off of my PR.

JackBoosY (approved):

strega-nil-ms:
Alright, great, thanks for the awesome work @longnguyen2004 :)

PR #17780, opened 2021-05-10
Strip "-l" suffix before removing "uuid" from libraries

375gnu:
**Describe the pull request**

My builds using mingw64 were failing due to missing `libuuid` while it's supposed that on windows it's removed from a list of libraries. The problem is that when the removal is invoked `ALL_LIBS_LIST` contains names with `-l` suffix so the actual value is `-luuid` and it's not removed. This small change fixes this issue.


JackBoosY:
cc @longnguyen2004  @Neumann-A  for review this PR.

longnguyen2004 (approved):

strega-nil-ms:
Closed as #17137 has the same change and was merged.

Thanks @375gnu !

PR #16706, opened 2021-03-15
[vcpkg_fixup_cmake_targets] Fix up OSX system development path

JackBoosY:
In OSX, some cmake configuration files contain the absolute path of the system development kits, such as:
```cmake
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/IOKit.framework

\$<\$<PLATFORM_ID:Darwin>:/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/CoreFoundation.framework>
```
And they should be:
```cmake
-framework IOKit

\$<\$<PLATFORM_ID:Darwin>:-framework CoreFoundation>
```
Fix this.

Fixes #16259.

PR #17790, opened 2021-05-11
[curl] Fix linker flags in curl-config and libcurl.pc for osx

dg0yt:
- #### What does your PR fix?  
  Apply same library path fixup as for Linux, plus:
  Instead of just listing the absolute framework location with no flags, the flags are now:
  `-framework CoreFoundation -framework Security`

  Needed to to fix gdal for osx (#9068).

  unchanged (but targeting osx), no

  yes

  yes

### DIFF
--- a/ports/curl/0020-fix-pc-file.patch
+++ b/ports/curl/0020-fix-pc-file.patch
@@ -5,7 +5,7 @@ diff -urw a/CMakeLists.txt b/CMakeLists.txt
        message(WARNING "Bad lib in library list: ${_libname}")
        continue()
      endif()
-+  elseif(CMAKE_VERSION VERSION_GREATER "3.8.99" AND _lib MATCHES "^(.*)/([^/]*)[.]framework$")
++  elseif(_lib MATCHES "^(.*)/([^/]*)[.]framework$")
 +    if(CMAKE_MATCH_1 IN_LIST CMAKE_C_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES)
 +      set(_lib "-framework ${CMAKE_MATCH_2}")
 +    else()
--- a/ports/curl/0020-fix-pc-file.patch
+++ b/ports/curl/0020-fix-pc-file.patch
@@ -1,0 +1,16 @@
+diff -urw a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt	2021-05-10 20:46:52.204346972 +0200
++++ b/CMakeLists.txt	2021-05-11 19:39:00.065235266 +0200
+@@ -1482,6 +1482,12 @@
+       message(WARNING "Bad lib in library list: ${_libname}")
+       continue()
+     endif()
++  elseif(CMAKE_VERSION VERSION_GREATER "3.8.99" AND _lib MATCHES "^(.*)/([^/]*)[.]framework$")
++    if(CMAKE_MATCH_1 IN_LIST CMAKE_C_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES)
++      set(_lib "-framework ${CMAKE_MATCH_2}")
++    else()
++      set(_lib "-framework ${_lib}")
++    endif()
+   endif()
+   if(_lib MATCHES ".*/.*" OR _lib MATCHES "^-")
+     set(LIBCURL_LIBS          "${LIBCURL_LIBS} ${_lib}")
--- a/ports/curl/portfile.cmake
+++ b/ports/curl/portfile.cmake
@@ -13,6 +13,7 @@ vcpkg_from_github(
         0010_fix_othertests_cmake.patch
         0011_fix_static_build.patch
         0012-fix-dependency-idn2.patch
+        0020-fix-pc-file.patch
 )
 
 string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" CURL_STATICLIB)
--- a/ports/curl/vcpkg.json
+++ b/ports/curl/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "curl",
   "version": "7.74.0",
-  "port-version": 7,
+  "port-version": 8,
   "description": "A library for transferring data with URLs",
   "homepage": "https://github.com/curl/curl",
   "dependencies": [
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -1546,7 +1546,7 @@
     },
     "curl": {
       "baseline": "7.74.0",
-      "port-version": 7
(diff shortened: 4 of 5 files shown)


cenit:
https://github.com/microsoft/vcpkg/pull/16706 might be a more generic and effective solution maybe?
Instead of a patch for a specific port, that's a more complete solution valid for all ports

dg0yt:
@cenit This patch is not targeting use with CMake at all, but use with libcurl.pc or with curl-config. I don't think #16706 will have any impact. At best, I could reuse the regex pattern if desired.

However, I would agree that some of the instructions from the portfile should perhaps be moved to vcpkg_fixup_pkgconfig. But this is out of scope for now.
(Note that I already created #17748 wrt port curl.)

cenit:
> Instead of just listing the absolute framework location with no flags, the flags are now:
-framework CoreFoundation -framework Security

I was referring to this.
#16706 will remove absolute paths from system frameworks and make flags like yours. What I meant is that the PR above will make https://github.com/microsoft/vcpkg/pull/17790/commits/77dfc1986afba076dc1902b0bd669d931ee15de8 unnecessary

cenit:
> Instead of just listing the absolute framework location with no flags, the flags are now:
-framework CoreFoundation -framework Security

I was referring to this ☝🏻 
#16706 will remove absolute paths from system frameworks and make flags like yours. Isn't ports/curl/0020-fix-pc-file.patch fixing CMake config file too? Maybe I just misunderstood and it's fixing pkgconfig, sorry

cenit:
but ok for your work, and sorry for the unnecessary troubles with my comments :)

dg0yt:
This PR intentionally does nothing about the missing debug variant of `curl-config`, or its missing executable bits. Related questions are already waiting for answers (#17607, #17616).

For the gdal port, I try to use `${PKG_CONFIG} libcurl` instead of `curl-config`, to fix debug builds. But without this PR, this breaks osx. (And CI didn't complain about curl-config before because potentially consuming ports ignore the non-executable script.)

ports/curl/portfile.cmake
```diff
@@ -155,7 +156,7 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "debug")
         vcpkg_replace_string(${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-dbg/libcurl.pc " -loptimized" "")
         vcpkg_replace_string(${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-dbg/libcurl.pc " -ldebug" "")
 
-    elseif(VCPKG_TARGET_IS_LINUX)
+    elseif(VCPKG_TARGET_IS_LINUX OR VCPKG_TARGET_IS_OSX)
```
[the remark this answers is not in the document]
dg0yt: After the last edit, I had an even simpler idea: `else()`  ;-)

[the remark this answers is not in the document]
dg0yt: Done.

JackBoosY (approved):

dg0yt:
This PR uses `CMAKE_SYSTEM_FRAMEWORK_PATH`, other than in https://github.com/JackBoosY/vcpkg/pull/2 (my proposal to the cmake config problem with the frameworks) where I used `CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES`. On my MBA, the values are as follows:
~~~
-- CMAKE_SYSTEM_FRAMEWORK_PATH: ~/Library/Frameworks;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/Library/Frameworks;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/Network/Library/Frameworks;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks;/Library/Developer/CommandLineTools/Library/Frameworks;/Library/Developer/CommandLineTools/Library/Frameworks;/Library/Frameworks;/Network/Library/Frameworks;/System/Library/Frameworks
-- CMAKE_C_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES: /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks
-- CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES: /Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks
~~~

In the end, it might make sense to use the same variable for pc files and cmake config, and possibly move these changes to `vcpkg_fixup_pkgconfig`.

dg0yt:
- Rebased on master.
- Switched to `CMAKE_C_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES`.

JackBoosY (approved):

ports/curl/0020-fix-pc-file.patch
```diff
@@ -0,0 +1,16 @@
+diff -urw a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt	2021-05-10 20:46:52.204346972 +0200
++++ b/CMakeLists.txt	2021-05-11 19:39:00.065235266 +0200
+@@ -1482,6 +1482,12 @@
+       message(WARNING "Bad lib in library list: ${_libname}")
+       continue()
+     endif()
++  elseif(CMAKE_VERSION VERSION_GREATER "3.8.99" AND _lib MATCHES "^(.*)/([^/]*)[.]framework$")
```
strega-nil-ms, line 8: I don't understand why this is checking for a cmake version..? what's going on here?

dg0yt: a) The following code uses `CMAKE_MATCH_<n>` which is new 3.9
b) I'm not aware that minimum version of CMake for vcpkg is *documented*. However, I'm aware that `VCPKG_FORCE_SYSTEM_BINARIES` is documented.
So I can not make assumptions about the cmake version which is used.

strega-nil-ms: we require 3.20 as of now.

dg0yt: I removed the version check now. IIRC now it was also added with upstreaming in mind.

dg0yt:
Ping @strega-nil-ms

dg0yt:
Ping @strega-nil-ms. This is blocking gdal improvement (#9068) and curl updates.

ras0219-msft:
This looks like a fine intermediate fix until we can deploy a more general version into `vcpkg_fixup_pkgconfig()`. Thanks for the PR!

Issue #2021, opened 2017-10-20
Header search path is not set when installing static triplets


olilarkin:
If I only install the openssl:x64-windows-static port, I cannot  #include <openssl/sha.h> in a new commandline project. 

the header search path is not set. If I also  install the openssl:x64-windows version, it is. This is true for all ports as far as i can tell

olilarkin:
if you have both static and dll versions installed, how do you specify what get's linked?

twMr7:
The following docs shall help:
[Triplet Files](https://github.com/Microsoft/vcpkg/blob/master/docs/users/triplets.md)
[Triplet Selection in MSBuild](https://github.com/Microsoft/vcpkg/blob/master/docs/users/integration.md#triplet-selection)

JackBoosY:
Hi @olilarkin , as @twMr7 said, have you solved this problem?

olilarkin:
this was an awful long time ago and i can't remember even what it was for. closing

PR #17856, opened 2021-05-12
[expat] Update to 2.4.1

### LINKED ISSUE #17842: [expat] update to 2.4.1, please (CVE-2013-0340)
# Library name
expat

# New version number
2.4.1

# Other information that may be useful (release notes, etc...)
The current expat port [packages a commit](https://github.com/microsoft/vcpkg/blob/3b4dd085b0f5b410d9f587b3c9f7ad08a07449aa/ports/expat/portfile.cmake#L4) halfway between 2.2.9 and 2.2.10 while upstream is at 2.4.1 [with security fixes](https://blog.hartwork.org/posts/cve-2013-0340-billion-laughs-fixed-in-expat-2-4-0/). It would be great if you could update to 2.4.1. Thank you!

Best, Sebastian

JonLiu1993:
Fix #17842 

Update expat to the latest version 2.4.1

Note: no feature need to test


ports/expat/portfile.cmake
```diff
@@ -1,11 +1,11 @@
+set (EX_VERSION 2.3.0)
+
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO libexpat/libexpat
-    REF c092d40c300c6d219cb3b111932a824022265370 #Head from commit 2020-08-18
-    SHA512 5a5d41b500f5602a32aea8f4e15593e639206bb3f97553497e80b2975360cac88ac90386f5efc11728614f24bbb620fb908a3c8ca71c9e7b312f6157b2477afe
+    REF 2d58706dad5a6b996ee143790f812bd0af4acde2 #v2.3.0
```
hartwork, line 6: - Would tag name `R_2_3_0` work here as well?
- Would accessing the release archive tarball instead of Git work?

hartwork (commented):
@JonLiu1993 thanks for picking this up, very nice! :+1: :+1:

hartwork:
@JonLiu1993 any idea what the fontconfig error at https://dev.azure.com/vcpkg/public/_build/results?buildId=53511&view=logs&j=c7e67a25-2047-5d01-7913-57de0573f534&t=0e07d234-e12f-5893-a7f1-2767a24f64ae&l=2815 is about? Is the CMake error log available somewhere?

dg0yt:
The logs are in the artifacts of that build.
Fontconfig doesn't find expat.  Since fontconfig is based autoconf, not cmake, I suspect that the pc files from expat are no longer working. You removed patches related to pkg-config on Windows.

hartwork:
> The logs are in the artifacts of that build.

I don't see any artifacts when I click on artifacts. Please not that I'm new to Azure and vcpkg.

> Fontconfig doesn't find expat. Since fontconfig is based autoconf, not cmake, I suspect that the pc files from expat are no longer working. The PR removed patches related to pkg-config on Windows.

That's a good hint. There were patches upstream for vcpkg in https://github.com/libexpat/libexpat/pull/414 precisely, so `-DEXPAT_BUILD_PKGCONFIG=ON` seems needed in vcpkg for Expat 2.3.0 now, regarding the `if(1)` part of the patch.
Regarding the `_EXPAT_OUTPUT_NAME` part of the patch, upstream code https://github.com/libexpat/libexpat/blob/0ee4a6a94e50fef516addcca066349b365614e5c/expat/CMakeLists.txt#L374-L375 makes me guess that that part of the patch is probably no longer needed.

To summarize, I would try adding `-DEXPAT_BUILD_PKGCONFIG=ON` and stick to dropping the patch and see how far that takes us.

CC @JonLiu1993

dg0yt:
> I don't see any artifacts when I click on artifacts. Please not that I'm new to Azure and vcpkg.

1. Your link pointed to a particular job in the build. You can click on the build number in the top line ("20210512.13") or use the back button before "Jobs in run \#2021..."
2. Now in the overview of the build, there is "12 published", which refers to, and links to, the artifacts. 
3. In the "failure logs for x64-windows" line, click the vertical three dots on the right side, than "Download artifacts".

hartwork:
> > I don't see any artifacts when I click on artifacts. Please not that I'm new to Azure and vcpkg.
> 
>     1. Your link pointed to a particular job in the build. You can click on the build number in the top line ("20210512.13") or use the back button before "Jobs in run # 2021..."
> 
>     2. Now in the overview of the build, there is "12 published", which refers to, and links to, the artifacts.
> 
>     3. In the "failure logs for x64-windows" line, click the vertical three dots on the right side, than "Download artifacts".

Thank you!

I would have never seen those three dots without your mention, given how far they are away in a full screen browser window…
In log file `config-x64-windows-dbg-out.log` I find confirmation that fontconfig's configure script does indeed fail to find Expat:
```
checking for EXPAT... no
checking expat.h usability... yes
checking expat.h presence... yes
checking for expat.h... yes
checking for XML_SetDoctypeDeclHandler... no
```

hartwork:
A quick note that I have updated issue #17842 to be about [Expat 2.4.1 with security fixes](https://blog.hartwork.org/posts/cve-2013-0340-billion-laughs-fixed-in-expat-2-4-0/) for CVE-2013-0340 now.

JonLiu1993:
@hartwork , @dg0yt .I am sorry to reply to you so late, I have seen your discussions and suggestions. And I will upgrade the issue assign of expat to 2.4.1 to myself. I will seriously consider your suggestions, thank you

hartwork:
@JonLiu1993 that is great news, thank you!

hartwork:
@JonLiu1993 any news?

ports/expat/portfile.cmake
JonLiu1993: @hartwork ,
R_2_3_0 and commit all work here,but we often use commit.
the release archive tarball also accessing

hartwork:
@JonLiu1993 I overlooked your changes updating to 2.4.1 — sorry! It seems like fontconfig is still failing. I wonder if that because of the…
```diff
+set(_EXPAT_OUTPUT_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${BUILD_TYPE_UPPER}_POSTFIX})
```
…bits now gone. We could either try reintroducing the patch cut down to that line or see if fontconfig is the better place to patch. What do you think?

JonLiu1993:
> @JonLiu1993 I overlooked your changes updating to 2.4.1 — sorry! It seems like fontconfig is still failing. I wonder if that because of the…
> 
> ```diff
> +set(_EXPAT_OUTPUT_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${BUILD_TYPE_UPPER}_POSTFIX})
> ```
> 
> …bits now gone. We could either try reintroducing the patch cut down to that line or see if fontconfig is the better place to patch. What do you think?

I made a mistake. After checking, I found that I deleted the required patch. I will re-add the patch and submit it.thanks for your reminder

dg0yt:
@hartwork Upstream must make sure that the pc file is correct at least for the build types where it modifies `CMAKE_<CONFIG>_POSTFIX`. And this modification happens for "debug" on WIN32:

https://github.com/libexpat/libexpat/blob/a28238bdeebc087071777001245df1876a11f5ee/expat/CMakeLists.txt#L342-L346

And that's why the patch remains necessary at the moment.

(Ideally upstream would always care about `CMAKE_<CONFIG>_POSTFIX` because it could be set by the user for custom config types.)

hartwork:
I'm happy to fix things upstream but I'll first need to understand things in detail. (Build system issues like this one was why I opened #17842 before the release of 2.4.0 but now things will have to go into 2.4.2.)

What I don't understand is:
- from a CMake point of view, is `WIN32 AND MINGW` what we have here?
- what is the expected postfix — is this about postfix "d" or postfix "w" or ..?

ports/expat/pkgconfig.patch
```diff
@@ -1,15 +1,13 @@
...
+diff --git a/expat/CMakeLists.txt b/expat/CMakeLists.txt
+index cd12a99..860548b 100644
+--- a/expat/CMakeLists.txt
++++ b/expat/CMakeLists.txt
+@@ -419,6 +419,8 @@ expat_install(FILES lib/expat.h lib/expat_external.h DESTINATION ${CMAKE_INSTALL
+ #
+ # pkg-config file
+ #
++
++set(_EXPAT_OUTPUT_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${BUILD_TYPE_UPPER}_POSTFIX})
```
hartwork, line 10: @JonLiu1993 I notice that fontconfig is still failing and I think I might know why:
existing [line 394](https://github.com/libexpat/libexpat/blob/a28238bdeebc087071777001245df1876a11f5ee/expat/CMakeLists.txt#L394) (`set_property(TARGET expat PROPERTY OUTPUT_NAME "${_EXPAT_OUTPUT_NAME}")`) propagates the current value of `_EXPAT_OUTPUT_NAME` to CMake target `expat` and the patch above adjusts the value of `_EXPAT_OUTPUT_NAME` _after_ that, after it has been copied, i.e. too late. Does that make sense?

ports/expat/portfile.cmake
```diff
@@ -1,11 +1,13 @@
...
     OUT_SOURCE_PATH SOURCE_PATH
     REPO libexpat/libexpat
-    REF c092d40c300c6d219cb3b111932a824022265370 #Head from commit 2020-08-18
-    SHA512 5a5d41b500f5602a32aea8f4e15593e639206bb3f97553497e80b2975360cac88ac90386f5efc11728614f24bbb620fb908a3c8ca71c9e7b312f6157b2477afe
+    REF a28238bdeebc087071777001245df1876a11f5ee #v2.4.1
+    SHA512 2c22f2dfab50644637a7777229849c91630ae8c50683df6ad6409d473690ce01b59ce9f98e66b0aeac8c650507f04edbb5d9738130c88b87bbc1adb7831c22a9
     HEAD_REF master
     PATCHES
-        pkgconfig.patch
+        pkgconfig.patch
```
hartwork, line 10: I think there that extra space was added by mistake

ports/expat/pkgconfig.patch
hartwork: PS: I'm also wondering where variable `BUILD_TYPE_UPPER` is set. With Expat I only find `_EXPAT_BUILD_TYPE_UPPER` but it also set further down.

dg0yt:
> What I don't understand is:
> 
>  * from a CMake point of view, is `WIN32 AND MINGW` what we have here?

I don't understand the question. Basically there is an issue on WIN32 targets, including MINGW.

>  * what is the expected postfix — is this about postfix "d" or postfix "w" or ..?

I guess this cannot be answered here. vcpkg policy is to follow upstream, assuming that upstream knows what it users want or are used to (mind users that don't use cmake config or pc files).
For the record, some upstreams use the same name for all toolchains on WIN32, but others prefer to treat MINGW as POSIX.

But I don't think the answer is needed. You are not asked to change the name. 
The point is, expat provides wrong information about the name in the pc files:
- in particular for build type Debug, due to its own POSTFIX settings (breaking the build of fontconfig here)
- in general for users which set `CMAKE_<CONFIG>_POSTFIX`.

Example from a quick build of expat 2.4.1 on Linux with mingw toolchain, CMAKE_BUILD_TYPE=Debug
~~~
$ ls <REMOVED>/install/lib
cmake  libexpatd.dll.a  pkgconfig
$ cat <REMOVED>/install//lib/pkgconfig/expat.pc 
prefix=<REMOVED>/install
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: expat
Version: 2.4.1
Description: expat XML parser
URL: http://www.libexpat.org
Libs: -L${libdir} -lexpat
Cflags: -I${includedir}
~~~
So the binary is `libexpatd.dll.a` but the pc file says `-lexpat`.

hartwork:
> So the binary is `libexpatd.dll.a`. The pc file says `-lexpat` when it should say `-lexpatd`.

@dg0yt I see, thanks!

hartwork:
@dg0yt @JonLiu1993 I'm still figuring this out. I have a draft pull request up at https://github.com/libexpat/libexpat/pull/495 now. I appreciate feedback and uncovering any remaining misunderstandings. Thanks :pray:

hartwork (approved):
@JonLiu1993 I understand the patch now and why it works — great to have things working now :tada: While a bit hacky, if the goal is _not_ to wait for Expat 2.4.2, this seems like a good short-term way forward to me. Thanks for your work on this topic! :pray:

ports/expat/pkgconfig.patch
```diff
@@ -1,15 +1,13 @@
...
++++ b/expat/CMakeLists.txt
+@@ -419,6 +419,8 @@ expat_install(FILES lib/expat.h lib/expat_external.h DESTINATION ${CMAKE_INSTALL
+ #
+ # pkg-config file
+ #
++string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_UPPER)
++set(_EXPAT_OUTPUT_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${BUILD_TYPE_UPPER}_POSTFIX})
+ if(EXPAT_BUILD_PKGCONFIG)
+     set(prefix ${CMAKE_INSTALL_PREFIX})
+     set(exec_prefix "\${prefix}")
```
NancyLi1013, line 13: Please keep the original format of EOL.

ports/expat/pkgconfig.patch
NancyLi1013: Are you still confused about this part? @hartwork

ports/expat/portfile.cmake
```diff
@@ -22,11 +24,12 @@ vcpkg_configure_cmake(
         -DEXPAT_BUILD_TESTS=OFF
         -DEXPAT_BUILD_TOOLS=OFF
         -DEXPAT_SHARED_LIBS=${EXPAT_LINKAGE}
+        -DEXPAT_BUILD_PKGCONFIG=ON
```
NancyLi1013, line 27:
```suggestion
```
The value of this option is decided by `_EXPAT_BUILD_PKGCONFIG_DEFAULT`. The value `_EXPAT_BUILD_PKGCONFIG_DEFAULT` varies from platforms. So we should respect this by upstream. 

https://github.com/libexpat/libexpat/blob/6a0584a59a963554ffc1d8dea98bfec25b354d44/expat/CMakeLists.txt#L119

https://github.com/libexpat/libexpat/blob/master/expat/CMakeLists.txt#L99-L103

ports/expat/portfile.cmake
```diff
@@ -22,11 +24,12 @@ vcpkg_configure_cmake(
...
 )
 
 vcpkg_install_cmake()
 
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/expat-2.2.9)
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/expat-${EX_VERSION})
 vcpkg_fixup_pkgconfig()
```
NancyLi1013, line 33: So this function is only needed on non Windows platform.

ports/expat/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "expat",
+  "version": "2.4.1",
```
NancyLi1013, line 3:
```suggestion
  "version-semver": "2.4.1",
```

NancyLi1013 (changes requested):

ports/expat/portfile.cmake
hartwork: If we'd rely on the default here, the reverse dependencies that need this file in vcpkg will miss it. The need for off-default behavior was the main point behind adding option `EXPAT_BUILD_PKGCONFIG` upstream, please see https://github.com/libexpat/libexpat/issues/413 for proof.

ports/expat/pkgconfig.patch
hartwork: > Are you still confused about this part? @hartwork

It's clear by now, because `BUILD_TYPE_UPPER` has been re-introduced here by commit 6b8f6aa708f21ab930f2052523ab33e8826e33c8.
For my upstream version of this change please see https://github.com/libexpat/libexpat/pull/495 if interested. Review is welcome.

ports/expat/portfile.cmake
Neumann-A: @NancyLi1013: NO! Always install PKG-CONFIG files! This way packages can correctly be consumed by e.g. meson/qmake/autotools. Maybe this should be added to the maintainers guidelines. If a port is installing pc files on any platform it should do so for every platform!

hartwork: @Neumann-A there is no need to shout, we can hear you

hartwork (approved):

NancyLi1013 (approved):

NancyLi1013:
Thanks for your help to review @hartwork and @dg0yt.

hartwork:
Is anything left to do before a merge?

strega-nil-ms:
Thanks @JonLiu1993 :)

PR #17883, opened 2021-05-12
[libbson/mongo-c-driver] Update to 1.17.5-1

### LINKED ISSUE #17751: new package name bson-1.0 and mongoc-1.0 not support in mongo c driver
There are two package name for bson and mongoc

The old one
```cmake
find_package(libbson-1.0)
find_package(libmongoc0-1.0)
```

The new one
```cmake
find_package(bson-1.0)
find_package(mongoc0-1.0)
```
If using the old package, always hit the warning:
```text
CMake Warning at /usr/local/lib/cmake/libbson-1.0/libbson-1.0-config.cmake:15 (message):
  This CMake target is deprecated.  Use 'mongo::bson_shared' instead.
  Consult the example projects for further details.
  CMakeLists.txt:6 (find_package)
```

[discuss](https://developer.mongodb.com/community/forums/t/confusion-about-using-libbson-in-cmake/105318)
>The libbson-1.0 package is an older package the required the consuming project to manually set each of the properties based on variables set by the package. This was not very CMake-like and so the new bson-1.0 package uses a proper CMake target. The libbson-1.0 package is deprecated and will be discontinued in a future release. So, you are better off to use the bson-1.0 package, which is planned to be supported long term.

However when install mongo c driver from vcpkg, CMake can't find the new package.

NancyLi1013:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #17751

Update these two ports to the latest version and also fix the usage.


ports/libbson/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: libbson
-Version: 1.16.1
-Port-Version: 2
+Version: 1.17.5-1
```
JackBoosY, line 2: Note this is the upstream published version number `1.17.5-1`.

ports/libbson/fix-uwp.patch
```diff
@@ -1,146 +1,16 @@
...
-index 7c479d8..d383424 100644
+index d708dc3..a63edf2 100644
 --- a/src/libbson/CMakeLists.txt
 +++ b/src/libbson/CMakeLists.txt
-@@ -226,6 +226,9 @@ set (HEADERS_FORWARDING
+@@ -230,6 +230,8 @@ set (HEADERS_FORWARDING
     ${PROJECT_SOURCE_DIR}/src/bson/forwarding/bson.h
  )
  
 +add_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
JackBoosY, line 9: Should be:
```cmake
if (MSVC)
    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
endif()
```

ports/libbson/portfile.cmake
```diff
@@ -42,11 +42,7 @@ vcpkg_copy_pdbs()
 
 set(PORT_POSTFIX "1.0")
 
-if (VCPKG_LIBRARY_LINKAGE STREQUAL static)
-    vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/libbson-static-${PORT_POSTFIX} TARGET_PATH share/bson-${PORT_POSTFIX})
-else()
-    vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/libbson-${PORT_POSTFIX} TARGET_PATH share/bson-${PORT_POSTFIX})
-endif()
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/bson-${PORT_POSTFIX} TARGET_PATH share/bson-${PORT_POSTFIX})
```
JackBoosY, line 45: Upstream behavor.

ports/libbson/usage
```diff
@@ -1,6 +1,8 @@
 The package libbson is compatible with built-in CMake targets:
```
JackBoosY, line 1: I think we should remove this usage now.

ports/mongo-c-driver/disable-static-when-dynamic-build.patch
```diff
@@ -1,61 +1,88 @@
-diff --git a/src/libmongoc/CMakeLists.txt b/src/libmongoc/CMakeLists.txt
```
JackBoosY, line 1: Please change this file's EOL.

ports/mongo-c-driver/usage
```diff
@@ -1,7 +1,8 @@
 The package mongo-c-driver is compatible with built-in CMake targets:
```
JackBoosY, line 1: Please remove this usage.

ports/mongo-c-driver/CONTROL
```diff
@@ -1,6 +1,6 @@
 Source: mongo-c-driver
```
JackBoosY, line 1: Please change to manifest.

ports/libbson/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: libbson
```
JackBoosY, line 1: Please switch to manifest.

JackBoosY (changes requested):

NancyLi1013:
In flavor of #18751.

PR #8455, opened 2019-10-04
[boringssl] Add new port

qis:
This is a new port for BoringSSL. As explained in https://github.com/microsoft/vcpkg/issues/8084#issuecomment-538029140, there are several issues I see with it:

1. It does not support UWP due to usage of desktop-only APIs like [RtlGenRandom](https://docs.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom).
2. It totally ignores `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` variables set by a MSVC toolchains.
3. When `CMakeLists.txt` is patched to honor flags set by a toolchain, `/EH` flags break compilation.
4. It uses its own gtest for tests which has the same `/EH` flags issues.
5. It uses Go for codegen.

It may be useful to people targeting Windows Desktop, Linux and possibly OSX who stick with the supported build system options.

It may be a good idea to patch out the tests as a workaround for the poorly designed `CMakeLists.txt`, or make tests optional.

PR #17949, opened 2021-05-15
[quictls] create a new port

luncliff:
### What does your PR fix?  

Create a port for https://github.com/quictls/openssl (fork of openssl/openssl)

Support https://github.com/microsoft/msquic/issues/1426.

#### Build check
MinGW, PowerPC, BSD is not in consideration for this port version.

* Windows (static only)
   * [x] `x64-windows`
   * [x] `x64-windows-static`
   * [x] `x64-windows-static-md`
   * [x] `x86-windows`
   * [x] `x86-windows-static`
   * [x] `x86-windows-static-md`
   * [x] `arm64-windows`
   * [x] `arm64-windows-static`
   * [x] `arm64-windows-static-md`
* UWP (static only)
   * [x] `x64-uwp`
   * [x] `x86-uwp`
   * [x] `arm64-uwp`
   * [x] `arm-uwp`
* Darwin
   * [x] `x64-osx`
   * [x] `arm64-osx`
* iOS
   * [x] `arm64-ios`
   * [x] `x64-ios`
* Linux
   * [x] `x64-linux`
* Android
   * [x] `arm64-android`
   * [x] `arm-android`
   * [x] `x64-android`
   * [x] `x86-android`

References

- https://github.com/quictls/openssl/tree/OpenSSL_1_1_1k+quic
    - ${SOURCE_PATH}/Configuration/README.design
    - ${SOURCE_PATH}/Configuration/10-main.conf

And see comments below :D

#### Import check

* `curl`: import via `find_package(OpenSSL REQUIRED)`  
  Replace "openssl" dependency to "quictls" and run `vcpkg install curl[openssl]` for all triplets above
   * [x] Windows
   * [x] UWP
   * [x] Darwin
   * [x] iOS
   * [x] Linux
   * [x] Android

Q. The port's install collides with `openssl`. Do I have to put `skip` in the ci.baseline.txt?

--> The port's build check is skipped. Need to test manually

Yes.

I read the existing configurations of `openssl` port and `microsoft/msquic`. Then reorganized in my style.  
I tried to simplify them with comments but additional suggestions will be helpful!


ports/quictls/portfile.cmake
```diff
@@ -0,0 +1,193 @@
+if(EXISTS ${CURRENT_INSTALLED_DIR}/include/openssl/ssl.h)
+    message(FATAL_ERROR "Can't build '${PORT}' if another SSL library is installed. Please remove existing one and try install '${PORT}' again if you need it.")
+endif()
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO quictls/openssl
+    REF 77248f1c1c8ea8dc57d624050b78eab0b3a10a04 # 2021-05-03
+    SHA512 881639f0bfd83858ce5d28aaf013dc34105cbc9c2fcc040c873c41ca45a0ea3dcb9dfd8e81821b21e92d4b9577f86d11ac010bb1f3746713892969c62d46fda6
+    HEAD_REF openssl-3.0.0-alpha15+quic
```
nibanks, line 10: MsQuic needs the latest 1.1.1 - https://github.com/quictls/openssl/tree/OpenSSL_1_1_1k+quic

ports/quictls/portfile.cmake
```diff
@@ -0,0 +1,193 @@
...
+    PATCHES
+        fix-http.patch
```
nibanks, line 12: Why is this needed? We don't need then when we build via submodule in msquic.

ports/quictls/portfile.cmake
```diff
@@ -0,0 +1,193 @@
...
+)
+
+# Option: shared/static
+set(OPENSSL_SHARED no-shared)
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic")
+    set(OPENSSL_SHARED shared)
+endif()
+
+# Option: feature / algorithms
+list(APPEND CONFIGURE_OPTIONS
```
nibanks, line 22: These config options don't match exactly with what we use in MsQuic for quictls it seems - https://github.com/microsoft/msquic/blob/main/submodules/CMakeLists.txt#L54

versions/baseline.json
```diff
@@ -5208,6 +5208,10 @@
       "baseline": "1.15.1-4",
       "port-version": 0
     },
+    "quictls": {
+      "baseline": "2021-05-03",
```
nibanks, line 5212: Should `baseline` be the version?

NancyLi1013:
Hi @luncliff 
Thanks for your PR.
Can you please help confirm the status on uwp? Does this port support uwp?

luncliff:
Seems like the branch doesn't. 😑 I will update "support" soon.

I think I can leave those uwp related platform names in detect_platform
script, but it can be confusing for users...
Do you think VC-WIN-UWP things need to be removed?

NancyLi1013:
> Seems like the branch doesn't. 😑 I will update "support" soon. I think I can leave those uwp related platform names in detect_platform script, but it can be confusing for users... Do you think VC-WIN-UWP things need to be removed?

 I noticed that the information from configure-perl-x64-uwp-dbg-out.log:
```
Configuring OpenSSL version 1.1.1k+quic (0x101010bfL) for VC-WIN64A-UWP
Using os-specific seed configuration
```
But I'm not familiar with these, so I cannot give you any suggestions. Sorry, I cannot help you. But I suggest to use support field in vcpkg if it doesn't support. If not, please try to fix the problems on uwp.

`Q. The port's install collides with openssl. Do I have to put skip in the ci.baseline.txt?`

Yes, we should add skip in ci.baseline.txt.

luncliff:
Understood. I will update this within days.

ports/quictls/portfile.cmake
luncliff: Nice point. Removed them.

ports/quictls/portfile.cmake
luncliff: What I intended is supporting SSL selection as a feature of future `ms-quic` port. It will be like `ms-quic[core]`(with schannel), `ms-quic[quictls]`(with quictls/opnessl), etc.

versions/baseline.json
luncliff: Yeah, when it's possible. Currently this PR uses `version-date`. Which means the date of the snapshot (not source version)

luncliff:
@NancyLi1013 @nibanks, I just checked some community triplet builds. I think It's fine now.

nibanks:
cc @ThadHouse 

This seems Ok to me, but I am not knowledgeable about vcpkg enough to be able to completely review.

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,93 @@
+#
+# References
+#   - ${VCPKG_ROOT}/scripts/cmake/vcpkg_common_definitions.cmake
+#   - https://github.com/quictls/openssl/tree/openssl-3.0.0-alpha16+quic
```
nibanks, line 4: This is incorrect, now that you pointed this to the 1.1.1 branch, right?

nibanks (approved):

ports/quictls/detect_platform.cmake
luncliff: Nice point! I forgot to update this part. 😱

luncliff:
For future reviewers, this port is **skipped** in CI, so manual check is required.

```
cd ${VCPKG_ROOT}
./vcpkg install \
    quictls:arm64-android \
    quictls:arm-android \
    quictls:x64-android \
    quictls:x86-android \
    quictls:x64-osx \
    quictls:arm64-osx \
    quictls:x64-ios \
    quictls:arm64-ios
...
```

ThadHouse:
UWP should work. We support UWP in MsQuic, and use this branch of quictls, so it shouldn't have any issues. But otherwise looks good to me.

NancyLi1013:
Have you tested these official triplets on your local? @luncliff 
CI will not check the build results since they have been skipped on our CI. So we should keep it work fine on our machine. For community triplets, I noticed you had checked them. As @ThadHouse said above, seems you need to update the changes to support uwp.

@nibanks and @ThadHouse thanks for your help and review.

luncliff:
@NancyLi1013 Some import test of `quictls` will be done in #18225 (`ms-quic`). Let me cross-check both project again for UWP.

luncliff:
I focused on "We support UWP in MsQuic". 

https://github.com/microsoft/msquic/blob/a3788df6b433272099c4af048b04604119d816c5/scripts/build.ps1#L302-L304

Currently, MsQuic's CI is checking UWP build with `"schannel"`, which is default value of `QUIC_TLS` in `WIN32`.

I think it will be great we can select between `"schannel"` and `"openssl"` for future `ms-quic` port works.
But https://github.com/quictls/openssl/tree/OpenSSL_1_1_1k+quic/Configurations doesn't provide configuration for UWP. 

If quictls/openssl have to support UWP platform build, it seems like those configurations in [openssl-3.0.0-alpha17+quic](https://github.com/quictls/openssl/tree/openssl-3.0.0-alpha17+quic/Configurations) need to be picked into the working branch(OpenSSL_1_1_1k+quic).

https://github.com/quictls/openssl/blob/e79f5fc69a4cefebac20c2c6642ffdeec430e6f7/Configurations/50-win-onecore.conf#L108-L117

I think thous work can be done later. Or I will convert this PR to draft until the work is done.

ThadHouse:
CI builds both schannel and openssl binaries of UWP.

https://github.com/microsoft/msquic/blob/main/submodules/CMakeLists.txt#L26

From my research, arm builds are always UWP compatible, and for x86 and x64 things just need to be suffixed with `-ONECORE` It doesn't seem like /APPCONTAINER is actually needed from my testing, and I've successfully ran the binaries built through the app store api checker and everything seems to work.

luncliff:
> CI builds both schannel and openssl binaries of UWP.
> 
> https://github.com/microsoft/msquic/blob/main/submodules/CMakeLists.txt#L26
> 
> From my research, arm builds are always UWP compatible, and for x86 and x64 things just need to be suffixed with `-ONECORE` It doesn't seem like /APPCONTAINER is actually needed from my testing, and I've successfully ran the binaries built through the app store api checker and everything seems to work.

So `/APPCONTAINER` is not necessary... that was the point I couldn't confirm. I couldn't sure it affects the project's CI. I will copy those platform fields and retry. Appreciate the check @ThadHouse 👍

luncliff:
Well, bad news.
"post-build validation" reports it doesn't contain App Container bit. 

```
PS D:\shared\vcpkg> .\vcpkg.exe install quictls:arm64-uwp quictls:x64-uwp quictls:x86-uwp quictls:arm-uwp
...
Starting package 1/4: quictls:arm-uwp
Building package quictls[core]:arm-uwp...
-- Using cached D:/shared/vcpkg/downloads/quictls-openssl-a6e9d76db343605dae9b59d71d2811b195ae7434.tar.gz
-- Cleaning sources at D:/shared/vcpkg/buildtrees/quictls/src/b195ae7434-aa88ae7c62.clean. Use --editable to skip cleaning for the packages you specify.
-- Extracting source D:/shared/vcpkg/downloads/quictls-openssl-a6e9d76db343605dae9b59d71d2811b195ae7434.tar.gz
-- Using source at D:/shared/vcpkg/buildtrees/quictls/src/b195ae7434-aa88ae7c62.clean
-- Configuring arm-uwp-rel
-- Building arm-uwp-dbg
-- Building arm-uwp-rel
-- Installing: D:/shared/vcpkg/packages/quictls_arm-uwp/share/quictls/copyright
The following DLLs do not have the App Container bit set:

    D:/shared/vcpkg/packages/quictls_arm-uwp/debug/bin/libcrypto-81_1_1-arm.dll
    D:/shared/vcpkg/packages/quictls_arm-uwp/debug/bin/libssl-81_1_1-arm.dll
    D:/shared/vcpkg/packages/quictls_arm-uwp/bin/libcrypto-81_1_1-arm.dll
    D:/shared/vcpkg/packages/quictls_arm-uwp/bin/libssl-81_1_1-arm.dll

This bit is required for Windows Store apps.
    D:\shared\vcpkg\ports\quictls\portfile.cmake
Error: Building package quictls:arm-uwp failed with: POST_BUILD_CHECKS_FAILED
```

luncliff:
I wanted to build in DLL but seems like there is no way :(

ThadHouse:
Ah, we exclusively use no-shared, which makes sense. We do pass /APPCONTAINER to the msquic dll which statically links to openssl. So that would explain why it wasn't needed for us.

luncliff:
I see. I think I can advance to next work after checking each triplets.

NancyLi1013:
Please ping me to review again if all things have been done on your sides. @luncliff

luncliff:
> Please ping me to review again if all things have been done on your sides. @luncliff

I think all done. Wish the port works well on your team's environments

scripts/ci.baseline.txt
```diff
@@ -1329,6 +1329,13 @@ qt5-x11extras:x86-windows=fail
 qt5-x11extras:x64-windows=fail
 qt5-x11extras:x64-windows-static=fail
 qt5-x11extras:x64-windows-static-md=fail
+# Conflicts with openssl
+quictls:arm64-windows=skip
+quictls:x64-linux=skip
+quictls:x64-osx=skip
+quictls:x64-windows-static=skip
+quictls:x64-windows=skip
+quictls:x86-windows=skip
```
NancyLi1013, line 1338:
```suggestion
quictls:x86-windows=skip
quictls:x64-windows-static-md=skip
quictls:x64-uwp=skip
quictls:arm-uwp=skip
```
We should set all triplets as skip in ci.baseline.txt.

NancyLi1013 (changes requested):

NancyLi1013:
Have you tested these triplets on your local machine? @luncliff

luncliff:
Build and find_package(Openssl). For all triplets above. I don't know if there is a way to test QUIC API

NancyLi1013:
> Build and find_package(Openssl). For all triplets above. I don't know whether there is a way to test QUIC related API

Can you build `quictls ` successfully on your machine?

luncliff:
> Can you build `quictls ` successfully on your machine?

Yes, I did. Was there any trouble?

NancyLi1013:
> > Can you build `quictls ` successfully on your machine?
> 
> Yes, I did. Was there any trouble?

Of course not. Since CI will not check the build status, we must make sure it can build successfully on our local.

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,93 @@
+#
+# References
+#   - ${VCPKG_ROOT}/scripts/cmake/vcpkg_common_definitions.cmake
+#   - https://github.com/quictls/openssl/tree/OpenSSL_1_1_1k+quic
+#       - ${SOURCE_PATH}/Configuration/README.design
+#       - ${SOURCE_PATH}/Configuration/10-main.conf
+#
```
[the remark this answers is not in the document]
luncliff: Applied in cdd1597 👍

NancyLi1013 (approved):

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+        set(PLATFORM "android-arm")
+    elseif(VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
+        set(PLATFORM "android-x86_64")
+    elseif(VCPKG_TARGET_ARCHITECTURE MATCHES "x86")
+        set(PLATFORM "android-x86")
+    endif()
+
+elseif(VCPKG_TARGET_IS_LINUX)
+    if(VCPKG_TARGET_ARCHITECTURE MATCHES "arm")
+        set(PLATFORM "linux-armv4")
```
strega-nil-ms, line 15: Is there a reason we're skipping arm64?

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "x64")
+        set(PLATFORM "linux-x86_64")
+    else()
+        set(PLATFORM "linux-generic32")
```
strega-nil-ms, line 19: this feels... wrong, what is `linux-generic32`? Shouldn't this check for STREQUAL "x86"?

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
+if(VCPKG_TARGET_IS_ANDROID)
+    # ${SOURCE_PATH}/Configuration/15-android.conf
+    if(VCPKG_TARGET_ARCHITECTURE MATCHES "arm64")
```
strega-nil-ms, line 3: In general, also, this should use `STREQUAL`

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+elseif(VCPKG_TARGET_IS_OSX)
+    if(VCPKG_TARGET_ARCHITECTURE MATCHES "arm64")
+        set(PLATFORM "darwin64-arm64-cc")
+    else()
+        set(PLATFORM "darwin64-x86_64-cc")
+    endif()
+
+elseif(VCPKG_TARGET_IS_FREEBSD OR VCPKG_TARGET_IS_OPENBSD)
+    set(PLATFORM "BSD-generic64")
```
strega-nil-ms, line 43: please make sure that this only sets up x64 on x64.

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+elseif(VCPKG_TARGET_IS_MINGW)
+    if(VCPKG_TARGET_ARCHITECTURE MATCHES "64")
+        set(PLATFORM "mingw64")
```
strega-nil-ms, line 47: probably wants to be x86_64 or x86, not doing this matching.

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+    if(VCPKG_TARGET_ARCHITECTURE STREQUAL "x86")
+        set(PLATFORM "VC-WIN32-ONECORE")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "x64")
+        set(PLATFORM "VC-WIN64A-ONECORE")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm")
+        set(PLATFORM "VC-WIN32-ARM")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
+        set(PLATFORM "VC-WIN64-ARM")
+    else()
+        message(FATAL_ERROR "Unknown UWP target architecture: ${VCPKG_TARGET_ARCHITECTURE}")
```
strega-nil-ms, line 63: We already have fatal errors at the bottom of this file.

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+    if(VCPKG_TARGET_ARCHITECTURE STREQUAL "x86")
+        set(PLATFORM "VC-WIN32")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "x64")
+        set(PLATFORM "VC-WIN64A")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm")
+        set(PLATFORM "VC-WIN32-ARM")
+    elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
+        set(PLATFORM "VC-WIN64-ARM")
+    else()
+        message(FATAL_ERROR "Unknown Windows target architecture: ${VCPKG_TARGET_ARCHITECTURE}")
```
strega-nil-ms, line 77: We already have a fatal error at the bottom

ports/quictls/detect_platform.cmake
```diff
@@ -0,0 +1,86 @@
...
+    # ${SOURCE_PATH}/Configuration/15-ios.conf
+    if(VCPKG_TARGET_ARCHITECTURE MATCHES "arm64")
+        set(PLATFORM "ios64-xcrun")
+    elseif(VCPKG_TARGET_ARCHITECTURE MATCHES "arm")
+        set(PLATFORM "ios-xcrun")
+    elseif(VCPKG_TARGET_ARCHITECTURE MATCHES "x86" OR
+           VCPKG_TARGET_ARCHITECTURE MATCHES "x64")
+        set(PLATFORM "iossimulator-xcrun")
+    else()
+        message(FATAL_ERROR "Unknown iOS target architecture: ${VCPKG_TARGET_ARCHITECTURE}")
```
strega-nil-ms, line 32: There's already a FATAL_ERROR at the bottom of this file

strega-nil-ms (changes requested):
Needs some changes in detect_platform! Thanks :)

ports/quictls/detect_platform.cmake
luncliff: No. It's sad I can't check build works with the triplet :(
Added it in e7e0901

ports/quictls/detect_platform.cmake
luncliff: Fixed in e7e0901. Changed the value to `linux-x86` and checks `VCPKG_TARGET_ARCHITECTURE`.

ports/quictls/detect_platform.cmake
luncliff: I habitually used the expression. Replaced them to `STREQUAL`  in e7e0901.

ports/quictls/detect_platform.cmake
luncliff: Added explicit check of `VCPKG_TARGET_ARCHITECTURE` like the other cases in e7e0901.

ports/quictls/detect_platform.cmake
luncliff: That was a mistype. Nice point :+1: Fixed in e7e0901.

ports/quictls/detect_platform.cmake
luncliff: I wanted to give more hint that it's a issue for the platform.
Made fallthrough to the bottom check.

ports/quictls/detect_platform.cmake
luncliff: --> https://github.com/microsoft/vcpkg/pull/17949#discussion_r650355039

ports/quictls/detect_platform.cmake
luncliff: --> https://github.com/microsoft/vcpkg/pull/17949#discussion_r650355039

ras0219-msft:
Because we're providing _zero_ testing coverage of this in CI (since it conflicts with the blessed official openssl), it seems like this would be more appropriately hosted in a community registry:

https://github.com/microsoft/vcpkg/discussions/17161

nibanks:
> Because we're providing _zero_ testing coverage of this in CI (since it conflicts with the blessed official openssl), it seems like this would be more appropriately hosted in a community registry.
> 
> See our pinned discussion of community registries: #17161

Is there a way to **not** conflict with openssl. BTW, this fork of openssl **is blessed** for Microsoft usage, explicitly used by MsQuic.

luncliff:
> Because we're providing _zero_ testing coverage of this in CI (since it conflicts with the blessed official openssl), it seems like this would be more appropriately hosted in a community registry.
> 
> See our pinned discussion of community registries: #17161

Thanks for review. I already have my registry in #17161, so that's an option for me to support `quictls` and `ms-quic`...
But at a glance this port looks like a similar case of boringssl(#8455).  And I did reference its portfile.

In #18225 ac64cb5, I found the port is can be built in CI environment. So if there is a way to check without harming openssl related check, I would like to try for this.

ras0219-msft:
>BTW, this fork of openssl is blessed for Microsoft usage, explicitly used by MsQuic.

We would love to support it, but we cannot fracture the entire catalog over it.

1. We do not want to add libraries to the curated catalog that we cannot build. This is a maintenance nightmare, because we have no way to validate that updates do not break everything.
2. We build a _consistent_ universe -- this means that if one package sees `include/openssl.h` from `openssl`, another package must not see `include/openssl.h` from `my-custom-ssl`.

> Is there a way to not conflict with openssl

Some scattered points:

1. If MsQuic is consuming only the official `openssl` interface, we could include msquic built on openssl in the main catalog. `quictls` can be hosted in a third party catalog and it's very easy to choose `quictls` to replace all uses of `openssl` via an overlay port.
2. If quictls had a build mode that _augmented_ the blessed copy of openssl instead of replacing it (such as not building any symbols that openssl provides and not providing `openssl.h`), then we would be happy to include that version in the public catalog. Note that just deleting `openssl.h` is insufficient, because static linking would still result in One Definition Rule (ODR) violations and Undefined Behavior (UB, nasal demons, very bad!).

>But at a glance this port looks like a similar case of boringssl(#8455).

BoringSSL was added to the catalog long before we had third party registry support -- today, it would be better served by a third-party registry with the exact arguments above.

nibanks:
@ras0219-msft there is a long story around why we **had** to create the quictls fork, but essentially this fork is the base OpenSSL + QUIC functionality. I don't want to onboard this without build coverage. My question is if there is a way to still onboard this with coverage.

The goal isn't to have quictls replace usage for anything that wants openssl. The goal is to make it available for anyone who expressly wants quictls.

ras0219-msft:
After looking further into quictls, it's possible we could express it as a feature in the existing openssl port (`openssl[quictls]`). When chosen by the user's dependency tree (off by default), this would activate using the `quictls` fork of the `openssl` codebase.

However, this still wouldn't fix the issue of build coverage -- in CI we would still need to build and deploy the upstream copy of openssl, which means the quictls feature would not be selected and not covered. From a pure _build_ perspective, we could add a test port that doesn't install any files but does build quictls; however test ports are not available for customers on the vcpkg cli (and it wouldn't help anyway -- the port installs no files).

Additionally, even if we solve the problem of coverage for quictls, there's no way we can add build coverage for msquic -- by design, the `openssl.h` we will install and make available in CI _must_ be the vanilla upstream `openssl.h`, not a fork.

There is a potentially extreme option of vendoring quictls _inside_ msquic's port and only supporting dynamic linkage (static linking quictls to hide the inner openssl symbols). This means users would see `openssl.h` and link against the upstream openssl, but could also link against msquic.

luncliff:
> There is a potentially extreme option of vendoring quictls inside msquic's port and only supporting dynamic linkage (static linking quictls to hide the inner openssl symbols). This means users would see openssl.h and link against the upstream openssl, but could also link against msquic.

Thanks for check @ras0219-msft. I wanted this port can be separated with `ms-quic`, but if so I will **nest** `quictls` inside `ms-quic`.

Let me keep this PR open until this weekend. I will sync this with #18225 and then close this one.

PR #17957, opened 2021-05-16
[luasec] new port

TheCycoONE:
Adds luasec, a lua library for networking over TLS.

https://github.com/brunoos/luasec/

all, No

Yes

Yes

### DIFF
--- a/ports/luasec/CMakeLists.txt
+++ b/ports/luasec/CMakeLists.txt
@@ -1,0 +1,38 @@
+cmake_minimum_required(VERSION 3.20.0)
+project(luasec)
+
+find_path(LUA_INCLUDE_DIR lua.h PATH_SUFFIXES lua)
+find_library(LUA_LIBRARY lua)
+find_package(OpenSSL)
+
+set(LUASEC_INCLUDES ${LUA_INCLUDE_DIR} src)
+set(LUASEC_LIBRARIES
+    ${LUA_LIBRARY}
+    OpenSSL::SSL
+    OpenSSL::Crypto
+    OpenSSL::applink
+    ws2_32)
+
+add_library(lua-ssl
+    src/config.c
+    src/ssl.c
+    src/context.c
+    src/x509.c
+    src/ec.c
+    src/options.c
+    src/luasocket/buffer.c
+    src/luasocket/io.c
+    src/luasocket/timeout.c
+    src/luasocket/wsocket.c)
+
+target_include_directories(lua-ssl PRIVATE ${LUASEC_INCLUDES})
+target_link_libraries(lua-ssl PRIVATE ${LUASEC_LIBRARIES})
+set_target_properties(lua-ssl PROPERTIES PREFIX "")
+
+install(TARGETS lua-ssl
+    RUNTIME DESTINATION bin
+    LIBRARY DESTINATION lib
+    ARCHIVE DESTINATION lib)
+
+install(FILES src/ssl.lua DESTINATION share/lua)
+install(FILES src/https.lua DESTINATION share/lua/ssl)
--- a/ports/luasec/portfile.cmake
+++ b/ports/luasec/portfile.cmake
@@ -1,0 +1,33 @@
+set(LUASEC_REVISION v1.0.1)
+set(LUASEC_HASH 21ae200e40f13a35eebca95cdac25ae13532a3008b2f9ec97a079033b9efde69e81954354453623eb77340368f482d3c69adb26881a6c0d7c4897df31301af93)
+
+if (VCPKG_TARGET_IS_WINDOWS)
+    vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
+endif()
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO brunoos/luasec
+    REF ${LUASEC_REVISION}
+    SHA512 ${LUASEC_HASH}
+    HEAD_REF master
+)
+
+file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
+
+vcpkg_configure_cmake(
+    SOURCE_PATH ${SOURCE_PATH}
+    PREFER_NINJA
+)
+
+vcpkg_install_cmake()
+vcpkg_copy_pdbs()
+
+# Remove debug share
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
+
+# Handle copyright
+file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+
+# Allow empty include directory
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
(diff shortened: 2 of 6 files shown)


ports/luasec/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+set(LUASEC_INCLUDES ${LUA_INCLUDE_DIR} src)
+set(LUASEC_LIBRARIES
+    ${LUA_LIBRARY}
+    OpenSSL::SSL
+    OpenSSL::Crypto
+    OpenSSL::applink
+    ws2_32)
+
+add_library(ssl
```
ras0219, line 16: This library name looks like it would conflict with other ssl libraries (`libssl.so`/`ssl.dll`). What's the official upstream name?

```suggestion
add_library(luasec
```

TheCycoONE: Upstream expects the ssl.dll/ssl.so name.

If the code is linked to the C/C++ project that should be alright to change it, however it's not uncommon to include through lua in which case the ssl.dll name needs to stay.

https://packages.debian.org/bullseye/amd64/lua-sec/filelist is how debian packages it:
liblua<lua_version>-sec.so.<luasec_version> for linking to C and lua/<lua_version>/ssl.so for including in lua.

Any preferences? I think lpeg, luasocket, luafilesystem, etc. should follow the same convention.

TheCycoONE: archlinux only installs it to `usr/lib/lua/5.4/ssl.so` - any opinion on bin/lua/ssl.(dll|so)

TheCycoONE: https://github.com/lua/lua/blob/681297187ec45268e872b26753c441586c12bdd8/luaconf.h#L212 may also be relevant to the discussion if we're concerned about the dynamic load from lua use case. We could recommend setting CPATH to include lua-?.dll and name the library lua-ssl.dll

Do you like that option?

JonLiu1993:
@ras0219,Could you help review this pr?

ports/luasec/portfile.cmake
```diff
@@ -0,0 +1,30 @@
+set(LUASEC_VERSION 1.0.1)
```
PhoebeHui, line 1:
```suggestion
```

ports/luasec/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+    src/x509.c
+    src/ec.c
+    src/options.c
+    src/luasocket/buffer.c
+    src/luasocket/io.c
+    src/luasocket/timeout.c
+    src/luasocket/wsocket.c)
+
+if(WIN32)
+    add_definitions("-DLSEC_API=__declspec(dllexport)")
```
JackBoosY, line 29: From https://github.com/brunoos/luasec/blob/master/luasec.vcxproj#L54
```xml
      <PreprocessorDefinitions>WIN32;_DEBUG;_WINDOWS;_USRDLL;LUASEC_EXPORTS;LUASEC_INET_NTOP;WINVER=0x0501;_WIN32_WINNT=0x0501;NTDDI_VERSION=0x05010300;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
And it was defined in https://github.com/brunoos/luasec/blob/d5df31561751ec0d4098dfc09c92ece215a56a5a/src/compat.h#L15-L19.
So we don't need this.

PhoebeHui (commented):
The LUASEC_VERSION looks not used.

ports/luasec/portfile.cmake
```diff
@@ -0,0 +1,30 @@
...
+set(LUASEC_REVISION v1.0.1)
+set(LUASEC_HASH 21ae200e40f13a35eebca95cdac25ae13532a3008b2f9ec97a079033b9efde69e81954354453623eb77340368f482d3c69adb26881a6c0d7c4897df31301af93)
+
```
JackBoosY, line 4: From https://github.com/brunoos/luasec/blob/d5df31561751ec0d4098dfc09c92ece215a56a5a/src/compat.h#L15-L19:
```cpp
#if defined(_WIN32)
#define LSEC_API __declspec(dllexport) 
#else
#define LSEC_API extern
#endif
```
I think the upstream only support dynamic build on Windows, so please add the following code here:
```cmake
if (VCPKG_TARGET_IS_WINDOWS)
endif()
```

JackBoosY (changes requested):

TheCycoONE:
Changed as suggested and removed the prefix from the so library. I'm still not clear on how we'd want to handle a possible name collision. I think the same approach should be applied to luasocket, lpeg, luafilesystem - whatever we pick.

TheCycoONE:
@JackBoosY any idea why I'm getting those errors "REGRESSION: libtasn1:x64-windows-static. If expected, add libtasn1:x64-windows-static=fail to .\scripts\ci.baseline.txt." - seems to happen every time, and I'm not sure what it has to do with my changes.

JackBoosY:
Waiting for https://github.com/microsoft/vcpkg/pull/18116 merge.

JackBoosY:
Ready for review now?

TheCycoONE:
Yes, thank you.

ports/luasec/CMakeLists.txt
JackBoosY: I'd like to change this name to `lua-ssl`.

TheCycoONE: Do you think we want or need any message letting people know to include lua-?.dll in their LUA_CPATH? I'm not sure the policy on messages in vcpkg.

JackBoosY: - If this library completely uses the code of other third-party libraries, we must instead find and use the third-party libraries in vcpkg.
- If this library only conflicts with other libraries in name, I think in order to prevent the library name from conflicting with other libraries, we have to change its name to `${PORT}-${LIBRARY}` to avoid that.

TheCycoONE: ok, I have changed the library to lua-ssl.

ports/luasec/portfile.cmake
```diff
@@ -0,0 +1,33 @@
...
+vcpkg_copy_pdbs()
+
+# Remove debug share
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
+
+# Handle copyright
+file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+
+# Allow empty include directory
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
JackBoosY, line 33: Can I ask why this port doesn't contain headers? Is it just a tool or something?

JackBoosY (changes requested):

ports/luasec/portfile.cmake
TheCycoONE: It's a lua C library, similar to:
https://github.com/microsoft/vcpkg/tree/master/ports/lpeg 
https://github.com/microsoft/vcpkg/tree/master/ports/luafilesytem
https://github.com/microsoft/vcpkg/tree/master/ports/luasocket

By convention they export luaopen_<name>
https://www.lua.org/pil/26.2.html describes the mechanism. As long as the export follows the luaopen_<name> convention and has the right name it can be found by the package searcher from a require statement in lua:
http://www.lua.org/manual/5.4/manual.html#pdf-package.searchers

JackBoosY (approved):

strega-nil-ms:
Thanks @TheCycoONE  :)

PR #17962, opened 2021-05-16
[vcpkg] [Android]Compile armv6 with arm mode

talregev:
Compile armv6 with arm mode


PhoebeHui:
@dan-shaw, could you help further review?

talregev:
I change this setting because when I compile with default `armeabi` setting curl for amrv6 it get exception on runtime.
from `android.toolchain.cmake` file on the ndk 15c:

```cmake
if(ANDROID_ABI MATCHES "armeabi")
	if(ANDROID_ARM_MODE STREQUAL thumb)
		list(APPEND ANDROID_COMPILER_FLAGS
			-mthumb)
	elseif(ANDROID_ARM_MODE STREQUAL arm)
		list(APPEND ANDROID_COMPILER_FLAGS
			
```

The default is `thumb` and it cause exception on run time when running curl command. 
```cmake
if(NOT ANDROID_ARM_MODE)
	set(ANDROID_ARM_MODE thumb)
endif()
```

talregev:
I will close this PR, because arm mode is good for our project, maybe for other project it not good:
https://stackoverflow.com/questions/10638130/what-is-the-arm-thumb-instruction-set
https://stackoverflow.com/questions/25720012/native-components-armeabi-v7a-and-thumb

I will not delete this PR, because to let other people know about this for their android ndk armeabi project.

talregev:
I was thinking to re-open this PR
I think it better to use arm mode then thumb mode,
Because even I try to compile all the client in our open source with thumb flag, I still get an exception in run time when call curl method.

When everything is compile with arm mode it don't happen.

talregev:
@PhoebeHui @dan-shaw

PhoebeHui:
@strega-nil-ms, could you help further review?

PR #17966, opened 2021-05-17
[glad] fix features

Neumann-A:
logic was the wrong way around


autoantwort:
I think all the features dependencies like 
```json
    "gl-api-11": {
            "description": "Imports extensions from OpenGL API specification version 1.1.",
            "dependencies": [
              {
                "name": "glad",
                "default-features": false,
                "features": [
                  "gl-api-10"
                ]
              }
            ]
          },
```
are then quite useless and can be removed.

Neumann-A:
> are then quite useless and can be removed.

Not really. If you depend on "gl-api-10" but the port was build with only "gl-api-11" then you need to rebuild glad. The features here basically say "gl-api-11" includes "gl-api-10" and this has to be expressed as a dependency to be correctly forwarded to other ports.

autoantwort:
Ahh yeah you are right.

ras0219 (approved):

PhoebeHui (approved):

PR #17974, opened 2021-05-18
[nanodbc] Allow specifing the ODBC version for nanodbc

AnthonyCalandra:
This patch adds the option to specify an ODBC version directly to the nanodbc library, which up to this point has been looking for ODBC versions itself (see [here](https://github.com/nanodbc/nanodbc/blob/master/nanodbc/nanodbc.cpp#L103)). It only passes in the `NANODBC_ODBC_VERSION` option to nanodbc's cmake script if this flag has been set by the user.

- #### What does your PR fix?  
  See the following comment on what this PR is fixing: https://github.com/nanodbc/nanodbc/issues/114#issuecomment-358361843
  I experienced this issue on a CentOS box with unixODBC 2.3.1. In summary: unixODBC 2.3.1 reports an incorrect ODBC version which nanodbc picks up, but should be overridden with the version passed in through `NANODBC_ODBC_VERSION`.

  Not sure I need to update the baseline file; please let me know if I do.

  Yes

  Yes

### DIFF
--- a/ports/nanodbc/portfile.cmake
+++ b/ports/nanodbc/portfile.cmake
@@ -1,7 +1,7 @@
 # Only static libraries are supported.
 # See https://github.com/nanodbc/nanodbc/issues/13
 vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
-	
+
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO nanodbc/nanodbc
@@ -10,6 +10,10 @@ vcpkg_from_github(
     HEAD_REF master
 )
 
+if(DEFINED NANODBC_ODBC_VERSION)
+    set(NANODBC_ODBC_VERSION -DNANODBC_ODBC_VERSION=${NANODBC_ODBC_VERSION})
+endif()
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
@@ -17,6 +21,7 @@ vcpkg_configure_cmake(
         -DNANODBC_DISABLE_EXAMPLES=ON
         -DNANODBC_DISABLE_TESTS=ON
         -DNANODBC_ENABLE_UNICODE=OFF
+        ${NANODBC_ODBC_VERSION}
 )
 
 vcpkg_install_cmake()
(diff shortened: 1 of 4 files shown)


ports/nanodbc/portfile.cmake
```diff
@@ -10,13 +10,18 @@ vcpkg_from_github(
...
+endif()
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
     OPTIONS
         -DNANODBC_DISABLE_EXAMPLES=ON
         -DNANODBC_DISABLE_TESTS=ON
         -DNANODBC_ENABLE_UNICODE=OFF
+        ${NANODBC_ODBC_VERSION}
```
AnthonyCalandra, line 24: I didn't pass this in directly as an option because nanodbc's cmake file only checks to see if `NANODBC_ODBC_VERSION` is defined -- unlike an option which can be either ON/OFF. See: https://github.com/nanodbc/nanodbc/blob/master/CMakeLists.txt#L78

dg0yt: Actually the value of NANODBC_ODBC_VERSION is passed on to the compilation:

https://github.com/nanodbc/nanodbc/blob/cbc4ec2418a9feec559e06ea5dc0a632f305d5d1/CMakeLists.txt#L81-L82

And how is it specified for building port nanodbc? Would I need to modify the triplet?

AnthonyCalandra: I think my original comment wasn't clear enough. Yes, it is passed to the compilation, but I'm talking about embedding the option directly, i.e.
```cmake
vcpkg_configure_cmake(
    SOURCE_PATH ${SOURCE_PATH}
    PREFER_NINJA
    OPTIONS
        -DNANODBC_DISABLE_EXAMPLES=ON
        -DNANODBC_DISABLE_TESTS=ON
        -DNANODBC_ENABLE_UNICODE=OFF
        -DNANODBC_ODBC_VERSION=${NANODBC_ODBC_VERSION}
)
```
I made this comment to make it clear I intentionally didn't do it this way.

As for your second question, that's correct, you could add a line to the triplet file, and that's what I did on my machine (another work-around without adding this patch would be setting the flag as part of `VCPKG_CXX_FLAGS` but this patch more appropriate IMO).

JackBoosY:
Depends on https://github.com/microsoft/vcpkg/pull/17983.

JackBoosY (approved):

PR #17978, opened 2021-05-18
[vcpkg] Update tls12-download.exe to 2021-05-05 version

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/17688

Update tls12-download.exe to 2021-05-05 version.

Related https://github.com/microsoft/vcpkg/pull/17734

PR #17983, opened 2021-05-18
[vcpkg baseline][many ports] Fix build error

JackBoosY:
Fix baseline issues, these changes need to fix togather, since they affected each other in CI testing.
 - [X] wxwidgets 
 - [x] mlpack
 - [X] embree2
 - [X] embree3
 - [x] opencv4
 - [x] zxing-cpp
 - [x] dbow2
 - [x] clamav

1. Fix wxwidgets bug:
```
F:\vcpkg\buildtrees\wxwidgets\x86-windows-dbg\libs\stc\cotire\wxscintilla_CXX_prefix.hxx(4): fatal error C1083: Cannot open include file: 'F:\vcpkg\buildtrees\wxwidgets\src\54f4f955a1-1de49a51d0\build\cmake\lib\stc\CONFIG': No such file or directory
```
It's becuase the parameter `wxSCINTILLA_PREC_HEADER` is empty:
```cmake
if(wxBUILD_PRECOMP)
    # The auto-generated header causes undefined members and identifiers in the
    # standard c++ headers when using clang.
    # Do not disable precompiled headers entirely but use the main Scintilla
    # header as prefix header so there is at least a small speedup.
    if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
        set(wxSCINTILLA_PREC_HEADER "${wxSOURCE_DIR}/src/stc/scintilla/include/Scintilla.h")
    endif()
    wx_target_enable_precomp(wxscintilla ${wxSCINTILLA_PREC_HEADER})
endif()
```
causes the leaked field `CONFIG` to be set in `set_target_properties` in function `wx_target_enable_precomp`.
Then, `get_property` obtained the property and read its file path, resulting in the generation of the wrong _wxscintilla_CXX_prefix.hxx_:
```cpp
/* cotire.cmake 1.8.0 generated file */
/* F:/vcpkg/buildtrees/wxwidgets/x86-windows-dbg/libs/stc/cotire/wxscintilla_CXX_prefix.hxx */
/* cotire.cmake 1.8.0 generated file */
/* F:/vcpkg/buildtrees/wxwidgets/x86-windows-dbg/libs/stc/cotire/wxscintilla_CXX_prefix.hxx */
#ifdef __cplusplus
#include "F:\vcpkg\buildtrees\wxwidgets\src\54f4f955a1-1de49a51d0\build\cmake\lib\stc\CONFIG"
#endif
```
which is not exist.

[...]

PR #17984, opened 2021-05-18
[vcpkg baseline] [embree2/embree3] Set cmake policy

PhoebeHui:
Related to https://github.com/microsoft/vcpkg/pull/17978

Set cmake policy for fixing failures:

```
CMake Error in CMakeLists.txt:
  You have set CMAKE_BACKWARDS_COMPATIBILITY to a CMake version less than
  2.4.  This version of CMake only supports backwards compatibility with
  CMake 2.4 or later.  For compatibility with older versions please use any
  CMake 2.8.x release or lower.
```

Reference https://cmake.org/cmake/help/latest/policy/CMP0001.html#policy:CMP0001

### DIFF
--- a/ports/embree2/CONTROL
+++ b/ports/embree2/CONTROL
@@ -1,5 +1,0 @@
-Source: embree2
-Version: 2.17.7
-Homepage: https://github.com/embree/embree
-Description: High Performance Ray Tracing Kernels.
-Build-Depends: tbb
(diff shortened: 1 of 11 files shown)


PhoebeHui:
Merge this changes to PR https://github.com/microsoft/vcpkg/pull/17983 since the baseline ports affected by each others.

PR #17592, opened 2021-04-30
[openh264] only build shared or static library

mcmtroffaes:
**Describe the pull request**

- #### What does your PR fix?  
  Currently, when installing openh264, both the static and shared library get built and potentially installed, even though only a static or shared one is requested by vcpkg. For example, on x64-linux:

  ```bash
  # before patch
  $ ./vcpkg install openh264
  $ ls installed/x64-linux/lib/
  libopenh264.a  libopenh264.so  libopenh264.so.2.1.0  libopenh264.so.6  pkgconfig
  # after patch
  $ ./vcpkg install openh264
  $ ls installed/x64-linux/lib/
  libopenh264.a  pkgconfig
  ```

  This PR incorporates an upstream patch from https://github.com/cisco/openh264/pull/3351 to ensure that openh264 respects meson's default_library option, resolving this issue (since this is the mechanism also used by vcpkg to request either the static or shared library).

- #### Which triplets are supported/not supported? Have you updated the [CI baseline](https://github.com/microsoft/vcpkg/blob/master/scripts/ci.baseline.txt)?  
  No changes.

- #### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?  
  Yes.

- #### If you have added/updated a port: Have you run `./vcpkg x-add-version --all` and committed the result?  
  Yes.

Issue #17481, opened 2021-04-24
Provide pkgconfig via the CMake toolchain file on Windows


glebm:
**Is your feature request related to a problem? Please describe.**

The Windows experience would be much nicer if vcpkg CMake toolchain file provided pkgconfig and set `PKG_CONFIG_EXECUTABLE`.

**Proposed solution**

Set PKG_CONFIG CMake variables in the CMake toolchain file.

**Additional context**

Currently users have to work around this by downloading and building pkgconfig themselves.
Internally, vcpkg already seems to have much of what would be needed to support it.

JackBoosY:
We do need this.
cc @Neumann-A

Neumann-A:
hmm added #17488 and #17487 so people can choose which one they want.

JackBoosY:
@Neumann-A Also, we should output the usage of pkgconfig I think.

PR #17985, opened 2021-05-18
[ffmpeg] use pkgconfig to get system dependencies for cmake

mcmtroffaes:
- #### What does your PR fix?  
  This PR uses the config.mak file generated during the build process to get all system dependencies. This ensures that only the required system dependencies are included, and future-proofs the port against changes in such dependencies, which are a pain to track. This substantially simplifies the FindFFMPEG.cmake script and makes it more robust and easier to maintain. It also ensures that FindFFMPEG.cmake lists the same dependencies as the generated pkg-config file. I've noticed that for many builds the dependencies are slightly different (sometimes more, sometimes less) from what the old FindFFMPEG.cmake did, so presumably this PR also might fix some subtle bugs with less often used functions that need those libraries.

  One change I do point out concerns non-static builds: some dependent libraries were only used in the FindFFMPEG.cmake script in static builds: see the ``set(ENABLE_* ${STATIC_LINKAGE})`` lines in the original portfile. This is not what the upstream pkgconfig file does, however. With this PR, dependencies will be listed exactly as in the pkgconfig file. I think it's a price worth paying for keeping things simple.

  Finally, I note that many past pull requests for the ffmpeg port have been concerned with fixing dependencies across platforms. My hope is that this PR will result in a more stable port in the long term, and less time wasted on chasing dependency fixes.

  I've thoroughly tested the PR on linux, osx, and windows triplets (see https://travis-ci.com/github/mcmtroffaes/ffmpeg-msvc-build/builds/225915669 - this has one failure but that's addressed in #17592 and also https://ci.appveyor.com/project/mcmtroffaes/ffmpeg-msvc-build/builds/39162659).

  No changes to baseline. No changes to supported triplets.

  Yes, to the best of my knowledge.

  Yes.


cenit:
Just a side note, not against your work of course

FindFFMPEG.cmake is slowly transforming irreversibly from a CMake module that is able to find FFMPEG, independently from how it's built, to a CMake config file tailored around the specific options.
This should be taken into consideration too: because at the beginning I was hoping to send the module upstream to Kitware, now it looks like the only way to export it from vcpkg will be sending it to ffmpeg (with a much lower possibilities of acceptance).

cenit:
but again, this is a route that the module took some time ago, and this is just the final step.
I'd rename it into FFMPEGConfig.cmake, and remove the unnecessary wrapper

mcmtroffaes:
Thanks for your useful and considerate feedback @cenit - I'm sorry that I didn't think of that, and in retrospect I definitely should have mentioned this in my PR as an important consideration (and maybe discussed it first via an issue). Definitely with this patch there's little reason indeed then not to go for a cmake config file. IMO, having looked for quite some time at the complexity of the configure script, I don't think it is feasible to have a fully functional FindFFMPEG script that would work reliably for all possible platforms and variations of ffmpeg builds. Anyway, I'm happy to have a discussion and if there are good reasons not to this route and thus reject this PR, I'd be interested to hear them.

I note that a cmake config could be generated as part of the configure process, which could be submitted to upstream ffmpeg after some testing here in vcpkg. Dealing with upstream ffmpeg is slow but I've managed to get some small patches applied. But I'm not sure to what extent support for generating a cmake config would be welcomed by upstream ffmpeg.

Another route might be for the vcpkg cmake config to be able to import the pkg-config file. I don't know how to do that however. If someone knows this is possible and can point me to some examples, I'd be happy to investigate.

One question though, if folks agree we should go for the config route: should this then be turned into an unofficial cmake config (with the unofficial prefix), as in for instance https://github.com/microsoft/vcpkg/blob/master/ports/libvpx/unofficial-libvpx-config.cmake.in and other ports that provide non-upstream cmake configs? Or should we stick with FFMPEGConfig.cmake so find_package(FFMPEG) keeps working? I assume FFMPEGConfig.cmake would be preferred to keep compatibility?

cenit:
to be totally honest, i never understood the unofficial prefix. it's just a way to ask for more patches and future work. Let's keep it this way so that you don't have to fix dozen of ports that would become broken (do not think only about ports here, but also private ports that expects to work with find_package(FFMPEG))

otoh, if you want to try to upstream the config after some testing here, for me that would be wonderful. I know it will be slow and difficult and maybe undesired by ffmpeg maintainers, but at least you had a chance of success and so you are in a better position :)

mcmtroffaes:
Sounds like there's a plan then. I'd prefer this to be merged first, and then work towards a proper config at a later time, if that's ok.

And thanks for rerunning CI. Looks like there's a failure of opencv4, fortunately only on linux. I tried to reproduce locally and this looks related to an issue with png detection, unrelated to this patch? Here's my local build log:

```
$ ./vcpkg install ffmpeg[core]:x64-linux opencv[core,ffmpeg]:x64-linux
...
-- Configuring x64-linux-dbg
    Command failed: /home/xxx/vcpkg/downloads/tools/cmake-3.20.2-linux/cmake-3.20.2-linux-x86_64/bin/cmake /home/xxx/vcpkg/buildtrees/opencv4/src/4.5.1-e92f755db3.clean [...]
    Working Directory: /home/xxx/vcpkg/buildtrees/opencv4/x64-linux-dbg
    Error code: 1
      /home/xxx/vcpkg/buildtrees/opencv4/config-x64-linux-dbg-out.log
      /home/xxx/vcpkg/buildtrees/opencv4/config-x64-linux-dbg-err.log

  scripts/cmake/vcpkg_configure_cmake.cmake:332 (vcpkg_execute_required_process)
  ports/opencv4/portfile.cmake:290 (vcpkg_configure_cmake)

Error: Building package opencv4:x64-linux failed with: BUILD_FAILED
  Package: opencv4:x64-linux
  Vcpkg version: 2021-01-13-unknownhash

$ tail /home/xxx/vcpkg/buildtrees/opencv4/config-x64-linux-dbg-out.log
-- Found Threads: TRUE
-- Found ZLIB: /home/xxx/vcpkg/installed/x64-linux/debug/lib/libz.a (found version "1.2.11")
-- Found ZLIB: /home/xxx/vcpkg/installed/x64-linux/debug/lib/libz.a (found version "1.2.11")
-- Found PNG: optimized;/home/xxx/vcpkg/installed/x64-linux/debug/lib/libpng.a;debug;/home/xxx/vcpkg/installed/x64-linux/debug/lib/libpng16d.a (found version "1.6.37")
-- Looking for /home/xxx/vcpkg/installed/x64-linux/include/libpng/png.h
-- Looking for /home/xxx/vcpkg/installed/x64-linux/include/libpng/png.h - not found
-- OpenCV Python: during development append to PYTHONPATH: /home/xxx/vcpkg/buildtrees/opencv4/x64-linux-dbg/python_loader
See also "/home/xxx/vcpkg/buildtrees/opencv4/x64-linux-dbg/CMakeFiles/CMakeOutput.log".
See also "/home/xxx/vcpkg/buildtrees/opencv4/x64-linux-dbg/CMakeFiles/CMakeError.log".
```

PhoebeHui:
Opencv:x64-linux failed on CI testing:

```
CMake Error at /mnt/vcpkg-ci/installed/x64-linux/share/ffmpeg/FindFFMPEG.cmake:66 (find_library):
  Could not find FFMPEG_DEPENDENCY_-Wl,--no-undefined_RELEASE using the
  following names: -Wl,--no-undefined
  /mnt/vcpkg-ci/installed/x64-linux/share/ffmpeg/FindFFMPEG.cmake:146 (append_dependencies)
  /mnt/vcpkg-ci/installed/x64-linux/share/ffmpeg/vcpkg-cmake-wrapper.cmake:6 (_find_package)
  /agent/_work/1/s/scripts/buildsystems/vcpkg.cmake:815 (include)
  modules/videoio/cmake/detect_ffmpeg.cmake:6 (find_package)
  modules/videoio/cmake/init.cmake:3 (include)
  modules/videoio/cmake/init.cmake:30 (add_backend)
  cmake/OpenCVModule.cmake:298 (include)
  cmake/OpenCVModule.cmake:361 (_add_modules_1)
  modules/CMakeLists.txt:7 (ocv_glob_modules)
```

[config-x64-linux-dbg-err[1].log](https://github.com/microsoft/vcpkg/files/6514222/config-x64-linux-dbg-err.1.log)
[config-x64-linux-dbg-out[1].log](https://github.com/microsoft/vcpkg/files/6514223/config-x64-linux-dbg-out.1.log)

mcmtroffaes:
Many thanks for the logs. Somehow I'm hitting some other errors and not this one, but I think I should be able to fix this error from the log message.

mcmtroffaes:
I still cannot build opencv4 locally on linux with vcpkg but I've found a way to reproduce some link errors on linux using a test application. I think it's related to the link order when linking against a large number of libraries; I'll put this back to draft status for the time being until I'm sure it's all ironed out.

cenit:
> I still cannot build opencv4 locally on linux with vcpkg

what's your error?
i don't think that the png message above is an error. It's just too verbose, because header has a different name and is verified after that one

cenit:
are you sure you are building the same opencv4 as CI?
please look at `vcpkg/scripts/test_ports/vcpkg-ci-opencv` to see the many features that are tested in CI

cenit:
I did that special port to test as many features as possible because otherwise they were getting rotten extremely quickly due to other updates.
Even now some triplets are not building opencv4 anymore due to cascaded failures, which means that a dependency has been accepted in master with a broken status in that triplet... unfortunately!

mcmtroffaes:
Yes, I tried precisely those features. However, I just found that after merging the latest master, opencv4[ffmpeg] locally builds successfully. 🤦 I've pushed the merge just now, hopefully that makes azure happy again. 🤞

PhoebeHui:
@mcmtroffaes, the opencv4:x64-linux still failed with same error as before in CI testing. I will try to build it in my VM.

mcmtroffaes:
Many thanks @PhoebeHui - it's odd that it gives the *exact* same error... I got exactly the same error compiling a test application against ffmpeg (with the same feature set as in vcpkg CI), and https://github.com/microsoft/vcpkg/pull/17985/commits/4d68c694d1a775d5857bc95c8d2df253fa54b972 solved that error for me. Then I ran into some linking errors which were fixed in https://github.com/microsoft/vcpkg/pull/17985/commits/2049bd6c22c67dd325e0c105ebd0fa728e164a00.

Anyway, I'm getting closer to building opencv4 locally with the same feature set as in vcpkg CI on linux. There are many system dependencies on linux for some of opencv4's requirements, which is the main thing slowing me down (and which might also explain my earlier problems). But hopefully that will tease out the error eventually. I will report once it's done.

cenit:
> There are many system dependencies on linux for some of opencv4's requirements, which is the main thing slowing me down (and which might also explain my earlier problems). But hopefully that will tease out the error eventually. I will report once it's done.

you can see here https://github.com/microsoft/vcpkg/blob/master/scripts/azure-pipelines/linux/provision-image.sh how linux image is provisioned in azure for the CI system (all system dependencies that are installed)

PhoebeHui:
I encountered same errors with latest commit in this PR, the same error also occurred in aubio and pangolin port.

The full failure logs from previous CI testing.
[failure logs for x64-linux (2).zip](https://github.com/microsoft/vcpkg/files/6521816/failure.logs.for.x64-linux.2.zip)

dg0yt:
> ...the same error also occurred in aubio and pangolin port.
> failure logs for x64-linux (2).zip

> ~~~
> CMake Error at /mnt/vcpkg-ci/installed/x64-linux/share/ffmpeg/FindFFMPEG.cmake:66 (find_library):
>  Could not find FFMPEG_DEPENDENCY_gcc_s_RELEASE using the following names:
>  gcc_s
>~~~

gcc_s? Isn't this meant to be used "implicitly"?
I don't see curl in the depend-info, but I know that curl requires gcc_s in its pc file, and I think it shouldn't. 
Maybe other ports create similar "requirements" when creating cmake config or pc files in the same way as curl: from discovered link libraries.

mcmtroffaes:
Thanks @PhoebeHui This is actually a new error (previous one was about a linker flag, "-Wl,..."). Aside as to whether this library should be included or not, it's easy to filter it out to (hopefully) let the regression test pass. I've added a commit to that effect, for now.

I'm a bit surprised as to why the library isn't found, though. I have it in my own installation and libgcc_s is present and found. Is it possible that libgcc_s (and libgcc) are missing from azure?

mcmtroffaes:
Scrap that last comment, I'm now also getting the gcc_s error on a completely fresh build with all dependencies installed as in azure.

dg0yt:
I still wonder if gcc and gcc_s really need to be linked explicitly. These are libraries gcc will link by default unless you give `-nostdlib`. Cf. https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#Link-Options
They are probably located in implicit link directories which are not accessed by cmake's find_library, and no one would look for them unless trying to go through implicit link libraries.

mcmtroffaes:
x64-linux is now succeeding 🎉 

Would it be possible to retrigger the arm64_windows build on azure? Seems like it failed due to a network connection error:

```
...
-- Downloading https://strawberryperl.com/download/5.30.0.1/strawberry-perl-5.30.0.1-32bit.zip... Failed. Status: 28;"Timeout was reached"
CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:184 (message):
      
      Failed to download file.
...
```

mcmtroffaes:
Concerning -lgcc and -lgcc_s: I agree @dg0yt that's how it looks, although I'm no gcc expert. If some -lgcc or -lgcc_s flags are passed redundantly, then that ought to be resolved by the pkgconfig of the concerned library. I checked and they definitely do not originate from ffmpeg itself. There might be reasons as to why they might be necessary in certain situations for certain platforms when linking certain libraries, who knows... For instance, the ffmpeg configure script adds an -lgcc flag (without -nostdlib as far as I can tell) when targeting symbian, showing that there are unusual situations.

Sibras:
> Another route might be for the vcpkg cmake config to be able to import the pkg-config file. I don't know how to do that however. If someone knows this is possible and can point me to some examples, I'd be happy to investigate.

See [https://cmake.org/cmake/help/latest/module/FindPkgConfig.html](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)

So for example:
```
find_package(PkgConfig REQUIRED)
pkg_check_modules (FF REQUIRED ffmpeg)
```

gives FF_LIBRARIES and FF_STATIC_LIBRARIES variables which can be checked depending on the current build configuration.

Given the portfile already requires pkgconfig then adding the above should be a non-issue and hopefully makes life a bit easier by handling this automatically.

mcmtroffaes:
Ah, nice, thanks @Sibras! I'll see what I can do with that. With the current design of the PR, indeed this could be called from the portfile at the end of the build process. Would there be a reason to prefer calling pkg_check_modules from the FindFFMPEG/FFMPEGConfig file instead?

dg0yt:
>  indeed this could be called from the portfile

You cannot use this from the portfile: It needs the target configuration. However, from the portfile you can use 
https://github.com/microsoft/vcpkg/blob/master/ports/vcpkg-pkgconfig-get-modules/x_vcpkg_pkgconfig_get_modules.cmake

And I want to warn that many pc files have issues (wrong prefix, missing dependencies, wrong dependencies).

mcmtroffaes:
Ah right, makes sense. Are you envisioning an FFMPEGConfig which basically wraps pkg_check_modules on the relevant ffmpeg libraries? The main problem is that, even if pkgconf is installed, find_package(PkgConfig) does not find the executable provided by pkgconf; seems related to #17481.

If this means fixing up some bad .pc files, still a worthwhile effort IMO. Since ffmpeg already uses pkg-config for many of its dependencies during its build process, I'd hope we wouldn't have too many breakages for ffmpeg.

dg0yt:
> Are you envisioning ...

Not sure who is asked ... I'm not using ffmpeg and didn't ask for it. I'm just struggling with getting gdal dependencies right, and this involves pc files ;-) 

You might want to `grep "^prefix=" installed/<triplet>/lib/pkgconfig/*.pc` to get an idea which files need work. 
- If they have an absolute prefix path, they need work.
- If they have `prefix=${pcfiledir}/../..`, they are probably okay. (At least vcpkg_fixup_pkgconfig seems to have run.)

mcmtroffaes:
Oh sure. I was asking everyone who may have an opinion on this. The ffmpeg .pc files don't require any other .pc files besides those for other ffmpeg libraries (although pkgconfig is used to populate Libs and Libs.private for quite many of its dependencies), so the situation is considerably simpler. The pc files are pretty correct except that they don't use the -l and -L convention for libraries and library paths, and I've seen this confuse pkgconf output for --libs-only-l and --libs-only-L, but that's easy to correct; I may post a separate PR for that or even try to get it fixed in upstream ffmpeg.

dg0yt:
I got the impression I should make more use of the Requires(.private) field when the dependency has its own pc file.

mcmtroffaes:
I'm not sure what you're trying to do that you need this specific field, but at the risk of telling something you already know: in an ideal scenario where all pc files are correct, a call to ``pkg-config --libs`` (or, from cmake, pkg_check_modules) can provide all dependencies in the correct link order. This will look for any modules listed in Requires. Anyway, this is getting a bit off topic. 😄

Sibras:
> 
> https://github.com/microsoft/vcpkg/blob/master/ports/vcpkg-pkgconfig-get-modules/x_vcpkg_pkgconfig_get_modules.cmake

That looks like it would make it alot easier as it saves having to manually set things like the path (which is what I was originally thinking)
> 
> And I want to warn that many pc files have issues (wrong prefix, missing dependencies, wrong dependencies).

ffmpeg already has vcpkg_fixup_pkgconfig() and as far as I know all of its dependencies should now have that aswell (as ffmpeg uses the .pc files for most dependencies internally). So the pc file should be correct - and if not then clearly thats something that needs fixing in vcpkg_fixup_pkgconfig().

mcmtroffaes:
Thanks for the feedback, everyone.

This should be pretty ready now, but let's wait for the CI to do its cycle. To small things:
* I've added the CURRENT_PACKAGES_DIR to the pkgconfig search path in x_vcpkg_pkgconfig_get_modules to enable parsing of pc files of the port itself during the portfile script phase.
* I noted a few small issues with .pc files consumed by ffmpeg. I've put those into separate PRs (see #18077 and #18078) but it doesn't affect anything here (since the libpaths are ignored at this point).

mcmtroffaes:
@PhoebeHui The PR is now ready for final review. Using pkg-config has the extra benefit that no unnecessary libraries are linked against - in particular the list of dependencies for the dynamic builds is now completely empty. That restores the original ``set(ENABLE_* ${STATIC_LINKAGE})`` behaviour (but without needing the variables), and it's an improvement from how it used to be: system libraries were always listed regardless of static or dynamic builds; now they are only listed for static builds. This PR should also help working towards supporting pkgconfig's ``pkg_check_modules`` for ffmpeg, giving more granular control over which of ffmpeg's libraries are linked against, and also making it easier to use ffmpeg's libraries in other projects that use pkg-config (via cmake or otherwise). I've already tested this for libavutil on windows and it worked at least for x86-windows (with the two PR's for fixing the two .pc files for openh264 and dav1d).

As far as I can tell, the CI failures are all related to a download problem with kf5i18n (mingw-w64-i686-libiconv-1.16-1-any.pkg.tar.xz cannot be found on any of the msys mirrors), and are therefore completely unrelated to this PR. Related packages, ffmpeg, and its dependencies such as opencv4 and aubio, are all succeeding as expected, on all CI triplets.

PhoebeHui:
I test this PR with aubio, it works well, thanks @mcmtroffaes for your effort! All features test default by CI pipeline, so the feature doesn't need test locally.

cenit:
> Related packages, ffmpeg, and its dependencies such as opencv4 and aubio, are all succeeding as expected, on all CI triplets.

so CI is now testing opencv4 on all triplets? good. it lost vcpkg-ci-opencv on a couple of triplets some days ago with a cascade failure that went unnoticed, strange it got also fixed without anyone noticing. Please double check (i will do the same)

PhoebeHui:
I will double check the opencv test status in CI pipeline:

- [ ] x86-windows
- [ ] x64-windows
- [ ] x64-windows-static
- [ ] x64-windows-static-md
- [ ] x64-uwp
- [X] arm64-windows
- [ ] arm-uwp
- [ ] x64-osx
- [ ] x64-linux

cenit:
ah ok, so uwp is still cascade failure unfortunately. So we are not testing also these changes in uwp?
@PhoebeHui can you please find out what broke opencv4 in uwp CI? it would be helpful for fixing it

mcmtroffaes:
Apologies @cenit - I thought because CI was green on uwp, opencv4 was passing, but I've now checked the azure output and indeed it's marked as "cascade", as is ffmpeg. I've just installed the uwp developer tools to do a local test against x64-uwp.

PhoebeHui:
Opencv4 cascade due to the dependency ports cascade or skip.

The dependency ports with x65-uwp in CI testing:
Cascade: ceres, ffmpeg, glog, jasper, ogre, qt5-base, tesseract, vtk
Skip: cudnn, gflags,hdf5, openexr, tbb, 
Doesn't support uwp: halide, gdcm

mcmtroffaes:
I found the culprit, it's openh264:

```
..\src\33cf500b7b-c619a10b3a.clean\meson.build:127:2: ERROR: Problem encountered: FIXME: Unhandled system windowsstore
```

cenit:
:(
better to disable that feature. If it's a cascade failure, not even core features are tested! And in fact in the past I saw opencv getting in a very bad situation if left untested.
I know it's a little bit out of scope for this PR, but would it make sense to unlock CI here so that we are sure it's working?

@mcmtroffaes I know it's misleading, but green badge sometimes just means "I had no opportunity to test so I won't tell you it's broken". With @BillyONeal I have an open issue on exactly this problem, because any cascade failure is not marked as a problem in the PR first triggering it

PhoebeHui:
The  'supports' filed only works in CI testing.

mcmtroffaes:
Ah, thanks for clarifying that!

PhoebeHui:
avcpp with x64-uwp and arm-uwp failed with same error:
```
  D:\buildtrees\avcpp\src\7f1e65bea1-0345499867.clean\src\packet.cpp(97,5): error C4996: 'av_init_packet': was declared deprecated
```
[failure logs for x64-uwp.zip](https://github.com/microsoft/vcpkg/files/6538417/failure.logs.for.x64-uwp.zip)

cenit:
@mcmtroffaes happy that you found out that opencv was cascaded by your own modifications and that now is verified (and working) in CI

not so happy that unblocking ffmpeg on uwp you now uncovered other ports that were dormant but broken

mcmtroffaes:
Not that I know. It looks like avcpp might be the cause; see https://github.com/h4tr3d/avcpp/blob/master/src/avutils.h#L27 I'm not certain yet this is the cause but I'm working on a patch.

mcmtroffaes:
Thanks so much @PhoebeHui for your patience and helping with the error logs: I appreciate your diligence. I've added a quick patch for the uwp issue, pleased to see it all green again on uwp.

PhoebeHui:
@mcmtroffaes @cenit, opencv4 with arm-uwp and x64-uwp passed in CI testing, appreciate your effort very much!

@strega-nil-ms, could you please help further review?

mcmtroffaes:
Merged master & bumped port version.

PhoebeHui (approved):

strega-nil-ms:
LGTM thanks @mcmtroffaes :)

mcmtroffaes:
Thanks so much for the merge!

PR #17986, opened 2021-05-18
[drogon] Update to 1.6.0

an-tao:


JonLiu1993 (approved):

PhoebeHui:
@an-tao, could you submit an issue or a PR  to Upstream about the new added patch?

an-tao:
> @an-tao, could you submit an issue or a PR to Upstream about the new added patch?

The problem is about the defination of ssize_t type, in hiredis library, it is defined as long long, but in my libs(drogon and trantor) it is defined as std::intptr_t, so there is a confilict.
I decide to use the same base type as hiredis does in the subsequent version, so the patch will be removed in the next porting.

strega-nil-ms:
Thanks @an-tao for supporting vcpkg :)

PR #17994, opened 2021-05-18
[sail] Update to 0.9.0-pre13

HappySeaFox:
**Describe the pull request**

- #### What does your PR fix?  
Update 0.9.0-pre12 -> 0.9.0-pre13

Everything except UWP is supported. CI baseline is not updated.

  `YES`

  `YES`


ports/sail/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sail",
-  "version-semver": "0.9.0-pre12",
+  "version-semver": "0.9.0-pre13",
   "port-version": 1,
```
JonLiu1993, line 4: Please delete "port-version"

versions/baseline.json
```diff
@@ -5449,7 +5449,7 @@
       "port-version": 0
     },
     "sail": {
-      "baseline": "0.9.0-pre12",
+      "baseline": "0.9.0-pre13",
       "port-version": 1
```
JonLiu1993, line 5453: Please delete "port-version"

ports/sail/vcpkg.json
JonLiu1993: When you update the port, you need to set the port-version to 0 or delete it

happy-sea-fox: I see. What about `versions/baseline.json` ? Should `port-version` also be deleted from there?

happy-sea-fox: Same question regarding `versions/s-/sail.json`

JonLiu1993: > > When you update the port, you need to set the port-version to 0 or delete it
> 
> 1. Are still sure about that? We have to delete a non-zero `port-version` from `port/vcpkg.json` when we update a port? It makes no sense for me. Could you please point to the docs describing this requirement?

@happy-sea-fox ,[This ](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#update-the-port-version-field-in-the-manifest-file-of-any-modified-ports)is our policy about port-version 
> 2. What about `versions/baseline.json` ? Should `port-version` also be deleted from there?
`versions/baseline.json`

To be precise, `versions/baseline.json` does not need to be manually modified by you, you only need to modify the `port-version` in port/vcpkg.json and then run the command "vcpkg x-add-version sail --overwrite-version" Then it will automatically remain consistent with the `port-version` in port/vcpkg.json

happy-sea-fox: Thanks for the clarification! 👍 Now it's clear that port-version is a field relative to the current upstream version. Previously I was sure it uses an absolute numbering not relative to anything.

JonLiu1993: > Thanks for the clarification! 👍 Now it's clear that port-version is a field relative to the current upstream version. Previously I was sure it uses an absolute numbering not relative to anything.

At the same time I thanks for your contribution to vcpkg 👍

happy-sea-fox:
CI failure is related to #17983. Will wait for a fix.

JackBoosY:
Ready for review now?

JonLiu1993 (approved):

PR #18013, opened 2021-05-19
[scripts-audit] z_vcpkg_apply_patches, z_vcpkg_escape_regex_control_characters

strega-nil:
fixes a minor bug as well, where `GIT_CONFIG_NOSYSTEM` is never reset.

### DIFF
--- a/scripts/cmake/z_vcpkg_apply_patches.cmake
+++ b/scripts/cmake/z_vcpkg_apply_patches.cmake
@@ -31,9 +31,13 @@ This should only be used for edge cases, such as patches that are known to fail 
 function(z_vcpkg_apply_patches)
     cmake_parse_arguments(PARSE_ARGV 0 "arg" "QUIET" "SOURCE_PATH" "PATCHES")
 
+    if(DEFINED arg_UNPARSED_ARGUMENTS)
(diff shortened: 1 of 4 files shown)


scripts/cmake/z_vcpkg_apply_patches.cmake
```diff
@@ -31,9 +31,13 @@ This should only be used for edge cases, such as patches that are known to fail
 function(z_vcpkg_apply_patches)
     cmake_parse_arguments(PARSE_ARGV 0 "arg" "QUIET" "SOURCE_PATH" "PATCHES")
 
+    if(DEFINED arg_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR "internal error: z_vcpkg_apply_patches was passed extra arguments: ${arg_UNPARSED_ARGUMENTS}")
```
JackBoosY, line 35: Should we just output it as a warning message instead?

JackBoosY (dismissed):

scripts/cmake/z_vcpkg_apply_patches.cmake
strega-nil: No, since this is an internal script; there's no need to make it a warning since ports don't call it.

strega-nil:
Closing for rollup (#18112)

PR #18025, opened 2021-05-20
[libsodium] Fix mingw builds

past-due:
**Describe the pull request**
Fix mingw builds by better matching the behavior of libsodium's autoconf build scripts.

### DIFF
--- a/ports/libsodium/vcpkg.json
+++ b/ports/libsodium/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "libsodium",
-  "version-string": "1.0.18",
-  "port-version": 4,
+  "version": "1.0.18",
+  "port-version": 5,
   "description": "A modern and easy-to-use crypto library",
   "homepage": "https://github.com/jedisct1/libsodium",
   "dependencies": [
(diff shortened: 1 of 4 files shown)


ports/libsodium/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "libsodium",
   "version-string": "1.0.18",
```
NancyLi1013, line 3:
```suggestion
  "version": "1.0.18",
```
This field should be `version `instead of `version-string`. Please see https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md.

NancyLi1013 (changes requested):

ports/libsodium/vcpkg.json
past-due: Fixed. 👍

past-due:
@NancyLi1013: Force-pushed the fix. 👍

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your fixing. @past-due

ras0219-msft:
LGTM, thanks for the fixes!

PR #18028, opened 2021-05-20
vcpkg_configure_make: Fixes for mingw

past-due:
### EDIT: Waiting on merge of #17137 and #17215 to rebase.

**Describe the pull request**

Various fixes / tweaks for `vcpkg_configure_make` and mingw.

- mingw: Do not use `compile` / `windres-rc` wrappers.
   - Fixes an issue where mingw environments that use LLVM-based toolchains (example: the msys2 `clang64` environment) fail to build ports because LLVM's windres does not support the (Windows) `-fo` option.
   - **Note: Most is not needed once #17215 is merged.**
- mingw: Use the vcpkg-acquired msys binaries for `objdump` / `nm`
   - Otherwise mingw environments that use LLVM-based toolchains (example: the msys2 `clang64` environment) cannot build ports as their `objdump` is preferred and its output differs enough that it isn't properly parsed by `libtool`.
- mingw: Remove `-luuid` from default libs to avoid `libtool` errors that break builds.
   - Note: This matches the behavior for regular Windows builds, which remove `uuid` to prevent adding `-luuid` to the libs.
   - **Note: Not needed once #17137 is merged. Will remove from this PR.**

  mingw community triplets


dg0yt:
This is mostly duplicating work which already has pending PR.
- wrappers: https://github.com/microsoft/vcpkg/pull/17215
- uuid: https://github.com/microsoft/vcpkg/pull/17137

dg0yt:
PS: I would have more changes to vcpkg_configure_make, but there is no point in stacking more PRs. CI runs take long for changes to this script, and quite often they run into unrelated problems.

past-due:
> This is mostly duplicating work which already has pending PR.
> 
> * wrappers: #17215
> * uuid: #17137

Thank you - had not noticed those yet. 🥳 I prefer #17137's changes for luuid fixes and can rebase the wrapper tweaks (ensuring `RC` and `WINDRES` are set) on top of #17215 once that's merged.

past-due:
@dg0yt I've gone ahead and pulled out the one main non-duplicated fix into #18055

Closing this in favor of your two PRs (#17137, #17215) and #18055.

PR #18033, opened 2021-05-20
[Doc] Update binarycaching.md with clearer description

AugP:
We received user feedback about the binary caching feature not being clear. I rewrote some of the opening section to try to describe it better. Feedback is appreciated.

**Describe the pull request**

- #### What does your PR fix?  
  Small documentation fix. 

  Yes

### DIFF
--- a/docs/users/binarycaching.md
+++ b/docs/users/binarycaching.md
@@ -2,9 +2,15 @@
 
 **The latest version of this documentation is available on [GitHub](https://github.com/Microsoft/vcpkg/tree/master/docs/users/binarycaching.md).**
 
-Binary caching is vcpkg's method for reusing package builds between projects and between machines. Think of it as a "package restore accelerator" that gives you the same results as though you built from source. Each build is packaged independently, so changing one library only requires rebuilding consuming libraries.
+Libraries installed with vcpkg are built from source. However, as builds can take time, it is inconvenient waiting for this to happen each time a library is installed for the first time in a new environment.
 
-If your CI provider offers a native "caching" function, we recommend using both methods for the most performant results.
+Binary caching is a vcpkg feature that saves copies of library binaries in a shared location that can be accessed by vcpkg for future installs. This means that, as a user, you should only need to build dependencies from source once, and only once. After that, if vcpkg is asked to install the same library with the same build configuration in the future, it will just copy the sources and binary from the cache and finish the operation in a few seconds.
+
+Binary caching can be very effective when working as part of a team on the same project, as only one team member needs to populate the cache the first time they install their dependencies with vcpkg. Other team members can then benefit from the caching feature to install the same dependencies on their own machines. Binary caching is also beneficial when both local builds and continuous integration builds are desired. For example, if you provision a new machine or container to perform a CI run, it can still restore binaries from the cache generated on a previous build by a developer working on their local machine. 
+
+Caches can be hosted in a variety of environments. The most basic examples are a network file share or another location on the local machine, though we recommend only doing this for testing, not as a production workflow. Caches can also be stored in any NuGet feed, Azure Blog Storage, or Google Cloud Storage (experimental). 
+
+If your CI provider offers a native "caching" function, we recommend using both vcpkg binary caching and the native method for the most performant results.
 
 In-tool help is available via `vcpkg help binarycaching`.
 
--- a/docs/users/binarycaching.md
+++ b/docs/users/binarycaching.md
@@ -2,13 +2,15 @@
 
 **The latest version of this documentation is available on [GitHub](https://github.com/Microsoft/vcpkg/tree/master/docs/users/binarycaching.md).**
 
-Libraries installed with vcpkg are built from source. However, as builds can take time, it is inconvenient waiting for this to happen each time a library is installed for the first time in a new environment.
+Libraries installed with vcpkg can always be built from source. However, that can duplicate work and waste time when working across multiple projects.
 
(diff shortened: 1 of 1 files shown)


docs/users/binarycaching.md
```diff
@@ -2,9 +2,15 @@
 
 **The latest version of this documentation is available on [GitHub](https://github.com/Microsoft/vcpkg/tree/master/docs/users/binarycaching.md).**
 
-Binary caching is vcpkg's method for reusing package builds between projects and between machines. Think of it as a "package restore accelerator" that gives you the same results as though you built from source. Each build is packaged independently, so changing one library only requires rebuilding consuming libraries.
+Libraries installed with vcpkg are built from source. However, as builds can take time, it is inconvenient waiting for this to happen each time a library is installed for the first time in a new environment.
```
ras0219, line 5:
```suggestion
Libraries installed with vcpkg can always be built from source. However, that can duplicate work and waste time when working across multiple projects.
```

docs/users/binarycaching.md
```diff
@@ -2,9 +2,15 @@
...
 
-If your CI provider offers a native "caching" function, we recommend using both methods for the most performant results.
+Binary caching is a vcpkg feature that saves copies of library binaries in a shared location that can be accessed by vcpkg for future installs. This means that, as a user, you should only need to build dependencies from source once, and only once. After that, if vcpkg is asked to install the same library with the same build configuration in the future, it will just copy the sources and binary from the cache and finish the operation in a few seconds.
```
ras0219, line 7:
```suggestion
Binary caching is a vcpkg feature that saves copies of library binaries in a shared location that can be accessed by vcpkg for future installs. This means that, as a user, you should only need to build dependencies from source once. If vcpkg is asked to install the same library with the same build configuration in the future, it will just copy the built binaries from the cache and finish the operation in seconds.
```

docs/users/binarycaching.md
```diff
@@ -2,9 +2,15 @@
...
+
+Binary caching can be very effective when working as part of a team on the same project, as only one team member needs to populate the cache the first time they install their dependencies with vcpkg. Other team members can then benefit from the caching feature to install the same dependencies on their own machines. Binary caching is also beneficial when both local builds and continuous integration builds are desired. For example, if you provision a new machine or container to perform a CI run, it can still restore binaries from the cache generated on a previous build by a developer working on their local machine.
```
ras0219, line 9:
```suggestion
Binary caching is especially effective when using Continuous Integration, since local developers can reuse the binaries produced during a CI run. It also greatly enhances the performance of "ephemeral" or "hosted" build agents, since all local changes are otherwise lost between runs. By using binary caching backed by a cloud service, such as GitHub, Azure, or many others, you can ensure your CI runs at maximum speed and only rebuilds your dependencies when they've changed.
```

ras0219: We should avoid recommending reuse of packages from developers into CI or between developers, since that is poor security practice.

docs/users/binarycaching.md
```diff
@@ -2,9 +2,15 @@
...
+
+Binary caching can be very effective when working as part of a team on the same project, as only one team member needs to populate the cache the first time they install their dependencies with vcpkg. Other team members can then benefit from the caching feature to install the same dependencies on their own machines. Binary caching is also beneficial when both local builds and continuous integration builds are desired. For example, if you provision a new machine or container to perform a CI run, it can still restore binaries from the cache generated on a previous build by a developer working on their local machine. 
+
+Caches can be hosted in a variety of environments. The most basic examples are a network file share or another location on the local machine, though we recommend only doing this for testing, not as a production workflow. Caches can also be stored in any NuGet feed, Azure Blog Storage, or Google Cloud Storage (experimental).
```
ras0219, line 11:
```suggestion
Caches can be hosted in a variety of environments. The most basic examples are a folder on the local machine or a network file share. Caches can also be stored in any NuGet feed (such as GitHub or Azure DevOps Artifacts), Azure Blob Storage*, or Google Cloud Storage*.

 * (experimental) 
```

JackBoosY (approved):

Osyotr:
There are few things unclear left:
1) In what cases library will be rebuilt from source? Compiler update, port update, etc.
2) Is there any way to clean "dead" cache? Or better, don't rebuld (and cache) libraries when compiler update doesn't break abi compatibility?

AugP:
> There are few things unclear left:
> 
> 1. In what cases library will be rebuilt from source? Compiler update, port update, etc.
> 2. Is there any way to clean "dead" cache? Or better, don't rebuld (and cache) libraries when compiler update doesn't break abi compatibility?

I agree, thank you for the feedback! I'll work with Robert to address these questions in the PR.

ras0219-msft:
>In what cases library will be rebuilt from source? Compiler update, port update, etc.

There is the implementation notes section, however the ABI is explicitly undocumented -- we fully intend to add additional fields and content to it in the future.

https://github.com/microsoft/vcpkg/blob/f1d44a97469401417687bd9dfe92fecc26ecf849/docs/users/binarycaching.md#implementation-notes-internal-details-subject-to-change-without-notice

>Is there any way to clean "dead" cache?

Unfortunately there's no way for vcpkg to determine what's truly "dead"; you could have a 3 month old instance of vcpkg on your disk that you've moved from its original location, so there's no way we could track it.

As for the actual operation of cleaning, it's backend specific. For files, they can just be deleted. For NuGet / Cloud providers, it's up to those specific providers on how they offer "cleaning" services.

> Or better, don't rebuld (and cache) libraries when compiler update doesn't break abi compatibility?

Unfortunately, abi compatibility is _extremely_ specific to the compilers and can even vary based on flags. For example, LTO/LTCG usually changes the abi compatibility computation. Because of this, we take a conservative approach to ensure correctness.

ras0219 (approved):

Issue #1745, opened 2017-08-31
v141 static 3p library using CRT DLL


VikingExplorer:
The biggest and most important use case for DLLs is for common code that changes rarely and is almost always backwardly compatible. In short, that's absolutely the windows API.  That's why the CRT DLL should always be linked. I don't think it makes any sense to increase the size of one's own binaries by statically linking CRT.

However, it's quite a different story for our own modules/libraries and third party libraries we may use. To make these into DLLs implies that multiple EXEs will take advantage of shared DLLs to save RAM while being used by the customer.  

A) customer using multiple EXEs from the same vendor is an edge case. 
B) RAM is so plentiful, it's hardly worth the effort to use a DLL
C) DLLs introduces a significant deployment nightmare, otherwise known as DLL Hell. 
D) using static libs excludes all unused/dead code from binaries, where DLLs include it all.

Given these two points, I naturally assumed that static 3p libraries linking to CRT DLL would be the default.

So, I'm surprised that the two triplets provided are:

DLL - CRT.DLL
Static - CRT-static

Neither of which seem particularly useful.  Any help would be greatly appreciated.

VikingExplorer:
I want to run vcpkg install again, but only after I figure out to make sure it correctly sets vc141.

2nd question: where is the syntax description in the documentation.  If I just want to generate static libs, would this be correct:

vcpkg install cpprestsdk:x64-windows-static

VikingExplorer:
In the triplets folder, I created: x64-windows-static-md.cmake

set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE static)
set(VCPKG_PLATFORM_TOOLSET v141)

In a VS2017 Cmd Prompt:  vcpkg install cpprestsdk:x64-windows-static-md

In the folder: vcpkg\installed\x64-windows-static-md\debug\lib

Files:  
boost_atomic-vc140-mt-gd-1_65.lib
boost_chrono-vc140-mt-gd-1_65.lib

As you can see, they are neither 141, nor are they md.

VikingExplorer:
To restore #pragma linking, bugs must be fixed in vcpkg to create the properly named lib files.

Tried changing this to 141:  
    set(Boost_COMPILER "-vc140")
vcpkg\scripts\buildsystems\vcpkg.cmake  line 74

[...]

Issue #1338, opened 2017-06-22
[Boost] static link is broken


cenit:
As from the subject, static linking of boost libraries is currently broken.
This commit https://github.com/Microsoft/vcpkg/commit/7811ef602ecd963ae333e3427c82629d1c8e9f34 is the source of the problems: why renaming static libraries, if those are the names that CMake is then really looking for?
As an example, now we rename 
```
x86-windows-static/libboost_thread-vc140-mt-s-1_64.lib
```
into
```
x86-windows-static/boost_thread-vc140-mt-1_64.lib.
```
But this is wrong: boost_thread-vc140-mt-1_64.lib is known to CMake to be the filename of an import library (allowing dynamic linking with boost_thread-vc140-mt-1_64.dll), while *libboost_thread-vc140-mt-s-1_64.lib* should be the name of the static library (also with the -s, meaning it's statically linked with CRT).
In this way, if I configure CMake to look for static linking with boost, it cannot find the libraries.

What is the original reason of this renaming? Since CMake looks for these filenames, while should we break it?

ras0219-msft:
Sorry for taking so long to respond, and thank you for bringing up this issue!

We're definitely aware of the various issues around boost here (#369, #392, #601, #694). The essence of the problem (and solution) is:
>vcpkg only allows one instance of each library to be installed at a time. Therefore, the names of the lib files should not matter; the "capability" of boost is either there or it is not.
>
>Furthermore, a particular triplet can be configured to build some libraries statically and some dynamically. For example, you may opt to create your own triplet that builds zlib as a dll, and then all other dependencies are built as static libs. A portfile of another library can't easily know what configuration the other libraries were built with.
>
>Therefore, it can't know whether boost is static or dynamic in order to pass the appropriate flag into FindBOOST. We tried stripping out the build-specific identifiers from the names, but it did not work out (#369). Therefore, we rename the static names to the dynamic names, and then always look for the dynamic names. Therefore, you always use FindBOOST in the same way and the lib file does the right thing.

So, the current intended behavior is for downstream projects is to _leave all boost settings at their defaults_. Then, you use the Vcpkg triplet to choose what settings you want for boost (static link/static crt/etc) at configure time. On other systems, this can be easily manipulated at configure time via `-DBoost_USE_STATIC_LIBS=ON`.

cenit:
Dear Robert,
thank you very much for your answer. At first I really struggled to understand it: for me, static linking is broken, but for you it is working.
[...]

PR #18047, opened 2021-05-21
[vcpkg.cmake] Cleanup every special case in the find_package override

Neumann-A:
and replace it with appropriate `vcpkg-cmake-wrappers.cmake`.

CI will run due to adding the wrappers or changing the portfile of the affected base ports


ports/gsl/vcpkg-cmake-wrapper.cmake
```diff
@@ -0,0 +1,10 @@
+_find_package(${ARGS})
+if(GSL_FOUND AND TARGET GSL::gsl)
```
Osyotr, line 2: Missing endif()?

ports/boost-uninstall/portfile.cmake
```diff
@@ -2,3 +2,5 @@ set(VCPKG_POLICY_EMPTY_PACKAGE enabled)
 
 message(STATUS "\nPlease use the following command when you need to remove all boost ports/components:\n\
     \"./vcpkg remove boost-uninstall:${TARGET_TRIPLET} --recurse\"\n")
+
+configure_file("${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake" "${CURRENT_PACKAGES_DIR}/share/boost/vcpkg-cmake-wrapper.cmake" @ONLY)
```
JackBoosY, line 6: This port is only used to recursively uninstall boost, so I think we don't need to add a wrapper here.\
So should we move it to `boost-core` or `boost-build` or `boost-modular-build-helper`?

JackBoosY (changes requested):

ports/boost-uninstall/portfile.cmake
Neumann-A: `boost-modular-build-helper` should be a host dependency as it seems so it is out
`boost-build` is also a host dependency and not all of boost depends in it, so it is out 
`boost-vcpkg-helpers` is only for headers boost libs?

`boost-core` hmm could be a possibility but, does it cover everything in boost?

`boost-uninstall` covers the following helper scripts, so for me it looks like the correct port currently unless `boost-core` is just as powerful? 
```
boost-build
boost-modular-build-helper
boost-vcpkg-helpers
```

JackBoosY: Maybe we should add a new port `boost-wrapper`?

Neumann-A: another one? I don't think so.... Maybe ask the person responsible for the boost ports where to put the wrapper

scripts/buildsystems/vcpkg.cmake
```diff
@@ -813,50 +813,6 @@ function("${VCPKG_OVERRIDE_FIND_PACKAGE_NAME}")
     z_vcpkg_start_parent_scope_export()
     if(EXISTS "${VCPKG_CMAKE_WRAPPER_PATH}")
         include("${VCPKG_CMAKE_WRAPPER_PATH}")
-    elseif("${PACKAGE_NAME}" STREQUAL "Boost" AND EXISTS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/include/boost")
```
ras0219-msft, line 816: These need to stick around since removing them would break users continuing to use the old ports.

ras0219-msft (changes requested):
Thanks for the PR, this is really great and much appreciated!

I think using `boost-uninstall` is fine for this; the naming isn't perfect but the semantics are correct.

Unfortunately, we can't remove the direct hacks inside `vcpkg.cmake` because existing users of older versions currently rely on them. For example, if someone has a version baseline from the current master and uses boost, removing the lines from `vcpkg.cmake` would suddenly break them when they next `git pull`.

Without removing those lines, this LGTM. I think it's still worth doing to move us to a cleaner world.

scripts/buildsystems/vcpkg.cmake
Neumann-A: done. But i still want them gone because they are visible in --trace everytime the toolchain is loaded. 
maybe have a `vcpkg-(v2|no-old-stuff|clean|minimal).cmake` at some time in the feature. 
Especially in manifest mode all ports get rebuild anyway. Only people using classic mode would be affected.

JackBoosY:

JackBoosY (approved):

strega-nil-ms:
This looks great to me! Thanks @Neumann-A :)

JonLiu1993:
@Neumann-A 
Today when I do some work with `highfivex:64-windows-static`, I met some errors related to boost-test:
```
FAILED: tests/unit/tests_high_five_base.exe 
cmd.exe /C "cd . && F:\test8\vcpkg\downloads\tools\cmake-3.21.1-windows\cmake-3.21.1-windows-i386\bin\cmake.exe -E vs_link_exe --intdir=tests\unit\CMakeFiles\tests_high_five_base.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\mt.exe --manifests  -- C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\link.exe  tests\unit\CMakeFiles\tests_high_five_base.dir\tests_high_five_base.cpp.obj  /out:tests\unit\tests_high_five_base.exe /implib:tests\unit\tests_high_five_base.lib /pdb:tests\unit\tests_high_five_base.pdb /version:0.0 /machine:x64 /nologo    /debug /INCREMENTAL /subsystem:console  F:\test8\vcpkg\installed\x64-windows-static\debug\lib\boost_unit_test_framework-vc140-mt-gd.lib  F:\test8\vcpkg\installed\x64-windows-static\debug\lib\libhdf5_D.lib  F:\test8\vcpkg\installed\x64-windows-static\debug\lib\zlibd.lib  F:\test8\vcpkg\installed\x64-windows-static\debug\lib\libszip_D.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib && cd ."
LINK Pass 1: command "C:\PROGRA~2\MICROS~1\2019\ENTERP~1\VC\Tools\MSVC\1429~1.301\bin\Hostx64\x64\link.exe tests\unit\CMakeFiles\tests_high_five_base.dir\tests_high_five_base.cpp.obj /out:tests\unit\tests_high_five_base.exe /implib:tests\unit\tests_high_five_base.lib /pdb:tests\unit\tests_high_five_base.pdb /version:0.0 /machine:x64 /nologo /debug /INCREMENTAL /subsystem:console F:\test8\vcpkg\installed\x64-windows-static\debug\lib\boost_unit_test_framework-vc140-mt-gd.lib F:\test8\vcpkg\installed\x64-windows-static\debug\lib\libhdf5_D.lib F:\test8\vcpkg\installed\x64-windows-static\debug\lib\zlibd.lib F:\test8\vcpkg\installed\x64-windows-static\debug\lib\libszip_D.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:tests\unit\CMakeFiles\tests_high_five_base.dir/intermediate.manifest tests\unit\CMakeFiles\tests_high_five_base.dir/manifest.res" failed (exit code 1120) with the following output:
LINK : warning LNK4098: defaultlib 'LIBCMTD' conflicts with use of other libs; use /NODEFAULTLIB:library
tests_high_five_base.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl boost::unit_test::lazy_ostream::~lazy_ostream(void)" (__imp_??1lazy_ostream@unit_test@boost@@UEAA@XZ) referenced in function "public: virtual __cdecl boost::unit_test::lazy_ostream_impl<class boost::unit_test::lazy_ostream,char [1],char const (&)[1]>::~lazy_ostream_impl<class boost::unit_test::lazy_ostream,char [1],char const (&)[1]>(void)" (??1?$lazy_ostream_impl@Vlazy_ostream@unit_test@boost@@$$BY00DAEAY00$$CBD@unit_test@boost@@UEAA@XZ)
tests_high_five_base.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: static class boost::unit_test::lazy_ostream & __cdecl boost::unit_test::lazy_ostream::instance(void)" (__imp_?instance@lazy_ostream@unit_test@boost@@SAAEAV123@XZ) referenced in function "public: void __cdecl HighFiveBasic::test_method(void)" (?test_method@HighFiveBasic@@QEAAXXZ)
tests_high_five_base.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) protected: __cdecl boost::unit_test::lazy_ostream::lazy_ostream(bool)" (__imp_??0lazy_ostream@unit_test@boost@@IEAA@_N@Z) referenced in function "public: __cdecl boost::unit_test::lazy_ostream_impl<class boost::unit_test::lazy_ostream,char [1],char const (&)[1]>::lazy_ostream_impl<class boost::unit_test::lazy_ostream,char [1],char const (&)[1]>(class boost::unit_test::lazy_ostream const &,char const (&)[1])" (??0?$lazy_ostream_impl@Vlazy_ostream@unit_test@boost@@$$BY00DAEAY00$$CBD@unit_test@boost@@QEAA@AEBVlazy_ostream@12@AEAY00$$CBD@Z)
```
Please note these symbols have prefix `__declspec(dllimport) public:` which is incorrect.
So I checked the declaration:
```cpp
class BOOST_TEST_DECL lazy_ostream {
public:
```
Macro `BOOST_TEST_DECL` is declared in _include/boost/test/detail/config.hpp_ line 107:
```cpp
#  ifdef BOOST_TEST_SOURCE
#    define BOOST_TEST_DECL BOOST_SYMBOL_EXPORT BOOST_SYMBOL_VISIBLE
#  else
#    define BOOST_TEST_DECL BOOST_SYMBOL_IMPORT BOOST_SYMBOL_VISIBLE
#  endif  // BOOST_TEST_SOURCE
#else
#  if defined(BOOST_TEST_INCLUDED)
#     define BOOST_TEST_DECL
#  else
#     define BOOST_TEST_DECL BOOST_SYMBOL_VISIBLE
#  endif
#endif
```
Further more:
```cpp
#if defined(BOOST_ALL_DYN_LINK) && !defined(BOOST_TEST_DYN_LINK)
#  define BOOST_TEST_DYN_LINK
#endif

// in case any of the define from cmake/b2 is set
#if !defined(BOOST_TEST_DYN_LINK) \
    && (defined(BOOST_UNIT_TEST_FRAMEWORK_DYN_LINK) \
        || defined(BOOST_TEST_EXEC_MONITOR_DYN_LINK) \
        || defined(BOOST_PRG_EXEC_MONITOR_DYN_LINK) )
#  define BOOST_TEST_DYN_LINK
#endif

#if defined(BOOST_TEST_INCLUDED)
#  undef BOOST_TEST_DYN_LINK
#endif

#if defined(BOOST_TEST_DYN_LINK)
#  define BOOST_TEST_ALTERNATIVE_INIT_API

#  ifdef BOOST_TEST_SOURCE
#    define BOOST_TEST_DECL BOOST_SYMBOL_EXPORT BOOST_SYMBOL_VISIBLE
#  else
#    define BOOST_TEST_DECL BOOST_SYMBOL_IMPORT BOOST_SYMBOL_VISIBLE
#  endif  // BOOST_TEST_SOURCE
#else
#  if defined(BOOST_TEST_INCLUDED)
#     define BOOST_TEST_DECL
#  else
#     define BOOST_TEST_DECL BOOST_SYMBOL_VISIBLE
#  endif
#endif
```
So `BOOST_TEST_DYN_LINK` must be defined in some where, then I checked highfive's CMakeLists.txt, in _HIGHFIVE_SOURCE/tests/unit/CMakeLists.txt_ line 6:
```cmake
if(NOT Boost_USE_STATIC_LIBS)
  add_definitions(-DBOOST_TEST_DYN_LINK=TRUE)
endif()
```
I think that's reasonable because we should use dynamic declaration when linking to  boost dynamic libraries.
However we built boost as static library before.

Then I found this PR.

I think the reason why force `Boost_USE_STATIC_LIBS` to `OFF` is because you fix the `FindBoost` usage, but it affect some other stuff such like this issue.

Then I changed `Boost_USE_STATIC_LIBS` to `ON` manually, and got the find boost errors.
After some investigations, I think if we not defined `Boost_LIB_PREFIX`, `FindBoost.cmake` will always find boost library with prefix `lib` when `Boost_USE_STATIC_LIBS` is `ON` and `WIN32` is defined. But the fact is we do NOT pick this prefix in boost build process. Even I set it to empty before, it didn't work.

At last, I think if we keep this changes, we also need to modify the related declarations in _include/boost/test/detail/config.hpp_, Or we should completely fix this issues by not remove the official prefix `lib` when building static.

What do you think about?

cc @ras0219-msft @JackBoosY @PhoebeHui 
Related issue: https://github.com/microsoft/vcpkg/issues/1338 https://github.com/microsoft/vcpkg/issues/1745

Neumann-A:
@JonLiu1993 The code here is just copied over from vcpkg.cmake: https://github.com/microsoft/vcpkg/blob/431d3810aa9241aed49b14367be417f225d2af5e/scripts/buildsystems/vcpkg.cmake#L746
to make vcpkg handling of these settings more consistent. The old code in vcpkg.cmake is kept for backward compatibility

JackBoosY:
@Neumann-A Alright, I think we may need more discussion with @ras0219-msft here to confirm the fix solution.

JackBoos:
I suddenly realized that we can do some "hack" in the wrapper as before:
```cmake
unset(BOOST_TEST_DYN_LINK)
unset(BOOST_TEST_DYN_LINK CACHE)
```

JackBoosY:
> I suddenly realized that we can do some "hack" in the wrapper as before:
> 
> ```cmake
> unset(BOOST_TEST_DYN_LINK)
> unset(BOOST_TEST_DYN_LINK CACHE)
> ```

In https://github.com/BlueBrain/HighFive/blob/master/tests/unit/CMakeLists.txt#L5-L8:
```cmake
# NOTE: These unit tests expect linking with boost test framework lib
if(NOT Boost_USE_STATIC_LIBS)
  add_definitions(-DBOOST_TEST_DYN_LINK=TRUE)
endif()
```
`BOOST_TEST_DYN_LINK` is a c++ macro so we can't do that.

PR #18054, opened 2021-05-21
[scripts-audit] vcpkg_fail_port_install

strega-nil:
Note that this change allows existing failures;
it changes MATCHES to STREQUAL, meaning that
vcpkg_fail_port_install(ON_ARCH "arm") now succeeds on arm64;
this change in behavior is fine (IMO) since it strictly succeeds
when it used to fail.


strega-nil-ms:
Marking requires:discussion since this is a change in behaviour.

Neumann-A:
just deprecate the script. It was just created because there was no support field. 
The new approach should be:
parse vcpkg.json somewhere in port.cmake. Issue a warning/error if the port does not support the current target.

strega-nil:
@Neumann-A sure, that is also a thing I'd like for us to do.

strega-nil:
Closing for rollup (#18112)

PR #18055, opened 2021-05-21
vcpkg_configure_make: mingw fix: Use the vcpkg-acquired msys binaries for objdump / nm

past-due:
**Describe the pull request**
- mingw: Use the vcpkg-acquired msys binaries for `objdump` / `nm`
   - Otherwise mingw environments that use LLVM-based toolchains (example: the msys2 `clang64`, `clang32` environments) cannot build ports as their `objdump` is preferred and its output differs enough that it isn't properly parsed by `libtool`.

Cherry-picked from #18028

  only mingw triplets are impacted

### DIFF
--- a/scripts/cmake/vcpkg_configure_make.cmake
+++ b/scripts/cmake/vcpkg_configure_make.cmake
@@ -384,8 +384,14 @@ function(vcpkg_configure_make)
         else()
             _vcpkg_append_to_configure_environment(CONFIGURE_ENV RANLIB ":")
         endif()
-        if(VCPKG_DETECTED_CMAKE_OBJDUMP) #Objdump is required to make shared libraries. Otherwise define lt_cv_deplibs_check_method=pass_all
+        if(VCPKG_DETECTED_CMAKE_OBJDUMP AND NOT VCPKG_TARGET_IS_MINGW) #Objdump is required to make shared libraries. Otherwise define lt_cv_deplibs_check_method=pass_all
             _vcpkg_append_to_configure_environment(CONFIGURE_ENV OBJDUMP "${VCPKG_DETECTED_CMAKE_OBJDUMP}") # Trick to ignore the RANLIB call
+        elseif(VCPKG_TARGET_IS_MINGW)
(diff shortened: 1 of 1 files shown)


Neumann-A:
A few things:
- should vcpkg even try to acquire MSYS in this case? With MinGW isn't there already a working MSYS environment? (or at least one which can run bash and so on)
- Shouldn't the toolchain be adjusted in this cases instead of special casing in `vcpkg_configure_make`? Maybe a way to control `lt_cv_deplibs_check_method=pass_all` via the triplet is more appropiate?

past-due:
> A few things:
> 
> * should vcpkg even try to acquire MSYS in this case? With MinGW isn't there already a working MSYS environment? (or at least one which can run bash and so on)

I was seeing issues in the clang msys2 environments when the actual environment (LLVM-based) tools were being selected. (Hence relying on the acquired msys, which is guaranteed to have the gnu tools, and thus have more consistent behavior between regular Windows triplets and mingw triplets without relying on specifics of the potential msys environment.)

> * Shouldn't the toolchain be adjusted in this cases instead of special casing in `vcpkg_configure_make`? Maybe a way to control `lt_cv_deplibs_check_method=pass_all` via the triplet is more appropiate?

This seems like the better place to make the fix because it is only required for make/automake builds. I debated the latter measure, but this proposal does not require bypassing the libtool checks.

JackBoosY:
Does anyone object to this change?

JackBoosY (commented):
One thing I need to confirm: why objdump and nm are used in the build process?

dg0yt:
objdump and nm are used by `libtool` which deals with building shared libraries correctly.

I'm not fully convinced if there is a point in extending `vcpkg_acquire_msys` from providing the generic POSIX tools (for windows triplets) to providing toolchain components (for mingw triplets). In particular when the user probably comes from a more complete mingw environment anyway (MSYS2).

JackBoosY (approved):

past-due:
After discussion in the vcpkg Discord, closing this in favor of #18132

PR #18065, opened 2021-05-22
[portaudio] Switch to vcpkg_from_github, minor format improvements

h3ndrk:
**Describe the pull request**

- #### What does your PR fix?  
  - Update line numbers of existing patch
  - Switch from `vcpkg_from_git()` to `vcpkg_from_github()`
  - Convert `portfile.cmake` from DOS line-endings to Unix line-endings

  No changes, left untouched

  Yes (apart from the portfile line-endings)

  Yes


PhoebeHui (commented):
LGTM, @h3ndrk, thanks for your PR!

PhoebeHui (approved):

Hoikas:
IMO `add-namespace.patch` is inappropriate. Upstream should be in charge of their exported targets. By adding this patch, we add a compatibility break between the vcpkg built portaudio and the upstream.

ports/portaudio/portfile.cmake
```diff
@@ -1,44 +1,46 @@
...
+vcpkg_fail_port_install(ON_TARGET "uwp")
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO PortAudio/portaudio
+    REF 147dd722548358763a8b649b3e4b41dfffbcfbb6
+    TAG v19.7.0
+    SHA512 0f56e5f5b004f51915f29771b8fc1fe886f1fef5d65ab5ea1db43f43c49917476b9eec14b36aa54d3e9fb4d8bdf61e68c79624d00b7e548d4c493395a758233a
+    PATCHES
+        fix-library-can-not-be-found.patch
+        add-namespace.patch
```
ras0219-msft, line 10: As @Hoikas commented, upstream is responsible for their target names -- vcpkg should not break the official usage guidance.
```suggestion
```

ras0219-msft (changes requested):
Thanks for the PR!

We would be happy to accept just the change to `vcpkg_from_github()`, but not modifying the targets.

ports/portaudio/portfile.cmake
h3ndrk: Understood. Thanks for the feedback.

h3ndrk:
Thanks for the feedback. I opened https://github.com/PortAudio/portaudio/pull/583 and reverted the namespace patch. The CI seems to hang...

PhoebeHui (approved):

ports/portaudio/portfile.cmake
```diff
@@ -1,44 +1,45 @@
...
-endif()
-
-# Handle copyright
-file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+vcpkg_fail_port_install(ON_TARGET "uwp")
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO PortAudio/portaudio
+    REF 147dd722548358763a8b649b3e4b41dfffbcfbb6
+    TAG v19.7.0
```
strega-nil-ms, line 6:
```suggestion
```

strega-nil-ms (commented):
Only minor changes required, thanks @h3ndrk :)

ports/portaudio/portfile.cmake
h3ndrk: Oops, done.

PhoebeHui (approved):

ras0219-msft:
Thanks!

PR #18091, opened 2021-05-24
[doc] document vcpkg_fixup_pkgconfig merging behaviour

mcmtroffaes:
This also documents the fact that pkg-config is never to be called with the --static flag. Maybe something should be added to the maintainer guide as well (i.e. .pc files should always work without ``--static``), but for now, this seems like a good pace to add it.

See discussion in #17748.

cc @Neumann-A


PhoebeHui (approved):
Thank you @mcmtroffaes for your PR!

ras0219 (approved):

strega-nil-ms:
To be merged as part of the Friday Rollup :)

Thanks @mcmtroffaes :)

mcmtroffaes:
Great, and you're very welcome!

PR #18092, opened 2021-05-24
[WIP][curl][Android] Fix compile curl with zlib from vcpkg

### LINKED ISSUE #18015: [Android][curl] Wrong zlib on android compilation
Robert Schumacher Can I ask improvement of libcurl.pc when compile android?
pkg-config libcurl.pc --libs
-Llibcurl.pc/../../lib -lcurl -lc -lgcc -ldl -pthread libcurl.pc/../../lib/libssl.a libcurl.pc/../../lib/libcrypto.a /home/tal/boinc/3rdParty/android/android-ndk-r15c/platforms/android-16/arch-arm/usr/lib/libz.so

It give libz.so Although it compile libcurl.a with libz.a from vcpkg. Can we fix libcurl.pc that it will give the correct libz flag location? zlib.pc/../../lib/libz.a?

@ras0219-msft @ras0219 

---
Edit:
I find the problem that vcpkg find the zlib from the ndk instead the zlib inside vcpkg.

talregev:
Force curl to compile with zlib from vcpkg when exist other zlib in the system
Fix #18015


Neumann-A:
Missing proper explanation why zlib from the android NDK is found. This is not a single port problem but a wider problem in my opinion.

talregev:
@Neumann-A Explanation and prof found in the issue. If you still not find, please write here again.

Neumann-A:
> Also the fix is general, not just for android.

The fix is not general..... If other ports depend on zlib the same issue appears again. Have you tried `set(CMAKE_FIND_USE_PACKAGE_ROOT_PATH FALSE)` (or `ZLIB_ROOT`) in your toolchain file?
Also if you don't want to use zlib from your NDK then your toolchain is wrong because it explicitly is setup in a way that finds the dependencies there. If you don't want that you have to setup your own toolchain instead.

talregev:
@Neumann-A 
I can do many things. Like you say it more general then this port.
Please read this comment:
https://github.com/microsoft/vcpkg/issues/18015#issuecomment-847041378

talregev:
I think more people agree with me that vcpkg first need to find the things from the vcpkg. For example like lib curl that depend on zlib and compile vcpkg zlib because of it.  Also you want your library will be up to date for security issues and bug fixing.
So it a problem for a lot of people and Like you suggested, vcpkg need to fix it in more global way.
I give hear an humble suggestion how to fix only lib curl. 
I hope more people will see my PR and have a suggestion how vcpkg should solve it in the general case. (for all ports).

talregev:
Don't merge yet.
There is a general fix for this:
https://github.com/microsoft/vcpkg/pull/17336

NancyLi1013 (commented):
Could you please also update `port-version` in vcpkg.json file?

NancyLi1013:
Please run ` vcpkg x-add-version curl` to update version files after you complete all works.

talregev:
@NancyLi1013 I am freezing this PR until I will find out the result of this PR #17336.
If they cancel it out, I will increase up the version inside the json file.
Thank you for your time and review.

NancyLi1013:
Why do you close this PR?

talregev:
Because they merge this PR https://github.com/microsoft/vcpkg/pull/17336 that fix my problem.

Issue #17578, opened 2021-04-29
[tensorflow-cc:x64-windows] build failure


alexreinking:
**Host Environment**
- OS: Windows x64
- Compiler: MSVC 19.28.29913
- Vcpkg version: 2021-01-13-d67989bce1043b98092ac45996a8230a059a2d7e

**To Reproduce**
`./vcpkg install tensorflow-cc:x64-windows`

**Failure logs** 
```
D:\vcpkg>vcpkg install tensorflow-cc:x64-windows
Computing installation plan...
The following packages will be built and installed:
    tensorflow-cc[core]:x64-windows -> 2.4.1
  * tensorflow-common[core]:x64-windows -> 2.4.1#1
Additional packages (*) will be modified to complete this operation.
Detecting compiler hash for triplet x64-windows...
Could not locate cached archive: C:\Users\Alex Reinking\AppData\Local\vcpkg\archives\25\25ba2e73e538d1e14ecc334e9c5ce57ab80ff5f5.zip
Could not locate cached archive: C:\Users\Alex Reinking\AppData\Local\vcpkg\archives\71\713c99c62fcb7821e52e80038b3b5f1b27e4ae65.zip
Starting package 1/2: tensorflow-common:x64-windows
Building package tensorflow-common[core]:x64-windows...
-- Installing: D:/vcpkg/packages/tensorflow-common_x64-windows/share/tensorflow-common/copyright
-- Performing post-build validation
-- Performing post-build validation done
Stored binary cache: C:\Users\Alex Reinking\AppData\Local\vcpkg\archives\25\25ba2e73e538d1e14ecc334e9c5ce57ab80ff5f5.zip
Building package tensorflow-common[core]:x64-windows... done
Installing package tensorflow-common[core]:x64-windows...
Installing package tensorflow-common[core]:x64-windows... done
Elapsed time for package tensorflow-common:x64-windows: 121.6 ms
Starting package 2/2: tensorflow-cc:x64-windows
Building package tensorflow-cc[core]:x64-windows...
-- Downloading https://github.com/bazelbuild/bazel/releases/download/3.7.0/bazel-3.7.0-windows-x86_64.zip -> bazel-3.7.0-windows-x86_64.zip...
-- Downloading https://repo.msys2.org/msys/x86_64/unzip-6.0-2-x86_64.pkg.tar.xz -> msys-unzip-6.0-2-x86_64.pkg.tar.xz...
[...]

PR #18111, opened 2021-05-25
[tensorflow-common] check for invalid username

jgehw:
The `bazel` build tool used to build tensorflow does not support spaces in the username of the user running it. However, Windows allows spaces in the username (not only in the display name, which would be ok, but also in the native username string, which causes trouble). Thus add check to portfile, as this happens to users every couple of months, e.g. cf. #17578.


pratikpc:
@jgehw would the output_user_root work well as a temporary work around if we could pass it?

CMake can be used to remove the space and if we can, then we pass output_user_root to Bazel or we can change location of Bazel output_user_root for all files for current build?

NancyLi1013 (commented):
Can you also update port-version in vcpkg.json file?

jgehw:
> @jgehw would the output_user_root work well as a temporary work around if we could pass it?

@pratikpc thanks for pointing me at this. I added the suggested work-around to the portfile.

ports/tensorflow-common/tensorflow-common.cmake
```diff
@@ -219,8 +221,9 @@ foreach(BUILD_TYPE dbg rel)
 			list(JOIN COPTS " " COPTS)
 			list(JOIN CXXOPTS " " CXXOPTS)
 			list(JOIN LINKOPTS " " LINKOPTS)
+			# use --output_user_root to work-around too-long-path-names issue and username-with-spaces issue
 			vcpkg_execute_build_process(
-				COMMAND ${BASH} --noprofile --norc -c "'${BAZEL}' build --verbose_failures ${BUILD_OPTS} ${COPTS} ${CXXOPTS} ${LINKOPTS} --python_path='${PYTHON3}' --define=no_tensorflow_py_deps=true ///tensorflow:${BAZEL_LIB_NAME} ///tensorflow:install_headers"
+				COMMAND ${BASH} --noprofile --norc -c "'${BAZEL}' --output_user_root='${CURRENT_BUILDTREES_DIR}/.bzl' build --verbose_failures ${BUILD_OPTS} ${COPTS} ${CXXOPTS} ${LINKOPTS} --python_path='${PYTHON3}' --define=no_tensorflow_py_deps=true ///tensorflow:${BAZEL_LIB_NAME} ///tensorflow:install_headers"
```
ras0219, line 226: Does `--output_user_root=` do what `TEST_TMPDIR` did? It is very important that vcpkg does not modify the global system state.

ras0219: Looks like yes: https://docs.bazel.build/versions/3.1.0/install-windows.html#step-4-optional-configure-output-directories

ras0219 (approved):

ports/tensorflow-common/tensorflow-common.cmake
jgehw: Almost: The output directory defaults to `~/.cache/bazel` on Linux, `/private/var/tmp` on macOS, and on Windows it defaults to `%HOME%` if set, else `%USERPROFILE%` if set, else the result of calling `SHGetKnownFolderPath()` with the `FOLDERID_Profile` flag set.
That was overridden by setting `$TEST_TMPDIR` because path names got too long on Windows. Removing this environment variable restores the defaults on non-Windows (there didn't seem to be a reason why this was set on all platforms). Providing the `--output_user_root=` parameter on Windows causes the output still to be placed beneath `${CURRENT_BUILDTREES_DIR}/.bzl` but with the difference that the output is placed directly beneath that folder whereas the `$TEST_TMPDIR` way caused further subfolders, among them `_bazel_${USERNAME}` which caused the trouble here.

In case using  `~/.cache/bazel` on Linux and `/private/var/tmp` on macOS should not be allowed, I could change it back to  `$TEST_TMPDIR` on these platforms.

NancyLi1013:
Can you run `vcpkg x-add-version --overwrite-version tensorflow-common` to update version files again? @jgehw

ports/tensorflow-common/tensorflow-common.cmake
jgehw: Update: I decided to also use `${CURRENT_BUILDTREES_DIR}/.bzl` on non-Windows platforms as the tmp build output easily sums up to 100GB - so users only have to clean up one place to free disk space.

NancyLi1013 (approved):

strega-nil-ms:
This is nasty, but it's the correct solution. Thanks @jgehw :)

JackBoosY:
Hi @jgehw, recently we received some regressions about bazel that bazel has not released the control of _buildtrees\tensorflow-cc\.bzl\install\3d1356cc22901f271d47ddd780d95b50\A-server.ja_ after building tensorflow-cc.
This also means that the bazel process is not terminated after tensorflow-cc is built.

Do you have any ideas about this?

PR #18120, opened 2021-05-26
[netcdf-c] Fix pc files; use features

dg0yt:
- #### What does your PR fix?  
  This PR improves the netcdf-c port:
  - Improves the portfile (quoting, use vcpkg-cmake).
  - Fixes invalid pc-files.
  - Adds features to control dependencies.
  - Fixes building with mingw on Windows.

  all, no

  yes

  I am still working on this PR. TODO:
  - [x] Rebase actual netcdf-c changes on master when #17941(szip and hdf5 changes) is merged.
  - ... ~Look into update to NetCDF 4.8.0 (released April 02)~ Edit: I will put the NetCDF 4.8 in a separate PR.
  - [x] Update version


ports/hdf5/szip.patch
```diff
@@ -1,13 +1,16 @@
...
+     find_package (SZIP NAMES ${SZIP_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS static shared)
      if (NOT SZIP_FOUND)
        find_package (SZIP) # Legacy find
-       if (SZIP_FOUND)
+-      if (SZIP_FOUND)
++      if (DEFINED SZIP_SHARED_LIBRARY AND NOT DEFINED SZIP_STATIC_LIBRARY)
++          set (SZIP_STATIC_LIBRARY "${SZIP_SHARED_LIBRARY}")
++      elseif (DEFINED SZIP_STATIC_LIBRARY AND NOT DEFINED SZIP_SHARED_LIBRARY)
++          set (SZIP_SHARED_LIBRARY "${SZIP_STATIC_LIBRARY}")
++      elseif (SZIP_FOUND)
```
ras0219, line 13: This patch confuses me -- why do we not need to set `LINK_COMP_LIBS` iff only one of the `SZIP_XYZ_LIBRARY` variables are set?

ports/netcdf-c/fix-pkgconfig.patch
```diff
@@ -0,0 +1,36 @@
...
++macro(replace_pkgconfig_module PATTERN MODULE)
++  if(NC_LIBS_PRIVATE MATCHES "${PATTERN}")
++    string(REPLACE "${CMAKE_MATCH_0}" "" NC_LIBS_PRIVATE "${NC_LIBS_PRIVATE}")
++    string(APPEND NC_REQUIRES_PRIVATE " ${MODULE}")
++  endif()
++endmacro()
++replace_pkgconfig_module("-lhdf5_hl"   "hdf5_hl")
++replace_pkgconfig_module("-lhdf5"      "hdf5")
++replace_pkgconfig_module("-lCURL[^ ]*" "libcurl")
++replace_pkgconfig_module("-lZLIB[^ ]*" "zlib")
```
ras0219, line 20: It seems this should get the bonus spaces treatment
```suggestion
+set(NC_LIBS_PRIVATE " ${LIBS} ")
+set(NC_REQUIRES_PRIVATE "")
+macro(replace_pkgconfig_module PATTERN MODULE)
+  if(NC_LIBS_PRIVATE MATCHES " ${PATTERN} ")
+    string(REPLACE "${CMAKE_MATCH_0}" " " NC_LIBS_PRIVATE "${NC_LIBS_PRIVATE}")
+    string(APPEND NC_REQUIRES_PRIVATE " ${MODULE}")
+  endif()
+endmacro()
+replace_pkgconfig_module("-lhdf5_hl"   "hdf5_hl")
+replace_pkgconfig_module("-lhdf5"      "hdf5")
+replace_pkgconfig_module("-lCURL[^ ]*" "libcurl")
+replace_pkgconfig_module("-lZLIB[^ ]*" "zlib")
```

ports/netcdf-c/portfile.cmake
```diff
@@ -23,31 +25,57 @@ else()
     set(NC_USE_STATIC_CRT OFF)
 endif()
 
-vcpkg_configure_cmake(
-    SOURCE_PATH ${SOURCE_PATH}
+set(FEATURE_OPTIONS "")
```
ras0219, line 28: This set of if statements could be made much clearer with `vcpkg_check_features()` -- note that the same feature can be checked multiple times to set multiple variables.

ports/netcdf-c/vcpkg.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    {
+      "name": "vcpkg-cmake-config",
+      "host": true
+    }
+  ],
+  "default-features": [
+    "dap",
+    "netcdf-4"
+  ],
```
ras0219, line 20: Can these be off by default? In general, minimizing on-by-default features is best.

ports/szip/portfile.cmake
```diff
@@ -1,30 +1,68 @@
...
 )
 
 vcpkg_extract_source_archive_ex(
     OUT_SOURCE_PATH SOURCE_PATH
-    ARCHIVE ${ARCHIVE}
-    REF szip-2.1.1
+    ARCHIVE "${ARCHIVE}"
+    REF szip-${SZIP_VERSION}
     PATCHES
-        fix-szip-config-to-set-szip-found.patch # This patch is required for linux on osx; It does not matter for windows
```
ras0219, line 12: Why is this patch no longer required?

ports/szip/mingw-lib-names.patch
```diff
@@ -0,0 +1,30 @@
+diff -urN a/config/cmake/SZIPMacros.cmake b/config/cmake/SZIPMacros.cmake
+--- a/config/cmake/SZIPMacros.cmake	2021-05-15 09:06:50.546455249 +0200
++++ b/config/cmake/SZIPMacros.cmake	2021-05-15 09:01:24.933765423 +0200
+@@ -126,7 +126,7 @@
+ macro (SZIP_SET_BASE_OPTIONS libtarget libname libtype)
+   # message (STATUS "${libname} libtype: ${libtype}")
+   if (${libtype} MATCHES "SHARED")
+-    if (WIN32)
++    if (WIN32 AND NOT MINGW)
```
ras0219, line 9: Ideally we want to use upstream's library naming choices as much as possible; what are the names before and after this change?

ras0219 (commented):
Overall this looks awesome! I just have a few questions about some of the changes.

ports/hdf5/szip.patch
dg0yt: This is from  #17941, and I mentioned it there:
> hdf5 has a somewhat strange way of linking "components". It collects "static" and "shared" libs (zlib, szip) but links only to the static ones. That's why the first CI runs failed after changing szip to single linkage.

It is the minimal patch to resove it, and its needed now because szip no longer builds both types of linkage.

ports/netcdf-c/fix-pkgconfig.patch
dg0yt: Yes, I was considering an extra space, too. Glad to see it being appreciated.

ports/netcdf-c/vcpkg.json
dg0yt: I didn't want to change the current set of features at the moment, but maybe it is okay with the update to the next version.
I would suggest a different set of features:
- dap opt-in
- **netcdf-4 opt-out** but:
- szip (new) opt-in
- parallel (new) opt-in

netctdf-4 is based on hdf5 and is probably needed most of the time. 
szip is opt-out in hdf5. It has special licence terms for commercial use of derived works IIUC.
I would make szip opt-in in hdf5 at the same time.

ports/szip/portfile.cmake
dg0yt: #17941 again: szip is no longer building both thypes of linkage, and another patch takes care of setting the default component to match the linkage.

ports/szip/mingw-lib-names.patch
dg0yt: Please read #17941 from the beginning.

dg0yt:
@ras0219 I hoped the draft state and the first TODO list item would make it clear that this PR carried changes which don't need review here. But I needed a CI run before attempting to look at the new version.

I want to put the version update into a separate PR now. It will need a cleanup of patches.

ports/netcdf-c/portfile.cmake
```diff
@@ -23,31 +25,46 @@ else()
...
-    SOURCE_PATH ${SOURCE_PATH}
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        dap       ENABLE_DAP
+        netcdf-4  ENABLE_NETCDF_4
+        netcdf-4  USE_HDF5
+        tools     BUILD_UTILITIES
+    INVERTED_FEATURES
+        dap       CMAKE_DISABLE_FIND_PACKAGE_CURL
+        netcdf-4  CMAKE_DISABLE_FIND_PACKAGE_HDF5
```
PhoebeHui, line 36: Does this need? if disable it, how it to find hdf5 and curl?

dg0yt: > if disable it, how it to find hdf5 and curl?

If you disable a feature, the port shall not find packages which happen to be installed for other reasons.
When the `dap` feature isn't requested, CURL shall not be found. (`find_package(CURL)` is unguarded.)
When the `netcdf-4` feature isn't requested, HDF5 shall not be found. (`find_package(HDF5)` is guarded, but there are some interactions between netcdf-4 and hdf5 options which will also change with 4.8.0. This is a safe-guard.)

PhoebeHui: I see, thanks for your clarification!

PhoebeHui (approved):

dg0yt:
Ping. Blocking gdal improvement (#9068).

ras0219-msft:
Thanks!

PR #18123, opened 2021-05-26
[openexr] fixup pkgconfig

mcmtroffaes:
- #### What does your PR fix?  
  The generated .pc file for openexr contains an absolute path. This PR fixes that.

  No changes.

  Yes, to the best of my knowledge.

  Yes.


JonLiu1993 (approved):

dg0yt:
Unfortunately, it is no enough. For debug, libraries are build with a `_d` suffix:
~~~
$ ls packages/openexr_x64-linux-dynamic/debug/lib
libHalf-2_5_d.so            libIlmImf-2_5_d.so.24.0.0
libHalf-2_5_d.so.24         libIlmImfUtil-2_5_d.so
libHalf-2_5_d.so.24.0.0     libIlmImfUtil-2_5_d.so.24
libIex-2_5_d.so             libIlmImfUtil-2_5_d.so.24.0.0
libIex-2_5_d.so.24          libIlmThread-2_5_d.so
libIex-2_5_d.so.24.0.0      libIlmThread-2_5_d.so.24
libIexMath-2_5_d.so         libIlmThread-2_5_d.so.24.0.0
libIexMath-2_5_d.so.24      libImath-2_5_d.so
libIexMath-2_5_d.so.24.0.0  libImath-2_5_d.so.24
libIlmImf-2_5_d.so          libImath-2_5_d.so.24.0.0
libIlmImf-2_5_d.so.24       pkgconfig
~~~

*Fortunately*, given the `libsuffix=-2_5` line, it is probably easy to fix by substituting `=-2_5` with `=-2_5_d` in files matching glob `${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig/*.pc`.

mcmtroffaes:
Many thanks for catching @dg0yt, I appreciate you looking over this. Any objection to instead simply setting CMAKE_DEBUG_POSTFIX to the empty string as in the commit I just added?

mcmtroffaes:
(Note https://github.com/AcademySoftwareFoundation/openexr/blob/master/cmake/OpenEXRSetup.cmake#L72 so it really needs to be explicitly set to the empty string.)

dg0yt:
> Any objection to instead simply setting CMAKE_DEBUG_POSTFIX to the empty string as in the commit I just added?

IMO it is fine, as long as library names match exported cmake and pkgconfig configuration. It seems hardly possibly to use these libs correctly without exported configuration. (I will know more when I add openexr support to port gdal.)

However, [vcpkg policy is to not rename binaries outside the names given by upstream](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-rename-binaries-outside-the-names-given-by-upstream). Decision left to approvers.

mcmtroffaes:
Thanks, and sure, I asked explicitly because of that policy. Using an upstream configurable setting to fix issues with the upstream exported .pc configuration would seem within reasonable boundaries to me and guarantees all configurations (cmake, pkgconfig) remain in sync; it's not like renaming the libraries post-installation.

But if it's not ok, I'd be happy to try to fix the openexr .pc generation with a patch that can be upstreamed. However, I really rather go for the currently proposed and much simpler solution, not requiring any special case regexp-ing of the .pc files, and not requiring any upstreamable patches.

Neumann-A:
> However, vcpkg policy is to not rename binaries outside the names given by upstream. Decision left to approvers.

Policy is policy. You don't know if a downstream port has a custom `Find<Whatever>` which depends on the upstream naming schema.

mcmtroffaes:
Sure. I've posted a patch upstream.

mcmtroffaes:
There's some non-trivial discussion upstream on how best to fix this. I'll put this back on draft until the outcome of that discussion is clear, so I can add the final upstream patch.

JackBoosY:

mcmtroffaes:
@JackBoosY Thanks, your message was perfectly timed: upstream just merged the patch! It's ready now for review.

JackBoosY (approved):

strega-nil-ms:
LGTM, thanks @mcmtroffaes :)

Issue #16630, opened 2021-03-09
[boost] linking failure on Mac M1 arm64-osx


cdcseacave:
**Host Environment**
- OS: MacOS arm64-osx
- Compiler: clang 12

**To Reproduce**
`./vcpkg install boost`
followed by trying to link to boost in any app.

**Failure logs** 
Many error of this type:
```
ld: warning: ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_program_options.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_program_options.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64

ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_iostreams.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_iostreams.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_system.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_system.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_regex.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file /Users/dancostin/Pro/vcpkg/installed/arm64-osx/lib/libboost_regex.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
```

cdcseacave:
I don't think that helps, as I forgot to say in the above description that I set the environment variable `export VCPKG_DEFAULT_TRIPLET=arm64-osx`

cdcseacave:
I have just removed boost and installed it again with explicit triplet command line parameter as suggested, but nothing changed. It is reported that boost is installed for arm64, but the errors are the same as above when trying to link:
```The following packages will be built and installed:
[...]

Issue #11132, opened 2020-05-02
[boost-modular-build-helper] B2 ignores C/CXX/LDFLAGS when toolchain is not MSVC or GCC


psfrolov:
# Problem
In `user-config.jam` we have:
```
if "@VCPKG_PLATFORM_TOOLSET@" != "external"
{
    using msvc : : cl.exe
        :
        <setup>"@NOTHING_BAT@"
        @TOOLSET_OPTIONS@
        ;
}
else
{
    using gcc : 5.4.1 : @CMAKE_CXX_COMPILER@
        :
        <ranlib>@CMAKE_RANLIB@
        <archiver>@CMAKE_AR@
        @CXXFLAGS@
        @CFLAGS@
        @LDFLAGS@
        # MINGW here causes b2 to not run cygpath
        <flavor>mingw
        ;
}
```
That said if toolset is not `msvc` we are going `else` route with toolset set to `gcc` `54.1`. This mostly works but have subtle side effects if compiler we are building with is not in fact GCC. For example, if we are building with Clang then `using gcc` makes B2 to ignore all the options in `using` block because `clang` is not `gcc`. Thus no compiler/linker options from Vcpkg code and user triplets are passed to build.

# Proposed solution
1. Make toolset and version in `user-config.jam` configurable:
```
using @VCPKG_B2_COMPILER_ID@ : @CMAKE_CXX_COMPILER_VERSION@ : @CMAKE_CXX_COMPILER@
```
2. In `CMakeLists.txt` map CMake toolchain ID to B2 toolchain ID:
```cmake
# Determine compiler ID to use in user-config.jam
# CMake compiler IDs: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html
# B2 compiler IDs: https://boostorg.github.io/build/manual/master/index.html#bbv2.reference.tools.compilers
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    set(VCPKG_B2_COMPILER_ID gcc)
elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")  # AppleClang, Clang, XLClang
    set(VCPKG_B2_COMPILER_ID clang)
elseif(...)
    ...
else()
    # Either hard fail or fallback to GCC and issue a warning.
    # Hard fail is better because alternative is undefined behaviour.
endif()
```
[...]

PR #18124, opened 2021-05-26
[boost-modular-build-helper] Bugfix/boost arm64 ios

StefAviator:
Building boost for arm64 ios created x86_64 .a libraries. This is fixed by using a different modular builder, using gcc 12.0.0 and adding compiler options to the user-config.jam file.

### DIFF
--- a/ports/boost-modular-build-helper/boost-modular-build.cmake
+++ b/ports/boost-modular-build-helper/boost-modular-build.cmake
@@ -99,7 +99,7 @@ function(boost_modular_build)
         vcpkg_install_cmake()
     endfunction()
 
-    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
+    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND NOT (VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64" AND (VCPKG_CMAKE_SYSTEM_NAME STREQUAL "iOS" OR VCPKG_CMAKE_SYSTEM_NAME STREQUAL "Darwin")))
         set(build_flag 0)
         if(NOT DEFINED VCPKG_BUILD_TYPE)
             set(build_flag 1)
--- a/ports/boost-modular-build-helper/boost-modular-build.cmake
+++ b/ports/boost-modular-build-helper/boost-modular-build.cmake
@@ -99,7 +99,7 @@ function(boost_modular_build)
         vcpkg_install_cmake()
     endfunction()
 
-    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
+    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
         set(build_flag 0)
         if(NOT DEFINED VCPKG_BUILD_TYPE)
             set(build_flag 1)
--- a/ports/boost-modular-build-helper/user-config.jam
+++ b/ports/boost-modular-build-helper/user-config.jam
@@ -8,7 +8,7 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
         @TOOLSET_OPTIONS@
         ;
 }
-else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64"
+else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64" && "@VCPKG_CMAKE_SYSTEM_NAME@" = "iOS"
 {
     using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
         :
--- a/ports/boost-modular-build-helper/user-config.jam
+++ b/ports/boost-modular-build-helper/user-config.jam
@@ -19,10 +19,23 @@ else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64" && "@VCPKG_CMAKE_SYSTEM_NAME@" =
         @LDFLAGS@
         # MINGW here causes b2 to not run cygpath
         <flavor>mingw
-        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
         ;        
 }
-
+else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64" && "@VCPKG_CMAKE_SYSTEM_NAME@" = "Darwin"
+{
+    using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
+        :
+        <ranlib>@CMAKE_RANLIB@
+        <archiver>@CMAKE_AR@
+        @CXXFLAGS@
+        @CFLAGS@
+        @LDFLAGS@
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
+        ; 
+}
 else
 {
     using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
--- a/ports/boost-modular-build-helper/user-config.jam
+++ b/ports/boost-modular-build-helper/user-config.jam
@@ -19,7 +19,7 @@ else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64"
         @LDFLAGS@
         # MINGW here causes b2 to not run cygpath
         <flavor>mingw
-        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk"
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
         ;        
 }
 
--- a/ports/boost-modular-build-helper/user-config.jam
+++ b/ports/boost-modular-build-helper/user-config.jam
@@ -8,9 +8,24 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
         @TOOLSET_OPTIONS@
         ;
 }
+else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64"
+{
+    using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
+        :
+        <ranlib>@CMAKE_RANLIB@
+        <archiver>@CMAKE_AR@
+        @CXXFLAGS@
+        @CFLAGS@
+        @LDFLAGS@
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk"
+        ;        
+}
+
 else
 {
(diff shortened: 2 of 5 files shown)


ports/boost-modular-build-helper/boost-modular-build.cmake
```diff
@@ -99,7 +99,7 @@ function(boost_modular_build)
         vcpkg_install_cmake()
     endfunction()
 
-    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
+    if(VCPKG_CMAKE_SYSTEM_NAME AND NOT VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore" AND NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
```
StefAviator, line 102: To compile arm64 for ios, we need to skip this builder and use the one below which allows the use of the user-config.jam file

ports/boost-modular-build-helper/user-config.jam
```diff
@@ -8,9 +8,24 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
...
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk"
+        ;        
+}
+
 else
 {
-    using gcc : 5.4.1 : @CMAKE_CXX_COMPILER@
+    using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
```
StefAviator, line 28: Updating to the latest gcc.

ports/boost-modular-build-helper/user-config.jam
```diff
@@ -8,9 +8,24 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
...
+        :
+        <ranlib>@CMAKE_RANLIB@
+        <archiver>@CMAKE_AR@
+        @CXXFLAGS@
+        @CFLAGS@
+        @LDFLAGS@
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk"
+        ;
```
StefAviator, line 23: This compilerflag is needed to compile arm64 for iOS

PhoebeHui:
@StefAviator, thanks for the PR! Since the triplet doesn't cover in CI testing, could you test it locally?

ports/boost-modular-build-helper/boost-modular-build.cmake
PhoebeHui: This may need to limit to ios?

PhoebeHui (changes requested):
Could you also update the port version to '10'  in https://github.com/microsoft/vcpkg/blob/master/ports/boost-modular-build-helper/vcpkg.json#L4? Submit the changes and execute './vcpkg x-add-version boost-modular-build-helper' to update the baseline version, then submit the changes again.

ports/boost-modular-build-helper/boost-modular-build.cmake
StefAviator: Ok agreed, I will check for VCPKG_TARGET_ARCHITECTURE.

StefAviator:
> @StefAviator, thanks for the PR! Since the triplet doesn't cover in CI testing, could you test it locally?

On my MacBook Pro, building boost components with commands such as 
`./vcpkg install boost-exception:arm64-ios`
now outputs .a files which show 'arm64' when checked with 'lipo -info'. Previously it would show as 'x86-64'. Is that enough for testing purposes? If I need to attach any log files let me know.

I don't have a device with M1 chipset currently but would appreciate if someone can confirm the output. I have asked someone in the original issue.

ports/boost-modular-build-helper/boost-modular-build.cmake
vejmartin: @StefAviator, thank you for this PR! The code path below seems to be tailored specifically for Windows, e.g., it doesn't seem to copy over any `.dylib` files as opposed to the `CMakeLists.txt` file. The `unix_build` function goes through the `CMakeLists.txt` file which does seem to use the `user-config.jam` file:
```
add_custom_target(boost ALL
    COMMAND "${B2_EXE}"
        --user-config=${CMAKE_CURRENT_BINARY_DIR}/user-config.jam
```

Were there any errors using the the unix build?

ports/boost-modular-build-helper/user-config.jam
vejmartin: There's a way to specify the iOS target via the `CMakeLists.txt` file, which could remove the need to change the `user-config.jam`. Specifically the line that specifies the `target-os` option:
```
if(APPLE)
    list(APPEND B2_OPTIONS target-os=darwin toolset=clang)
elseif(WIN32)
    list(APPEND B2_OPTIONS target-os=windows toolset=gcc)
elseif(ANDROID)
```
For iOS, that would be `target-os=iphone` [docs](https://www.boost.org/doc/libs/1_76_0/tools/build/doc/html/index.html#bbv2.builtin.features.target-os)

It would also be nice to make this work for both arm64 iOS and macOS builds, i.e. the `arm64-ios` and `arm64-osx` triplets:
```
set(VCPKG_TARGET_ARCHITECTURE arm64)
set(VCPKG_CMAKE_SYSTEM_NAME iOS)
```
and
```
set(VCPKG_TARGET_ARCHITECTURE arm64)
set(VCPKG_CMAKE_SYSTEM_NAME Darwin)
```

Boosts's b2 doesn't natively support `arm64`, so the extra `-arch arm64` can be added to the `CMakeLists.txt` config like so:
```
if(APPLE AND VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
  set(CFLAGS "${CFLAGS} <compileflags>-arch <compileflags>arm64")
  set(CXXFLAGS "${CXXFLAGS} <compileflags>-arch <compileflags>arm64")
  set(LDFLAGS "${LDFLAGS} <linkflags>-arch <linkflags>arm64")  
endif()
```

vejmartin (changes requested):

ports/boost-modular-build-helper/boost-modular-build.cmake
wintergt: Thanks for the feedback, I will try out your suggestions.

ports/boost-modular-build-helper/user-config.jam
StefAviator: I have tried adding these compilerflags to CMakeLists.txt but it keeps outputting x86-64 library files. I tried this change using both unix_build and the second builder. If you know how to get this to work to cross-compile on an Apple PC/laptop with Intel chip to arm64 by adjusting the CMakeLists file, please share.

StefAviator:
I have added a check for either 'iOS' or 'Darwin' system (the latter per @vejmartin's suggestion) to use this build fix. Now it also supports cross-compiling for arm64 osx. I was not able to get it to work using adjustments to the CMakeLists.txt file that @vejmartin suggested so it is necessary to keep using the user-config.jam file update. The functionality can be confirmedted as such:

```
./vcpkg install boost-exception:arm64-ios boost-atomic:arm64-osx
-- (build)
~/vcpkg $ lipo -info installed/arm64-ios/lib/libboost_exception.a 
Non-fat file: installed/arm64-ios/lib/libboost_exception.a is architecture: arm64
~/vcpkg $ lipo -info installed/arm64-osx/lib/libboost_atomic.a   
Non-fat file: installed/arm64-osx/lib/libboost_atomic.a is architecture: arm64
```

We have had feedback that using the current implementation (vcpkg master) works for compiling on an arm64 system (M1 chip macbook) to create arm64 boost library files. So this fix is limited to fixing cross-compiling from intel Apple PC/laptop (x86-64) to arm64 boost libraries. As such it might be interesting to check if the host Apple system is arm64, in which case this fix of using user-config.jam file can be bypassed. I would appreciate some feedback if this is deemed necessary and how this can be done. Thanks.

ports/boost-modular-build-helper/user-config.jam
```diff
@@ -8,9 +8,37 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
...
+        <archiver>@CMAKE_AR@
+        @CXXFLAGS@
+        @CFLAGS@
+        @LDFLAGS@
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk"
+        ;        
+}
+else if "@VCPKG_TARGET_ARCHITECTURE@" = "arm64" && "@VCPKG_CMAKE_SYSTEM_NAME@" = "Darwin"
```
wintergt, line 25: "Darwin" specifies is used in the triplet for unix / osx / MacOS.

ports/boost-modular-build-helper/user-config.jam
```diff
@@ -8,9 +8,37 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
...
+    using gcc : 12.0.0 : @CMAKE_CXX_COMPILER@
+        :
+        <ranlib>@CMAKE_RANLIB@
+        <archiver>@CMAKE_AR@
+        @CXXFLAGS@
+        @CFLAGS@
+        @LDFLAGS@
+        # MINGW here causes b2 to not run cygpath
+        <flavor>mingw
+        <compileflags>"-std=c++14 -stdlib=libc++ -DNDEBUG -arch arm64 -fembed-bitcode -Wno-unused-local-typedef -Wno-nullability-completeness -mmacosx-version-min=10.12 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
```
wintergt, line 36: We use the default MacOS sdk on the system.

StefAviator, line 36: Here we are using the default MacOS sdk on the system. "Darwin" above is used in the triplet for unix / osx / MacOS.

StefAviator:
> Could you also update the port version to '10' in https://github.com/microsoft/vcpkg/blob/master/ports/boost-modular-build-helper/vcpkg.json#L4? Submit the changes and execute './vcpkg x-add-version boost-modular-build-helper' to update the baseline version, then submit the changes again.

Ok done.

vejmartin:
@StefAviator, I think the reason why the options weren't being picked up by the user-config might have been that the `CMakeLists.txt` file sets the toolset to clang:
```
list(APPEND B2_OPTIONS target-os=darwin toolset=clang)
``` 
while, the user-config doesn't define a clang toolset (only gcc):
```
using gcc : 5.4.1 : @CMAKE_CXX_COMPILER@
```

By conditionally converting the `gcc` toolset to `clang`, the options from `CMakeLists.txt` are picked up again:
```
if "@CMAKE_CXX_COMPILER_ID@" in "Clang" "AppleClang" 
{
    compiler_name = "clang" ;
    compiler_version = "@CMAKE_CXX_COMPILER_VERSION@" ;
}
else
{
    compiler_name = "gcc" ;
    compiler_version = "5.4.1" ;
}
using $(compiler_name) : $(compiler_version) : @CMAKE_CXX_COMPILER@
```
Please also note that the user might specify a custom SDK root via the `SDKROOT` environment variable.
Here's a working example with CMakeLists.txt for arm64 iOS/macOS: https://github.com/vejmartin/vcpkg/commit/18d39b98dc05222bbbb35c764c476d185995a80e

StefAviator:
> Here's a working example with CMakeLists.txt for arm64 iOS/macOS: [vejmartin@18d39b9](https://github.com/vejmartin/vcpkg/commit/18d39b98dc05222bbbb35c764c476d185995a80e)

@vejmartin Thanks for the code. Unfortunately it doesn't work on my system. I have both gcc and clang installed. If I use your code as written, it still uses the gcc compiler and outputs x86_64 libraries, the original problem. If I remove the 'if' statement and force it to use clang in boost-modular-build.cmake, it ends with error. I can see in the logs (attached) that it starts off using clang, but then proceeds with gcc anyway. Possibly B2 uses gcc, but that part is unclear. 

[build-arm64-ios-rel-out.log](https://github.com/microsoft/vcpkg/files/6631020/build-arm64-ios-rel-out.log)

So far my implementation using the user-config file is the only solution that allows cross-compiling from x86 system to output boost arm64 libs on my system. If you have any suggestions feel free to share.

vejmartin:
@StefAviator, thank you for looking into and testing my example. 

Thank you also for the build log. I'm assuming `build-arm64-ios-rel-out.log` is from using `boost-modular-build.cmake` on macOS, as the `CMakeLists.txt` method produces `install-arm64-ios-rel-out.log` instead.
> Unfortunately it doesn't work on my system.

Could you please attach the `install-arm64-ios-rel-out.log` log when using the `CMakeLists.txt` method or any other errors?

If we could work out what's causing the error when using the `CMakeLists.txt` method, we could ensure there won't be two separate files/methods for building `arm64` and `x86_64` on macOS.

As `boost-modular-build.cmake` isn't currently used to build on x64-osx*, it doesn't support dynamic libraries. You can verify that by building `boost-system:arm64-osx-dynamic` (please note that `boost-exception` always produces a static lib). This should produce the following file: `install/arm64-osx-dynamic/lib/libboost_system.dylib`.

I also found the following comment useful for debugging: https://github.com/boostorg/build/issues/716#issuecomment-825196282:
> run b2 with -d+2 flag to make logs useful. A log with --debug-building may later be required to debug too.

You could add these two flags to the build command in `CMakeLists.txt` (add_custom_target) to get more info in the logs.

vejmartin:
These issues also seem to be related, https://github.com/microsoft/vcpkg/issues/16630, https://github.com/microsoft/vcpkg/issues/11132

PhoebeHui:
@StefAviator @vejmartin, is work still being done for this PR?

StefAviator:
@PhoebeHui Yes this is still ongoing. Unfortunately I was not able to get @vejmartin 's suggestions to work. That is to say, while I would prefer to do it his way, it still only outputs x86_64 libraries instead of arm64. I will do one more update this week to address an issue and then let david-antiteum test it (he has arm64 chip mac) and then it can be included as i believe there is currently no other working version that outputs iOS arm64.

vejmartin:
@StefAviator, a fix for the toolchain options was recently merged, https://github.com/microsoft/vcpkg/pull/18529, which also fixes arm64-osx/ios builds for me.

StefAviator:
@vejmartin @PhoebeHui This is indeed fixed by [#18529](https://github.com/microsoft/vcpkg/pull/18529) for arm64-osx/ios builds, confirmed as well on my system, they produce arm64 libraries. We can close this PR as it's not needed anymore.

PhoebeHui:
@vejmartin @StefAviator, thanks for conforming this!

PR #18145, opened 2021-05-27
[x265] x265 Main10 and Main12 support

### LINKED ISSUE #13875: [x265] add multilib support
Add x265 compilation options with multilib support for 8, 10 and 12 bit color depths.

a1q123456:
Added two new features to support main10 and main12 profile of x265, tested on windows and linux

- #### What does your PR fix?  
  Fixes #13875

x64-linux x86-windows x86-windows-static

  Yes

  Yes

### DIFF
--- a/ports/x265/CONTROL
+++ b/ports/x265/CONTROL
@@ -1,6 +1,12 @@
 Source: x265
 Version: 3.4
-Port-Version: 4
+Port-Version: 5
 Homepage: https://github.com/videolan/x265
 Description: x265 is a H.265 / HEVC video encoder application library, designed to encode video or images into an H.265 / HEVC encoded bitstream.
 Supports: !(uwp | arm)
+
+Feature: bit10
+Description: 10-bit support
+
+Feature: bit12
+Description: 12-bit support
--- a/ports/x265/portfile.cmake
+++ b/ports/x265/portfile.cmake
@@ -1,3 +1,6 @@
+vcpkg_internal_get_cmake_vars(OUTPUT_FILE _VCPKG_CMAKE_VARS_FILE)
+include("${_VCPKG_CMAKE_VARS_FILE}")
+
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO videolan/x265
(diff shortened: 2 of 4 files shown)


ports/x265/CONTROL
```diff
@@ -1,6 +1,12 @@
...
+Port-Version: 5
 Homepage: https://github.com/videolan/x265
 Description: x265 is a H.265 / HEVC video encoder application library, designed to encode video or images into an H.265 / HEVC encoded bitstream.
 Supports: !(uwp | arm)
+
+Feature: bit10
+Description: 10-bit support
+
+Feature: bit12
+Description: 12-bit support
```
PhoebeHui, line 12: Have you test the 2 features? I suspect that './vcpkg install x265[bit10,bit12]' would install one of the feature, which break our policy https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-implement-alternatives.

PhoebeHui (changes requested):

ports/x265/CONTROL
a1q123456: I think  './vcpkg install x265[bit10,bit12]' installs a x265 with 8bit + 10bit + 12bit support. So if a library requires x265 main10 profile and another library requires main12 profile, install x265[bit10,bit12] would make both main10 and main12 available in x265

This portfile actually compiles every version of x265 selected by user, then links them together, just like 'build/vcxx-x86_64/multilib.bat' in the source tree of x265 does

a1q123456: This is the help message printed by x265 installed with bit10 and bit12

PS D:\my_vcpkg\vcpkg\packages\x265_x64-windows\tools\x265> .\x265.exe --version
x265 [info]: HEVC encoder version 0.0
x265 [info]: build info [Windows][MSVC 1927][64 bit] ***8bit+10bit+12bit***
x265 [info]: using cpu capabilities: MMX2 SSE2Fast LZCNT SSSE3 SSE4.2 AVX FMA3 BMI2 AVX2

a1q123456: And if users only install bit12, only main profile(core feature) and main12(bit12 feature) profile is available for x265

PhoebeHui: I see, thanks for clarifying this!

ports/x265/portfile.cmake
```diff
@@ -18,17 +21,173 @@ endif ()
...
+        ${FEATURE_OPT_REL}
+        -DEXTRA_LIB=${EXTRA_LIB_REL}
     OPTIONS_DEBUG
         -DENABLE_CLI=OFF
+        ${FEATURE_OPT_DBG}
+        -DEXTRA_LIB=${EXTRA_LIB_DBG}
 )
 
-vcpkg_install_cmake()
+vcpkg_build_cmake()
```
JackBoosY, line 130: Can you please write a macro or function and foreach the features?

PhoebeHui: The VCPKG_TARGET_SHARED_LIBRARY_SUFFIX and VCPKG_TARGET_STATIC_LIBRARY_PREFIX can be used to improve this.

a1q123456: Okay, I’ll rewrite these stuff

a1q123456: I've made some change for this file. Can you please review it again? @JackBoosY @PhoebeHui

ports/x265/portfile.cmake
```diff
@@ -18,17 +31,181 @@ endif ()
...
     OPTIONS_DEBUG
-        -DENABLE_CLI=OFF
+        ${OPT_DBG}
+    )
+    vcpkg_build_cmake()
+    
+    set(x265_lib "x265-${build_x265_BIT}")
+    
+    if (NOT BUILDING_CORE_FEATURE)
+        foreach(BUILDTYPE "debug" "release")
```
JackBoosY, line 119: Why is it divided into debug and release? In fact, each call of `vcpkg_*_cmake` includes debug and release.

ports/x265/portfile.cmake
```diff
@@ -18,17 +31,181 @@ endif ()
...
+
+                file(COPY ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${SHORT_BUILDTYPE}/${x265_lib_name} 
+                    DESTINATION ${CURRENT_BUILDTREES_DIR}/tmp-${TARGET_TRIPLET}-${SHORT_BUILDTYPE}/${x265_lib})
+                list(APPEND ${FEATURE_OPT_BUILDTYPE_VAR} "-DLINKED_${build_x265_BIT}BIT=ON")
+                file(REMOVE_RECURSE ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${SHORT_BUILDTYPE})
+                list(APPEND ${EXTRA_LIB_BUILDTYPE_VAR} "${CURRENT_BUILDTREES_DIR}/tmp-${TARGET_TRIPLET}-${SHORT_BUILDTYPE}/${x265_lib}/${x265_lib_name}")
+            endif()
+        endforeach()
+    elseif(NOT ENABLE_SHARED)
+        foreach(BUILDTYPE "debug" "release")
```
JackBoosY, line 138: See comment above.

JackBoosY (changes requested):

ports/x265/portfile.cmake
```diff
@@ -18,17 +31,181 @@ endif ()
...
+                    if(VCPKG_TARGET_IS_WINDOWS)
+                        execute_process(COMMAND ${VCPKG_DETECTED_CMAKE_AR} /ignore:4006 /ignore:4221 /OUT:${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${SHORT_BUILDTYPE}/${x265_lib_name} ${LIB_FILES}  RESULT_VARIABLE AR_EXIT_CODE OUTPUT_VARIABLE AR_STDOUT ERROR_VARIABLE AR_STDERR)
+                    elseif(VCPKG_TARGET_IS_LINUX)
+                        string(REPLACE ";" "\nADDLIB " LIB_FILES "${LIB_FILES}")
+                        execute_process(COMMAND "sh" "-c" "${VCPKG_DETECTED_CMAKE_AR} -M <<EOF
+CREATE ${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${SHORT_BUILDTYPE}/${x265_lib_name}
+ADDLIB ${LIB_FILES}
+SAVE
+END
+EOF" RESULT_VARIABLE AR_EXIT_CODE OUTPUT_VARIABLE AR_STDOUT ERROR_VARIABLE AR_STDERR)
```
JackBoosY, line 161: I think we should patch the CMakeLists.txt instead of add code here.

JackBoosY (changes requested):

JackBoosY:
I will help you to fix that.
According to our policy, we should patch x265 cmake build system (`CMakeLists.txt` and `*.cmake`) to implement that.
And we should push the changes to the upstream, wait for the upstream approval first.

But first, you need to answer my questions above.

ports/x265/portfile.cmake
```diff
@@ -8,8 +11,18 @@ vcpkg_from_github(
...
+endif()
+
+set(HAS_12_BIT OFF)
+if("bit12" IN_LIST FEATURES)
+    set(HAS_12_BIT ON)
+endif()
+
 set(ENABLE_ASSEMBLY OFF)
-if (VCPKG_TARGET_IS_WINDOWS)
+if (VCPKG_TARGET_IS_WINDOWS AND (NOT HAS_12_BIT AND NOT HAS_10_BIT))
```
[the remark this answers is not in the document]
a1q123456: This is because if we enable the option HIGH_BIT_DEPTH and assembly code, x265 will build an assembly file `intrapred16.asm`, and this file uses ymm16 register(at line 18657) which is not supported in avx x86 mode. So I think it's a bug of x264, maybe we should remove it and let upstream to fix this issue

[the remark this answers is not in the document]
a1q123456: > Can you explain why?

This is because if we enable the option HIGH_BIT_DEPTH and assembly code, x265 will build an assembly file intrapred16.asm, and this file uses ymm16 register(at line 18657) which is not supported in avx x86 mode. So I think it's a bug of x264, maybe we should remove it and let upstream to fix this issue

ports/x265/portfile.cmake
a1q123456: > Why is it divided into debug and release? In fact, each call of `vcpkg_*_cmake` includes debug and release.

I think replace the code below by functions from vcpkg_*.cmake maybe difficult, because this macro calls vcpkg_configure_cmake and vcpkg_build_cmake,  to generate different version of x265, and merge them together later, so we still need paths of different version of x265 in order to merge them.

ports/x265/portfile.cmake
a1q123456: > I think we should patch the CMakeLists.txt instead of add code here.

I'm not sure we should patch CMakeLists.txt to add some more build steps because these code is just a translation of build command line scripts from x265. But the portfile could remain just simple if we patch the CMakeLists.txt, this is a pro

JackBoosY:
According to our policy, conflicting features should not be added.
So we can only add 10bit OR 12bit as feature.

a1q123456:
> According to our policy, conflicting features should not be added.
> So we can only add 10bit OR 12bit as feature.

Hi JackBoosY,

I think they are not conflicting, users can install both 10-bit version and 12-bit version, and they will get only what they installed. For example, they'll not have 10-bit support even if 12-bit version had installed.

JackBoosY:
In _SOURCE/CMakeLists.txt_:
```cmake
elseif(HIGH_BIT_DEPTH)
    if(MAIN12)
        set(X265_NS x265_12bit)
    else()
        set(X265_NS x265_10bit)
    endif()
```

So I think that only the tool can use all bit options, not the library. 
Since the tool is only a feature, we can only let the tool use the features activated by the current port.

a1q123456:
> In _SOURCE/CMakeLists.txt_:
> 
> ```cmake
> elseif(HIGH_BIT_DEPTH)
>     if(MAIN12)
>         set(X265_NS x265_12bit)
>     else()
>         set(X265_NS x265_10bit)
>     endif()
> ```
> 
> So I think that only the tool can use all bit options, not the library.
> Since the tool is only a feature, we can only let the tool use the features activated by the current port.

Actually, the portfile.cmake performs build operation multiple times, then archive the binaries together. So the code above may run multiple times, one for MAIN12 branch, one for the else branch, and one for the if branch corresponding to the elssif (HIGH_BIT_DEPTH) branch

JackBoosY:
@a1q123456  Yes, however, the user can only use one generated library. is it?

a1q123456:
> @a1q123456 Yes, however, the user can only use one generated library. is it?
They can use any version they installed, x265 uses an api entry `x265_api_get(int bit_depth)` to select different bit depth version of x265. For example, when a user call it with parameter 10, this function loads the 10-bit version of x265, And x265 uses namespace to distinguish different bit-depth version of x265 libraries.

a1q123456:
> @a1q123456 Yes, however, the user can only use one generated library. is it?

They can use any version they installed, x265 uses an api entry x265_api_get(int bit_depth) to select different bit depth version of x265. For example, when a user call it with parameter 10, this function loads the 10-bit version of x265, And x265 uses namespace to distinguish different bit-depth version of x265 libraries.

JackBoosY:
@a1q123456 
1. As I said, the upstream clearly indicated in cmake that these features are conflicting **for building the library**.
2. Even in terms of tool, these libraries with different features can be used to build it at the same time, we can only install one of them into vcpkg. Because installing all libraries will cause link problems when integrating msbuild. That's why we shouldn't add conflict features.

a1q123456:
@JackBoosY Okay, understood. So should I change something or just close this PR?

JackBoosY:
Close this PR since it violates our policy.

PR #18183, opened 2021-05-29
[libvpx] Update to latest and add support for arm64 on macOS

vejmartin:
Updates libvpx to latest and adds support for arm64 on macOS

- #### What does your PR fix?  
  N/A

  arm64-osx, arm64-osx-dynamic, No

  Yes

  Yes


ports/libvpx/0002-Fix-nasm-debug-format-flag.patch
```diff
@@ -1,21 +1,21 @@
...
+           EXE_SFX=.exe
+           ;;
+         win64)
+           add_asflags -f win64
+-          enabled debug && add_asflags -g cv8
++          enabled debug && [ "${AS}" = yasm ] && add_asflags -g cv8
++          enabled debug && [ "${AS}" = nasm ] && add_asflags -gcv8
+           EXE_SFX=.exe
+           ;;
+         linux*|solaris*|android*)
```
NancyLi1013, line 21: Please keep the same format of EOL as before in this patch file. Maybe LF or CRLF.

ports/libvpx/0004-remove-library-suffixes.patch
```diff
@@ -1,13 +1,13 @@
...
++++ b/build/make/gen_msvs_vcxproj.sh
+@@ -394,7 +394,7 @@ generate_vcxproj() {
+               else
+                 config_suffix=""
+               fi
+-              tag_content TargetName "${name}${lib_sfx}${config_suffix}"
++              tag_content TargetName "${name}"
+             fi
+             close_tag PropertyGroup
+         done
```
NancyLi1013, line 13: The same as above.

ports/libvpx/portfile.cmake
```diff
@@ -165,7 +164,11 @@ else()
 	elseif(VCPKG_TARGET_IS_LINUX)
         set(LIBVPX_TARGET "${LIBVPX_TARGET_ARCH}-linux-gcc")
     elseif(VCPKG_TARGET_IS_OSX)
-        set(LIBVPX_TARGET "${LIBVPX_TARGET_ARCH}-darwin17-gcc") # enable latest CPU instructions for best performance and less CPU usage on MacOS
+        if(VCPKG_TARGET_ARCHITECTURE STREQUAL arm64)
+            set(LIBVPX_TARGET "arm64-darwin20-gcc")
```
NancyLi1013, line 168: Why do you use `drawin20 `instead of `darwin17 `here?

NancyLi1013 (changes requested):

ports/libvpx/portfile.cmake
vejmartin: Originally the arm64 triplet was used for iOS, however, with Big Sur, it now also applies to macOS. `arm64-darwin20-gcc` is arm64 macOS and `arm64-darwin-gcc` is arm64 iOS. There's no `darwin17` support for arm64 macs.
See: https://github.com/webmproject/libvpx/commit/979e27c9700448bded96fb1d4ba31441a8da82db

ports/libvpx/0002-Fix-nasm-debug-format-flag.patch
vejmartin: Thank you for catching that, I'll update it.

ports/libvpx/portfile.cmake
NancyLi1013: Thanks for your clarification.

NancyLi1013 (approved):

ports/libvpx/portfile.cmake
```diff
@@ -165,7 +164,11 @@ else()
 	elseif(VCPKG_TARGET_IS_LINUX)
         set(LIBVPX_TARGET "${LIBVPX_TARGET_ARCH}-linux-gcc")
     elseif(VCPKG_TARGET_IS_OSX)
-        set(LIBVPX_TARGET "${LIBVPX_TARGET_ARCH}-darwin17-gcc") # enable latest CPU instructions for best performance and less CPU usage on MacOS
+        if(VCPKG_TARGET_ARCHITECTURE STREQUAL arm64)
```
strega-nil-ms, line 167:
```suggestion
        if(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
```
the quotes are necessary.

strega-nil-ms: this is also really suspect given the above setting of LIBBPX_TARGET_ARCH

vejmartin: @strega-nil-ms, thank you, I just learned about this policy, https://cmake.org/cmake/help/latest/policy/CMP0054.html#policy:CMP0054 :)

NancyLi1013:
Please run `vcpkg x-add-version --overwrite-version libvpx` to solve the failures on x86-windows.

ports/libvpx/portfile.cmake
NancyLi1013: There are also some other places that have the same problem. Might also require quotes.

NancyLi1013:
LGTM, except for the comments here https://github.com/microsoft/vcpkg/pull/18183#discussion_r644447947. 
@strega-nil-ms please help confirm if it's necessary to update.

Thanks.

PR #18199, opened 2021-05-31
[xmlsec]Update to 1.2.32

SvenPStarFinanz:
- #### What does your PR fix?  
Updates xmlsec to 1.2.32. Among other minor fixes that version removes the need to patch UWP build.

no change

Yes

Yes


ports/xmlsec/vcpkg.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "xmlsec",
-  "version-string": "1.2.31",
-  "port-version": 1,
+  "version-string": "1.2.32",
```
JonLiu1993, line 3: @SvenPStarFinanz ,We should use "version" replace "version-string"

versions/x-/xmlsec.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "369cbf67956067b49be4de045129b9cfbfed9431",
+      "version-string": "1.2.32",
```
JonLiu1993, line 5: We should use "version" replace "version-string"

JonLiu1993 (changes requested):

ports/xmlsec/vcpkg.json
SvenPStarFinanz: Ok, did so.

versions/x-/xmlsec.json
SvenPStarFinanz: Did so.

SvenPStarFinanz:
Apart from the version update I found/fixed a problem resulted by #16908 on Windows build of xmlsec. In src\dl.c of xmlsec there was constructed the name of the crypto library, which always have to begin with "lib", so I turned back to the old file name on Windows.

JonLiu1993 (approved):

ras0219-msft:
That makes sense to me, thanks for the update & fix!

PR #18246, opened 2021-06-02
[google-cloud-cpp] update to latest (v1.28.0) version

coryan:
- #### What does your PR fix?  

N/A. It updates `google-cloud-cpp` to the latest release (v1.28.0). One of the new things in this release is the `-DGOOGLE_CLOUD_CPP_ENABLE_EXAMPLES=OFF` option, which eliminates the need for `disable-benchmarks.patch`.

No change.

Yes.

Yes.


JonLiu1993 (changes requested):
@coryan ,Thanks for your pr,please add 'vcpkg_fixup_pkgconfig()' below '
````
vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake TARGET_PATH share)
````
 in portfile.cmake

coryan:
Done, please take another look.

ras0219-msft:
It looks like the grpc pkgconfig files are not fixed up; we'll need to add similar commands to that and abseil before google-cloud-cpp will be able to properly support pkgconfig.

coryan:
Ugh. I have reverted the `pkgconfig` change for now.  I can send PRs for Abseil and gRPC if you want me to and then send a PR for `google-cloud-cpp`.  I could create a PR will all the changes, but that sounds complicated and more risky.

ras0219-msft:
Thanks a bunch; I think doing all three at the same time would be small enough to be put into a single PR, but separate is definitely the "proper" solution :)

PR #18258, opened 2021-06-03
[vcpkg docs] Updated list of off-by-default features

meastp:
According to [Microsoft's blog post](https://devblogs.microsoft.com/cppblog/all-vcpkg-enterprise-features-now-generally-available-versioning-binary-caching-manifests-and-registries) :

> We also turned registries and versioning on by default, so you no longer need to specify a feature flag to use them.

So the documentation should be updated accordingly :)


PhoebeHui (approved):
Thanks @meastp for your PR!

ras0219-msft:
Thanks!

PR #18274, opened 2021-06-05
Add vcpkg_completions.zsh

autoantwort:
Needed for microsoft/vcpkg-tool#87

### DIFF
--- a/scripts/vcpkg_completion.zsh
+++ b/scripts/vcpkg_completion.zsh
@@ -1,4 +1,3 @@
-#/usr/bin/env bash
 
 _vcpkg_completions()
 {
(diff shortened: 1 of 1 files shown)


autoantwort:
A lot of random failures.

scripts/vcpkg_completion.zsh
```diff
@@ -0,0 +1,10 @@
+#/usr/bin/env bash
```
strega-nil-ms, line 1: Is there a reason that this vcpkg_completion function is getting run in bash?

strega-nil-ms (changes requested):

scripts/vcpkg_completion.zsh
autoantwort: It is not getting run in bash. The script is sourced and therefore the shebang can be removed because it is ignored. 
 ```suggestion
```

ras0219-msft:
LGTM, thanks!

PR #18277, opened 2021-06-05
[cpp-httplib] Upgrade to v0.8.9

tosone:
Update cpp-httplib to v0.8.9


versions/c-/cpp-httplib.json
```diff
@@ -1,8 +1,13 @@
 {
   "versions": [
     {
-      "git-tree": "fb5ea9d8cb5c86ca6924539fb32a8d9c43e0faa7",
-      "version-string": "0.8.6",
```
JonLiu1993, line 5: @tosone ,Thanks for your pr,please do not modify the historical version of git-tree, which may lead to the inability to trace the historical version

JonLiu1993 (changes requested):

versions/c-/cpp-httplib.json
tosone: Done

JonLiu1993 (approved):

strega-nil-ms:
Thanks @tosone!

PR #18285, opened 2021-06-05
[vcpkg_download_distfile] Pad the SHA to re-enable testing the hash

ras0219:


autoantwort (approved):

NancyLi1013 (approved):

scripts/cmake/vcpkg_download_distfile.cmake
```diff
@@ -80,6 +80,11 @@ function(vcpkg_download_distfile VAR)
             message(FATAL_ERROR "vcpkg_download_distfile must not be passed both SHA512 and SKIP_SHA512.")
         endif()
     endif()
+    string(LENGTH "${vcpkg_download_distfile_SHA512}" vcpkg_download_distfile_SHA512_length)
+    if(vcpkg_download_distfile_SHA512_length LESS 128)
+        # Pad with zeroes to make a valid SHA512
+        string(SUBSTRING "${vcpkg_download_distfile_SHA512}00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" 0 128 vcpkg_download_distfile_SHA512)
+    endif()
```
strega-nil-ms, line 87:
```suggestion
    if(vcpkg_download_distfile_SHA512 STREQUAL "0")
        string(REPEAT "0" 128 vcpkg_download_distfile_SHA512)
    endif()
    string(LENGTH "${vcpkg_download_distfile_SHA512}" vcpkg_download_distfile_SHA512_length)
    if(NOT vcpkg_download_distfile_SHA512_length EQUAL 128)
        message(FATAL_ERROR "Invalid SHA512: ${vcpkg_download_distfile_SHA512}")
    endif()
```

can we do something like this instead?

strega-nil-ms: or maybe:
```suggestion
    string(LENGTH "${vcpkg_download_distfile_SHA512}" vcpkg_download_distfile_SHA512_length)
    if(NOT vcpkg_download_distfile_SHA512_length EQUAL "128")
            string(REPEAT "0" 128 vcpkg_download_distfile_SHA512)
    endif()
```

strega-nil-ms (approved):

autoantwort:
Yeah but now it doesn't work anymore if I use `1` as SHA, or something else

vicroms (approved):

ras0219-msft:
That's true, but I think for the purposes of local development it is not a burden to require use of the literal 0 (compared with the burden of synthesizing _exactly_ 128 `0`'s).

autoantwort:
> That's true, but I think for the purposes of local development it is not a burden to require use of the literal 0 (esp. compared with the burden of synthesizing _exactly_ 128 `0`'s). 

The code to fill up the hash with zeros up to a string length of 128 was not that complicated. 

> Most importantly, the error messages should make the path forward clear.

No, it only shows that I have entered an invalid sha, e.g. for the input `1`. Imho not very user friendly. Imho it should just do it like in your first version, it is the easiest one for the users.

strega-nil-ms:
Thanks again robert!

autoantwort:
When you stick with the only `0` way you should update the documentation [here](https://github.com/microsoft/vcpkg/blob/master/docs/examples/packaging-github-repos.md), because there you say that users should use `1`

PR #18293, opened 2021-06-06
[tgui] Fixed sfml feature

texus:
No matter whether the `sfml` or `sdl2` feature was selected, only the SDL backend would be build. The backend name contained quotes around it which caused it to not match with any backend and silently fall back to the SDL one.

- #### What does your PR fix?  
  The SFML backend can be used again.

  No changes.

  Yes

  Yes


JackBoosY (approved):

ras0219-msft:

PR #18295, opened 2021-06-06
[bullet3]Update to 3.17

rioki:
Bump bullet3 to 3.17

- #### What does your PR fix?  

 all (no change)

Note: I had to remove some of the patch, as the verbatim change breaks CMake. It appears to target APPLE; I will amend the change if the CI breaks, I have no way to check.

I strongly hope so

yes


rioki:
@stephenjust I don't understand the changes to the install targets and really don't want to break something. If you can give me a quick look over the patch file.

ports/bullet3/cmake-fix.patch
```diff
@@ -73,14 +66,13 @@ index 68e0a93bc..872e14361 100644
 -					${CMAKE_CURRENT_BINARY_DIR}/BulletConfig.cmake
 -					@ONLY ESCAPE_QUOTES
 -				)
-+
++set (BULLET_CONFIG_CMAKE_PATH share/bullet3 )
```
NancyLi1013, line 69:
```suggestion
```
This has been set before. It's unnecessary to set it again.

ports/bullet3/cmake-fix.patch
```diff
@@ -90,188 +82,3 @@ index 68e0a93bc..872e14361 100644
...
- 					ARCHIVE DESTINATION lib${LIB_SUFFIX})
-@@ -75,5 +78,8 @@ DESTINATION ${INCLUDE_INSTALL_DIR} FILES_MATCHING PATTERN "*.h"  PATTERN
- 			SET_TARGET_PROPERTIES(LinearMath PROPERTIES FRAMEWORK true)
- 			SET_TARGET_PROPERTIES(LinearMath PROPERTIES PUBLIC_HEADER "${LinearMath_HDRS}")
- 		ENDIF (APPLE AND BUILD_SHARED_LIBS AND FRAMEWORK)
-+
-+		INSTALL(EXPORT LinearMathTargets
-+			DESTINATION ${BULLET_CONFIG_CMAKE_PATH})
- 	ENDIF (NOT INTERNAL_CREATE_DISTRIBUTABLE_MSVC_PROJECTFILES)
- ENDIF (INSTALL_LIBS)
```
NancyLi1013, line 277: Why do you remove these parts in the patch? Can you please help make it clarify?

ports/bullet3/portfile.cmake
```diff
@@ -5,8 +5,8 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO bulletphysics/bullet3
-    REF 2.89
-    SHA512 3c4ba6a3b3623ef44dd4a23e0bc2e90dec1f2b7af463edcb886e110feac1dfb4a91945f0ed640052cac228318539e275976d37238102fb10a0f78aef065a730b
+    REF 3.17
+    SHA512 a5105bf5f1dd365a64a350755c7d2c97942f74897a18dcdb3651e6732fd55cc1030a096f5808cf50575281f05e3ac09aa50a48d271a47b94cd61f5167a72b7cc
     HEAD_REF master
     PATCHES cmake-fix.patch
 )
```
NancyLi1013, line 12: Can you add the keyword `FEATURES `for the function to avoid the deprecated warning?
```
    FEATURES    
        multithreading       BULLET2_MULTITHREADING
)

```

NancyLi1013 (changes requested):

ports/bullet3/cmake-fix.patch
rioki: As far as I understand, this is not necessary, BUT I have no way to check, since I don't have an OSX machine available. The problem is that the patch did not apply cleanly and I needed to redo the change. The key thing is here that is was not applied consistently across all modules. 
I tried to contact @stephenjust who originally wrote this code to understand what the issue was, but had no luck so far. The CI had no problem building with downstream modules, so either I am missing something subtle or that change is not needed.
I will admit that I don't feel confident committing the change as is or even with the changes reapplied. Guidance would be favorable.

ports/bullet3/cmake-fix.patch
rioki: As far as I understand, this is not necessary, BUT I have no way to check, since I don't have an OSX machine available. The problem is that the patch did not apply cleanly and I needed to redo the change. The key thing is here that is was not applied consistently across all modules.
I tried to contact @stephenjust who originally wrote this code to understand what the issue was, but had no luck so far. The CI had no problem building with downstream modules, so either I am missing something subtle or that change is not needed.
I will admit that I don't feel confident committing the change as is or even with the changes reapplied. Guidance would be favorable.

rioki: I reapplied the change, but this time consistently across all modules. I still don't 100% understand what the change is supposed to do.

stephenjust: In previous versions of Bullet, it just didn't install usable Targets at all without this patch. Upstream may have fixed this, I know a couple of people had PRs open around the time that I added these patches, but I haven't had a chance to review upstream bullet to see if this is still needed or not.

rioki: Thank you for the input.

NancyLi1013 (approved):

NancyLi1013:
Have you tested the feature `multithreading`? @rioki

ports/bullet3/cmake-fix.patch
NancyLi1013: Seems there are no changes about this part in upstream. So this should be also needed in current version. @rioki 

Can you check this again?

rioki:
I privately tested x86-windows and x64-windows with multithreading. x64-windows multithreading is even in a shipped product.

NancyLi1013:
Can you run  `vcpkg x-add-version bullet3` to update versions? @rioki

NancyLi1013:
LGTM now, thanks for your update @rioki.

strega-nil-ms:
Thanks @rioki :)

PR #18300, opened 2021-06-06
[vcpkg] Add static triplet for arm-windows

ysc3839:


NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your PR @ysc3839.

strega-nil-ms:
Is there a reason you need this? As far as I know, there is exactly one machine that runs Windows on ARM32, which makes it seem to me like there isn't a real use to this...

ysc3839:
@strega-nil-ms
I want to add arm/arm64 support for my project [FontMod](https://github.com/ysc3839/FontMod), which is a GDI font replacing library.
Although it's nearly no ARM-only device, there still some ARM apps. So my project need to support these apps.

PR #18308, opened 2021-06-07
[rapidjson] Remove usage

### LINKED ISSUE #18292: [rapidjson] port usage reports wrong variable
****

RapidJSON provides its install path as `RapidJSON_INCLUDE_DIR(S)`, the port adds `RAPIDJSON_INCLUDE_DIRS` for backwards-compatibility with older versions. It seems like the usage output catches only the last variable being set:
```
The package rapidjson provides CMake integration:

    find_package(RapidJSON CONFIG REQUIRED)
    target_include_directories(main PRIVATE ${RAPIDJSON_INCLUDE_DIRS})
```

**Proposed solution**
I am not certain what is the correct way to fix the usage message - if it's possible to set the include variable explicitly, that would probably be ideal. I would be happy to do it if somebody can point me to the correct knob to turn :)

Alternatively, trick the usage message generation heuristic? But that does sound a bit hacky.

JackBoosY:
Remove the usage since the upstream export the target `rapidjson`.

Fixes #18292.

### DIFF
--- a/ports/rapidjson/usage
+++ b/ports/rapidjson/usage
@@ -1,9 +1,0 @@
-The package rapidjson provides CMake integration:
-
-    find_package(RapidJSON CONFIG REQUIRED)
-
-    target_link_libraries(main PRIVATE rapidjson)
-    # Or
-    target_include_directories(main PRIVATE ${RAPIDJSON_INCLUDE_DIRS})
-    # Or
-    target_include_directories(main PRIVATE ${RapidJSON_INCLUDE_DIRS})
--- a/ports/rapidjson/usage
+++ b/ports/rapidjson/usage
@@ -1,4 +1,9 @@
 The package rapidjson provides CMake integration:
 
     find_package(RapidJSON CONFIG REQUIRED)
+
+    target_link_libraries(main PRIVATE rapidjson)
+    # Or
     target_include_directories(main PRIVATE ${RAPIDJSON_INCLUDE_DIRS})
+    # Or
+    target_include_directories(main PRIVATE ${RapidJSON_INCLUDE_DIRS})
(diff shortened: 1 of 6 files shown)


JackBoosY:
@upsj  Is that what you need?

ports/rapidjson/usage
```diff
@@ -1,4 +1,9 @@
 The package rapidjson provides CMake integration:
 
     find_package(RapidJSON CONFIG REQUIRED)
+
+    target_link_libraries(main PRIVATE rapidjson)
+    # Or
     target_include_directories(main PRIVATE ${RAPIDJSON_INCLUDE_DIRS})
```
upsj, line 7: Does it make sense to advertise the old usage which is no longer supported by upstream? They only export the mixed-case variable.

upsj (approved):
@JackBoosY Thanks, that looks good! Somehow I didn't even notice the usage file 🤦‍♂️  I thought it was somehow autogenerated based on the exported cmake config.

ports/rapidjson/usage
JackBoosY: Normally, we should only provide the target name `Rapidjson` here.
But for compatibility, we still need to keep it.

JackBoosY: Also, for the improvement, I've create an upstream PR https://github.com/Tencent/rapidjson/pull/1901 to change the default behavor.

upsj: Advertising it here means that people might use it in their projects, which could then break if they use RapidJSON installed outside of vcpkg. Is that intended? The port still provides backwards-compatibility for people using the upper-case variable, but maybe we shouldn't point people to a deprecated usage by default?
I'm basically suggesting
```suggestion
```

JackBoosY: Maybe we should add a comment here to suggest users to use the target name instead of the macros?

JackBoosY:
Now, the usage is printed automatically:
```cmake
The package rapidjson:x86-windows provides CMake targets:

    find_package(RapidJSON CONFIG REQUIRED)
    target_link_libraries(main PRIVATE rapidjson)

```

PhoebeHui (approved):
LGTM

strega-nil-ms:
great, thanks :)

PR #18310, opened 2021-06-07
[openssl] Fix build on ARM64 macos by updating ninja to 1.10.2

### LINKED ISSUE #15741: OpenSSL for M1/arm Mac creates a library that can only be linked with iOS platforms
**Describe the bug**

Building OpenSSL (version 1.1.1i) through vcpkg results in a complete compilation with failures, but the resulting binary is marked as being only able to be linked against iOS.

The error from the linker when I'm trying to link against libcrypto (through grpc, if that's relevant) is:

```
ld: in /Users/heckj/bin/vcpkg/installed/arm64-osx/debug/lib/libcrypto.a(a_strex.o), building for macOS, but linking in object file built for iOS
```

**Environment**
- OS: macOS 11.1, M1 native arm64
- Compiler: Clang 12 (Xcode 12.3 toolchain)

**To Reproduce**
1. `./vcpkg install grpc --triplet arm64-osx`
2. `otool -lv ./installed/arm64-osx/lib/libcrypto.a  | grep -A5 LC_`

If the output includes `LC_VERSION_MIN_IPHONEOS`, then the binary is encoded with iOS as it's platform. A binary encoded for macOS should have output that looks something like this:

```
      cmd LC_BUILD_VERSION
  cmdsize 24
 platform MACOS
    minos 11.0
      sdk 11.1
   ntools 0
```

I found this after some debugging, the full story and detail of which is available at <https://rhonabwy.com/2021/01/18/m1-arm64-native-openssl-with-vcpkg/>.

**Expected behavior**
- I'd hoped to have a libcrypto.a that was linkable to macOS binary.

**Failure logs** 
- attached as `buildtree_openssl_logs.zip`
- [buildtree_openssl_logs.zip](https://github.com/microsoft/vcpkg/files/5832781/buildtree_openssl_logs.zip)

**Additional context**
One element of note that I was able to determine: if you build the OpenSSL code through a rosetta2 enabled terminal (using `x86_64` emulation), then you'll get this same issue. To compile it locally, you need to run the compilation with an `arm64` native terminal. I'm not clear on what in OpenSSL's build is inferring platform details from the terminal session, but apparently something is.

I did replicate the cmake build steps that vcpkg ran, and ran them manually in a terminal session to debug. From that I found this detail about needing to run in arm64 native mode for OpenSSL. When I ran those steps from a rosetta enabled terminal, I got the same result as vcpkg directly.

I have tried completely removing and compiling vcpkg as arm64 native, but that didn't appear to make a difference in how everything got build and resulted in the same issues. I do have x86_64 (rosetta2 translation) installed tooling in my path using Homebrew.

ghost:
Update the ninja version used for vcpkg to one that includes native arm64 macos support. This fixes the build for openssl, and maybe makes all builds that use ninja faster.

- #### What does your PR fix?  
  Fixes #15741 

  arm64-osx should be the only affected triplet. No, since CI does not support that target.

  Yes.

  No Ports updated here


JackBoosY (approved):

strega-nil-ms:
cool, thanks @Lucius-Q-User :)

PR #18319, opened 2021-06-07
[scripts-audit] add guidelines for cmake

strega-nil:


docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,55 @@
...
+    - Exception: positional parameters that are optional should be given a name via
+      `set(argument_name "${ARG<N>}") after checking `${ARGC}`.
+- There are no unparsed or unused arguments. Always check for `ARGN` or `arg_UNPARSED_ARGUMENTS`,
+  and either `FATAL_ERROR`, or `WARN` if necessary for backwards compatibility.
+- All `cmake_parse_arguments` use `PARSE_ARGV` for resistance to embedded semicolons.
+- All `foreach` loops use `IN LISTS` for resistance to embedded semicolons.
+- The variables `${ARGV}` and `${ARGN}` are unreferenced, except in helpful messages to the user.
+  - (i.e., `message(FATAL_ERROR "blah was passed extra arguments: ${ARGN}")`)
+- We always use functions, not macros or top level code.
+  - Exception: `vcpkg.cmake`'s `find_package`
```
Neumann-A, line 32:
```suggestion
  - Exception: `vcpkg.cmake`'s `find_package`. The reason for this exceptions is that `find_package` is basically a special `include` statement. Parent scoping all newly set variables within the scope of the override would severely litter users `--trace/--trace-expand` output making it unusable. 

```

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,55 @@
...
+- Scripts in the scripts tree should not be expected to need changes as part of normal operation. (For example, `vcpkg_acquire_msys` has hard coded specific packages and versions thereof used which we believe is unacceptable)
+- All non-splat variable expansions are in quotes "".
+- There are no "pointer" parameters (where a user passes a variable name rather than the contents) except for out parameters.
+- Undefined names are not referenced.
+- Out parameters only set `PARENT_SCOPE`.
+- `CACHE` variables are not used, except for internal global variables which are used to avoid duplicating work.
+- `include()`s are removed and fixes to `port.cmake` et al. are made as necessary to avoid this.
+- `foreach(RANGE)`'s arguments _must always be_ natural numbers, and `<start>` _must always be_ less than or equal to `<stop>`.
+  - This should be checked.
+
```
Neumann-A, line 42: I am missing usage of https://cmake.org/cmake/help/git-stage/command/include_guard.html
and also a note weather or not scripts should include all other required scripts for operation (which is currently not the case for some reason).

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,55 @@
...
+- We always use `cmake_parse_arguments` rather than function parameters, or referring to `${ARG<N>}`.
+  - This doesn't need to be followed for "script-local helper functions"
+  - Exception: exclusively positional parameters, like out variables.
+    - In this case, positional parameters should be put in the function declaration
+      (rather than using `${ARG<N>}`), and should be named according to local rules
+      (i.e. `snake_case`).
+    - Exception: positional parameters that are optional should be given a name via
+      `set(argument_name "${ARG<N>}") after checking `${ARGC}`.
+- There are no unparsed or unused arguments. Always check for `ARGN` or `arg_UNPARSED_ARGUMENTS`,
```
Neumann-A, line 25: This in generally leads to a lot of duplication of parameters in chained function calls

e.g. f1 calling f2 calling f3; f1 basically needs to know all possible options for f2 and f3.  
Having `arg_UNPARSED_ARGUMENTS` also requires to have an option or a one value parameter in the function, The last argument is a multi value parameter there will never be `arg_UNPARSED_ARGUMENTS` because it will be consumed by the multi value parameter.

Neumann-A: So basically function calls need to be `functioncall(<multi_var_args> <one_value_args> <options>)`

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,55 @@
...
+- Undefined names are not referenced.
+- Out parameters only set `PARENT_SCOPE`.
+- `CACHE` variables are not used, except for internal global variables which are used to avoid duplicating work.
+- `include()`s are removed and fixes to `port.cmake` et al. are made as necessary to avoid this.
+- `foreach(RANGE)`'s arguments _must always be_ natural numbers, and `<start>` _must always be_ less than or equal to `<stop>`.
+  - This should be checked.
+
+### Naming Variables
+
+- `cmake_parse_arguments`: set prefix to `"arg"`
```
Neumann-A, line 45: Sorry but the prefix should be unique for each function. Otherwise you have a disaster with `arg_OPTIONS` in f1 calling f2 with possibly the same OPTIONS parameter. 
I would suggest using `CMAKE_CURRENT_FUNCTION` instead and using it as `set(z_arg ${CMAKE_CURRENT_FUNCTION})`
`cmake_parse_arguments(PARSE_ARGV <N> "${z_arg}" "OPTIONS" `and using `${${z_arg}_OPTIONS}`

Neumann-A: i would also suggest to use `vcpkg_parse_arguments` here as a macro which automatically does the setting of `z_arg`

docs/maintainers/cmake-guidelines.md
strega-nil-ms: This doesn't make sense to me. `functioncall(... MULTIVALUE asdf asdf asdf)` has no unparsed arguments.

docs/maintainers/cmake-guidelines.md
strega-nil-ms: No this doesn't make any sense.

docs/maintainers/cmake-guidelines.md
strega-nil-ms: TIL about `include_guard`

strega-nil-ms: From latest:
 - `include()`s are only allowed in `ports.cmake` or `vcpkg-port-config.cmake`.

docs/maintainers/cmake-guidelines.md
Neumann-A: > No this doesn't make any sense. This is not how CMake's function scoping works

No that is wrong! But you could say that is not how `cmake_parse_arguments` works
```
include(CMakePrintHelpers)
function(foo)
    cmake_parse_arguments(PARSE_ARGV 0 arg "OPTION" "ONE_VALUE" "MULTI_VALUE")
    bar()
endfunction()

function(bar)
    cmake_print_variables(arg_OPTION arg_ONE_VALUE arg_MULTI_VALUE)
    cmake_parse_arguments(PARSE_ARGV 0 arg "OPTION" "ONE_VALUE" "MULTI_VALUE")
    cmake_print_variables(arg_OPTION arg_ONE_VALUE arg_MULTI_VALUE)
endfunction()

foo(OPTION "TRUE" ONE_VALUE "Hello" MULTI_VALUE "World" "!")
```
Output
```
$ cmake.exe -P function_scope.cmake
-- arg_OPTION="TRUE" ; arg_ONE_VALUE="Hello" ; arg_MULTI_VALUE="World;!"
-- arg_OPTION="FALSE" ; arg_ONE_VALUE="" ; arg_MULTI_VALUE=""

```

docs/maintainers/cmake-guidelines.md
Neumann-A: > functioncall(... MULTIVALUE asdf asdf asdf) has no unparsed arguments.

That is the problem. Let it simply be `functioncall(... MULTIVALUE asdf asdf OUTDATED_OPTION)`. Now you logically would have a `arg_UNPARSED_ARGUMENTS` but cannot detect it. 
Ah I see I meant `functioncall(<options> <one_value_args> <multi_var_args>)` should be the rule

dg0yt:
Any hint about the min version of CMake scripts must be compatible with / the version scripts may assume without checks?

docs/maintainers/cmake-guidelines.md
```diff
@@ -64,6 +64,23 @@ We hope that they will make both forwards and backwards compatibility easier.
 - All port-based scripts must use `include_guard(GLOBAL)`
   to avoid being included multiple times.
 
+### CMake Versions to Require
+
+- All CMake scripts, except for `vcpkg.cmake`,
+  may assume the version of CMake that is present in the
+  `cmake_minimum_required` of `ports.cmake`.
```
Neumann-A, line 71: Are scripts allowed to bump the version up? E.g. it would be helpful to have access to string(JSON) which is 3.19 to parse the version of a port from the manifest in the portfile.

strega-nil-ms: We'll bump it up as we update; that's a good point though.

scripts/buildsystems/vcpkg.cmake
```diff
@@ -46,6 +46,8 @@ mark_as_advanced(VCPKG_VERBOSE)
 
 option(VCPKG_APPLOCAL_DEPS "Automatically copy dependencies into the output directory for executables." ON)
 option(X_VCPKG_APPLOCAL_DEPS_SERIALIZED "(experimental) Add USES_TERMINAL to VCPKG_APPLOCAL_DEPS to force serialization." OFF)
+
+# requires CMake 3.14
 option(X_VCPKG_APPLOCAL_DEPS_INSTALL "(experimental) Automatically copy dependencies into the install target directory for executables." OFF)
```
ras0219, line 51:
```suggestion
option(X_VCPKG_APPLOCAL_DEPS_INSTALL "(experimental) Automatically copy dependencies into the install target directory for executables. Requires CMake 3.14." OFF)
```

dg0yt:
How about not allowing 
~~~
set(VAR )
~~~
because it is a common trap (in particular with initialization of lists) that the real intention is 
~~~
set(VAR "")
~~~
but the actual effect is
~~~
unset(VAR)
~~~
which unhides cache variables. So there are two explicit alternatives which are not ugly.

I would also encourage use of 
~~~
string(APPEND VAR " xxx")
~~~
over
~~~
set(VAR "${VAR} xxx")
~~~
because it makes accidental
~~~
set(VAR "${VAR};xxx")
~~~
stand out more clearly.

Neumann-A:
`set(VAR)` is used to create an empty variable in the local/current scoped thus hiding any higher scoped variables. Setting it to an empty string instead is rarely meant and doesn't play nice with `list(APPEND)`

dg0yt:
> `set(VAR)` is used to create an empty variable in the local/current scoped thus hiding any higher scoped variables. Setting it to an empty string instead is rarely meant and doesn't play nice with `list(APPEND)`

This is what I though until recently. But it is wrong, and that's why I'm bringing it up:
~~~
$ cat > demo.cmake << END_CMAKE
set(FOO "Initial")
set(FOO "Cached value" CACHE STRING "doc")
set(FOO )
message(STATUS "\${FOO}")
list(APPEND FOO "bar")
message(STATUS "\${FOO}")
set(FOO "")
list(APPEND FOO "bar")
message(STATUS "\${FOO}")
END_CMAKE
$ cmake -P demo.cmake 
-- Cached value
-- Cached value;bar
-- bar
~~~

dg0yt:
After adding one more `message(STATUS "${FOO}")`, it turns out that (at least) in script mode, the non-cache variable (aka "set" binding) doesn't even hide the cached variable (aka cache entry). This is in contradiction to [cmake documentation](https://cmake.org/cmake/help/git-stage/manual/cmake-language.7.html#cmake-language-variables). Oh no.

Neumann-A:
1) your example should use at least one scope change
2) cache variables in script mode are kund of funky since there is no concept of cache in script mode(no CMakeCache.txt to store them). So all Cache variables become priority read access.

dg0yt:
>  1. your example should use at least one scope change

The result doesn't change when I wrap the code in `function(...) ... endfunction()`.

>  2. cache variables in script mode are kund of funky since there is no concept of cache in script mode(no CMakeCache.txt to store them). So all Cache variables become priority read access.

I'm aware that there is no *persistent* cache in script mode but there are still the different types of variables. 
Variable evaluation is documented cmake behaviour, regardless of script mode. 
The documented behaviour of `set(VAR )` is unsetting (actually creating an "unset" binding), regardless of scope. 

I opened https://gitlab.kitware.com/cmake/cmake/-/issues/22285.

dg0yt:
> I opened https://gitlab.kitware.com/cmake/cmake/-/issues/22285.

A fix is already merged for CMake 3.21, and there will be a new policy, CMP0126.

My original point stands: `set(VAR )` does NOT create an empty variable, and so it doesn't hide anything.

docs/maintainers/cmake-guidelines.md
```diff
@@ -63,6 +63,7 @@ We hope that they will make both forwards and backwards compatibility easier.
   - This must be checked if necessary.
 - All port-based scripts must use `include_guard(GLOBAL)`
   to avoid being included multiple times.
+- `set(VAR )` should not be used. Use `set(VAR)` to unset a variable.
```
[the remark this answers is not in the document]
strega-nil-ms: I dunno, it seems fine to me?

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+- Never remove arguments in non-internal functions;
+  if they should no longer do anything, just take them as normal and warn on use.
+- Never add a new mandatory argument.
+
+### Naming Variables
+
+- `cmake_parse_arguments`: set prefix to `"arg"`
+- local variables are named `snake_case`
+- Internal global variable names are named `Z_VCPKG_`.
+- External experimental global variable names are named `X_VCPKG_`.
```
ras0219-msft, line 94:
```suggestion
- External experimental global variable names are prefixed with `X_VCPKG_`.

```

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+- Internal functions are named `z_vcpkg_*`
```
ras0219-msft, line 95:
```suggestion
- Internal functions are prefixed with `z_vcpkg_`
```

ras0219-msft (approved):

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+- `CACHE` variables are not used.
+  - Exception: internal global variables to avoid duplicating work.
+- `include()`s are only allowed in `ports.cmake` or `vcpkg-port-config.cmake`.
+- `foreach(RANGE)`'s arguments _must always be_ natural numbers,
+  and `<start>` _must always be_ less than or equal to `<stop>`.
+  - This must be checked if necessary.
+- All port-based scripts must use `include_guard(GLOBAL)`
+  to avoid being included multiple times.
+- `set(VAR )` should not be used. Use `unset(VAR)` to unset a variable,
+  and `set(VAR "")` to set it to an empty string.
```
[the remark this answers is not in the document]
Neumann-A: personally I think `set(VAR [[]])` would be cleaner but if `set(VAR "")` works for you I am fine with it,

PhoebeHui:
- [ ] REGRESSION: opencv4:arm-uwp: There is not enough space on the disk. 
- [ ] REGRESSION: qt5-sensors:x64-windows-static-md: Source download issue
- [ ] REGRESSION: qt5-scxml:x64-windows-static-md: Source download issue
- [ ] REGRESSION: otl:x64-linux: The hash changed, need update it.

Neumann-A:
what about guidelines for vcpkg-cmake-wrapper.cmake?
https://github.com/microsoft/vcpkg/pull/18393#issuecomment-861263579 shows that it seems to be required

dg0yt:
> what about guidelines for vcpkg-cmake-wrapper.cmake?

Fair, but the dominant problem is that documentation is completely missing vcpkg-cmake-wrapper.cmake and vcpkg-cmake-config.cmake.

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,103 @@
...
+  and `<start>` _must always be_ less than or equal to `<stop>`.
+  - This must be checked if necessary.
+- All port-based scripts must use `include_guard(GLOBAL)`
+  to avoid being included multiple times.
+- `set(VAR )` should not be used. Use `unset(VAR)` to unset a variable,
+  and `set(VAR "")` to set it to the empty value. _Note: this works for use as a list and as a string_
+
+### CMake Versions to Require
+
+- All CMake scripts, except for `vcpkg.cmake`,
```
dg0yt, line 72: `vcpkg-cmake-wrapper.cmake` scripts run in the same context as `vcpkg.cmake` (aka project mode), so they fall under the same exception.

docs/maintainers/cmake-guidelines.md
```diff
@@ -0,0 +1,103 @@
...
+  may assume the version of CMake that is present in the
+  `cmake_minimum_required` of `ports.cmake`.
+  - This `cmake_minimum_required` should be bumped every time a new version
+    of CMake is added to `vcpkgTools.xml`, as should the
+    `cmake_minimum_required` in all of the helper `CMakeLists.txt` files.
+- `vcpkg.cmake` must assume a version of CMake back to 3.1 in general
```
dg0yt, line 78: For `vcpkg-cmake-wrapper.cmake` scripts, version 3.1 even excludes `if("value" IN_LIST FEATURES)`. This makes it more complicated to deal with port features or even with arguments to `find_package` (e.g. port tiff).

Neumann-A: > `vcpkg.cmake` must assume a version of CMake back to 3.1 in general

Is sentence is nonsense and should be ignored. I think the correct phrasing would be that `vcpkg.cmake` must assume that the toplevel `CMakeLists.txt` might not have a `cmake_minimum_required` or a very low `cmake_minimum_required`. As such, setting appropriate cmake policies are a requirement. vcpkg assume that it is at least executed with a cmake binary of <some cmake version|preferly the cmake version used by vcpkg>. 
If there is somebody really arguing for a cmake binary of 3.1 I would like to have him/her invent `vcpkg.old.cmake` .....

strega-nil-ms:
@dg0yt @Neumann-A that would be good guidelines, but these are meant more for the scripts under the scripts tree.

BillyONeal (approved):

docs/maintainers/cmake-guidelines.md
strega-nil-ms: Ports can require a version of CMake that is later than what `vcpkg.cmake` cares about.

docs/maintainers/cmake-guidelines.md
strega-nil-ms: No, we really do support CMake version back to CMake 3.1 in vcpkg.cmake. There are extra features that require more recent versions, but we do support back to CMake 3.1 (or there's a bug.)

dg0yt: What is the reason to keep it at 3.1?
What is the threshold for moving it to another version?
As I already wrote:
> For vcpkg-cmake-wrapper.cmake scripts, version 3.1 even excludes if("value" IN_LIST FEATURES).

which is relevant for dealing with transitive dependencies.

strega-nil-ms: we do not consider ports as having to support any specific version of CMake; they can support whatever they want. All we need is that, if a port supports CMake 3.1, and the user uses 3.1 for their build, it works.

docs/maintainers/cmake-guidelines.md
ras0219-msft: Just to wrap this back up: the case where this scoped overriding would be a problem is if the caller expected `arg_OPTION` to _dereference_ correctly within the child function's context. However, this is explicitly disallowed above due to

> - There are no "pointer" parameters
  (where a user passes a variable name rather than the contents)
  except for out parameters.

ras0219-msft (approved):

PhoebeHui:
Tensorflow:x64-osx failure would be fixed by PR https://github.com/microsoft/vcpkg/pull/18812

strega-nil-ms:
Closed for rollup PR #18838

PR #18358, opened 2021-06-10
[SLikeNet] Update and fix building with x64-mingw-static

Pospelove:
This PR makes `slikenet` port buildable with the `x64-mingw-static` triplet. However there are some problems with merging it as I can see:

1. The patch introduces tons of changes. Obviously, this violates [the maintainer guide.](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-implement-features-in-patches)
The reason for patching code this way is pretty simple. As far as I know, MinGW doesn't have some essential functions support like `inet_pton`/`inet_ntop`. Building SLikeNet requires them.
2. `TwoWayAuthentication.cpp` breaks compilation. Investigating. For now, I remove this file via `portfile.cmake`.

I would really appreciate your feedback, especially options on how to workaround missing functions. My current idea is that we should add something like `vcpkg-platform-helper` as a port that would implement functions missing on some platforms.


NancyLi1013:
Please ping me if this PR is read for review @Pospelove.

Pospelove:
> This PR makes `slikenet` port buildable with the `x64-mingw-static` triplet. However there are some problems with merging it as I can see:
> 
> 1. The patch introduces tons of changes. Obviously, this violates [the maintainer guide.](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-implement-features-in-patches)
>    The reason for patching code this way is pretty simple. As far as I know, MinGW doesn't have some essential functions support like `inet_pton`/`inet_ntop`. Building SLikeNet requires them.
> 2. `TwoWayAuthentication.cpp` breaks compilation. Investigating. For now, I remove this file via `portfile.cmake`.
> 
> I would really appreciate your feedback, especially options on how to workaround missing functions. My current idea is that we should add something like `vcpkg-platform-helper` as a port that would implement functions missing on some platforms.

1. I have read the guide again and now I think that there is no violation. I do not implement anything new, just fix compatibility. Patched license file since the third-party implementation is being used.
2. Figured out that it's just an optional plugin. So, disabling it via compile definition when building under MinGW.

Pospelove:
@NancyLi1013 I think it's ready

ports/slikenet/portfile.cmake
```diff
@@ -1,43 +1,54 @@
...
-
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include ${CURRENT_PACKAGES_DIR}/debug/share)
-
-configure_file("${CMAKE_CURRENT_LIST_DIR}/slikenet-config.cmake" "${CURRENT_PACKAGES_DIR}/share/slikenet/slikenet-config.cmake" COPYONLY)
-configure_file("${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake" "${CURRENT_PACKAGES_DIR}/share/slikenet/vcpkg-cmake-wrapper.cmake" COPYONLY)
-file(INSTALL ${CMAKE_CURRENT_LIST_DIR}/usage DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT})
-file(INSTALL ${SOURCE_PATH}/license.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+vcpkg_fail_port_install(ON_TARGET "UWP")
+
+set(SLIKENET_PATCHES fix-install.patch)
```
NancyLi1013, line 3: Since there are many changes in this patch. Could you please submit these changes to upstream?

ports/slikenet/portfile.cmake
```diff
@@ -1,43 +1,54 @@
...
+# Uses an outdated OpenSSL version and is in an experimental namespace any way. As such we delete it here
+file(REMOVE_RECURSE "${SOURCE_PATH}/Source/src/crypto" "${SOURCE_PATH}/Source/include/slikenet/crypto")
+
+if (VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+    set(SLIKENET_ENABLE_STATIC TRUE)
+    set(SLIKENET_ENABLE_DLL FALSE)
+else()
+    set(SLIKENET_ENABLE_STATIC FALSE)
+    set(SLIKENET_ENABLE_DLL TRUE)
+endif()
```
NancyLi1013, line 32:
```suggestion
string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" SLIKENET_ENABLE_STATIC)
string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "dynamic" SLIKENET_ENABLE_DLL)
```
Please see the reference here https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#choose-either-static-or-shared-binaries

NancyLi1013 (changes requested):
Seems you rewrite portfile.cmake instead of only updating the changes based on the original file. Can you update this?

ports/slikenet/vcpkg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "slikenet",
+  "version-string": "2021-06-07",
```
NancyLi1013, line 3:
```suggestion
  "version-date": "2021-06-07",
```
This should be `version-date` here. Please see the doc here. https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md#version-date

NancyLi1013 (changes requested):

NancyLi1013:
Please `vcpkg x-add-version slikenet --overwrite-version` to solve the failures on x86-windows.

NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your updates and fixes. @Pospelove

PR #18156, opened 2021-05-27
[vcpkg] Expand Architecture list with escape chars

sandercox:
**Describe the pull request**
This expands the architecture lists with escape characters. When
building FAT binaries for macos using multiple architectures in the
values they need to be escaped otherwise they are passed on to CMake
incorrectly #14932

- #### What does your PR fix?  
  Fixes #14932

- #### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?  
  `yes`

PR #18397, opened 2021-06-11
[vcpkg_list] add new function

strega-nil:
This is a more general fix to #18156


dg0yt:
I wish there were unit tests for scripts like this one... IMO it is feasible, and its reasonable.

scripts/cmake/vcpkg_list.cmake
```diff
@@ -0,0 +1,98 @@
...
+        # getting a value from PUSH_BACK/POP_BACK is unsupported
+        message(FATAL_ERROR "vcpkg_list sub-command ${operation} requires two arguments.")
+    elseif(operation MATCHES "^(FILTER|TRANSFORM)$")
+        # regexes are hard
+        message(FATAL_ERROR "vcpkg_list sub-command ${operation} is currently unsupported.")
+    endif()
+
+    list("${operation}" lst ${args})
+    set("${out_var}" "${lst}" PARENT_SCOPE)
+endfunction()
```
dg0yt, line 98: In case of another commit, please add the trailing newline.

docs/maintainers/vcpkg_list.md
```diff
@@ -0,0 +1,52 @@
...
+
+A replacement for CMake's `list()` function, which correctly handles elements
+with internal semicolons (in other words, escaped semicolons).
+Use `vcpkg_list()` instead of `list()` whenever possible.
+
+```cmake
+vcpkg_list(<COMMAND> <out-var> <list-value> [<other-arguments>...])
+```
+
+Unlike CMake's `list()` function, since this is written in CMake,
```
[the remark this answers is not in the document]
dg0yt: Can you rewrite your example with a list, with item that have embedded semicolons (and perhaps spaces, too)?
(This would be a nice case for test-first programming.)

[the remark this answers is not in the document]
strega-nil-ms: @Pospelove it is literally not possible; let's give an example:

```cmake
function(vcpkg_list out)
  message(STATUS "${${out}}")
endfunction()

set(out "a;b;c")
vcpkg_list(out)
```

will give you

```
-- out
```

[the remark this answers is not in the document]
strega-nil-ms: We could in theory give every variable a `z_vcpkg_list_` prefix, but that still doesn't fix being able to write `vcpkg_list(ARGV1)` or w/e.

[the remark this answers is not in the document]
dg0yt: The picture is not complete without:
~~~cmake
function(vcpkg_list)
  message(STATUS "${${ARGV0}}")
endfunction()

set(out "a;b;c")
vcpkg_list(out)
~~~
producing:
~~~
-- a;b;c
~~~

docs/maintainers/vcpkg_list.md
```diff
@@ -0,0 +1,68 @@
+# vcpkg_list
+
+The latest version of this document lives in the [vcpkg repo](https://github.com/Microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_list.md).
+
+A replacement for CMake's `list()` function, which correctly handles elements
+with internal semicolons (in other words, escaped semicolons).
+Use `vcpkg_list()` instead of `list()` whenever possible.
```
dg0yt, line 7: I somewhat disagree with this recommendation, at least until unit tested.  Any deviation from cmake is an entry barrier. Is there an issue reported at CMake?

scripts/cmake/vcpkg_list.cmake
```diff
@@ -0,0 +1,128 @@
...
+
+```cmake
+if(NOT list STREQUAL "")
+    vcpkg_list(GET end "${list}" -1)
+    vcpkg_list(POP_BACK list "${list}")
+endif()
+```
+#]===]
+
+function(vcpkg_list operation out_var)
```
BillyONeal, line 67: This needs to vcpkg_parse_arguments to comply with our guidelines.

scripts/cmake/vcpkg_list.cmake
```diff
@@ -0,0 +1,128 @@
+#[===[.md:
+# vcpkg_list
+
+A replacement for CMake's `list()` function, which correctly handles elements
+with internal semicolons (in other words, escaped semicolons).
+Use `vcpkg_list()` instead of `list()` whenever possible.
```
BillyONeal, line 6: Since this a team wide policy change I think it needs to be discussed at a meeting or similar

scripts/cmake/vcpkg_list.cmake
```diff
@@ -0,0 +1,128 @@
...
+This also means that for sub-commands like `GET`, the out-parameter
+is placed _before_ the list, not at the end of the argument list.
+
+Otherwise, the `vcpkg_list()` function is the same as the built-in
+`list()` function, with the following restrictions:
+
+- `GET` supports only one index
+- `POP_BACK` and `POP_FRONT` do not support getting the value into
+  another out variable. Use C++ style `GET` then `POP_(BACK|FRONT)`.
+- `FILTER` and `TRANSFORM` are unsupported.
```
BillyONeal, line 32: Why not dispatch to the ordinarly list() for them?

scripts/cmake/vcpkg_list.cmake
```diff
@@ -0,0 +1,128 @@
...
+        endif()
+        list("${operation}" lst "${ARGV3}" out)
+        set("${out_var}" "${out}" PARENT_SCOPE)
+        return()
+    endif()
+    if(operation STREQUAL "SUBLIST")
+        if(NOT ARGC EQUAL "5")
+            message(FATAL_ERROR "vcpkg_list sub-command SUBLIST requires four arguments.")
+        endif()
+        # sublist removes a `\`, so add an additional backslash
```
BillyONeal, line 95: Does this do the right thing with lists of lists of lists?

BillyONeal (changes requested):
I think this needs some kind of test port or similar to verify functionality.

docs/maintainers/vcpkg_list.md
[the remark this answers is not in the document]
strega-nil-ms: The issue is:

```cmake
function(vcpkg_list)
  message(STATUS "${${ARGV0}}")
endfunction()

set(ARGV0 "hi")
vcpkg_list(ARGV0)
```
producing
```
-- ARGV0
```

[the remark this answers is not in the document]
dg0yt: This makes it clear. So there is a trade-off:
a) Support `ARGV<N>` as list name. 
b) Mimic the original `list` signature. (Possibly throw an error when the list name matches `^ARGV`.) 
AFAIU all other parameters are passed by value, not by variable name.

Is it really important to support (a), in particular when (b) can warn port authors? You can always assign to another variable, and it most cases you will already have that variable. This is much less error prone than remembering to pass the list as the list variable twice for appending, as in
~~~
vcpkg_list(APPEND ARGV1 "${ARGV1}" foo bar)
~~~
(which looks like an artificial case anyway).

docs/maintainers/vcpkg_list.md
strega-nil-ms: Yeah, https://gitlab.kitware.com/cmake/cmake/-/issues/20317

scripts/cmake/vcpkg_list.cmake
strega-nil-ms: yeah

scripts/cmake/vcpkg_list.cmake
strega-nil-ms: This is a special function, since it's a replacement for an existing CMake function.

scripts/cmake/vcpkg_list.cmake
strega-nil-ms: Because `list()` is broken for `FILTER` and `TRANSFORM` and we need to do Special Things to support them.

docs/maintainers/vcpkg_list.md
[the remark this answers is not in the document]
strega-nil-ms: @dg0yt thank you; you've made this a much better function, and now it has unit tests.

scripts/test_ports/unit-test-cmake/test-vcpkg_list.cmake
```diff
@@ -0,0 +1,753 @@
...
+    lst [[e;c\;d]] 
+)
+unit_test_check_variable_equal(
+    [[vcpkg_list(REMOVE_AT lst -1)]]
+    lst [[e;a\;b]] 
+)
+
+set(lst [[c;a\\;b;c]])
+unit_test_check_variable_equal(
+    [[vcpkg_list(REMOVE_AT lst 0)]]
```
BillyONeal, line 747: Add test of removing an escaped list?

scripts/test_ports/unit-test-cmake/test-vcpkg_list.cmake
```diff
@@ -0,0 +1,753 @@
...
+    lst "a;b;c"
+)
+unit_test_check_variable_equal(
+    [[vcpkg_list(INSERT lst -2 c)]]
+    lst "c;a;b"
+)
+unit_test_check_variable_equal(
+    [[vcpkg_list(INSERT lst 1 "c\\;d" e)]]
+    lst [[a;c\\;d;e;b]]
+)
```
BillyONeal, line 454: Add a test of inserting a list into another list?

scripts/test_ports/unit-test-cmake/test-vcpkg_list.cmake
```diff
@@ -0,0 +1,753 @@
...
+unit_test_check_variable_equal(
+    [[vcpkg_list(APPEND lst b)]]
+    lst ";;b"
+)
+unit_test_check_variable_equal(
+    [[vcpkg_list(APPEND lst "b;c" d)]]
+    lst [[;;b\;c;d]]
+)
+
```
BillyONeal, line 303: Add test of appending a list to another list as an element?

BillyONeal (approved):

BillyONeal:
I cancelled this world rebuild and will requeue on Ubuntu 20.04

strega-nil-ms:

PR #18409, opened 2021-06-12
[rmlui] Add Rmlui library

mikke89:
**Describe the pull request**

- #### What does your PR fix?  
  Adds a new port: [RmlUi](https://github.com/mikke89/RmlUi)

  Hopefully all platforms supported, tested on x64-windows, x64-windows-static, x86-windows.

  Yes

  Yes


ports/rmlui/vcpkg.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "rmlui",
+  "version": "4.0",
+  "description": "RmlUi is the C++ user interface library based on the HTML and CSS standards, designed as a complete solution for any project's interface needs.",
+  "homepage": "https://github.com/mikke89/RmlUi",
+  "documentation": "https://mikke89.github.io/RmlUiDoc/",
```
JackBoosY, line 6: Since you are the owner, can you please add `maintainers` here? See docs:
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/manifest-files.md#maintainers

JackBoosY (changes requested):

mikke89:
I noticed some issues when consuming the library with static builds. I'm fixing things upstream so that the necessary define `RMLUI_STATIC_LIB` is properly exported in CMake.

My (unpublished) changes seem to work well for CMake builds now. However, if I may ask, is it correct that MSBuild does not make use of the exported defines from CMake, so they need to be added manually by users? What is the best course of action in vcpkg ports to make sure macros are included, should we patch the library source to add the macro definition in-source?

Please don't merge before this is resolved, thanks!

JackBoosY:
> My (unpublished) changes seem to work well for CMake builds now. However, if I may ask, is it correct that MSBuild does not make use of the exported defines from CMake, so they need to be added manually by users? What is the best course of action in vcpkg ports to make sure macros are included, should we patch the library source to add the macro definition in-source?

If the macro exists in the header file, you should use `vcpkg_replace_string` to replace this macro in the header file with `#if 1` according to the conditions. See example:
https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/libflac/portfile.cmake#L56-L74

ports/rmlui/portfile.cmake
```diff
@@ -0,0 +1,37 @@
...
+    PREFER_NINJA
+    OPTIONS
+        ${FEATURE_OPTIONS}
+)
+
+vcpkg_install_cmake()
+vcpkg_fixup_cmake_targets(
+    CONFIG_PATH    lib/RmlUi/cmake
+)
+vcpkg_copy_pdbs()
```
noam-c, line 29: @mikke89 This function has been deprecated, unfortunately.
Luckily I've been spending some time this week _also_ trying to create a port for RmlUi. :joy:

Here's what worked for me:
```suggestion
vcpkg_cmake_install()
vcpkg_cmake_config_fixup(
    CONFIG_PATH lib/${PORT}/cmake
)

vcpkg_copy_pdbs()
```

noam-c: (note for the new commands I needed to add `vcpkg-cmake-config` and `vcpkg-cmake` as dependencies in the `vcpkg.json`)

mikke89: Cool, thanks! I'll make sure to get those changes in.

ports/rmlui/portfile.cmake
```diff
@@ -0,0 +1,57 @@
...
+		"#if !defined RMLUI_STATIC_LIB"
+		"#if 0"
+	)
+	if ("lua" IN_LIST FEATURES)
+		vcpkg_replace_string(${SOURCE_PATH}/Include/RmlUi/Lua/Header.h
+			"#if !defined RMLUI_STATIC_LIB"
+			"#if 0"
+		)
+	endif()
+endif()
```
JackBoosY, line 31: I think this should be moved to `vcpkg_cmake_install` and modify the content in _${CURRENT_PACKAGES_DIR}/include_. Modifications here will affect the entire build.

ports/rmlui/portfile.cmake
```diff
@@ -0,0 +1,57 @@
...
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup(
+	CONFIG_PATH  lib/RmlUi/cmake
+)
+vcpkg_copy_pdbs()
+
+file(REMOVE_RECURSE 
+	${CURRENT_PACKAGES_DIR}/debug/include
+	${CURRENT_PACKAGES_DIR}/debug/lib/RmlUi
+	${CURRENT_PACKAGES_DIR}/lib/RmlUi
```
JackBoosY, line 54: Is this a header-only library? The _lib_ folder should exist when building static.

JackBoosY (changes requested):

ports/rmlui/portfile.cmake
mikke89: You're right, thanks. I've changed it now.

ports/rmlui/portfile.cmake
mikke89: This just removes some empty directories which originally contained the CMake target files. The library files are still there with the names `RmlCore` and `RmlDebugger`.

ports/rmlui/portfile.cmake
```diff
@@ -0,0 +1,57 @@
...
+	FEATURES 
+		lua             BUILD_LUA_BINDINGS
+	INVERTED_FEATURES
+		freetype        NO_FONT_INTERFACE_DEFAULT
+)
+
+# Replace built-in thirdparty header
+file(COPY
+	${CURRENT_INSTALLED_DIR}/include/robin_hood.h
+	DESTINATION ${SOURCE_PATH}/Include/RmlUi/Core/Containers
```
JackBoosY, line 19: Which port provide this header? I think we should use `find_path` in `CMakeLists.txt`, remove the old one and add this file path to `target_include_directories` instead.

JackBoosY (changes requested):

ports/rmlui/portfile.cmake
mikke89: Alright, let me know if I understood correctly with the changes I made. It required a bit of patch-work, but if you prefer it this way then it's all good to me.

ports/rmlui/portfile.cmake
```diff
@@ -0,0 +1,64 @@
...
+# Replace built-in header with vcpkg version (from robin-hood-hashing port)
+file(REMOVE ${SOURCE_PATH}/Include/RmlUi/Core/Containers/robin_hood.h)
+vcpkg_replace_string( ${SOURCE_PATH}/Include/RmlUi/Config/Config.h
+	"#include \"../Core/Containers/robin_hood.h\""
+	"#include <robin_hood.h>"
+)
+vcpkg_replace_string( ${SOURCE_PATH}/CMake/FileList.cmake
+	"\${PROJECT_SOURCE_DIR}/Include/RmlUi/Core/Containers/robin_hood.h"
+	""
+)
```
JackBoosY, line 27: Can you please also patch these changes into `add-robin-hood.patch`?

JackBoosY (changes requested):

JackBoosY (approved):

JackBoosY:
When building rmlui:x64-linux:
```
CMake Error at scripts/cmake/z_vcpkg_apply_patches.cmake:61 (message):
  Applying patch failed: Checking patch CMakeLists.txt...

  Checking patch cmake/FileList.cmake...

  error: cmake/FileList.cmake: No such file or directory

  Checking patch include/RmlUi/Config/Config.h...

  error: include/RmlUi/Config/Config.h: No such file or directory

  scripts/cmake/vcpkg_extract_source_archive_ex.cmake:144 (z_vcpkg_apply_patches)
  scripts/cmake/vcpkg_from_github.cmake:154 (vcpkg_extract_source_archive_ex)
  ports/rmlui/portfile.cmake:1 (vcpkg_from_github)
```

JackBoosY (approved):

ras0219-msft:
This LGreatTM, thanks for the awesome library and making a PR into vcpkg!

mikke89:
Thank you!

PR #18411, opened 2021-06-12
[daw-json-link] New port

mheyman:
**Describe the pull request**

- #### What does your PR fix?  
  New port for **json-link** (perhaps the fastest JSON deserializer/serializer). Because we don't yet have enough JSON libraries...

  all (I hope - only checked windows and linux). No need to update ci.baseline.txt.

 Yes.

  Yes.


ports/json-link/portfile.cmake
```diff
@@ -0,0 +1,29 @@
...
+
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/daw_header_libraries DO_NOT_DELETE_PARENT_CONFIG_PATH)
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/daw_utf_range)
+vcpkg_cmake_config_fixup(CONFIG_PATH share/json_link/cmake$)
+vcpkg_cmake_config_fixup(CONFIG_PATH share/json_link/cmake)
+
+file(REMOVE_RECURSE
+    ${CURRENT_PACKAGES_DIR}/debug
+    ${CURRENT_PACKAGES_DIR}/lib
+    ${CURRENT_PACKAGES_DIR}/share/json_link
```
JonLiu1993, line 25: @mheyman ，Could you clarify why these two directories were deleted

JonLiu1993 (changes requested):

ports/json-link/portfile.cmake
mheyman: The contents of those directories seem to land in `share/json-link` and those directories are left empty at the end of the port install. Note: something is still amiss because I cannot seem to use the *...config.cmake* packages placed in `json-link` but, as this is a header-only library, those package definitions are not needed in projects that use json-link.

I see the windows build on github is failing but haven't had an opportunity to dig into it (it succeeds at home).

JonLiu1993:
@mheyman ,Could you please take a look?
[config-x86-windows-out.log](https://github.com/microsoft/vcpkg/files/6667099/config-x86-windows-out.log)

mheyman:
The failure looks like it occurs in a cmake `FetchContent_MakeAvailable()` call. This call is identical for linux and windows. The failure occurs after a `find_package(Git QUIET)` fails to find `git`. I suspect `git` is available in the path for linux and osx but not for windows builds.

Darrell Wright's *json-link* library pulls in headers from two other of his projects. That is what the `FetchContent_MakeAvailable()` is trying to do. 

If `FetchContent_MakeAvailable()` (with a requirement on git) is unavailable for vcpkg builds, it is possible that I can rework this port into three ports, one for each project. I'd have to patch-out the calls to `FetchContent...()`. If I do that, I think I should rename the port from *json-link* to *daw-json-link* with a dependency on *daw-utf-range* and *daw-header-libraries* to make it more obvious they are related (*daw-utf-range* is also dependent on *daw-header-libraries*). Much of the header files already land under a `daw/` directory (I'm a little dissapointed they don't all land under `daw/`).

Is that how I should proceed? Make three ports that never call `FetchContent_MakeAvailable()`? Or, is there an easier solution?

JackBoosY:
@mheyman  For dependent third-party libraries, we have two ways:
1. Use the port in vcpkg if it exists.
2. Add `vcpkg_from_github` or other download methods to pre-download in `portfile.cmake`, and copy the decompressed folder to source path.

JonLiu1993:
@mheyman ，Are you still working on fixing this pr?

mheyman:
Been away - still planning on it. I started the 3-port solution but the
patches required were not quick to finish and I need to set up some tests
too.

mheyman:
Renamed the port to `daw-json-link` which now depends on `daw-utf-range` which depends on `daw-header-libraries`.

`daw-utf-range` had been including a modified version of `utfcpp` - I moved the files under `include/daw` and the `utf8` namespace under `daw::` to keep from conflicting with the `utfccp` port. The modifications are constexpr/noexcept changes.

All of `daw-json-link` tests pass with the changes. Whew.

JonLiu1993:
@mheyman ,Can you please resolve the conflicts against master?

JackBoosY:

mheyman:
Merged to master...

versions/j-/json-link.json
```diff
@@ -0,0 +1,9 @@
+{
```
JackBoosY, line 1: Please remove this file.

ports/daw-header-libraries/portfile.cmake
```diff
@@ -0,0 +1,16 @@
...
+    SHA512 ae4f1c8fa5c2ca5e0fe7e21ba0f43cf42d746a324b69c8de5bbc2c89d7dd5f31f76302a9f713ac51533df2ace703e20b35291e30577449df1565f6a3e14ab659
+    HEAD_REF master
+)
+
+file(
+    COPY 
+        ${SOURCE_PATH}/include/daw
+    DESTINATION 
+        ${CURRENT_PACKAGES_DIR}/include/
+)
```
JackBoosY, line 14: I found the upstream provides the CMakeLists.txt, can we use `vcpkg_cmake_configure` and `vcpkg_cmake_install` instead of this?

ports/daw-json-link/portfile.cmake
```diff
@@ -0,0 +1,41 @@
...
+        ${SOURCE_PATH}/include/daw
+    DESTINATION 
+        ${CURRENT_PACKAGES_DIR}/include/
+)
+file(
+    COPY 
+        ${SOURCE_PATH}/include/third_party
+    DESTINATION 
+        ${CURRENT_PACKAGES_DIR}/include/daw
+)
```
JackBoosY, line 21: See above.

ports/daw-utf-range/portfile.cmake
```diff
@@ -0,0 +1,19 @@
...
+)
+
+file(
+    COPY 
+        ${SOURCE_PATH}/include/utf8.h
+        ${SOURCE_PATH}/include/utf8
+        ${SOURCE_PATH}/include/daw/utf_range
+    DESTINATION 
+        ${CURRENT_PACKAGES_DIR}/include/daw
+)
```
JackBoosY, line 17: See above.

ports/daw-utf-range/my_utfcpp.patch
```diff
@@ -0,0 +1,119 @@
...
+diff --git a/include/utf8/checked.h b/include/utf8/checked.h
+index c74171e..4d6f211 100644
+--- a/include/utf8/checked.h
++++ b/include/utf8/checked.h
+@@ -33,7 +33,7 @@ DEALINGS IN THE SOFTWARE.
+ #include <ciso646>
+ #include <stdexcept>
+ 
+-namespace utf8 {
++namespace daw::utf8 {
```
JackBoosY, line 72: I think we shouldn't modify the source code here.

JackBoosY (changes requested):

mheyman:
The first route I went down was simply using CMakeLists.txt. When that failed in the github vcpkg build system I had to split the single port into the three, *daw-json-link*, *daw-utf-range*, and *daw-header-libraries*. And, I tried to go the route of patching the CMakeLists.txt in each of those along with modifying the sources as needed. After going through multiple debug cycles, I thought "hey, this is *header-only*, why am I putting myself through all this grief when I can simply do a `file(COPY...)`? So, I switched tack and in less than an hour, I had a working install.

If you really want me to go back to the CMakeList.txt solution, I think it wasn't too far from a working solution using patched versions of those. If you simply want a *daw-json-link-config.cmake* and the like, I can certainly through some of those in there but I generally don't use them for header-only libraries (header-only-library CMake config files are not too complicated to hand code).

JackBoosY:
@mheyman I will give you some help.

mheyman:
Now uses `vcpkg_cmake_configure()`, `vcpkg_cmake_install()`, `vcpkg_cmake_config_fixup()` instead of the simple header-file-copy solution. :-)

Not sure I like the highly-patched CMakeLists.txt solution over the header file copy. I understand it is idiomatic vcpkg to use the cmake helpers but the vcpkg port maintainability suffers - if Darrell Wright ever appreciably changes the CMakeLists.txt files in the source repositories, maintaining the patches becomes more difficult.

ports/daw-utf-range/utfcpp_and_portname.patch
```diff
@@ -0,0 +1,187 @@
...
+diff --git a/include/utf8/checked.h b/include/utf8/checked.h
+index c74171e..4d6f211 100644
+--- a/include/utf8/checked.h
++++ b/include/utf8/checked.h
+@@ -33,7 +33,7 @@ DEALINGS IN THE SOFTWARE.
+ #include <ciso646>
+ #include <stdexcept>
+ 
+-namespace utf8 {
++namespace daw::utf8 {
```
JackBoosY, line 137: I still don't know why you added an extra namespace `daw`?

JackBoosY (changes requested):

ports/daw-utf-range/utfcpp_and_portname.patch
mheyman: The daw::utf8 namespace is there to keep from conflicting with the `utfcpp` vcpkg port. The source from that port is the same source that Darrell Wright modified with `constexpr`s and `noexcept`s (among other things) to work with his JSON library. I didn't want somebody to not be able to use both `utfcpp` *and* `json-link` in the same project. I considered just changing the namespace to `daw_utf8` but since I moved it under the `daw/` directory, I thought making namespace `daw::utf8` more appropriate.

JackBoosY: Any changes related to the source code require upstream approval, can you report it upstream?

Thanks.

mheyman: Done!

beached: I have gone through the changes and updated upstream for daw-json-link, daw-header-libraries, and daw-utf-range.  I think that should reduce/remove the need to patch the source

beached:
One comment on the testing.  It is primarily tested on x64 linux/macos/windows in c++17, in develop C++20 too.  It has less primarily tested on 32bit arm/32bit windows but not as a first class.  When travis had free CI it was tested on BE ppc and/or S390 I think.  It's tested in constant expressions so should work on all the triplets, but I don't know that with certainty

mheyman:
The latest pull request includes the excellent source updates from [beached](https://github.com/beached) that removed any need for patch files.

ports/daw-header-libraries/portfile.cmake
```diff
@@ -0,0 +1,16 @@
...
+    SHA512 0b13bec450ba68e2b2ac2e76c9c529fa84a95df092b104660c88e9cfda91df9e5b70f0f7f6bcbb4f7ea05c1a5613734ceff8d84929b37b115c2bc86c232d9261
+    HEAD_REF master
+)
+
+vcpkg_cmake_configure(SOURCE_PATH ${SOURCE_PATH})
+
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/${PORT})
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug" "${CURRENT_PACKAGES_DIR}/lib")
```
JackBoosY, line 14: Why delete debug and lib?

JackBoosY (changes requested):

ports/daw-header-libraries/portfile.cmake
mheyman: At one point, that was required because the build created empty directories that vcpkg did not approve of. I admit I didn't recheck that after updating to the new daw code. I will recheck as soon as I can

JackBoosY: If they are header-only library, please add comment at top of portfile.cmake.

mheyman:
Header-only comments applied. Sorry for the delay. Life intervened but I did find a second flock of chickens in the neighborhood :-)

versions/d-/daw-header-libraries.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+      "git-tree": "6687948f2f2f4c3ba578a8ac41624c58a541f1a2",
+      "version-semver": "1.21.1",
+      "port-version": 0
+    },
+    {
+      "git-tree": "ac2e26b64cc9c547c9d78fa19cf733d2c6b6a064",
+      "version-date": "2021-06-28",
+      "port-version": 0
+    }
```
JackBoosY, line 12:
```suggestion
    }
```

versions/d-/daw-json-link.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+      "git-tree": "04d0df53fc3b69309aa10f53cc92f0933e8139b8",
+      "version-semver": "2.9.11",
+      "port-version": 0
+    },
+    {
+      "git-tree": "0285ca0ade967334ccf07c69e803c1435da01147",
+      "version-date": "2021-05-16",
+      "port-version": 0
+    }
```
JackBoosY, line 12:
```suggestion
    }
```

versions/d-/daw-utf-range.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+      "git-tree": "e1c08fdf12013384e9a86e5233245321fea58793",
+      "version-semver": "2.1.0",
+      "port-version": 0
+    },
+    {
+      "git-tree": "c92a96715337b9019c1c12b2dae9596ec57c6dcc",
+      "version-date": "2021-06-15",
+      "port-version": 0
+    }
```
JackBoosY, line 12:
```suggestion
    }
```

JackBoosY (approved):

PR #18413, opened 2021-06-12
[Pkgconf] pkg-config compatibility feature

barcharcraz:
Cmake uses the "findPkgConfig module to find pkg-config and define functions to use it to find libraries and create imported targets. pkgconf is a drop in replacement for pkg-config that's easier to build on windows (pkg-config itself _can_ actually build on windows, with msvc just fine, however there are problems with static linking (that are being worked on), and most annoyingly, pkg-config depends on glib which depends on pkg-config, meaning you need to do a bootstrap process to get pkg-config to use the "system" glib.

Anyway, while cmake's FindPkgConfig module will work fine if you have pkgconf on the path and set PKG_CONFIG_EXECUTABLE to "pkgconf" or if you set the full path in PKG_CONFIG_EXECUTABLE, it won't look for pkgconf(.exe) using find_library, only pkg-config(.exe). pkgconf recommends creating a symlink named pkg-config to pkgconf if you're using pkgconf to replace pkg-config. Because symlinks are somewhat fought on windows I just created a copy, pkgconf isn't that big.

I came across this problem when using ffmpeg, I don't want to use the vcpkg specific cmake config file provided by vcpkg, and I'd rather not vendor a find module, so the best way to find it is to use pkg-config, as upstream supports. With this patch my program can just depend on pkgconf[pkg-config-compat] and everything works with no vcpkg specific code at all in my cmake.

### DIFF
--- a/ports/pkgconf/portfile.cmake
+++ b/ports/pkgconf/portfile.cmake
@@ -18,6 +18,11 @@ file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
 
 vcpkg_copy_tools(TOOL_NAMES pkgconf AUTO_CLEAN)
+if("pkg-config-compat" IN_LIST FEATURES)
+    configure_file("${CURRENT_PACKAGES_DIR}/tools/${PORT}/pkgconf${VCPKG_TARGET_EXECUTABLE_SUFFIX}" 
+                   "${CURRENT_PACKAGES_DIR}/tools/${PORT}/pkg-config${VCPKG_TARGET_EXECUTABLE_SUFFIX}" COPYONLY)
+endif()
+
 
 # Handle copyright
 file(INSTALL ${SOURCE_PATH}/COPYING DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
--- a/ports/pkgconf/vcpkg.json
+++ b/ports/pkgconf/vcpkg.json
@@ -1,7 +1,13 @@
 {
   "name": "pkgconf",
   "version": "1.7.4",
+  "port-version": 1,
   "description": "pkgconf is a program which helps to configure compiler and linker flags for development libraries. It is similar to pkg-config from freedesktop.org.",
   "homepage": "https://github.com/pkgconf/pkgconf",
-  "supports": "!uwp"
+  "supports": "!uwp",
+  "features": {
+    "pkg-config-compat": {
+      "description": "Install a copy of pkgconf with the name pkg-config. Useful for programs (like cmake) that only search for pkg-config and not pkgconf"
+    }
+  }
 }
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -4838,7 +4838,7 @@
     },
     "pkgconf": {
       "baseline": "1.7.4",
-      "port-version": 0
+      "port-version": 1
     },
     "platform-folders": {
       "baseline": "4.0.0",
--- a/versions/p-/pkgconf.json
+++ b/versions/p-/pkgconf.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "84ca32b8c04d4bb972fe5bc5800573b69bc74baf",
+      "version": "1.7.4",
+      "port-version": 1
+    },
+    {
       "git-tree": "75fe71378e6521fe0e14a82218a9430c831b9809",
       "version": "1.7.4",
       "port-version": 0


Neumann-A:
Personally I see this PR as invalid.

It is up to upstream CMake to add `pkgconf` into the `find_program` call done by `FindPkgConf` if you ask me. 
vcpkg could add a `vcpkg-cmake-wrapper` for it and do the `find_program` call with `pkgconf` before the `_find_package` call. Adding it as feature for it in pkgconf itself seems wrong.

barcharcraz:
This is a common thing that happens when pkgconf is built.

From the pkgconf readme:

## pkg-config symlink

If you want pkgconf to be used when you invoke `pkg-config`, you should install a
symlink for this.  We do not do this for you, as we believe it is better for vendors
to make this determination themselves.

    $ ln -sf pkgconf /usr/bin/pkg-config


We could do it in a separate package that depends on pkgconf, but that's kinda veering into alternative territory. Now that I think about it this is a bit problematic since vcpkg doesn't support incompatible packages.

Here's fedora's package definition for the same, they do the feature thing, but they have "conflicts" dependencies
https://src.fedoraproject.org/rpms/pkgconf/blob/rawhide/f/pkgconf.spec

barcharcraz:
on further inspection the fedora folks have pkgconf marked to obsolete pkg-config and don't maintain a pkg-config package anymore.

PhoebeHui:
I think It would be reasonable to create a overlay/custom ports for this.

JackBoosY:
Since pkgconf is not pkg-config but a replacement for it, so I don’t think it can completely replace pkg-config now unless cmake or linux maintainers switch to support it.

ras0219-msft:
Some ground thoughts:

1. I would _really_ prefer to use `pkgconf` on windows internally rather than building `pkg-config` (with all the complexities and bloat) and rather than downloading binaries from msys (and dealing with mirrors + expiring packages)
2. Given the significant investment by others in the ecosystem (Fedora), I think it's reasonable to expect it to be a sufficient drop-in replacement for our needs.
3. From the perspective of separating concerns, it is better to isolate the "choice" of what `.pc` tool to use to be separate from the definition of `pkgconf`. I can think of 4 reasonable alternatives off the top of my head:
    1. pkgconf:host
    2. pkg-config:host
    3. pkg-config/pkgconf from the system (APT)
    4. pkg-config from MSYS2 (find_acquire_program)
4. While it's clear we need to provide control and customization within vcpkg for these things, for the final consumer I think it's reasonable for them to run their own `find_program(PKG_CONFIG_EXECUTABLE pkgconf)` before calling anything using pkgconfig if they know they want to use pkgconf. Any better solution here requires first firming up a solution to expose host ports to consumers.

For internal use, I think the right path forward here is to introduce a metaport that controls the selection of the pc tool. By default, we will probably want to use `pkgconf` on Windows and system pkg-config on non-Windows. In order to switch dependencies based on the host, this metaport must be a host-only dependency (simply require `HOST_TRIPLET STREQUAL TARGET_TRIPLET`).

Then, we will also need to add another port which modifies the behavior of `FindPkgConfig`. The most straightforward way to do this would be to have a `vcpkg-cmake-wrapper` for it as a target-triplet port, but this would also affect end consumers. I know affecting the end-user is what your motivating desire is for this PR, but at the moment I'd consider it a weak anti-goal :).

If we took the metaport + findpkgconfig wrapper approach, it would look something like:

#### ports/tool-pkg-config-select/vcpkg.json
```json
{
    "dependencies": [ { "name": "pkgconf", "platform": "windows" } ]
}
```
#### ports/tool-pkg-config-select/portfile.cmake
```cmake
if(NOT HOST_TRIPLET STREQUAL TARGET_TRIPLET)
    message(FATAL_ERROR "${PORT} must be marked as a host dependency")
endif()
if(VCPKG_TARGET_IS_WINDOWS)
  file(WRITE "${CURRENT_PACKAGES_DIR}/share/tool-pkg-config-select/tool-pkg-config-select.cmake"
    "set(PKG_CONFIG_EXECUTABLE \"${CURRENT_PACKAGES_DIR}/share/tools/pkgconf/pkgconf.exe\"")
else()
  file(WRITE "${CURRENT_PACKAGES_DIR}/share/tool-pkg-config-select/tool-pkg-config-select.cmake"
    "find_program(PKG_CONFIG_EXECUTABLE pkg-config)")
endif()
```
#### ports/vcpkg-find-pkg-config/vcpkg.json
```json
{
    "dependencies": [ { "name": "tool-pkg-config-select", "host": true } ]
}
```
#### ports/vcpkg-find-pkg-config/portfile.cmake
```cmake
file(WRITE "${CURRENT_PACKAGES_DIR}/share/pkgconfig/vcpkg-cmake-wrapper.cmake"
    "include(\${CMAKE_CURRENT_LIST_DIR}/../../../${HOST_TRIPLET}/share/tool-pkg-config-select/tool-pkg-config-select.cmake)\n"
    "_find_package(${ARGN})")
```

I'm sure I've made some mistakes with the sketch above, but hopefully it gets my suggestion across.

barcharcraz:
why not just get the find module PR accepted upstream and then pull in just the new find module from upstream until the new release of cmake is out?

I don't completely understand how the cmake wrappers work though

PhoebeHui:
> why not just get the find module PR accepted upstream and then pull in just the new find module from upstream until the new release of cmake is out?

@barcharcraz, please submit a reqest or a PR to Upstream instead. Is work still being done for this PR?

Neumann-A:
upstream pr https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6301

Issue #17557, opened 2021-04-28
[fontconfig] nuget packaging failed


cenit:
**Describe the bug**
Unable to package fontconfig as nupkg. Might be due to links from etc subfolder to share subfolder inside packages/fontconfig

**Environment**
Log is for linux, but it happens also on windows

**To Reproduce**
1. define a nuget binary cache
2. ./vcpkg install fontconfig
3. See error

**Expected behavior**
nuget package to be created

**Failure logs** 

```
Starting package 1/1: fontconfig:x64-linux
Building package fontconfig[core]:x64-linux...
[DEBUG] popen(/mnt/src/vcpkg/downloads/tools/cmake-3.19.2-linux/cmake-3.19.2-Linux-x86_64/bin/cmake -DALL_FEATURES= -DCURRENT_PORT_DIR=/mnt/src/vcpkg/ports/fontconfig -D_HOST_TRIPLET=x64-linux -DFEATURES=core -DPORT=fontconfig -DVCPKG_USE_HEAD_VERSION=0 -D_VCPKG_DOWNLOAD_TOOL=BUILT_IN -D_VCPKG_EDITABLE=0 -D_VCPKG_NO_DOWNLOADS=0 -DCMD=BUILD -DDOWNLOADS=/mnt/src/vcpkg/downloads -DTARGET_TRIPLET=x64-linux -DTARGET_TRIPLET_FILE=/mnt/src/vcpkg/triplets/x64-linux.cmake -DVCPKG_BASE_VERSION=2021-01-13 -DVCPKG_CONCURRENCY=9 -DVCPKG_PLATFORM_TOOLSET=external -DGIT=/opt/git/git-2.28.0/bin/git -DVCPKG_ROOT_DIR=/mnt/src/vcpkg -DPACKAGES_DIR=/mnt/src/vcpkg/packages -DBUILDTREES_DIR=/mnt/src/vcpkg/buildtrees -D_VCPKG_INSTALLED_DIR=/mnt/src/vcpkg/installed -DDOWNLOADS=/mnt/src/vcpkg/downloads -DVCPKG_MANIFEST_INSTALL=OFF -P /mnt/src/vcpkg/scripts/ports.cmake 2>&1)
-- Using cached /mnt/src/vcpkg/downloads/fontconfig-fontconfig-844d8709a1f3ecab45015b24b72dd775c13b2421.tar.gz
[...]

PR #18414, opened 2021-06-12
Test symbolic links

Neumann-A:
Problem:
`vcpkg export <someport> --nuget` not working

Probably reason:
Broken symlinks..... So this PR test for this.

https://github.com/microsoft/vcpkg/issues/17557

How to fix:
Maybe - https://techbit.ca/2021/01/fixing-broken-symlinks-with-find-and-replace/
Maybe - Just run https://unix.stackexchange.com/a/220947
Maybe - Fix Makefile install. Search for LN_S and make symlinks relativ?


Neumann-A:
seems like only fontconfig is affected.

PR #18418, opened 2021-06-13
[xlsxio] new port

playgithub:
**Describe the pull request**

- #### What does your PR fix?  
port xlsxio

All
No

Yes

Yes

### DIFF
--- a/ports/xlsxio/0001-fix-build-type.patch
+++ b/ports/xlsxio/0001-fix-build-type.patch
@@ -1,0 +1,32 @@
+From 83eb9f6ed9cb37b2ba25f87d07453a6a3dc35820 Mon Sep 17 00:00:00 2001
+From: abc <a@b.com>
+Date: Sat, 12 Jun 2021 12:16:45 +0800
+Subject: [PATCH] fix-build-type
+
+---
+ CMakeLists.txt | 9 +++++++--
+ 1 file changed, 7 insertions(+), 2 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 3630123..a7fb924 100644
+--- a/CMakeLists.txt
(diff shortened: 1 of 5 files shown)


cenit:
the patch is unnecessary

better to pass (in configure cmake

```
-DBUILD_SHARED=$BUILD_SHARED
-DBUILD_STATIC=$BUILD_STATIC
```

after having defined those symbols beforehand with something like

```
set(BUILD_STATIC OFF)
set(BUILD_SHARED OFF)

if (VCPKG_LIBRARY_LINKAGE STREQUAL static)
   set(BUILD_STATIC ON)
else()
   set(BUILD_SHARED ON)
endif()
```

In this way we save some space in vcpkg and guarantee future proof to patch

playgithub:
> some



> the patch is unnecessary
> 
> better to pass (in configure cmake
> 
> ```
> -DBUILD_SHARED=$BUILD_SHARED
> -DBUILD_STATIC=$BUILD_STATIC
> ```
> 
> after having defined those symbols beforehand with something like
> 
> ```
> set(BUILD_STATIC OFF)
> set(BUILD_SHARED OFF)
> 
> if (VCPKG_LIBRARY_LINKAGE STREQUAL static)
>    set(BUILD_STATIC ON)
> else()
>    set(BUILD_SHARED ON)
> endif()
> ```
> 
> In this way we save some space in vcpkg and guarantee future proof to patch

In fact, I've tried this, but with `if(BUILD_SHARED_LIBS)` instead of `if (VCPKG_LIBRARY_LINKAGE STREQUAL static)`, and failed.
I'll try it, thanks.

cenit:
BUILD_SHARED_LIBS is not defined when processing a port file, that's why it didn't work

playgithub:
I've re-forked vcpkg

PR #18432, opened 2021-06-13
[angle] Update to chromium/4472

past-due:
Update the `angle` port (libANGLE) to `chromium/4472` (corresponding to the current Chromium Stable release).

(This follows the recommendations in the libANGLE "[Choosing an ANGLE branch](https://github.com/google/angle/blob/master/doc/ChoosingANGLEBranch.md#matching-a-chromium-release-to-an-angle-branch)" guide, matching the current Chromium Stable 91.0.4472.101.)

  Same triplets as existing


ports/angle/CMakeLists.txt
```diff
@@ -39,27 +39,42 @@ add_definitions(
...
+    add_definitions(-DANGLE_IS_WIN)
+endif()
+if(LINUX)
+    add_definitions(-DANGLE_IS_LINUX)
+endif()
+if(ANGLE_IS_64_BIT_CPU)
+    add_definitions(-DANGLE_IS_64_BIT_CPU)
+elseif(ANGLE_IS_32_BIT_CPU)
+    add_definitions(-DANGLE_IS_32_BIT_CPU)
+endif()
```
Neumann-A, line 55: Are those definitions only used when building or are they also needed by a consumer? (are the some header ifdefs pragmas)

past-due: They do not appear to be exposed in any public headers. 👍

past-due:
The x64_osx/x64_linux CI runs are failing on the `opencl` port, which appears to be unrelated to this PR?

JackBoosY:
@past-due That's related. Some files installed by angle are conflicts with opencl.

ports/angle/CMakeLists.txt
```diff
@@ -435,7 +683,7 @@ SET_TARGET_PROPERTIES(libANGLE PROPERTIES PREFIX "")
 SET_TARGET_PROPERTIES(libGLESv2 PROPERTIES PREFIX "")
 SET_TARGET_PROPERTIES(libEGL PROPERTIES PREFIX "")
 
-install(TARGETS libEGL libGLESv2 libANGLE angle_common angle_image_util angle_gpu_info_util angle_translator angle_preprocessor GL_VISIBILITY EXPORT ANGLEExport
+install(TARGETS libEGL libGLESv2 libANGLE angle_common angle_compression_utils angle_image_util angle_gpu_info_util angle_translator angle_preprocessor GL_VISIBILITY EXPORT ANGLEExport
```
Neumann-A, line 686: I still wonder if this is correct or if the main target  `libEGL libGLESv2 libANGLE` should use `$<TARGET_OBJECTS:helper_targets>` instead so that they don't need to get installed

past-due: I'm afraid I have not had a chance to test this suggestion. As-is it is working fine for our purposes. @JackBoosY, your thoughts?

past-due:
Updated to pick up https://github.com/google/angle/commit/d15be77864e18f407c317be6f6bc06ee2b7d070a which was pushed to the `chromium/4472` libANGLE branch.

strega-nil-ms (approved):

strega-nil-ms:
Thanks @past-due!

PR #18435, opened 2021-06-14
[libxml2] android and ios

m-kuhn:
**Add android and ios support to libxml2**


ports/libxml2/config.android.h
```diff
@@ -0,0 +1,288 @@
+/* config.h.  Generated from config.h.in by configure.  */
```
ras0219, line 1: It would be nice to remove all the chaff from this file (comments, newlines) since it's being checked in.

ras0219 (approved):

Neumann-A:
@ras0219 don't forget #17945 which removes the custom CMakeLists.txt

ports/libxml2/config.android.h
m-kuhn: It's a generated file, I think there is value in leaving this (e.g. to compare it with files generated in other environments). But if it's desired I can remove it.

m-kuhn:
> @ras0219 don't forget #17945 which removes the custom CMakeLists.txt

Feel free to close this one in favor of  #17945 anytime

PhoebeHui:
Closing in favor of https://github.com/microsoft/vcpkg/pull/17945

PR #18457, opened 2021-06-15
[seal]Updated to version 3.6.6. Replaced CONTROL with manifest file.

WeiDaiWD:
**Describe the pull request**

- #### What does your PR fix?  
  Updated `ports/seal` to version 3.6.6. Added a new feature "hexl" that depends on the `ports/hexl`. Replaced `CONTROL` file with `vcpkg.json` manifest file. Removed deprecated vcpkg commands.
  Fixed `ports/hexl` by adding another patch. Previously "hexl" does not look for its dependency in config file, causing downstream linking failures.

  x64-windows/x86-windows are not supported as the library only supports static linkage in Windows.
  No change to the CI baseline.

  I've tried my best.

  Yes.


versions/s-/seal.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "b15a8a466694df183d4d36ce369f59d93d3cd326",
+      "version-string": "3.6.6",
```
JonLiu1993, line 5: we should use "version-semver" instead of "version-string"

ports/seal/vcpkg.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "seal",
+  "version-string": "3.6.6",
```
JonLiu1993, line 3: @WeiDaiWD ,Thanks for your pr,we should use "version-semver" instead of "version-string"

JonLiu1993 (changes requested):

JonLiu1993:
@WeiDaiWD,Have you tested the feature locally on x64-linux, x64-windows, and x64-windows-static？

ports/seal/vcpkg.json
WeiDaiWD: What's the difference between `"version-semver"` and `"version"`? I was told to use `"version"` when adding/updating other ports, and wonder how to decide in future.

versions/s-/seal.json
WeiDaiWD: Should I change all appearances of `"version-string"` in this file? I mean for all versions? And is it supposed to be the same with whatever used in `/ports/seal/vcpkg.json`?

WeiDaiWD:
> @WeiDaiWD,Have you tested the feature locally on x64-linux, x64-windows, and x64-windows-static？

x64-linux and x64-windows-static work fine locally. x64-windows throws errors. `seal` does not support dynamic linking on windows, neither does `hexl`.

versions/s-/seal.json
JonLiu1993: In fact, I just pointed out the modification. You only need to modify the "version-string" to "version-semver" in vcpkg.json, and the versions/s-/seal.json file will be automatically modified

ports/seal/vcpkg.json
JonLiu1993: Please refer to the official [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#22-package-versions)

JonLiu1993:
If you have modified everything, please run the "./vcpkg x-add-version seal --overwrite-version " command and resubmit the changes

JonLiu1993 (approved):

ports/seal/portfile.cmake
```diff
@@ -3,49 +3,39 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
...
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     FEATURES
-    ms-gsl SEAL_USE_MSGSL
-    zlib SEAL_USE_ZLIB
-    zstd SEAL_USE_ZSTD
-
+        ms-gsl SEAL_USE_MSGSL
+        zlib SEAL_USE_ZLIB
+        zstd SEAL_USE_ZSTD
```
ras0219-msft, line 15: I see no feature macro added to respect the `hexl` feature; perhaps you meant to add something like
```suggestion
        zstd SEAL_USE_ZSTD
    INVERTED_FEATURES
        hexl CMAKE_DISABLE_FIND_PACKAGE_HEXL
```

ras0219-msft:
Thanks for the PR!

ports/seal/portfile.cmake
WeiDaiWD: Thanks for catching this. I've added `SEAL_USE_INTEL_HEXL` to the latest commits.

WeiDaiWD:
Do not merge. Debugging a linking error.

WeiDaiWD:
> Do not merge. Debugging a linking error.

This is resolved and tested locally.

WeiDaiWD:
I've also edited the PR description to cover changes to `ports/hexl`.

ports/seal/portfile.cmake
JonLiu1993: @WeiDaiWD,INVERTED_FEATURES No need?

WeiDaiWD: Nope, it is controlled by a single SEAL_USE_XXX argument.

JonLiu1993 (approved):

ras0219-msft:
Thanks!

PR #18511, opened 2021-06-17
[tiff] Fix build on arm64-linux

### LINKED ISSUE #18504: tiff:arm64-linux build failure
**Host Environment**
- OS: Ubuntu 18.04 (inside a Docker container)
- Compiler: aarch64-linux-gnu-gcc (Ubuntu/Linaro 7.5.0-3ubuntu1~18.04) 7.5.0

**To Reproduce**
`./vcpkg install tiff:arm64-linux`

**Failure logs** 
- [console.log](https://github.com/microsoft/vcpkg/files/6670041/console.log)
- [config-arm64-linux-dbg-out.log](https://github.com/microsoft/vcpkg/files/6670043/config-arm64-linux-dbg-out.log)
- [config-arm64-linux-dbg-err.log](https://github.com/microsoft/vcpkg/files/6670045/config-arm64-linux-dbg-err.log)

**Additional context**
I noticed this in the context of installing opencv and other requirements (jpeg, lzma, png16, protobuf, webp, zlib) built without issues.

dg0yt:
- #### What does your PR fix?  
  Fixes #18504. Might help also with other community triplets.

  all, no

  yes

  yes

### DIFF
--- a/ports/tiff/FindCMath.patch
+++ b/ports/tiff/FindCMath.patch
@@ -1,0 +1,13 @@
+diff --git a/cmake/FindCMath.cmake b/cmake/FindCMath.cmake
+index 2da3b7f..e6d5145 100644
+--- a/cmake/FindCMath.cmake
++++ b/cmake/FindCMath.cmake
+@@ -32,7 +32,7 @@ include(CheckLibraryExists)
+ 
+ check_symbol_exists(pow "math.h" CMath_HAVE_LIBC_POW)
+ if(NOT CMath_HAVE_LIBC_POW)
+-    find_library(CMath_LIBRARY NAMES m)
++    find_library(CMath_LIBRARY NAMES m PATHS ${CMAKE_C_IMPLICIT_LINK_DIRECTORIES})
+ 
+     set(CMAKE_REQUIRED_LIBRARIES_SAVE ${CMAKE_REQUIRED_LIBRARIES})
+     set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${CMath_LIBRARY})
--- a/ports/tiff/portfile.cmake
+++ b/ports/tiff/portfile.cmake
@@ -13,6 +13,7 @@ vcpkg_extract_source_archive_ex(
     PATCHES
         cmakelists.patch
         fix-pkgconfig.patch
+        FindCMath.patch
 )
 
 set(EXTRA_OPTIONS "")
--- a/ports/tiff/vcpkg.json
+++ b/ports/tiff/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "tiff",
   "version": "4.3.0",
+  "port-version": 1,
   "description": "A library that supports the manipulation of TIFF image files",
   "homepage": "https://libtiff.gitlab.io/libtiff/",
   "dependencies": [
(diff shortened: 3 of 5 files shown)


dg0yt:
More information:
`libm.{so,a}` is located in `/usr/aarch64-linux-gnu/lib` for Ubuntu, but this directory is not in the search list for `find_library` on arm64-linux. Instead of this patch to port tiff, it might make more sense to adjust vcpkg's toolchain setup in a way that makes the directory part of the search list.

JonLiu1993:
@dg0yt ,Could you please take a look?
[config-x64-windows-static-md-out.log](https://github.com/microsoft/vcpkg/files/6684788/config-x64-windows-static-md-out.log)

dg0yt:
@JonLiu1993 IMO the error is unrelated.

~~~
Installing in D:/packages/kf5plotting_x64-windows-static-md/debug. Run D:/buildtrees/kf5plotting/x64-windows-static-md-dbg/prefix.sh to set the environment for KPlotting.
-- Looking for __GLIBC__
-- Looking for __GLIBC__ - not found
CMake Error at D:/installed/x64-windows-static-md/share/ECM/kde-modules/KDEClangFormat.cmake:67 (configure_file):
  No such file or directory
  D:/installed/x64-windows-static-md/share/ECM/kde-modules/KDEFrameworkCompilerSettings.cmake:79 (include)
  CMakeLists.txt:15 (include)

dg0yt:
Needs CI re-run.

ras0219-msft:
> it might make more sense to adjust vcpkg's toolchain setup in a way that makes the directory part of the search list.

It seems a shame that CMake doesn't already handle this correctly; is the implication really that you can't "simply" build TIFF on Ubuntu arm64? Or is this in the cross-compilation case? (Still unfortunate -- but more understandable that it's less well trodden).

fh-igd-mueller-roemer:
> It seems a shame that CMake doesn't already handle this correctly; is the implication really that you can't "simply" build TIFF on Ubuntu arm64? Or is this in the cross-compilation case? (Still unfortunate -- but more understandable that it's less well trodden).

#18504 is specifically about cross compilation, I haven't tried aarch64 directly yet. Considering `FindCMath.cmake` is a custom find script bundled with `libtiff`, I don't see how CMake could handle this automatically, but as pointed out in https://github.com/microsoft/vcpkg/pull/18511#issuecomment-863496714 this could potentially also be addressed how the toolchain is set up.

dg0yt:
I also used cross compilation. But I don't think it makes a difference. This might be an interplay of different factors:
- CMake's `find_library` not looking into `CMAKE_<LANG>_IMPLICIT_LINK_DIRECTORIES`.
- Tiff having issues with its (new) cmake build system. (I can't even build with the minimum required version of cmake.)
- Ubuntu/Debian having a particular file system layout with regard to architectures.
- Ubuntu/Debian/Linux packaging still prefering autotools over cmake.

fh-igd-mueller-roemer:
I just tried a native build (actually Docker + QEMU, but that shouldn't matter) via `VCPKG_FORCE_SYSTEM_BINARIES=1 ./vcpkg/vcpkg install tiff` which worked, so it is a cross compilation issue only (but very relevant to me anyway, as QEMU is slow and space on my real device is severely limited)

PhoebeHui (approved):

PR #18516, opened 2021-06-17
[libass] fix fontconfig dependency in .pc file

mcmtroffaes:
**Describe the pull request**

- #### What does your PR fix?  
  This PR fixes link errors with fontconfig when linking against libass using pkgconfig on x64-linux. Ran into this problem when linking against ffmpeg[libass] (for instance, see link errors here: https://github.com/mcmtroffaes/ffmpeg-msvc-build/runs/2851854810?check_suite_focus=true#step:15:42).

  No changes.

  Yes.

  Yes.


JackBoosY (approved):

ras0219-msft:
Thanks for the fix, this looks great to me!

mcmtroffaes:
Lovely, many thanks for merging so quickly! For the record, upstream has plans for supporting meson (there's a long standing PR on their repo which appears to be in good shape), so hopefully we can then get rid of the custom CMakeLists.txt in due time.

Neumann-A:
> so hopefully we can then get rid of the custom CMakeLists.txt in due time.

you probably already can by switching to `vcpkg_configure_make`

mcmtroffaes:
> you probably already can by switching to `vcpkg_configure_make`

I wish you were right. Unfortunately upstream doesn't support windows builds with configure/make. It looks like this is the reason the port was created with a custom CMakeLists.txt here on vcpkg, and also the reason why upstream is moving to support meson.

Issue #17440, opened 2021-04-22
[ boost-math ] build failure: cross-compile from mingw to arm-linux


fairlyblank:
**Host Environment**
- OS: Windows10 & MingW64
- Compiler: 
  - Vcpkg version: 2021-01-13-d67989bce1043b98092ac45996a8230a059a2d7e
  - cross compiler: linaro-gcc-10.2.1

**To Reproduce**

I'm using my own triplet, almost same as `community/arm-linux.cmake` except passing throgh some env and chainloading  toolchain file, locating linaro gcc compilers.
`vcpkg install boost --triplet=arm-linux --host-triplet=x64-mingw-dynamic --overlay-triplets=[path-to-my-triplets]`

Most boost packages building successed , until boost-math.

**Failure logs** 

        Building package boost-filesystem[core]:arm-linux... done
        Installing package boost-filesystem[core]:arm-linux...
        Installing package boost-filesystem[core]:arm-linux... done
        Elapsed time for package boost-filesystem:arm-linux: 21.52 s
        Starting package 4/101: boost-math:arm-linux
        Building package boost-math[core]:arm-linux...
        -- [OVERLAY] Loading triplet configuration from: [path-to-my-triplets]\arm-linux.cmake
        -- Using cached E:/vcpkg/downloads/boostorg-math-boost-1.75.0.tar.gz
        -- Cleaning sources at E:/vcpkg/buildtrees/boost-math/src/ost-1.75.0-eaa6889798.clean. Use --editable to skip cleaning for the packages you specify.
        -- Extracting source E:/vcpkg/downloads/boostorg-math-boost-1.75.0.tar.gz
        -- Using source at E:/vcpkg/buildtrees/boost-math/src/ost-1.75.0-eaa6889798.clean
        -- Building release...
        -- Configuring arm-linux
        -- Building arm-linux-rel
        CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:146 (message):
            Command failed: D:/msys64/mingw64/bin/cmake.exe --build . --config Release --target install -- -v -j9
            Working Directory: E:/vcpkg/buildtrees/boost-math/arm-linux-rel
            See logs for more information:
              E:\vcpkg\buildtrees\boost-math\install-arm-linux-rel-out.log
        
        Call Stack (most recent call first):
          scripts/cmake/vcpkg_build_cmake.cmake:105 (vcpkg_execute_build_process)
[...]

PR #18529, opened 2021-06-18
[boost-modular-build-helper] Fix boost build toolchain options not being used

omartijn:
The issue was due to the options only being set for the gcc toolchain,
and then only for a specific version. On platforms defaulting to a
different toolchain (e.g. macOS) this didn't work at all.

Additionally, some missing flags were not propagated, in particular the
CMAKE_OSX_DEPLOYMENT_TARGET, CMAKE_OSX_SYSROOT and CMAKE_OSX_ARCHITECTURES

Some branching was also removed from the user-config.jam, as this was
unnecessary. The options are only relevant to the selected toolset and
ignored otherwise, so there is no need to artificially restrict scope.

- #### What does your PR fix?  
Building ports using Boost.build on platforms not using gcc 5.4.1. These were using their default compiler options, completely ignoring whatever was set from vcpkg.

This PR improves triplet support on linux and macOS by correctly propagating triplet options to Boost.build.

Yes

Yes. Port-version for boost-modular-build-helper is updated.

### DIFF
--- a/ports/boost-modular-build-helper/CMakeLists.txt
+++ b/ports/boost-modular-build-helper/CMakeLists.txt
@@ -46,13 +46,17 @@ else()
 endif()
 
 if(APPLE)
-    list(APPEND B2_OPTIONS target-os=darwin toolset=clang)
+    set(B2_TOOLSET clang)
+    list(APPEND B2_OPTIONS target-os=darwin)
 elseif(WIN32)
-    list(APPEND B2_OPTIONS target-os=windows toolset=gcc)
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=windows)
 elseif(ANDROID)
-    list(APPEND B2_OPTIONS target-os=android toolset=gcc)
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=android)
 else()
-    list(APPEND B2_OPTIONS target-os=linux toolset=gcc)
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=linux)
 endif()
 
 if(WIN32)
@@ -78,6 +82,28 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Debug")
     else()
         set(LDFLAGS "${CMAKE_STATIC_LINKER_FLAGS} ${CMAKE_STATIC_LINKER_FLAGS_DEBUG}")
     endif()
+endif()
+
+if(APPLE)
+    if(CMAKE_OSX_DEPLOYMENT_TARGET)
+        set(CXXFLAGS "-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET} ${CXXFLAGS}")
+        set(CFLAGS "-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET} ${CFLAGS}")
+        set(LDFLAGS "-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET} ${LDFLAGS}")
+    endif()
+
+    if(CMAKE_OSX_SYSROOT)
+        set(CXXFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${CXXFLAGS}")
+        set(CFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${CFLAGS}")
+        set(LDFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${LDFLAGS}")
+    endif()
+
+    # if specific architectures are set, configure them,
+    # if not set, this will still default to current arch
+    foreach(ARCH ${CMAKE_OSX_ARCHITECTURES})
+        set(CXXFLAGS "-arch ${ARCH} ${CXXFLAGS}")
+        set(CFLAGS "-arch ${ARCH} ${CFLAGS}")
+        set(LDFLAGS "-arch ${ARCH} ${LDFLAGS}")
+    endforeach()
 endif()
 
 string(STRIP "${CXXFLAGS}" CXXFLAGS)
@@ -122,7 +148,7 @@ foreach(INCDIR ${CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES})
 endforeach()
 
 if(APPLE)
-  set(CXXFLAGS "${CXXFLAGS} <compileflags>-D_DARWIN_C_SOURCE <compileflags>-std=c++11 <compileflags>-stdlib=libc++")
+  set(CXXFLAGS "${CXXFLAGS} <compileflags>-D_DARWIN_C_SOURCE <cxxflags>-std=c++11 <cxxflags>-stdlib=libc++")
   set(LDFLAGS "${LDFLAGS} <linkflags>-stdlib=libc++")
 endif()
 
@@ -157,6 +183,7 @@ endif()
 
 add_custom_target(boost ALL
     COMMAND "${B2_EXE}"
+        toolset=${B2_TOOLSET}
         --user-config=${CMAKE_CURRENT_BINARY_DIR}/user-config.jam
         --stagedir=${CMAKE_CURRENT_BINARY_DIR}/stage
         --build-dir=${CMAKE_CURRENT_BINARY_DIR}
--- a/ports/boost-modular-build-helper/CMakeLists.txt
+++ b/ports/boost-modular-build-helper/CMakeLists.txt
@@ -99,7 +99,7 @@ if(APPLE)
 
     # if specific architectures are set, configure them,
     # if not set, this will still default to current arch
-    foreach(ARCH ${CMAKE_OSX_ARCHITECTURES})
+    foreach(ARCH IN LISTS CMAKE_OSX_ARCHITECTURES)
         set(CXXFLAGS "-arch ${ARCH} ${CXXFLAGS}")
         set(CFLAGS "-arch ${ARCH} ${CFLAGS}")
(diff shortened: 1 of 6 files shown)


omartijn:
@PhoebeHui How can I debug these CI build failures? I've tested this locally without any issues. I've found this page which lists the artifacts which supposedly contain the build logs: https://dev.azure.com/vcpkg/public/_build/results?buildId=55081&view=artifacts&pathAsName=false&type=publishedArtifacts

But I can't do anything with this list. Clicking any of the items does nothing.

PhoebeHui:
@omartijn, from https://github.com/microsoft/vcpkg/pull/18529/checks?check_run_id=2857119376, you can see the regression ports in CI testing, you can verify them locally. I also attached the logs, let me know if you also need the failure logs for other triplets.

[failure logs for x64-windows (9).zip](https://github.com/microsoft/vcpkg/files/6676433/failure.logs.for.x64-windows.9.zip)

dg0yt:
> But I can't do anything with this list. Clicking any of the items does nothing.

On the right side of each line, there is a vertical three dots menu which lets you download the artifact.

omartijn:
> @omartijn, from https://github.com/microsoft/vcpkg/pull/18529/checks?check_run_id=2857119376, you can see the regression ports in CI testing, you can verify them locally. I also attached the logs, let me know if you also need the failure logs for other triplets.
> 
> [failure logs for x64-windows (9).zip](https://github.com/microsoft/vcpkg/files/6676433/failure.logs.for.x64-windows.9.zip)

Thanks. This is helpful. As I said, I didn't get the same failures locally by running those builds, so the build logs are essential to solve them.

> > But I can't do anything with this list. Clicking any of the items does nothing.
> 
> On the right side of each line, there is a vertical three dots menu which lets you download the artifact.

Ah, such a subtle design. Too subtle for me, I guess. This does exactly what I want. Thanks.

ports/boost-modular-build-helper/boost-modular-build.cmake
```diff
@@ -208,6 +221,13 @@ function(boost_modular_build)
          "-sZSTD_LIBPATH=${CURRENT_INSTALLED_DIR}/lib"
     )
 
+    # mac os allows setting the deployment target
+    if (VCPKG_TARGET_IS_OSX AND VCPKG_OSX_DEPLOYMENT_TARGET)
+        set(VCPKG_C_FLAGS "-mmacosx-version-min=10.11 ${VCPKG_C_FLAGS}")
+        set(VCPKG_CXX_FLAGS "-mmacosx-version-min=10.11 ${VCPKG_CXX_FLAGS}")
+        set(VCPKG_LINKER_FLAGS "-mmacosx-version-min=10.11 ${VCPKG_LINKER_FLAGS}")
+    endif()
```
PhoebeHui, line 229: It is wrong place to set the options for osx. Since unix_build has finished, it doesn't apply these options.

ports/boost-modular-build-helper/CMakeLists.txt
```diff
@@ -157,6 +183,7 @@ endif()
 
 add_custom_target(boost ALL
     COMMAND "${B2_EXE}"
+        toolset=${B2_TOOLSET}
```
PhoebeHui, line 186: I don't think this is needed, since the toolset has append to B2_OPTIONS which already passed.

ports/boost-modular-build-helper/boost-modular-build.cmake
omartijn: Thanks. This is a leftover from when I started and was trying things to get it to work. It shouldn't be there, as it's already handled in the `unix_build` function now. I'll remove it.

ports/boost-modular-build-helper/CMakeLists.txt
```diff
@@ -46,13 +46,17 @@ else()
...
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=windows)
 elseif(ANDROID)
-    list(APPEND B2_OPTIONS target-os=android toolset=gcc)
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=android)
 else()
-    list(APPEND B2_OPTIONS target-os=linux toolset=gcc)
+    set(B2_TOOLSET gcc)
+    list(APPEND B2_OPTIONS target-os=linux)
```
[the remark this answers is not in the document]
omartijn: It _used_ to, yes. Here I am changing it since I need to have the `B2_TOOLSET` variable to use for configuring `user-config.jam`. Otherwise that file only sets our configuration for the gcc toolset - which is not used on e.g. macOS.

ports/boost-modular-build-helper/CMakeLists.txt
omartijn: Yes, it is since I also remove it from those options and instead set it in a variable. I can change it so it sets the variable _and_ adds it to the options if that is preferred.

omartijn:
@PhoebeHui Ready for re-review.

ports/boost-modular-build-helper/CMakeLists.txt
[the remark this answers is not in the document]
PhoebeHui: user-config.jam only appiled on windows, so I think the changes doesn't need.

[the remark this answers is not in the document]
omartijn: It's actually used for all platforms, but on Linux it only worked if the user used gcc 5.4.1 and on macOS it didn't work at all. But it's been fixed so it works on macOS now as well and the options are pushed there.

PhoebeHui:
cc @yurybura @longnguyen2004 @autoantwort

longnguyen2004:
Aside from the macOS specific bits, which I have no experience in, this LGTM

omartijn:
@longnguyen2004 I agree that it makes sense to set the toolset dynamically, depending on the compiler. I didn't want to do this for two reasons, the first is that I believe the idea is to keep PRs as small as possible and to break them up in separate PRs if possibly. The second is that I'm not entirely sure what the implications are of changing the toolset.

For example, I'm on linux and have clang set as my default compiler, which CMake will then set in the user-config.jam to be used, but it now uses toolset gcc. If I change it to base the toolset on the compiler, it'd be toolset clang instead. What exactly does that change in boost.build? I have no idea. I'd be hesitant to change it given that I don't know the implications.

ports/boost-modular-build-helper/boost-modular-build.cmake
```diff
@@ -84,6 +84,19 @@ function(boost_modular_build)
...
+            endif()
+            if(VCPKG_OSX_SYSROOT)
+                list(APPEND configure_options "-DCMAKE_OSX_SYSROOT=${VCPKG_OSX_SYSROOT}")
+            endif()
+            if(VCPKG_OSX_ARCHITECTURES)
+                # note: VCPKG_OSX_ARCHITECTURES is a list and must be enclosed in quotes
+                list(APPEND configure_options "-DCMAKE_OSX_ARCHITECTURES=\"${VCPKG_OSX_ARCHITECTURES}\"")
+            endif()
+        endif()
+
```
ras0219-msft, line 99:
```suggestion
```

This is handled by `vcpkg_configure_cmake()` already (line 273)

ports/boost-modular-build-helper/CMakeLists.txt
```diff
@@ -80,6 +84,28 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Debug")
...
+
+    if(CMAKE_OSX_SYSROOT)
+        set(CXXFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${CXXFLAGS}")
+        set(CFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${CFLAGS}")
+        set(LDFLAGS "-isysroot ${CMAKE_OSX_SYSROOT} ${LDFLAGS}")
+    endif()
+
+    # if specific architectures are set, configure them,
+    # if not set, this will still default to current arch
+    foreach(ARCH ${CMAKE_OSX_ARCHITECTURES})
```
ras0219-msft, line 102:
```suggestion
    foreach(ARCH IN LISTS CMAKE_OSX_ARCHITECTURES)
```

ras0219-msft (approved):
I applied my suggestions to your branch -- LGTM and thanks for the PR!

ports/boost-modular-build-helper/user-config.jam
```diff
@@ -10,7 +10,7 @@ if "@VCPKG_PLATFORM_TOOLSET@" != "external"
 }
 else
 {
-    using gcc : 5.4.1 : @CMAKE_CXX_COMPILER@
+    using @B2_TOOLSET@ : : @CMAKE_CXX_COMPILER@
```
yurybura, line 13: After this change version will be lost. I think it would be good to add version like MSVC_VERSION. What do you think?

omartijn: We could do that. I'm unsure what the advantage would be, though.

AFAIK, the version there is useful if you declare multiple entries there, with different options for different version of a toolset, e.g. you want to set a `-std=c++11` option for gcc version 5 and up, but not for older versions.

Since we're only defining the one option, and not differentiating between versions I don't think it'd have any effect. I'm not a boost.build expert at all, though, and I may be completely wrong here. Feel free to correct me.

fairlyblank:
This PR fixed my issue #17440, thanks @omartijn

christophe-calmejane:
Hi, FYI, this patch broke iOS triplet.
Forcing `-mmacosx-version-min` for all Apple target, without checking `VCPKG_CMAKE_SYSTEM_NAME` causes compilation error if an iOS triplet tries to set `-mios-version-min`

PR #18549, opened 2021-06-19
[fdk-aac] Update to version 2.0.2.

Sibras:
Updated fdk to use latest tagged version.
Converted to using fdk's internal cmake file


ports/fdk-aac/portfile.cmake
```diff
@@ -3,21 +3,21 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
...
+    OPTIONS_RELEASE -DBUILD_PROGRAMS=OFF
 )
 
 vcpkg_install_cmake()
+vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/${PORT})
+vcpkg_fixup_pkgconfig()
 vcpkg_copy_pdbs()
-vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/tools/fdk-aac)
+vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/tools/${PORT})
 file(INSTALL ${SOURCE_PATH}/NOTICE DESTINATION ${CURRENT_PACKAGES_DIR}/share/fdk-aac RENAME copyright)
```
JonLiu1993, line 22:
```suggestion
file(INSTALL ${SOURCE_PATH}/NOTICE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
```

ports/fdk-aac/portfile.cmake
```diff
@@ -3,21 +3,21 @@ vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
 vcpkg_from_github(
```
JonLiu1993, line 3:
```suggestion
vcpkg_fail_port_install(ON_ARCH "arm"  ON_TARGET "uwp")
vcpkg_from_github(
```

JonLiu1993 (changes requested):

ports/fdk-aac/portfile.cmake
Sibras: The Ci passed on arm-uwp so im not sure what the issue is, am i missing something?

Sibras: Nevermind, I missed the supports field in the control file.

Sibras: The new cmake build file seems to support arm and uwp so this is no longer required

ports/fdk-aac/vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "fdk-aac",
+  "version-string": "2.0.2",
```
JonLiu1993, line 3: We should use "version-semver" instesd of "version-string"

versions/f-/fdk-aac.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "c5735b2f5cd0fd265adf4749b1e87d97df75b48c",
+      "version-string": "2.0.2",
```
JonLiu1993, line 5: We should use "version-semver" instesd of "version-string"

JonLiu1993 (changes requested):

ports/fdk-aac/vcpkg.json
Sibras: done

versions/f-/fdk-aac.json
Sibras: done

JonLiu1993 (approved):

ras0219-msft:
LGTM, thanks for the fix & update!

cenit:
can you please add the feature as a default one in vcpkg-ci-ffmpeg also for these new valid configs, so that we test it automatically?

PR #18551, opened 2021-06-20
[teemo] Update version to v2.2.

winsoft666:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes Build-Depends library.

  `Yes`

  <Yes / I am still working on this PR>


NancyLi1013 (commented):
Could you please run `vcpkg x-add-version teemo` to update version files and commit the changes?

winsoft666:
I have modified version to v2.2, please recheck again.

NancyLi1013:
> Could you please run `vcpkg x-add-version teemo` to update version files and commit the changes?

Please update version files via the above command.

winsoft666:
Sorry, It's my mistake.
I have updated again.

NancyLi1013 (approved):

ports/teemo/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: teemo
-Version: 2.1
-Build-Depends: curl[non-http]
+Version: 2.2
+Build-Depends: curl[non-http,openssl]
```
ras0219-msft, line 3: Why does this require `curl[openssl]`?

1. We default to using the Win32 APIs on Windows, not openssl. At a minimum, this needs to be qualified to non-windows platforms.
2. My understanding is that this is a behavior, not an API. `teemo` shouldn't be able to notice whether curl is built with or without SSL support and it's reasonable that a user might want to use `teemo` on top of curl without openssl.

ras0219-msft (commented):
Thanks for the PR!

ports/teemo/CONTROL
winsoft666: teemo is cross-platfrom, not noly support windows.
If you want libcurl support https, openssl is not optional.
Now https is very popular.

ras0219-msft: Our default build of libcurl does support https on windows -- but via the Win32 APIs, not openssl.

Could you try using:
```suggestion
Build-Depends: curl[core], openssl
```
and let us know specifically what does not work for you?

dg0yt: Port curl has the feature 'ssl'  which chooses an implementation (another feature) depending on the platform. So there is no need to hardcode 'openssl'.

winsoft666: @dg0yt @ras0219-msft 
I think both of yours opinion are right, I have updated to:
`Build-Depends: curl[core]`

NancyLi1013:
@winsoft666 
Could you please run`vcpkg x-add-version --overwrite-version teemo to solve the failures on x86-windows?
`

PR #17246, opened 2021-04-12
[openssl] fix cross compilation with mingw

d12fk:
When cross compiling openssl on Linux, e.g.

  ./vcpkg install openssl:x64-mingw-dynamic

failed, because 'windres' was not found. The call to
openssl's Configure was missing the WINDRES environment
variable, which is needed in case it is prefixed. On
Ubuntu the binutils-mingw-w64-x86-64 package installed
it as /usr/bin/x86_64-w64-mingw32-windres.

Since mingw uses openssl-unix to build, it's worth mentioning
that the WINDRES env has no impact for non-mingw and is ignored
there.

PR #18554, opened 2021-06-20
Fix mingw (cross) build issues in vcpkg_configure_make

dg0yt:
- #### What does your PR fix?  
  Fixes resource compilation on mingw. (Similar to #17246, but a) also sets RC, and b) without the unrelated changes to OpenSSL.)
  Fixes the incorrect passing of `${prefix}` to the shell where it would be expanded to an empty string before reaching `configure` (i.e. actual: `--bindir=/tools/gettext/bin`, expected: `--bindir=${prefix}/tools/gettext/bin` etc., logged in `config.log`).

  all (but made for mingw), no
  (Tested with libiconv:x64-mingw-static on Ubuntu 18.04.)

  yes

  not needed


JackBoosY:
Waiting for https://github.com/microsoft/vcpkg/pull/18579 merge to resolve the regression.

ras0219-msft:
Thanks!

PR #18590, opened 2021-06-22
[nuspell] Update to version 5.0.0

dimztimz:
- #### What does your PR fix?  
  Port update

  all

  `yes`

  yes


ports/nuspell/vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "nuspell",
+  "version-string": "5.0.0",
```
JonLiu1993, line 3: We should use "version-semver" instead of "version-string"

versions/n-/nuspell.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "c8fc12009a4d995ec09f86ab57ca78f33476669a",
+      "version-string": "5.0.0",
```
JonLiu1993, line 5: We should use "version-semver" instead of "version-string"

JonLiu1993 (changes requested):

JonLiu1993 (approved):

ras0219-msft:
Thanks for the update!

PR #18248, opened 2021-06-02
[CGNS|VTK-M|VTK|ParaView] Update ports

Neumann-A:
Update ports to match versions used by ParaView.
The update is required to maybe switch these ports to qt6 in the future. 
(only tested x64-windows-static; rest needs to be done by CI)

- [x] Fix missing svg icons in (static!) builds?

PR #18592, opened 2021-06-22
[cgns] Update to 4.2.0

Neumann-A:
extracted from #18248


Neumann-A:
seems to require #18248 so I'll close this

PR #18621, opened 2021-06-24
 [libdatachannel] Fix usage

### LINKED ISSUE #18616: [libdatachannel] It can find dependent libraries on its own.
My program depends on libdatachannel. Then libdatachannel depends on Threads::Threads and LibJuice::juice.
In my program CMakeLists.txt use:

find_package(libdatachannel)

The error message:
Target "RabbitProxy" links to target "Threads::Threads" but the target was not found.  Perhaps a find_package() call is missing for an IMPORTED target, or an ALIAS target is missing?

Of cause. can to add find_package(Threads).
But my expectation is that it can find dependent libraries on its own.
Is there a way to solve?

NancyLi1013:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #18616

Add `find_dependency() ` in libdatachannel-config.cmake to enable `libdatachannel` found successfully via `find_package(libdatachannel CONFIG REQUIRED)`.

### DIFF
--- a/ports/libdatachannel/portfile.cmake
+++ b/ports/libdatachannel/portfile.cmake
@@ -34,4 +34,12 @@ file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/cmake/libdatachannel)
 vcpkg_fixup_pkgconfig()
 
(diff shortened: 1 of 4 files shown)


NancyLi1013:
Related PR from upstream https://github.com/paullouisageneau/libdatachannel/pull/446

PR #18622, opened 2021-06-24
[fmt] Fix an easily avoidable compilation warning

PazerOP:
In the fmt package, there is a patch that adds `#pragma warning(disable:4189)` near the top of the main include file. This does not account for non-MSVC compilers, so I added a check for that.

- #### What does your PR fix?  
  Fixes fmt not being able to compile on non-MSVC compilers that treat unknown pragmas as errors (whether due to warnings-as-errors, or otherwise).

  Only a compile issue fix.

  Yes, only a compile issue fix.

  Yes.

### DIFF
--- a/ports/fmt/fix-warning4189.patch
+++ b/ports/fmt/fix-warning4189.patch
@@ -2,11 +2,13 @@ diff --git a/include/fmt/format.h b/include/fmt/format.h
 index 4e96539..0f1d179 100644
 --- a/include/fmt/format.h
 +++ b/include/fmt/format.h
-@@ -33,6 +33,7 @@
+@@ -33,6 +33,9 @@
  #ifndef FMT_FORMAT_H_
  #define FMT_FORMAT_H_
- 
+
++#ifdef _MSC_VER
 +#pragma warning(disable:4189)
++#endif
  #include <algorithm>
  #include <cerrno>
  #include <cmath>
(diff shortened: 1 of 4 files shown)


NancyLi1013 (commented):
Could you please also update` port-version` as 5 in vcpkg.json?

Hoikas:
IMO this is inappropriate for inclusion into vcpkg because it will change the warning state of consumer code.

PazerOP:
> Could you please also update` port-version` as 5 in vcpkg.json?

Done.

> Despite what OP says, this patch seems to add a new warning pragma where there was not one previously. IMO this is inappropriate for inclusion into vcpkg because it will change the warning state of consumer code.

It does not add a new warning pragma. See https://github.com/microsoft/vcpkg/blob/master/ports/fmt/fix-warning4189.patch. Although I agree with you it's not the best idea in the first place.

Hoikas:
Ah, I see. GitHub's diff views of diffs is playing with my eyes... IMO that whole patch should be deleted.

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your fixing @PazerOP.

ras0219-msft:
This PR is a strict upgrade, so I'm going to merge it as-is -- however I do agree that this should eventually be changed to a push+pop and upstreamed :)

Thanks for the PR!

PR #18625, opened 2021-06-24
[fontconfig] Fix x86-osx -> arm64-osx cross-compilation

ghost:
Pass host system cflags instead of target system cflags to tools that are supposed to run on host system.
Run host system version of fc-cache instead of target system.

- #### What does your PR fix?  

  x86-osx, arm64-osx No

  Yes

  Yes

### DIFF
--- a/ports/fontconfig/build2.patch
+++ b/ports/fontconfig/build2.patch
@@ -7,7 +7,7 @@ index f0fa0ec50..24e5afd16 100644
  
  $(TOOL): $(TSRC) $(ALIAS_FILES)
 -	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS)
-+	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS) $(LIBINTL) $(CFLAGS)
++	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS) $(LIBINTL) $(CFLAGS_FOR_BUILD)
  
  $(TARG): $(TMPL) $(TSRC) $(DEPS)
  	$(AM_V_GEN) $(MAKE) $(TOOL) && \
--- a/ports/fontconfig/portfile.cmake
+++ b/ports/fontconfig/portfile.cmake
@@ -98,13 +98,18 @@ file(INSTALL "${SOURCE_PATH}/COPYING" DESTINATION "${CURRENT_PACKAGES_DIR}/share
 if(NOT VCPKG_TARGET_IS_WINDOWS)
     set(ENV{FONTCONFIG_PATH} "${CURRENT_PACKAGES_DIR}/etc/fonts")
     set(ENV{FONTCONFIG_FILE} "${CURRENT_PACKAGES_DIR}/etc/fonts/fonts.conf")
-    vcpkg_execute_required_process(COMMAND "${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin/fc-cache${VCPKG_TARGET_EXECUTABLE_SUFFIX}" --verbose
+    if("${TARGET_TRIPLET}" STREQUAL "${HOST_TRIPLET}")
+        set(FC_CACHE_BASE "${CURRENT_PACKAGES_DIR}")
+    else()
+        set(FC_CACHE_BASE "${_VCPKG_INSTALLED_DIR}/${HOST_TRIPLET}")
+    endif()
+    vcpkg_execute_required_process(COMMAND "${FC_CACHE_BASE}/tools/${PORT}/bin/fc-cache${VCPKG_TARGET_EXECUTABLE_SUFFIX}" --verbose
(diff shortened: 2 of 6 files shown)


Neumann-A:
could you add:
```
    list(APPEND OPTIONS "as_ln_s=ln -sr")
endif()
```
before `vcpkg_configure_make`

This also fixes the broken symlinks in fontconfig

ports/fontconfig/build2.patch
```diff
@@ -7,7 +7,7 @@ index f0fa0ec50..24e5afd16 100644
  
  $(TOOL): $(TSRC) $(ALIAS_FILES)
 -	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS)
-+	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS) $(LIBINTL) $(CFLAGS)
++	$(AM_V_GEN) $(CC_FOR_BUILD) -o $(TOOL) $< $(AM_CPPFLAGS) $(LIBINTL) $(CFLAGS_FOR_BUILD)
```
dg0yt, line 10: Where does CFLAGS_FOR_BUILD come from, before https://github.com/microsoft/vcpkg/pull/15605?

Lucius-Q-User:
> could you add:
> 
> ```
> if(NOT VCPKG_TARGET_IS_WINDOWS)
>     list(APPEND OPTIONS "as_ln_s=ln -sr")
> endif()
> ```
> 
> before `vcpkg_configure_make`
> 
> This also fixes the broken symlinks in fontconfig

That breaks the build, macos version of ln does not support -r

Lucius-Q-User:
Looks like simage was broken even before the patch.

PhoebeHui:
simage failures would be fixed by PR https://github.com/microsoft/vcpkg/pull/18645

ports/fontconfig/portfile.cmake
```diff
@@ -98,13 +98,18 @@ file(INSTALL "${SOURCE_PATH}/COPYING" DESTINATION "${CURRENT_PACKAGES_DIR}/share
 if(NOT VCPKG_TARGET_IS_WINDOWS)
     set(ENV{FONTCONFIG_PATH} "${CURRENT_PACKAGES_DIR}/etc/fonts")
     set(ENV{FONTCONFIG_FILE} "${CURRENT_PACKAGES_DIR}/etc/fonts/fonts.conf")
-    vcpkg_execute_required_process(COMMAND "${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin/fc-cache${VCPKG_TARGET_EXECUTABLE_SUFFIX}" --verbose
+    if(NOT "${TARGET_TRIPLET}" STREQUAL "${HOST_TRIPLET}")
+        set(FC_CACHE_BASE "${_VCPKG_INSTALLED_DIR}/${HOST_TRIPLET}")
+    else()
+        set(FC_CACHE_BASE "${CURRENT_PACKAGES_DIR}")
+    endif()
```
PhoebeHui, line 105:
```suggestion
    if("${TARGET_TRIPLET}" STREQUAL "${HOST_TRIPLET}")
        set(FC_CACHE_BASE "${CURRENT_PACKAGES_DIR}")
    else()
        set(FC_CACHE_BASE "${_VCPKG_INSTALLED_DIR}/${HOST_TRIPLET}")
    endif()
```

ports/fontconfig/build2.patch
PhoebeHui: @Lucius-Q-User, I didn't find the CFLAGS_FOR_BUILD in fontconfig, so why do you change it?

Lucius-Q-User: CFLAGS refers to target system cflags, which can break the build, (e.g -arch flag for clang, or -march/-mcpu flags)

dg0yt: `CFLAGS_FOR_BUILD` might be set for cross builds after #15605, but for now you miss `CFLAGS` for native builds.
In addition, `$(LIBINTL)` can go wrong as well for cross builds.

Lucius-Q-User: Alternatively, we can update the version of fontconfig that is currently in the tree. Newer versions made fc-case, which is the source of cross-compile problems into a python script.

Lucius-Q-User: would it be ok to conditionally include cflags depending on whether we are cross-compiling or not?

PhoebeHui:
@Lucius-Q-User, the dependency PR has been closed, does this PR still be valid? can I close it?

Lucius-Q-User:
The dependency PR was closed because fontconfig was updated via a different PR. Since fontconfig was updated, it now uses a completely different build system (meson instead of automake) and this PR no longer applies.
The cross-compile is unfortunately still broken, but i believe that https://github.com/microsoft/vcpkg/pull/21772 should be able to fix it.

PhoebeHui:
@Lucius-Q-User, thanks for your response! It looks PR https://github.com/microsoft/vcpkg/pull/21772 aim to solve the problem.

Closing in favor of https://github.com/microsoft/vcpkg/pull/21772.

PR #18661, opened 2021-06-26
[stb] Add support for stb_vorbis and update port.

Colon-D:
**Describe the pull request**

- #### What does your PR fix?  
  I wanted to use the stb_vorbis library from stb.
  The current stb port does not have support for the stb_vorbis library.

  I have added support for stb_vorbis.

  I have also added some updates to the port.
  I have replaced the CONTROL file with a `vcpkg.json` manifest file since I needed to add some dependencies on vcpkg features.
  In the replacement file, I have updated the version string to "2021-06-20". That is the data of the latest commit in the stb repository.
  I updated the portfile's `REF` to the latest commit as well.

  To add stb_vorbis I used the compilation guide at https://nothings.org/stb_vorbis/.

  I made the portfile first cut the header out from the source file and write it to a seperate file.
  Then, the portfile needs to compile the source file as a library.
  First, it adds an export macro to the source file along with an include to find the header.
  Then it copies a `CMakeLists.txt` to the source path. I wrote this file to configure and install a library for stb_vorbis.
  Next, vcpkg configures and installs the library.

  The portfile installs the rest of the stb libraries as normal.

  I have no clue what `vcpkg-cmake-wrapper.cmake` is, and I do not think `FindStb.cmake` is still needed. If I do not install them, I can still use the other libraries, but I have kept them to be safe. At least keeping the `FindStb.cmake` should ensure that projects that use stb don't break.

  This is my first pull request. I hope that there are no issues. I updated this port for my personal use and it works for me so feel free to deny the request if there are issues.

  x64-windows, No
  Other triplets should be support but I have not tested them.

  Yes. (hopefully)

  Yes. It seemed to also update the formatting of `libsmb2` in `baseline.json`, I reverted that change since it is not relevant to this.

### DIFF
--- a/ports/stb/portfile.cmake
+++ b/ports/stb/portfile.cmake
@@ -1,10 +1,56 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO nothings/stb
-    REF b42009b3b9d4ca35bc703f5310eedc74f584be58 # accessed on 2020-09-14
-    SHA512 a9ba80d19dae4e527171bb02e1caa4d3eb0704cdf7f8fef1a7a09e6b05c65b829b0aa580f469f158a39bf13018289f40c1680ab5c22bfa5e932bff94eced475d
+    REF 80c8f6af0304588b9d780a41015472013b705194 # accessed on 2021-06-24
+    SHA512 0a1d8ad6b606a0904fce29855693a6fed6de1979272edf5bcb9837904dff79abaaf8f1ef9606cd5933396daa56c0851e6c1e3fc24547ab91accc7a1f17ace9f4
     HEAD_REF master
 )
+
+# create a header file for stb_vorbis
+file(READ "${SOURCE_PATH}/stb_vorbis.c" stb_vorbis)
+#   find beginning and length of header
+string(FIND "${stb_vorbis}" "//  HEADER BEGINS HERE" stb_vorbis_header_begin)
+string(FIND "${stb_vorbis}" "//  HEADER ENDS HERE" stb_vorbis_header_end)
+math(EXPR stb_vorbis_header_length "${stb_vorbis_header_end} - ${stb_vorbis_header_begin}")
+#   create a substring for the header
+string(SUBSTRING "${stb_vorbis}" "${stb_vorbis_header_begin}" "${stb_vorbis_header_length}" stb_vorbis_header)
+#   create a string for the source
+string(SUBSTRING "${stb_vorbis}" "0" "${stb_vorbis_header_begin}" stb_vorbis_source_before)
+string(SUBSTRING "${stb_vorbis}" "${stb_vorbis_header_end}" "-1" stb_vorbis_source_after)
+string(CONCAT stb_vorbis_source "${stb_vorbis_source_before}" "#include \"stb_vorbis.h\"\n" "${stb_vorbis_source_after}")
+#   write source to file, replacing the original source file
+file(WRITE "${SOURCE_PATH}/stb_vorbis.c" "${stb_vorbis_source}")
+#   prepend export macro
+string(PREPEND stb_vorbis_header [[
+#ifdef _WIN32
+    #ifdef STB_VORBIS_EXPORTS
+        #define STB_VORBIS __declspec(dllexport)
+    #else
+        #define STB_VORBIS __declspec(dllimport)
+    #endif
+#else
+    #define STB_VORBIS
+#endif
+]])
+#   replace all instances of 'extern' beyond 'extern "C"' with 'STB_VORBIS extern'
+#     split string into before 'extern "C"' and after 'extern "C"'
+string(FIND "${stb_vorbis_header}" "extern \"C\"" extern_c_pos)
+math(EXPR after_extern_c_pos "${extern_c_pos} + 10")
+string(SUBSTRING "${stb_vorbis_header}" "0" "${after_extern_c_pos}" before_extern_c)
+string(SUBSTRING "${stb_vorbis_header}" "${after_extern_c_pos}" "-1" after_extern_c)
+#     replace
+string(REPLACE "extern" "STB_VORBIS extern" after_extern_c_modified "${after_extern_c}")
+#     concatenate
+string(CONCAT stb_vorbis_header_modified "${before_extern_c}" "${after_extern_c_modified}")
+#   write header to file
+file(WRITE "${SOURCE_PATH}/stb_vorbis.h" "${stb_vorbis_header_modified}")
+# configure
+file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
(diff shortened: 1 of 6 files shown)


ports/stb/vcpkg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "stb",
+  "version-string": "2021-06-20",
```
NancyLi1013, line 3:
```suggestion
  "version-date": "2021-06-26",
```
This should be [version-date](https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md#version-date). We should use the date that the commit was accessed by you.

ports/stb/vcpkg.json
```diff
@@ -0,0 +1,10 @@
...
+  "description": "public domain header-only libraries",
+  "homepage": "https://github.com/nothings/stb",
+  "dependencies": [
+    "vcpkg-cmake",
+    "vcpkg-cmake-config"
```
NancyLi1013, line 8:
```suggestion
    {
      "name": "vcpkg-cmake",
      "host": true
    },
    {
      "name": "vcpkg-cmake-config",
      "host": true
    }
```

ports/stb/portfile.cmake
```diff
@@ -1,11 +1,57 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO nothings/stb
-    REF b42009b3b9d4ca35bc703f5310eedc74f584be58 # accessed on 2020-09-14
-    SHA512 a9ba80d19dae4e527171bb02e1caa4d3eb0704cdf7f8fef1a7a09e6b05c65b829b0aa580f469f158a39bf13018289f40c1680ab5c22bfa5e932bff94eced475d
+    REF 80c8f6af0304588b9d780a41015472013b705194 # accessed on 2021-06-24
```
NancyLi1013, line 4:
```suggestion
    REF 80c8f6af0304588b9d780a41015472013b705194 # accessed on 2021-06-26
```
The date should be consistent with that in vcpkg.json.

ports/stb/portfile.cmake
```diff
@@ -1,11 +1,57 @@
...
+# configure
+file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
+vcpkg_cmake_configure(SOURCE_PATH ${SOURCE_PATH})
+# install
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup()
+vcpkg_copy_pdbs()
+
 file(GLOB HEADER_FILES ${SOURCE_PATH}/*.h)
 file(COPY ${HEADER_FILES} DESTINATION ${CURRENT_PACKAGES_DIR}/include)
```
NancyLi1013, line 57: Since you have exported stb-config.cmake, FindStb.cmake and vcpkg-cmake-wrapper.cmake should be not needed now. Could you please help confirm this?

NancyLi1013 (changes requested):

ports/stb/portfile.cmake
Colon-D: I can confirm from testing that, at least for me, they are not needed. I thought that removing them would cause compatibility issues for those still using `find_package(stb CONFIG REQUIRED)` and `target_link_libraries(main PRIVATE stb::stb)` but it appears that I was wrong.

NancyLi1013: If the usage has changed, we should apply the usage to the ports that need to depend on `stb`. We should only keep one kind of usage. So, it's necessary to remove these two files.

ports/stb/vcpkg.json
NancyLi1013: The dependencies should be like the suggested one.

NancyLi1013 (changes requested):

ports/stb/portfile.cmake
Colon-D: Good to know. All other changes look good to me 👍

jesta88:
All stb libs have been updated a few days ago for stability. Could we update the portfiles in this PR or should I create a new one?

NancyLi1013:
There are still some review suggestions not updated now. So we still need to wait for the response from @Colon-D.

Colon-D:
It seems I have made some mistakes. I thought I had confirmed the review suggestions but this is not the case.
I have made another attempt at resolving the review suggestions.
As this is my first attempt at a contribution to a public repository I am making lots of mistakes. Let me know if there are any other issues.

NancyLi1013:
For the failures on x86-windows, you can run `vcpkg x-add-version --overwrite-version stb` and then submit the changes to solve it.

versions/s-/stb.json
```diff
@@ -1,5 +1,15 @@
...
+    {
+      "git-tree": "ac9f688b619776c2923b958c3e49d87e4601498f",
+      "version-date": "2021-06-26",
+      "port-version": 0
+    },
+    {
+      "git-tree": "177d66234c94ae95be4a7e8725b2f4b1d86a9386",
+      "version-string": "2021-06-20",
+      "port-version": 0
+    },
```
NancyLi1013, line 12:
```suggestion
```

NancyLi1013 (changes requested):

ports/stb/portfile.cmake
```diff
@@ -1,11 +1,57 @@
...
+string(FIND "${stb_vorbis_header}" "extern \"C\"" extern_c_pos)
+math(EXPR after_extern_c_pos "${extern_c_pos} + 10")
+string(SUBSTRING "${stb_vorbis_header}" "0" "${after_extern_c_pos}" before_extern_c)
+string(SUBSTRING "${stb_vorbis_header}" "${after_extern_c_pos}" "-1" after_extern_c)
+#     replace
+string(REPLACE "extern" "STB_VORBIS extern" after_extern_c_modified "${after_extern_c}")
+#     concatenate
+string(CONCAT stb_vorbis_header_modified "${before_extern_c}" "${after_extern_c_modified}")
+#   write header to file
+file(WRITE "${SOURCE_PATH}/stb_vorbis.h" "${stb_vorbis_header_modified}")
```
JackBoosY, line 46: I don't think we can do this because you modified the upstream code/added untracked files that were not included in the upstream.

JackBoosY (changes requested):

NancyLi1013:
Hi @Colon-D 

Could you please address the review suggestions?

Colon-D:
I am afraid not. I wanted to share the changes I had made incase anyone else wanted stb's vorbis library. Doing so became much too complicated so I gave up and switched to another language.

NancyLi1013:
> I am afraid not. I wanted to share the changes I had made incase anyone else wanted stb's vorbis library. Doing so became much too complicated so I gave up and switched to another language.

It's a bit pity. But thanks for your contribution to vcpkg anyway.

NancyLi1013:
You can reopen this PR if you want to continue to work on this PR in the future.

PR #18672, opened 2021-06-28
[glew] Fix the dependency

### LINKED ISSUE #18662: GLEW build failure on Ubuntu
**Host Environment**
- OS: Linux (Ubuntu 20.04 via Github Actions)
- Compiler: GCC and Clang

**To Reproduce**
(in Github Actions)
' git submodule add https://github.com/microsoft/vcpkg vcpkg/'
'pushd vcpkg'
'./bootstrap-vcpkg.sh -disableMetrics'
'./vcpkg install catch2 glew glfw3 --triplet x64-linux'
(permalink to yaml file https://github.com/Journeyman-dev/TUIC/blob/b97438671742a044c69f11a2814b5ef60fd37982/.github/workflows/Tests.yml)
(actions report https://github.com/Journeyman-dev/TUIC/actions/runs/974265097)

**Failure logs** 
(logs extracted from Github Actions using actions/archive)
```
CMake Error at /home/runner/work/TUIC/TUIC/vcpkg/downloads/tools/cmake-3.20.2-linux/cmake-3.20.2-linux-x86_64/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenGL (missing: OPENGL_opengl_LIBRARY OPENGL_glx_LIBRARY
  OPENGL_INCLUDE_DIR)
  /home/runner/work/TUIC/TUIC/vcpkg/downloads/tools/cmake-3.20.2-linux/cmake-3.20.2-linux-x86_64/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /home/runner/work/TUIC/TUIC/vcpkg/downloads/tools/cmake-3.20.2-linux/cmake-3.20.2-linux-x86_64/share/cmake-3.20/Modules/FindOpenGL.cmake:443 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  /home/runner/work/TUIC/TUIC/vcpkg/scripts/buildsystems/vcpkg.cmake:771 (_find_package)
  CMakeLists.txt:40 (find_package)
```

**Additional context**
it is apparent that vcpkg recognizes that opengl[core] is needed for glew to work and downloads it as a dependency:
```
Run ./vcpkg install catch2 glew glfw3 --triplet x64-linux
    catch2[core]:x64-linux -> 2.13.6
    glew[core]:x64-linux -> 2.1.0#10
    glfw3[core]:x64-linux -> 3.3.4
  * opengl[core]:x64-linux -> 0.0#8
  ```
 It should be noted that  opengl[core] is not needed for glfw3 or catch2, so it must be installing it for glew unless there is a mistake in one of those two packages. 
 
 Despite the fact that vcpkg successfully installs opengl[core], it fails to link it with glew correctly. Also note that instead of using the built in vcpkg install that ships with GitHub Actions server, I cloned it as a submodule to ensure that I am using the latest version of vcpkg. As you can see from the Actions report, this error does not occur on Windows and Mac operating systems.

### PR DESCRIPTION
**Pull request by PhoebeHui**
Fixes https://github.com/microsoft/vcpkg/issues/18662

glew depends on port opengl, which only built for windows, so limit to platform for dependencies. 
It requires libraries libxmu-dev libxi-dev libgl-dev which should be installed via system package manager on linux, providing the message with user.

### DIFF
--- a/ports/glew/portfile.cmake
+++ b/ports/glew/portfile.cmake
@@ -1,5 +1,5 @@
 if(VCPKG_TARGET_IS_LINUX)
-    message(WARNING "${PORT} currently requires the following libraries from the system package manager:\n    libxmu-dev\n    libxi-dev\n    libgl-dev\n\nThese can be installed on Ubuntu systems via apt-get install libxmu-dev libxi-dev libgl-dev.")
+    message(WARNING "${PORT} requires the following libraries from the system package manager:\n    libxmu-dev\n    libxi-dev\n    libgl-dev\n\nThese can be installed on Ubuntu systems via apt-get install libxmu-dev libxi-dev libgl-dev.")
 endif()
 
 # Don't change to vcpkg_from_github! The sources in the git repository (archives) are missing some files that are distributed inside releases.
--- a/ports/glew/portfile.cmake
+++ b/ports/glew/portfile.cmake
@@ -1,3 +1,7 @@
+if(VCPKG_TARGET_IS_LINUX)
+    message(WARNING "${PORT} currently requires the following libraries from the system package manager:\n    libxmu-dev\n    libxi-dev\n    libgl-dev\n\nThese can be installed on Ubuntu systems via apt-get install libxmu-dev libxi-dev libgl-dev.")
+endif()
+
 # Don't change to vcpkg_from_github! The sources in the git repository (archives) are missing some files that are distributed inside releases.
 # More info: https://github.com/nigels-com/glew/issues/31 and https://github.com/nigels-com/glew/issues/13
 vcpkg_download_distfile(ARCHIVE
--- a/ports/glew/vcpkg.json
+++ b/ports/glew/vcpkg.json
@@ -1,0 +1,13 @@
+{
+  "name": "glew",
+  "version-string": "2.1.0",
+  "port-version": 11,
+  "description": "The OpenGL Extension Wrangler Library (GLEW) is a cross-platform open-source C/C++ extension loading library.",
+  "homepage": "https://github.com/nigels-com/glew",
+  "dependencies": [
+    {
+      "name": "opengl",
+      "platform": "windows"
+    }
+  ]
+}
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by JackBoosY — APPROVED** [review]

File: ports/glew/vcpkg.json [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+  "name": "glew",
+  "version-string": "2.1.0",
+  "port-version": 11,
+  "description": "The OpenGL Extension Wrangler Library (GLEW) is a cross-platform open-source C/C++ extension loading library.",
+  "homepage": "https://github.com/nigels-com/glew",
+  "dependencies": [
+    {
+      "name": "opengl",
+      "platform": "windows"
+    }
```
ras0219-msft, line 11:
```suggestion
    "opengl"
```

The port `opengl` is intended to represent `find_package(OpenGL)`, so that is a true dependency on all platforms.

ports/glew/portfile.cmake
```diff
@@ -1,3 +1,7 @@
+if(VCPKG_TARGET_IS_LINUX)
+    message(WARNING "${PORT} currently requires the following libraries from the system package manager:\n    libxmu-dev\n    libxi-dev\n    libgl-dev\n\nThese can be installed on Ubuntu systems via apt-get install libxmu-dev libxi-dev libgl-dev.")
```
ras0219-msft, line 2:
```suggestion
    message(WARNING "${PORT} requires the following libraries from the system package manager:\n    libxmu-dev\n    libxi-dev\n    libgl-dev\n\nThese can be installed on Ubuntu systems via apt-get install libxmu-dev libxi-dev libgl-dev.")
```

ras0219-msft (changes requested):

ports/glew/vcpkg.json
PhoebeHui: Updated, thanks for your review!

ports/glew/portfile.cmake
PhoebeHui: Updated, thanks!

PR #18706, opened 2021-06-29
[crow] update to v0.3+1 (0.3.1)

luca-schlecker:
Signed-off-by: Luca Schlecker <luca.schlecker@hotmail.com>

**Describe the pull request**
Updates crow to version `v0.3+1`. In vcpkg, `0.3.1` is used though, because of the problems described here:
https://github.com/CrowCpp/crow/issues/159.

Yes


NancyLi1013 (approved):

ras0219-msft:
Cross-post from https://github.com/CrowCpp/crow/issues/159#issuecomment-872298971:

>What about using 0.3.1 and vcpkg's version scheme? It is explicitly not semver (that's version-semver) and has no semantic meaning other than lexicographic ordering.
>
>Using version-string is really unfortunately, because it almost completely prevents us from performing version analysis.

ras0219-msft:
Thanks!

PR #18707, opened 2021-06-29
[clapack] Fix the CLAPACK_FOUND issue and F2C_FOUND issue

### LINKED ISSUE #18679: CLAPACK_FOUND used before initialized?
I don't think there's anything setting CLAPACK_FOUND in the file below, so I don't believe it's setting-up the proper targets. I'm actually getting a build error when trying to use the lapack target after a successful call to find_package(LAPACK REQUIRED) on a Windows build.

Additionally, the targets are only correctly set-up the first time you call find_package(LAPACK). Subsequent calls will have all the oF2C_LIBRARY_RELEASE etc. variables not set and hence the targets will typically be -NOTFOUND.

https://github.com/microsoft/vcpkg/blob/bb002cdd1a2c9eacbe60d8383fb5d437752b1e4e/ports/clapack/FindLAPACK.cmake#L171

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/18679

CLAPACK_FOUND used before initialized in currently FindLAPACK.cmake, the CLAPACK_FOUND variable will be set to TRUE if all the variables are valid via find_package_handle_standard_args.


cenit:
I'd also unify targets between clapack and lapack port, modifying the ones in clapack (now that you are touching the module, it's the best time). 
The two cmake modules define different targets and those from clapack should be aligned with those from lapack-reference)

ports/clapack/FindLAPACK.cmake
```diff
@@ -167,6 +167,12 @@ set(F2C_LIBRARY_DEBUG "${F2C_LIBRARY_DEBUG}" CACHE STRING "" FORCE)
...
 mark_as_advanced(LAPACK_INCLUDE_DIR LAPACK_LIBRARY)
 
+find_package_handle_standard_args(CLAPACK DEFAULT_MSG CLAPACK_LIBRARY CLAPACK_INCLUDE_DIR)
+mark_as_advanced(CLAPACK_INCLUDE_DIR CLAPACK_LIBRARY)
+
+find_package_handle_standard_args(F2C DEFAULT_MSG F2C_LIBRARY F2C_INCLUDE_DIR)
+mark_as_advanced(F2C_INCLUDE_DIR F2C_LIBRARY)
+
 #TARGETS
 if(CLAPACK_FOUND AND NOT TARGET clapack::clapack)
```
bucurb, line 177: if you find_package(clapack) above, then all the code from here to the end could be removed, right? It looks to me like it was trying to setup the clapack target from scratch, but that should be redundant

JackBoosY:
Please reopen this PR when you continue it.

PR #18711, opened 2021-06-29
[drogon] Added database support features

Greisby:
**Added features to allow removing all database access features and dependencies.**

- #### What does your PR fix?  
  - Adds **mysql**, **postgresql**, **redis** and **sqlite3** features to allow enabling/disable what's needed
  - Fixes brotli libraries detection when the triplet is static (eg. x64-windows-static), since they don't have a the same name as in dynamic (brotli*XXX*-static instead of brotli*XXX*).

  No changes, tested with *x64-windows-static* and *x64-osx*

  Not sure, it's my first pull request.

  No, since I converted the CONTROL to vcpkg.json


ports/drogon/vcpkg.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "drogon",
+  "version": "1.6.0",
+  "port-version": 2,
```
JonLiu1993, line 4: "port-version" should set to 1

ports/drogon/vcpkg.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "drogon",
+  "version": "1.6.0",
```
JonLiu1993, line 3: We should use "version-semver" instead of "version"

JonLiu1993 (changes requested):

ports/drogon/vcpkg.json
Greisby: Sorry, I discover vcpkg, I don't yet master its packages.
I set it to 2 because I though that the previous CONTROL was considered as version 1.

JonLiu1993: If "port-version" is not written, it defaults to 0

ports/drogon/vcpkg.json
Greisby: You mean version-string?

JonLiu1993: Here we should use version-semver, you can refer to the [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/specifications/versioning.md#version-semver)

Greisby: Well, I don't see the difference between version and version-semver since this package version is dot-separated numeric, but ok, I amended my last commit.
I'm a brand new git user, and I just discovered that I can squash commits. Do you want me to merge my two commits or is it ok so?

Greisby: Well, after my last modification & my reply, it's still "requires:author-response"
Is there something I have to do?

JonLiu1993:
@Greisby ,Can   you please resolve the conflicts against master?

Greisby:
Yes, of course. But can you tell me how I do that?
Here the Resolve conflicts button is greyed out, and in my fork I can only create a new Pull request.
Is that what Is have to do?

JonLiu1993:
> Yes, of course. But can you tell me how I do that?
> Here the Resolve conflicts button is greyed out, and in my fork I can only create a new Pull request.
> Is that what Is have to do?

You can manually change the conflict between your branch portfile.cmake file and the CONTROL file to keep it consistent with the current master branch. Then add this pr modification on this basis

Greisby:
I followed this guide https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/syncing-a-fork I hope I dit it right.

JonLiu1993:
- x86-windows
- x64-windows
- x64-windows-static

JonLiu1993 (approved):

Greisby:
Latest changes:
- added new feature: redis
- added feature dependencies (msyql, postgres, redis, sqlite3 depend on orm)

Greisby:
Sad, I thought I could trigger the tests :)

PhoebeHui (approved):

ports/drogon/vcpkg.json
```diff
@@ -0,0 +1,101 @@
...
+    "redis": {
+      "description": "Support reading and writing from/to Redis databases.",
+      "dependencies": [
+        {
+          "name": "drogon",
+          "features": [
+            "orm"
+          ]
+        },
+        "redis"
```
PhoebeHui, line 85:
```suggestion
```

PhoebeHui: We don't have redis port.

JonLiu1993:
- x86-windows
- x64-windows
- x64-windows-static
- x64-linux

PhoebeHui (approved):

Greisby:
Hi, just a warning: I integrated the resolv.patch upstream.
It should fail with **vcpkg install --head**
Not sure if this breaks the port.

PhoebeHui:
@Greisby, it does not block this PR.

ports/drogon/vcpkg.json
```diff
@@ -0,0 +1,100 @@
...
+            "ctl",
+            "mysql",
+            "orm",
+            "postgres",
+            "redis",
+            "sqlite3"
+          ]
+        }
+      ]
+    },
```
vicroms, line 35: I think we should remove this feature since users can already do `vcpkg install drogon[*]` if they want all features.

vicroms (changes requested):
Please remove feature `all`

JonLiu1993 (approved):

PR #18736, opened 2021-06-30
[meson] Fix operator precedence in if statement

abique:
This pull request fixes a precedence issue with the NOT that was actually applied to the whole expression.

### DIFF
--- a/scripts/cmake/vcpkg_configure_meson.cmake
+++ b/scripts/cmake/vcpkg_configure_meson.cmake
@@ -316,7 +316,7 @@ function(vcpkg_internal_meson_generate_cross_file _additional_binaries) #https:/
         string(APPEND CROSS "cpu = '${BUILD_CPU}'\n")
     endif()
 
-    if(NOT BUILD_CPU_FAM MATCHES "${HOST_CPU_FAM}" OR VCPKG_TARGET_IS_ANDROID OR VCPKG_TARGET_IS_IOS OR VCPKG_TARGET_IS_UWP)
+    if((NOT BUILD_CPU_FAM MATCHES "${HOST_CPU_FAM}") OR VCPKG_TARGET_IS_ANDROID OR VCPKG_TARGET_IS_IOS OR VCPKG_TARGET_IS_UWP)
         set(_file "${CURRENT_BUILDTREES_DIR}/meson-cross-${TARGET_TRIPLET}.log")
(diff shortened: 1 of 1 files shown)


JackBoosY (approved):

scripts/cmake/vcpkg_configure_meson.cmake
```diff
@@ -316,7 +316,7 @@ function(vcpkg_internal_meson_generate_cross_file _additional_binaries) #https:/
         string(APPEND CROSS "cpu = '${BUILD_CPU}'\n")
     endif()
 
-    if(NOT BUILD_CPU_FAM MATCHES "${HOST_CPU_FAM}" OR VCPKG_TARGET_IS_ANDROID OR VCPKG_TARGET_IS_IOS OR VCPKG_TARGET_IS_UWP)
+    if((NOT BUILD_CPU_FAM MATCHES "${HOST_CPU_FAM}") OR VCPKG_TARGET_IS_ANDROID OR VCPKG_TARGET_IS_IOS OR VCPKG_TARGET_IS_UWP)
```
dg0yt, line 319: I don't see that this is needed. It is common syntax, and cmake documentation defines a clear order: 
MATCHES, then NOT, then OR. 
https://cmake.org/cmake/help/v3.21/command/if.html#condition-syntax

I could not reproduce a configuration where NOT "was actually applied to the whole expression".
~~~cmake
set(BUILD_CPU_FAM "x64")
set(HOST_CPU_FAM "x64")
set(VCPKG_TARGET_IS_ANDROID 1)
if(NOT BUILD_CPU_FAM MATCHES "${HOST_CPU_FAM}" OR VCPKG_TARGET_IS_ANDROID OR VCPKG_TARGET_IS_IOS OR VCPKG_TARGET_IS_UWP)
  message(OOPS)
endif()
~~~

abique: Ah yes you're correct, I got confused.
That being said the parenthesis make the expression clearer.
I'm OK to discard this pull request or change the commit message to "improve expression readability".

dg0yt: Actually the unneeded parenthesis make it less clear/readable to me. I will always wonder if they are obsolete or if the author had the intention of placing one of them in a different location.

PR #18741, opened 2021-06-30
[protobuf] Fixed protoc path

### LINKED ISSUE #18689: protoc target not available in cross-compile builds
It looks like the protobuf port file is trying to add a protoc target even for the case where the host is different from the target triplet. That's all great, but I don't understand what is supposed to call the protobuf-targets-vcpkg-protoc.cmake file. The toolchain does not, and neither do any of the config or targets .cmake files.

https://github.com/microsoft/vcpkg/blob/6cdc26c884ac1c7c4cf2d480b364b773765622a3/ports/protobuf/portfile.cmake#L123

bobuc-havok:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #18689 

  all previously supported, No

  Yes

  Yes

### DIFF
--- a/ports/protobuf/vcpkg-cmake-wrapper.cmake
+++ b/ports/protobuf/vcpkg-cmake-wrapper.cmake
@@ -11,6 +11,6 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.3)
     cmake_policy(POP)
 endif()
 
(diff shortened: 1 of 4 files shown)


PhoebeHui (approved):
LGTM, thanks for the fix.

The tools copy to target triplet, see
https://github.com/microsoft/vcpkg/blob/master/ports/protobuf/portfile.cmake#L85

ras0219-msft:
Thanks for the fix!

This is a fine fix for now, though we would eventually like to stop copying the tools across to the target package. The eventual fix in that case is actually to replace the `COPYONLY` from portfile.cmake:123 with `@ONLY`, which will substitute in the correct variable.

PR #18768, opened 2021-07-02
[socket-io-client]fix socket-io-client use 

JonLiu1993:
Fix socket-io-client use, add target_link_libraries() for target sioclient_tls

### DIFF
--- a/ports/socket-io-client/portfile.cmake
+++ b/ports/socket-io-client/portfile.cmake
@@ -11,8 +11,6 @@ vcpkg_from_github(
         fix-error-C3321.patch
 )
 
-file(REMOVE "${SOURCE_PATH}/cmake/modules/Findzstd.cmake")
-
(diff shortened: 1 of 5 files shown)


NancyLi1013 (approved):

ports/socket-io-client/portfile.cmake
```diff
@@ -11,8 +11,6 @@ vcpkg_from_github(
         fix-error-C3321.patch
 )
 
-file(REMOVE "${SOURCE_PATH}/cmake/modules/Findzstd.cmake")
-
```
ras0219-msft, line 15: Why does this module no longer need to be removed?

JonLiu1993: The source path does not exist

ras0219-msft:
LGTM

PR #18782, opened 2021-07-02
[scripts-audit] vcpkg_build_ninja

strega-nil:
See issue #17691

### DIFF
--- a/scripts/cmake/vcpkg_build_ninja.cmake
+++ b/scripts/cmake/vcpkg_build_ninja.cmake
@@ -15,26 +15,33 @@ vcpkg_build_ninja(
 Only build the specified targets.
 #]===]
 
+function(z_vcpkg_build_ninja_build config targets)
+    message(STATUS "Building (${config})...")
+    vcpkg_execute_build_process(
+        COMMAND "${NINJA}" -C "${CURRENT_BUILDTREES_DIR}/${config}" ${targets}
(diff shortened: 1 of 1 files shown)


scripts/cmake/vcpkg_build_ninja.cmake
```diff
@@ -15,26 +15,33 @@ vcpkg_build_ninja(
...
-    cmake_parse_arguments(PARSE_ARGV 0 _vbn "" "" "TARGETS")
+    cmake_parse_arguments(PARSE_ARGV 0 arg "" "" "TARGETS")
 
-    vcpkg_find_acquire_program(NINJA)
+    if(DEFINED arg_UNPARSED_ARGUMENTS)
+        message(WARNING "${CMAKE_CURRENT_FUNCTION} was passed extra arguments: ${arg_UNPARSED_ARGUMENTS}")
+    endif()
+    if(NOT DEFINED arg_TARGETS)
+        message(FATAL_ERROR "TARGETS must be specified")
+    endif()
```
ras0219-msft, line 36: Previously, targets being undefined would have the effect of `ninja -C /path/to/dir`, which is a valid way of invoking ninja. I think this mode needs to continue to be supported

ras0219-msft (approved):

strega-nil-ms:

PR #18785, opened 2021-07-02
[scripts-audit] vcpkg_replace_string

strega-nil:
See issue #17691


strega-nil-ms:

PR #18787, opened 2021-07-02
[Qt6] Update to 6.1.2

Neumann-A:
Testing in CI because I cannot test it locally with (VS 16.10.1):
`
INTERNAL COMPILER ERROR in 'C:\PROGRA~2\MIB055~1\2019\COMMUN~1\VC\Tools\MSVC\1429~1.300\bin\Hostx64\x64\cl.exe'
`
in the release build of qtbase

updating to VS 16.10.3 fixed it so I can locally test again


Neumann-A:
don't know whats going on with x64-windows-static-md but its probably a CI regression and not related to this pr

Neumann-A:
So no drama here just one patch adjustment....

JackBoosY (approved):

FranciscoPombal:
@Neumann-A 

It seems vcpkg is always building Qt6 with the zstd-feature even when I explicitly list all features except the one:

https://github.com/FranciscoPombal/qBittorrent/runs/2985850646?check_suite_focus=true (workflow flie: https://github.com/FranciscoPombal/qBittorrent/actions/runs/999478507/workflow).

We previously found that a valid workaround would be to make sure we don't do `set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)`, but I'd also like to prevent this problem by simply building Qt6 without zstd support, as qBittorrent does not use it anyway.

Neumann-A:
@FranciscoPombal: You seem to be missing the "core" feature which disables building default-features in classic mode.

FranciscoPombal:
@Neumann-A 

> @FranciscoPombal: You seem to be missing the "core" feature which disables building default-features in classic mode.

OK, I'm trying that out now.

Question: "classic" mode as opposed to ... ? Is there a new way to install packages other than `vcpkg install foo[feat-bar,feature-baz]:triplet`?

Neumann-A:
> Question: "classic" mode as opposed to ... ? Is there a new way to install packages other than vcpkg install foo[feat-bar,feature-baz]:triplet?

Manifest mode: https://github.com/microsoft/vcpkg/blob/master/docs/specifications/manifests.md
Just have a `vcpkg.json` in your `CMAKE_SOURCE_DIR` and vcpkg automatically installs the dependencies. 
Add a nuget binary cache and you are good to go (although that also works in classic mode).

FranciscoPombal:
@Neumann-A 

Interesting, I thought manifest mode was largely an implementation detail that simply meant "the requested packages for installation happen to use a `vcpkg.json` file instead of a legacy `CONTROL` file", and that the behavior when installing a package would be the same.

Thanks for the clarification.

FranciscoPombal:
@Neumann-A 

It seems to work, but is it normal that it's saying it's actually installing both variants? https://github.com/FranciscoPombal/qBittorrent/runs/2991297503?check_suite_focus=true#step:7:1618

Neumann-A:
@FranciscoPombal You need to set `VCPKG_HOST_TRIPLET` (CMake); manifest) or pass `--host-triplet=<your_triplet>`. Otherwise it will use the default host triplet for the detected host.

ras0219-msft:
LGTM, thanks as always @Neumann-A!

Issue #18809, opened 2021-07-05
[tensorflow] Build failure on x64-osx


JackBoosY:
When building tensorflow on x64-osx:
```
Starting package 2/2: tensorflow:x64-osx
Building package tensorflow[core]:x64-osx...
CMake Error at scripts/cmake/vcpkg_execute_required_process.cmake:105 (message):
    Command failed: /usr/bin/python3 -c "import numpy"
    Working Directory: /Users/vagrant/Data/buildtrees/tensorflow
    Error code: 1
    See logs for more information:
      /Users/vagrant/Data/buildtrees/tensorflow/prerequesits-numpy-x64-osx-err.log

Call Stack (most recent call first):
  /Users/vagrant/Data/installed/x64-osx/share/tensorflow-common/tensorflow-common.cmake:55 (vcpkg_execute_required_process)
  ports/tensorflow/portfile.cmake:7 (include)
  scripts/ports.cmake:139 (include)


Error: Building package tensorflow:x64-osx failed with: BUILD_FAILED
Elapsed time for package tensorflow:x64-osx: 1.311 s
```

`/Users/vagrant/Data/buildtrees/tensorflow/prerequesits-numpy-x64-osx-err.log` shows:
```
init_dgelsd failed init
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/vagrant/Library/Python/3.8/lib/python/site-packages/numpy/__init__.py", line 391, in <module>
    raise RuntimeError(msg)
RuntimeError: Polyfit sanity test emitted a warning, most likely due to using a buggy Accelerate backend. If you compiled yourself, more information is available at https://numpy.org/doc/stable/user/building.html#accelerated-blas-lapack-libraries Otherwise report this to the vendor that provided NumPy.
RankWarning: Polyfit may be poorly conditioned
```

JackBoosY:
cc @jgehw

JackBoosY:
Same reason with https://github.com/microsoft/vcpkg/issues/18597.

PR #18695, opened 2021-06-29
[vcpkg] Update the minimum version of vcpkg

JackBoosY:
Since microsoft/vcpkg and microsoft/vcpkg-tool are synchronized to update `x-download` related content, update to `2021-05-05` (contains `x-download`) version to avoid being unable to find this command:
```
E:\Lib\vcpkg>vcpkg install sundials --triplet x64-windows
Computing installation plan...
The following packages will be built and installed:
    sundials[core]:x64-windows -> 5.7.0#1
Detecting compiler hash for triplet x64-windows...
Could not locate cached archive: C:\Users\ch6535\AppData\Local\vcpkg\archives\df\df61b63fc4d86a443bbedaa77be748372df4bfe
c.zip
Starting package 1/1: sundials:x64-windows
Building package sundials[core]:x64-windows...
-- Downloading https://github.com/LLNL/sundials/archive/73c280cd55ca2b42019c8a9aa54af10e41e27b9d.tar.gz -> LLNL-sundials
-73c280cd55ca2b42019c8a9aa54af10e41e27b9d.tar.gz...
[DEBUG] Feature flag 'binarycaching' unset
[DEBUG] Feature flag 'manifests' = off
[DEBUG] Feature flag 'compilertracking' unset
[DEBUG] Feature flag 'registries' unset
[DEBUG] Feature flag 'versions' unset
[DEBUG] BuiltinRegistry initialized with: ""
[DEBUG] Using vcpkg-root: E:\Lib\vcpkg
[DEBUG] Using installed-root: E:\Lib\vcpkg\installed
[DEBUG] BuiltinRegistry initialized with: ""
[DEBUG] Using buildtrees-root: E:\Lib\vcpkg\buildtrees
[DEBUG] Using downloads-root: E:\Lib\vcpkg\downloads
[DEBUG] Using packages-root: E:\Lib\vcpkg\packages
[DEBUG] Using scripts-root: E:\Lib\vcpkg\scripts
[DEBUG] Using ports-root: E:\Lib\vcpkg\ports
[DEBUG] Using versions-root: E:\Lib\vcpkg\versions
invalid command: x-download
```

Related: https://github.com/microsoft/vcpkg/pull/13639#discussion_r659941583 https://github.com/microsoft/vcpkg/issues/18353 etc.

PR #18812, opened 2021-07-05
[vcpkg baseline][tensorflow] Skip osx check in baseline

JackBoosY:
`tensorflow:x64-osx` has blocked some PR testing and cannot fix it in the short term (same reason with https://github.com/microsoft/vcpkg/issues/18597). Disable it in the pipeline test.

Related: https://github.com/microsoft/vcpkg/issues/18809  https://github.com/microsoft/vcpkg/pull/18695


PhoebeHui (approved):

strega-nil-ms:
Thanks @JackBoosY :)

PR #18816, opened 2021-07-05
[yasm/yasm-tool] Incorporate yasm-tool into yasm

### LINKED ISSUE #18746: [yasm] Dependent library of yasm binary is missing
**Host Environment**
- OS: Windows 8.1 x64, up-to-date
- Compiler: MSVC 19.29.30038.1

**To Reproduce**
vcpkg install mpg123:x86-windows --> fails
vcpkg install mpg123:x64-windows --> fails
vcpkg install mpg123:x86-windows-static --> succeeds
vcpkg install mpg123:x64-windows-static --> succeeds

**Failure logs** 
**x86**
```
    mpg123[core]:x86-windows -> 1.28.0
Detecting compiler hash for triplet x86-windows...
Could not locate cached archive: C:\Users\Daniel\AppData\Local\vcpkg\archives\7f\7ff2d3bab8cfe01abcb3de9fd52e93f808c52c7bc03a32808457f3d82d7cc346.zip
Starting package 1/1: mpg123:x86-windows
Building package mpg123[core]:x86-windows...
-- Downloading https://sourceforge.net/projects/mpg123/files/mpg123/1.28.0/mpg123-1.28.0.tar.bz2/download;https://sourceforge.net/projects/mpg123/files/mpg123
et;https://sourceforge.net/projects/mpg123/files/mpg123/1.28.0/mpg123-1.28.0.tar.bz2/download?use_mirror=freefr;https://sourceforge.net/projects/mpg123/files/
s://sourceforge.net/projects/mpg123/files/mpg123/1.28.0/mpg123-1.28.0.tar.bz2/download?use_mirror=netix;https://sourceforge.net/projects/mpg123/files/mpg123/1
-- Extracting source D:/Code/Repos/vcpkg/downloads/mpg123-1.28.0.tar.bz2
-- Applying patch 0001-fix-checkcpuarch-path.patch
-- Using source at D:/Code/Repos/vcpkg/buildtrees/mpg123/src/4e333ee4f3-f9ae7d9a6e.clean
-- Configuring x86-windows
-- Building x86-windows-dbg
CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:146 (message):
    Command failed: "C:/Program Files/CMake/bin/cmake.exe" --build . --config Debug --target install -- /p:VCPkgLocalAppDataDisabled=true /p:UseIntelMKL=No /m
    Working Directory: D:/Code/Repos/vcpkg/buildtrees/mpg123/x86-windows-dbg
      D:\Code\Repos\vcpkg\buildtrees\mpg123\install-x86-windows-dbg-out.log

  scripts/cmake/vcpkg_build_cmake.cmake:104 (vcpkg_execute_build_process)
  scripts/cmake/vcpkg_install_cmake.cmake:44 (vcpkg_build_cmake)
  ports/mpg123/portfile.cmake:33 (vcpkg_install_cmake)


Error: Building package mpg123:x86-windows failed with: BUILD_FAILED
  Package: mpg123:x86-windows
  Vcpkg version: 2021-05-05-e8977e69d9a3fb462d9ad42013d83a7682706659
```
**x64**
```
D:\Code\Repos\vcpkg>vcpkg install mpg123:x64-windows
    mpg123[core]:x64-windows -> 1.28.0
Could not locate cached archive: C:\Users\Daniel\AppData\Local\vcpkg\archives\7e\7ef4e1a094c74c70480583abe8b85b6aabc42d923dc1a5d95c544d2fe4667652.zip
Starting package 1/1: mpg123:x64-windows
Building package mpg123[core]:x64-windows...
-- Using D:/Code/Repos/vcpkg/downloads/mpg123-1.28.0.tar.bz2
-- Cleaning sources at D:/Code/Repos/vcpkg/buildtrees/mpg123/src/4e333ee4f3-f9ae7d9a6e.clean. Use --editable to skip cleaning for the packages you specify.
-- Extracting source D:/Code/Repos/vcpkg/downloads/mpg123-1.28.0.tar.bz2
-- Applying patch 0001-fix-checkcpuarch-path.patch
-- Using source at D:/Code/Repos/vcpkg/buildtrees/mpg123/src/4e333ee4f3-f9ae7d9a6e.clean
-- Configuring x64-windows
-- Building x64-windows-dbg
CMake Error at scripts/cmake/vcpkg_execute_build_process.cmake:146 (message):
    Command failed: "C:/Program Files/CMake/bin/cmake.exe" --build . --config Debug --target install -- /p:VCPkgLocalAppDataDisabled=true /p:UseIntelMKL=No /m
    Working Directory: D:/Code/Repos/vcpkg/buildtrees/mpg123/x64-windows-dbg
      D:\Code\Repos\vcpkg\buildtrees\mpg123\install-x64-windows-dbg-out.log

  scripts/cmake/vcpkg_build_cmake.cmake:104 (vcpkg_execute_build_process)
  scripts/cmake/vcpkg_install_cmake.cmake:44 (vcpkg_build_cmake)
  ports/mpg123/portfile.cmake:33 (vcpkg_install_cmake)


Error: Building package mpg123:x64-windows failed with: BUILD_FAILED
  Package: mpg123:x64-windows
  Vcpkg version: 2021-05-05-e8977e69d9a3fb462d9ad42013d83a7682706659
```

**Additional context**
The issue was introduced in commit 5edbbf1a70a0a502e6362f51bf875b96469ed807, which updated the mpg123 version from 1.26.5  to 1.28.0. This obviously means that both the mpg123 source code and multiple port-related vcpkg files got changed, and I am not 100% if we have an actual vcpkg port bug here or if the src itself is the culprit. I mention this as I tried to manually compile the src using latest VS 2019, and the process terminated with exactly the same - long - error message you'll find in one of the logs attached. 

Very interesting is the fact that building static libs works perfectly fine, so I the issue has to be connected to the building of _dynamic_ libraries somehow.

Ah, before you ask, yes, I tested with my "external" Ninja and CMake installations removed from %PATH% and Program Files as well, with exactly the same results.

Attached you'll find all logs generated during the operations mentioned above.

[mpg123_vcpkg_logs.zip](https://github.com/microsoft/vcpkg/files/6745472/mpg123_vcpkg_logs.zip)

JackBoosY:
Since port `yasm` and `yasm-tool` repo is same and generate the same files, remove port `yasm-tool` and switch the related dependency item `yasm-tool` to `yasm`.
Add option `DYNAMIC_DEPENDS` to `vcpkg_copy_tool` and `vcpkg_copy_tool_dependencies` to copy the dynamic dependencies (Dynamically load dependencies in code) too _tools/${PORT}_.

Fixes #18746.


JackBoosY:
cc @BillyONeal

JackBoosY:
Note that if the `DYNAMIC_DEPENDS` change is approved, all ports that call `vcpkg_copy_tool_dependencies` should be modified.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -20,6 +26,7 @@ This command should always be called by portfiles after they have finished rearr
 #]===]
 
 function(vcpkg_copy_tool_dependencies TOOL_DIR)
+    cmake_parse_arguments(PARSE_ARGV 0 _vctd "" "TOOL_DIR" "DYNAMIC_DEPENS")
```
dg0yt, line 29: Now there is a `TOOL_DIR` parameter *and* a `_vctd_TOOL_DIR` variable?
The preferred prefix will be `arg` soon ([guidelines PR](https://github.com/microsoft/vcpkg/pull/18319/files#diff-c07ff3d526bf3803125de60a85e94b82fa5479002c37439a6e3c8307b7eff4d6R103)).

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -43,5 +50,16 @@ function(vcpkg_copy_tool_dependencies TOOL_DIR)
         endmacro()
         search_for_dependencies("${CURRENT_PACKAGES_DIR}/bin")
         search_for_dependencies("${CURRENT_INSTALLED_DIR}/bin")
+        
+        if (_vctd_DYNAMIC_DEPENS)
+            foreach (SEARCH_ITEM ${_vctd_DYNAMIC_DEPENS})
```
dg0yt, line 55: `foreach (SEARCH_ITEM IN LISTS _vctd_DYNAMIC_DEPENS)`

scripts/cmake/vcpkg_copy_tools.cmake
```diff
@@ -63,9 +67,9 @@ function(vcpkg_copy_tools)
         endif()
     endforeach()
 
+    vcpkg_copy_tool_dependencies(TOOL_DIR "${_vct_DESTINATION}" DYNAMIC_DEPENS "${_vct_DYNAMIC_DEPENS}")
```
dg0yt, line 70: IIUC `_vct_DYNAMIC_DEPENS` may be undefined or a list. In other spots, no double quotes are used when passing lists.

docs/maintainers/vcpkg_copy_tool_dependencies.md
```diff
@@ -6,11 +6,17 @@ Copy all DLL dependencies of built tools into the tool folder.
...
+    TOOL_DIR <${CURRENT_PACKAGES_DIR}/tools/${PORT}>
+    [DYNAMIC_DEPENS <dep1>...]
+)
 ```
-## Parameters
+## TOOL_DIR
 The path to the directory containing the tools.
 
+### DYNAMIC_DEPENS
+A list of tool's dynamic dependency library names.
```
dg0yt, line 18: If this targets "plugins", it would be much clearer to me if this would be mentioned explicitly.
(Feeling even more reminded of CMake's BundleUtilities than before.)

cenit:
note also the DEPENS typo (should be corrected into DEPENDS)

docs/maintainers/vcpkg_copy_tool_dependencies.md
BillyONeal: How about:

> A list of libraries a tool is likely to load at runtime, such as plugins, or other [Run-Time Dynamic Linking](https://docs.microsoft.com/en-us/windows/win32/dlls/run-time-dynamic-linking) mechanisms like `LoadLibrary` or `dlopen`.

BillyONeal (commented):
I agree with @dg0yt 's feedback

docs/maintainers/vcpkg_copy_tool_dependencies.md
```diff
@@ -6,11 +6,20 @@ Copy all DLL dependencies of built tools into the tool folder.
 
 ## Usage
 ```cmake
-vcpkg_copy_tool_dependencies(<${CURRENT_PACKAGES_DIR}/tools/${PORT}>)
+vcpkg_copy_tool_dependencies(
+    TOOL_DIR <${CURRENT_PACKAGES_DIR}/tools/${PORT}>
```
dg0yt, line 10: With regard to the most frequent usage, how about making the parameter optional now, with a default value of `<${CURRENT_PACKAGES_DIR}/tools/${PORT}>`?

BillyONeal (approved):

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -19,7 +28,13 @@ This command should always be called by portfiles after they have finished rearr
...
 #]===]
 
-function(vcpkg_copy_tool_dependencies TOOL_DIR)
+function(vcpkg_copy_tool_dependencies)
+    cmake_parse_arguments(PARSE_ARGV 0 arg "" "TOOL_DIR" "DEPENDENCIES")
+    
+    if (NOT DEFINED arg_TOOL_DIR)
+        set(arg_TOOL_DIR "${CURRENT_PACKAGES_DIR}/tools/${PORT}")
+    endif()
+
```
ras0219-msft, line 37: This is a breaking change to the function's signature, which is not allowed.

```suggestion
function(vcpkg_copy_tool_dependencies TOOL_DIR)
    cmake_parse_arguments(PARSE_ARGV 1 arg "" "" "DEPENDENCIES")
```

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -43,5 +58,16 @@ function(vcpkg_copy_tool_dependencies TOOL_DIR)
         endmacro()
         search_for_dependencies("${CURRENT_PACKAGES_DIR}/bin")
         search_for_dependencies("${CURRENT_INSTALLED_DIR}/bin")
+        
+        if (arg_DEPENDENCIES)
+            foreach (SEARCH_ITEM IN_LIST ${arg_DEPENDENCIES})
```
ras0219-msft, line 63:
```suggestion
            foreach (SEARCH_ITEM IN_LIST arg_DEPENDENCIES)
```

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -43,5 +58,16 @@ function(vcpkg_copy_tool_dependencies TOOL_DIR)
         endmacro()
         search_for_dependencies("${CURRENT_PACKAGES_DIR}/bin")
         search_for_dependencies("${CURRENT_INSTALLED_DIR}/bin")
+        
+        if (arg_DEPENDENCIES)
```
ras0219-msft, line 62: Because this happens _after_ the `search_for_dependencies` above, the "DEPENDENCIES" DLLs will not have their transitive dependencies deployed.

To fix this, this `if()` block must be moved above `search_for_dependencies("${CURRENT_PACKAGES_DIR}/bin")` so we will properly call `applocal.ps1` on them.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -43,5 +58,16 @@ function(vcpkg_copy_tool_dependencies TOOL_DIR)
...
+            foreach (SEARCH_ITEM IN_LIST ${arg_DEPENDENCIES})
+                if (EXISTS "${CURRENT_PACKAGES_DIR}/bin/${SEARCH_ITEM}")
+                    debug_message("Copying file ${CURRENT_PACKAGES_DIR}/bin/${SEARCH_ITEM} to ${arg_TOOL_DIR}")
+                    file(COPY "${CURRENT_PACKAGES_DIR}/bin/${SEARCH_ITEM}" DESTINATION "${arg_TOOL_DIR}")
+                else()
+                    message(WARNING "Dynamic dependency ${SEARCH_ITEM} not found in ${CURRENT_PACKAGES_DIR}/bin.")
```
ras0219-msft, line 68:
```suggestion
                    message(FATAL_ERROR "Dynamic dependency ${SEARCH_ITEM} not found in ${CURRENT_PACKAGES_DIR}/bin.")
```
This should not be a warning, because the port should have perfect information about whether it calls `LoadLibrary()`. For transitive cases where the port doesn't have perfect information, `applocal.ps1` should be extended to correctly handle those calls.

versions/y-/yasm-tool.json
```diff
@@ -1,14 +0,0 @@
-{
```
ras0219-msft, line 1: `versions/y-/foo.json` files must _never_ be removed.

ports/yasm-tool/portfile.cmake
```diff
@@ -1,47 +0,0 @@
-vcpkg_fail_port_install(ON_TARGET "uwp")
```
ras0219-msft, line 1: Removing `yasm-tool` has nothing to do with fixing the actual linked issue. Please separate this change into a separate PR that clearly states what problem is solved.

ras0219-msft (changes requested):
The issue that this PR intends to fix (#18746) only requires a fraction of the changes made here -- just the addition of `DEPENDENCIES` to the `vcpkg_copy_tool*()` functions.

Please remove the unneeded changes from this PR and, if there is an issue solved by removing vcpkg-tool, open another PR with a linked issue.

JackBoosY:
> The issue that this PR intends to fix (#18746) only requires a fraction of the changes made here -- just the addition of `DEPENDENCIES` to the `vcpkg_copy_tool*()` functions.

What `vcpkg_copy_tool` implements is only to copy the executables to the specified directory. Finding and copying dependencies is implemented by `vcpkg_copy_tool_dependencies`.
In many ports, `vcpkg_copy_tool_dependencies` is directly called, which is why `DEPENDENCIES` is added to `vcpkg_copy_tool_dependencies`.

ports/yasm-tool/portfile.cmake
JackBoosY: In fact, this is the original modification of the PR.

I did the following steps:
1. Fix yasm binary ->
2. Found that the yasm binary is the same as the yasm-tool binary ->
3. Found that the port yasm does the same thing to yasm-tool, except for the cmake wrapper ->
4. Merged port yasm and port yasm-tool ->
5. Add parameters to vcpkg_copy_tool_dependencies to handle dynamic link items.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
JackBoosY: I'm worried that due to some configuration reasons, some dynamic libraries will not be generated in the _bin_ folder and this error will be reported in the future.

JackBoosY:
@ras0219-msft 
Let me explain more clearly:
All the modifications of this PR are to solve one issue: find the same binary yasm generated by different ports in vcpkg.
Port yasm is the same as port yasm-tool's repo, and the build process is the same, except that yasm-tool deletes lib and include.
So we don't need to add the yasm tool as a separate port, but should fix the tool of yasm itself.

Christsnatcher:
> Port yasm is the same as port yasm-tool's repo, and the build process is the same, except that yasm-tool deletes lib and include.

That's not 100% true, port yasm-tool only depends on system .runtime .dlls, while yasm.exe of port yasm directly depends on yasm.dll (which is placed correctly next to the .exe) AND requires yasmstd.dll for building dynamic libraries. And as I stated in my initial mpg123 issue which started all this "movement", this .dll gets "misplaced" inside the resp. "bin" folder where it is not found by the yasm.exe, which causes building of all dynamic libraries via yasm.exe to fail.

That being said, if you want to add the tool feature to the yasm port, you must alter its building conditions in a way that makes yasmstd.dll end up next to yasm.exe (same as yasm.dll), otherwise the whole action is pretty pointless.

JackBoosY:
> That's not 100% true, port yasm-tool only depends on system .runtime .dlls, 

That's because yasm-tool force to build static yasm, so the executable doesn't need the dynamic library.

> That being said, if you want to add the tool feature to the yasm port, you must alter its building conditions in a way that makes yasmstd.dll end up next to yasm.exe (same as yasm.dll), otherwise the whole action is pretty pointless.

Already done that.

Christsnatcher:
> Already done that.

Perfect, I didn't mean to lecture or bother you, only reminded you just in case. But you have everything under control as I see, so I'll just wait for the result and be happy then. Thank you!

JackBoosY:
Successfully tested the following affected ports:
1. `mpg123` requires `yasm_tool_helper`
2. `alac` requires `vcpkg_copy_tool_dependencies`
3. `cpuinfo` requires `vcpkg_copy_tools`

JackBoosY:
Ping @dg0yt @ras0219-msft @BillyONeal for review again.

PhoebeHui (approved):

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -5,10 +5,19 @@ Copy all DLL dependencies of built tools into the tool folder.
 
 ## Usage
 ```cmake
-vcpkg_copy_tool_dependencies(<${CURRENT_PACKAGES_DIR}/tools/${PORT}>)
+vcpkg_copy_tool_dependencies(
+    <${CURRENT_PACKAGES_DIR}/tools/${PORT}>
+    [DEPENDENCIES <dep1>...]
```
BillyONeal, line 10: This documentation suggests that the function uses named arguments but the implementation appears to use positional arguments.

Why can't the `yasm` port issue the right `file(COPY` calls itself rather than modifying a shared helper like this?

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -42,7 +64,8 @@ function(z_vcpkg_copy_tool_dependencies_search tool_dir path_to_search)
 endfunction()
 
 function(vcpkg_copy_tool_dependencies tool_dir)
-    if(ARGC GREATER 1)
+    cmake_parse_arguments(PARSE_ARGV 0 arg "" "" "DEPENDENCIES")
+    if(ARGC GREATER 2)
```
BillyONeal, line 68: ARGC is going to be greater than 2 if `tool_dir` and `DEPENDENCIES` are both used; I don't think we should be mixing up positional and named parameters.

ports/yasm-tool/portfile.cmake
```diff
@@ -1,47 +1 @@
...
-    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
-    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/include")
-else()
-    set(YASM_BUILD_BINARY OFF)
-    set(VCPKG_POLICY_EMPTY_PACKAGE enabled)
-endif()
-
-configure_file("${CMAKE_CURRENT_LIST_DIR}/vcpkg-port-config.cmake.in"
-    "${CURRENT_PACKAGES_DIR}/share/${PORT}/vcpkg-port-config.cmake" @ONLY)
+SET(VCPKG_POLICY_EMPTY_PACKAGE enabled)
```
BillyONeal, line 1: Should this print a warning that it's now defunct?

BillyONeal (changes requested):
I think this is a good change in concept; I didn't realize that we already had a yasm port that produced the yasm binaries when I added yasm-tool and friends, and smashing that all together is great. However, I don't understand the value we gain by adding things to a shared helper over it, particularly in a way that mixes positional and named parameters.

Can the yasm-tool port just contain the right renames itself?
If that is unreasonable for some reason or we expect this to be common, can we add a new function that copies these dynamic DLL dependencies rather than piggybacking on `vcpkg_copy_tool_dependencies` and its associated `applocal.ps1` baggage?

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
JackBoosY: The yasm example shows that there are still other ports that use the same method (`load_library`) to depend on other libraries, which cann't be analysis by `dumpbin`.
So this is a common change.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
JackBoosY: `tool_dir` should only be passed one value. what is your opinion?

ports/yasm-tool/portfile.cmake
JackBoosY: According to Robert's comment, we should not remove any ports.
So I think that's okay, both `yasm-tool` or `yasm[tools]` can be used.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
BillyONeal: I mean, there are ports that do that, yes, but I'm not seeing what's made clearer about this vs. just issuing the `file(COPY`s, and doing it like this creates the problems I listed below about mixing positional and named parameters.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
BillyONeal: I don't understand this response. `tool_dir` is only passed one value, but given

`vcpkg_copy_tool_dependencies("C:/this/is/tool_dir" DEPENDENCIES "foo.dll")`

`ARGC` is 3 so we're going to print that warning, which seems unintended.

ports/yasm-tool/portfile.cmake
BillyONeal: We can't remove it because that would break people's scripts and stuff, but we can warn that they aren't using the canonical one.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
BillyONeal: image

These can't be correct

JackBoosY:
In favor of #23218.

PR #17698, opened 2021-05-06
[gdal] Fix build for non-Windows targets

dg0yt:
- #### What does your PR fix?  
  This PR fixes building gdal for non-Windows systems (#9068).
  It also restructures the portfile for better maintainability.
  It includes significant patching to configure.ac. There seems to be no better way to handle vcpkg's release+debug filesystem layout.

- #### Which triplets are supported/not supported? Have you updated the [CI baseline](https://github.com/microsoft/vcpkg/blob/master/scripts/ci.baseline.txt)?  
  all, no

- #### Does your PR follow the [maintainer guide](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md)?  
  yes

- #### If you have added/updated a port: Have you run `./vcpkg x-add-version --all` and committed the result?  
  yes

PR #18825, opened 2021-07-06
[hdf5] Fix linking order in pc file

dg0yt:
- #### What does your PR fix?  
  This PR fixing the linking order in the hdf5 pc files where the hdf5 library of interest must be placed before the libraries it depends on. For example, `-ldl` must come after `-lhdf5`.

  all (observable change on linux with #17698), no

  yes

  yes


ras0219-msft:

PR #18837, opened 2021-07-06
[scripts-audit] z_vcpkg_get_cmake_vars

strega-nil-ms:
Rename `vcpkg_internal_get_cmake_vars` -> `z_vcpkg_get_cmake_vars`, and do an audit. Also add a cache variable so we don't rerun, and remove the `OPTIONS` variable since it probably shouldn't be used.


dg0yt:
> remove the `OPTIONS` variable since it probably shouldn't be used.

`OPTIONS` allowed to query additional variable without changing this script. A pending use is in PR #16706.
However, I would agree that this must eventually be achieved in another way - `OPTIONS` requires knowledge of implementation details.

dg0yt:
I wish there was a way to silence the output of this function. It causes an extra "Configuring " which is quite confusing.

strega-nil-ms:
Closing because this is the wrong account -.-

PR #18843, opened 2021-07-06
[scripts-audit] CMake buildsystem

strega-nil:
This does an actual audit of the vcpkg_*_cmake functions, rather than just replacing them like #16377

Depends on PR #18397


scripts/cmake/vcpkg_build_cmake.cmake
```diff
@@ -44,73 +44,88 @@ function(vcpkg_build_cmake)
...
-                set(SHORT_BUILDTYPE "rel")
-                set(CONFIG "Release")
+                set(short_buildtype "rel")
+                set(config "Release")
             endif()
 
-            message(STATUS "Building ${TARGET_TRIPLET}-${SHORT_BUILDTYPE}")
+            message(STATUS "Building ${TARGET_TRIPLET}-${short_buildtype}")
 
             if(arg_ADD_BIN_TO_PATH)
```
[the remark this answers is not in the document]
dg0yt: Ignore, this is not `tools`.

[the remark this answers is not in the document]
Neumann-A: arg_ADD_BIN_TO_PATH needs still to be disabled in crossbuilds.

scripts/cmake/vcpkg_configure_cmake.cmake
```diff
@@ -67,95 +67,126 @@ This command supplies many common arguments to CMake. To see the full list, exam
...
+    set(ninja_host ON) # Ninja as parallel configurator
+
     if(CMAKE_HOST_WIN32)
         if(DEFINED ENV{PROCESSOR_ARCHITEW6432})
-            set(arg_HOST_ARCHITECTURE $ENV{PROCESSOR_ARCHITEW6432})
+            set(host_architecture "$ENV{PROCESSOR_ARCHITEW6432}")
         else()
-            set(arg_HOST_ARCHITECTURE $ENV{PROCESSOR_ARCHITECTURE})
+            set(host_architecture "$ENV{PROCESSOR_ARCHITECTURE}")
+        endif()
```
Neumann-A, line 118: this should be its own helper function because it is used more often.

scripts/cmake/vcpkg_configure_cmake.cmake
```diff
@@ -166,76 +197,72 @@ function(vcpkg_configure_cmake)
     # If we use Ninja, make sure it's on PATH
     if(GENERATOR STREQUAL "Ninja" AND NOT DEFINED ENV{VCPKG_FORCE_SYSTEM_BINARIES})
         vcpkg_find_acquire_program(NINJA)
-        get_filename_component(NINJA_PATH ${NINJA} DIRECTORY)
+        get_filename_component(NINJA_PATH "${NINJA}" DIRECTORY)
         vcpkg_add_to_path("${NINJA_PATH}")
```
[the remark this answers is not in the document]
Neumann-A: probably not since it is passed as `CMAKE_MAKE_PROGRAM`

scripts/cmake/vcpkg_configure_cmake.cmake
```diff
@@ -166,76 +197,72 @@ function(vcpkg_configure_cmake)
...
+        elseif(VCPKG_TARGET_IS_FREEBSD)
             set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE "${SCRIPTS}/toolchains/freebsd.cmake")
-        elseif(VCPKG_CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+        elseif(VCPKG_TARGET_IS_OPENBSD)
             set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE "${SCRIPTS}/toolchains/openbsd.cmake")
-        elseif(VCPKG_CMAKE_SYSTEM_NAME STREQUAL "MinGW")
+        elseif(VCPKG_TARGET_IS_MINGW)
             set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE "${SCRIPTS}/toolchains/mingw.cmake")
         endif()
     endif()
```
Neumann-A, line 260: Move into own function ? `z_vcpkg_select_default_toolchain` I still think this should be defined in the triplet instead or selected earlier outside  of `vcpkg_configure_cmake` because it has nothing to do with this function

scripts/cmake/vcpkg_configure_cmake.cmake
Neumann-A: also wondering if this could use:
https://cmake.org/cmake/help/latest/command/cmake_host_system_information.html

strega-nil-ms: I don't believe so, since it doesn't seem to read through `PROCESSOR_ARCHITEW6432`, it'll use the architecture of the process rather than the chip.

PR #18898, opened 2021-07-10
[vcpkg] Fix toolchain compatibility with cmake < 3.15

dg0yt:
Since 876e67c2, the toolchain used `list(PREPEND ...)` by default. But this cmake command was introduced in cmake 3.15, and the toolchain is meant to work with 3.1 [!].

- #### What does your PR fix?  
  This PR restores compatibility with cmake < 3.15 for user projects, fixing:
  ~~~
  CMake Error at /home/dg0yt/Projekte/vcpkg/vcpkg/scripts/buildsystems/vcpkg.cmake:390 (list):
    list does not recognize sub-command PREPEND
  ~~~
  Tested with CMake 3.10 (Ubuntu 18.04).

  all, no

  yes

  not needed


dg0yt:
CI didn't rebuild a single port, despite touching `vcpkg.cmake`.
There is no automated test of building at least a dummy project with the vcpkg toolchain and a supported ancient version of cmake.

dg0yt:
CC @autoantwort This amends #17336.

cenit:
> CI didn't rebuild a single port, despite touching `vcpkg.cmake`.
> 
> There is no automated test of building at least a dummy project with the vcpkg toolchain and a supported ancient version of cmake.

a long standing very serious issue
(it has already let slip some bad bugs into master...)

@strega-nil

autoantwort:
> a long standing very serious issue
> (it has already let slip some bad bugs into master...)
> 
> @strega-nil

There is already https://github.com/microsoft/vcpkg-tool/pull/80

dg0yt:
> There is already [microsoft/vcpkg-tool#80](https://github.com/microsoft/vcpkg-tool/pull/80)

Not sure if you really want to invalidate binary artifacts for each change, even documentation typos.

And this won't add testing for cmake versions which are claimed to be supported (cf. https://github.com/microsoft/vcpkg/pull/18319/files#r652632414).
(Nobody noticed the current issue until now. Is cmake 3.1 support really needed?)

---

FTR, you can't work around the current bug with VCPKG_PREFER_SYSTEM_LIBS=ON because this is not passed from the user project to cmake ports built by vcpkg, including `detect_compiler` and `get_cmake_vars`.

autoantwort:
> (Nobody noticed the current issue until now. Is cmake 3.1 support really needed?)

I thought the same, I searched for an issue complaining about it but couldn't find one.

autoantwort (approved):

scripts/buildsystems/vcpkg.cmake
```diff
@@ -380,35 +380,40 @@ set(_VCPKG_INSTALLED_DIR "${VCPKG_INSTALLED_DIR}"
...
-    set(Z_VCPKG_PATH_LIST_OP APPEND)
+    list(LENGTH CMAKE_PREFIX_PATH Z_VCPKG_PREFIX_PATH_POS)
+    list(LENGTH CMAKE_LIBRARY_PATH Z_VCPKG_LIBRARY_PATH_POS)
+    list(LENGTH CMAKE_FIND_ROOT_PATH Z_VCPKG_FIND_ROOT_PATH_POS)
 else()
-    set(Z_VCPKG_PATH_LIST_OP PREPEND)
+    set(Z_VCPKG_PREFIX_PATH_POS 0)
+    set(Z_VCPKG_LIBRARY_PATH_POS 0)
+    set(Z_VCPKG_FIND_ROOT_PATH_POS 0)
 endif()
```
strega-nil, line 392: One could probably do:

```
function(z_vcpkg_path_list_op out_var lst paths)
    if(VCPKG_PREFER_SYSTEM_LIBS)
        set("${out_var}" "${lst};${paths}" PARENT_SCOPE)
    else()
        set("${out_var}" "${paths};${lst}" PARENT_SCOPE)
    endif()

Comment (dg0yt) [reply scripts/buildsystems/vcpkg.cmake:392]: Does "one could probably do" mean I should update the PR? I would prefer to stick to plain cmake. If there shall be a function I would prefer a name which catures the specific operation "add" or "merge" instead of the generic "op", e.g. `z_vcpkg_add_to_system_libs` in analogy to `VCPKG_PREFER_SYSTEM_LIBS`.

Comment (strega-nil) [reply scripts/buildsystems/vcpkg.cmake:392]: I would prefer to switch to this; I don't like how this is currently working. Doing this weird INSERT bodging feels bad.

Comment (dg0yt) [reply scripts/buildsystems/vcpkg.cmake:392]: `set("${out_var}" "${paths};${lst}" PARENT_SCOPE)` doesn't work correctly for empty lists. 
And `list(INSERT ...)` is unable to append. 
:confused:

Comment (Neumann-A) [reply scripts/buildsystems/vcpkg.cmake:392]: > Doing this weird INSERT bodging feels bad.

@strega-nil  INSERT is also used by VTK to insert into the beginning of CMAKE_MODULE_PATH. It is the correct way to PREPEND. 

> And list(INSERT ...) is unable to append.
@dg0yt: have you tried using -1 as an index? Otherwise that is the reason why list(APPEND) existed from the beginning while PREPEND was added much later (since INSERT 0 just worked for this instead)

Comment (dg0yt) [reply scripts/buildsystems/vcpkg.cmake]: >  have you tried using -1 as an index? Otherwise that is the reason why list(APPEND) existed from the beginning while PREPEND was added much later (since INSERT 0 just worked for this instead)

> To add: From the 3.0 docs:

>    If is -1 or lesser, it is indexed from the end of the list, with -1 representing the last list element.

Yes, I tried in a separate cmake script. `INSERT` inserts *before* the given list element, and `INSERT -1` inserts before *the last list element.* 
CMake lists aren't C++ containers...

Comment (strega-nil) [reply scripts/buildsystems/vcpkg.cmake]: You can do:

```cmake
if(VCPKG_PREFER_SYSTEM_LIBS)
    list(APPEND lst "${paths}")
else()
    list(INSERT lst 0 "${paths}")
endif()
set("${out_var}" "${lst}" PARENT_SCOPE)

Comment (dg0yt) [reply scripts/buildsystems/vcpkg.cmake]: > You can do ...

That's what I already did.

Comment (strega-nil-ms) [reply scripts/buildsystems/vcpkg.cmake]: oh, sorry, I clearly didn't reread the PR.

File: scripts/buildsystems/vcpkg.cmake [hunk 2]
```diff
@@ -380,39 +380,24 @@ set(_VCPKG_INSTALLED_DIR "${VCPKG_INSTALLED_DIR}"
...
-    list(${Z_VCPKG_PATH_LIST_OP} CMAKE_FIND_ROOT_PATH
-        "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}"
-        "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug"
-    )
-endif()
+    if(NOT DEFINED CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE MATCHES "^[Dd][Ee][Bb][Uu][Gg]$")
+        list(REVERSE vcpkg_paths) # Debug build: Put Debug paths before Release paths.
+    endif()
+    if(VCPKG_PREFER_SYSTEM_LIBS)
+        list(APPEND ${list} ${vcpkg_paths})
```
strega-nil-ms, line 392:
```suggestion
        list(APPEND "${list}" "${vcpkg_paths}")
```

scripts/buildsystems/vcpkg.cmake
```diff
@@ -380,39 +380,24 @@ set(_VCPKG_INSTALLED_DIR "${VCPKG_INSTALLED_DIR}"
...
+    else()
+        list(INSERT ${list} 0 ${vcpkg_paths}) # CMake 3.15 is required for list(PREPEND ...).
```
strega-nil-ms, line 394:
```suggestion
        list(INSERT "${list}" 0 "${vcpkg_paths}") # CMake 3.15 is required for list(PREPEND ...).
```

scripts/buildsystems/vcpkg.cmake
```diff
@@ -380,39 +380,24 @@ set(_VCPKG_INSTALLED_DIR "${VCPKG_INSTALLED_DIR}"
...
+    endif()
+    set(${list} "${${list}}" PARENT_SCOPE)
```
strega-nil-ms, line 396:
```suggestion
    set("${list}" "${${list}}" PARENT_SCOPE)
```

strega-nil-ms (approved):

vicroms (approved):

vicroms:
Will be bundled in next roll-up to trigger world rebuild

strega-nil-ms:
Closed for rollup

PR #18905, opened 2021-07-10
[hdf5] Link zlib using the ZLIB::ZLIB target

dg0yt:
- #### What does your PR fix?  
  This PR fixes linking of the debug version of hdf5. Before the changes, it was linking to the *release* version of libz, and this was passed on via the exported cmake config files. This PR resolves the issue by always using the imported target `ZLIB::ZLIB` so that config resolution can happen as needed. 
  This was found when validating the extended gdal cmake wrapper/the linker flags used for a consuming executable. Before the change, it contained both debug and release version of libz:
  ~~~
  ...
  vcpkg_installed/x64-linux/debug/lib/libhdf5_hl_debug.a
  vcpkg_installed/x64-linux/debug/lib/libhdf5_debug.a
  vcpkg_installed/x64-linux/lib/libz.a
  ...
  vcpkg_installed/x64-linux/debug/lib/libjpeg.a
  vcpkg_installed/x64-linux/debug/lib/libz.a
  ...
  ~~~

  all, no.

  yes.

  yes.

### DIFF
--- a/ports/hdf5/portfile.cmake
+++ b/ports/hdf5/portfile.cmake
@@ -14,6 +14,7 @@ vcpkg_from_github(
         mingw-import-libs.patch
         pkgconfig-requires.patch
         pkgconfig-link-order.patch
+        zlib.patch
 )
 
 if ("parallel" IN_LIST FEATURES AND "cpp" IN_LIST FEATURES)
--- a/ports/hdf5/vcpkg.json
+++ b/ports/hdf5/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "hdf5",
   "version": "1.12.0",
-  "port-version": 4,
+  "port-version": 5,
   "description": "HDF5 is a data model, library, and file format for storing and managing data",
   "homepage": "https://www.hdfgroup.org/downloads/hdf5/",
   "supports": "!uwp",
--- a/ports/hdf5/zlib.patch
+++ b/ports/hdf5/zlib.patch
@@ -1,0 +1,27 @@
+diff --git a/CMakeFilters.cmake b/CMakeFilters.cmake
+index 1ae05eb..3e5824f 100644
+--- a/CMakeFilters.cmake
++++ b/CMakeFilters.cmake
+@@ -53,6 +53,7 @@ if (HDF5_ENABLE_Z_LIB_SUPPORT)
+       if (NOT ZLIB_FOUND)
+         find_package (ZLIB) # Legacy find
+         if (ZLIB_FOUND)
++          set (ZLIB_LIBRARIES ZLIB::ZLIB)
+           set (LINK_COMP_LIBS ${LINK_COMP_LIBS} ${ZLIB_LIBRARIES})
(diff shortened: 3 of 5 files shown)


Neumann-A:
> it contained both debug and release version of libz:

Has nothing to to with hdf5. But has everything to do with `FindZLIB` which only returns a single zlib library in `ZLIB_LIBRARIES` instead of debug/optimized keywords on linux/osx. 
So please add a wrapper to zlib instead which fixes that issue. 

I am also pretty sure that the ZLIB target ist not correctly setup on linux (e.g. no _DEBUG and _RELEASE IMPORTED_LOCATION)

dg0yt:
> I am also pretty sure that the ZLIB target ist not correctly setup on linux (e.g. no _DEBUG and _RELEASE IMPORTED_LOCATION)

Well, it seems to be okay since cmake 3.4:
https://github.com/Kitware/CMake/commit/11097f52311f5605c545e9ca4a25ee03a811d523

I'm testing on Ubuntu 18.04 (CMake 3.10).

Neumann-A:
@dg0yt You don't need to test it. Just look at the module: It will have only _RELEASE set due to it being unable to differentiate between RELEASE and DEBUG and debug only searching for suffixed names. As such a wrapper needs to set the ZLIB_LIBRARY_(RELEASE|DEBUG) cache vars before the find_package call. This will automatically fix ZLIB_LIBRARIES and the linkage in the hdf5 targets.

dg0yt:
> You don't need to test it.

If nobody tests, I don't need to wonder why so many stuff is broken. vcpkg CI just scratches the surface.

dg0yt:
This PR should become obsolete with #18914.

PhoebeHui:
@dg0yt, thanks for your contribution! ping me if this PR ready for review.

dg0yt:
With #18914 merged, this PR is obsolete now. `hdf5-targets.cmake` shows proper zlib dependencies (x64-mingw-static):
~~~
set_target_properties(hdf5::hdf5-static PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "_GNU_SOURCE;_FILE_OFFSET_BITS=64;_LARGEFILE64_SOURCE;_LARGEFILE_SOURCE"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:m>;\$<LINK_ONLY:ws2_32>;\$<LINK_ONLY:wsock32>;\$<\$<NOT:\$<CONFIG:DEBUG>>:${_IMPORT_PREFIX}/lib/libzlib.a>;\$<\$<CONFIG:DEBUG>:${_IMPORT_PREFIX}/debug/lib/libzlibd.a>;\$<LINK_ONLY:szip-static>;\$<LINK_ONLY:szip-static>;\$<LINK_ONLY:\$<\$<BOOL:OFF>:>>;\$<\$<NOT:\$<PLATFORM_ID:Windows>>:>"
)
~~~

PR #18920, opened 2021-07-12
[assimp] Use contrib polyclipping

waych:
It appears that the assimp port is linking polyclipping twice: once as a
dependency port via `find_package()`, and again by linking in a local copy
in `contrib/`.  This results in colliding symbols at link.

Given that assimp upstream doesn't seem to be directly compatible with
the polyclipping port version, use the one that works in `contrib/`.

Discovered while trying to fix building russimp tests on Windows here: https://github.com/jkvargas/russimp/pull/16

- #### What does your PR fix?  
  Fixes the many link errors that come about from two copies of polyclipper being linked (many, similar to):
```
  = note: rust-lld: error: duplicate symbol: bool __cdecl ClipperLib::Orientation(class std::vector<struct ClipperLib::IntPoint, class std::allocator<str       uct ClipperLib::IntPoint>> const &)
          >>> defined at C:\Users\mike\Windows-Code\sprinkle\target\vcpkg\buildtrees\polyclipping\src\ffc88818c4-06b887c850.clean\cpp\clipper.cpp:386
          >>>            polyclipping.lib(clipper.cpp.obj)
          >>> defined at assimp-vc142-mt.lib(clipper.cpp.obj)
```

  No change in triplets supported.

  This breaks the **Do not use vendored dependencies** rule.

  Justification is that the upstream `assimp` has been stuck on a really old version of `polyclipping` for a number of years and has yet to be updated.  The version in ports doesn't work if used directly and it seems non-trivial to update upstream `assimp` to the new version.  See also: 
       - https://github.com/assimp/assimp/issues/2401 and
       - https://github.com/assimp/assimp/issues/788

  Further justification: without this change, the contrib version is already being used.  The change performed here is to drop in dependency of the `polyclipping` port, which is superfluous. This superfluous dependency likely went unnoticed as it silently works on Linux, with the library being ignored, while failing on Windows due to causing collisions.

  Yes

### DIFF
--- a/ports/assimp/CONTROL
+++ b/ports/assimp/CONTROL
@@ -1,6 +1,6 @@
 Source: assimp
 Version: 5.0.1
-Port-Version: 4
+Port-Version: 5
 Homepage: https://github.com/assimp/assimp
 Description: The Open Asset import library
-Build-Depends: zlib, rapidjson, minizip, stb, kubazip, irrlicht, polyclipping, utfcpp, poly2tri
+Build-Depends: zlib, rapidjson, minizip, stb, kubazip, irrlicht, utfcpp, poly2tri
--- a/ports/assimp/build_fixes.patch
+++ b/ports/assimp/build_fixes.patch
@@ -257,7 +257,7 @@ diff --git a/code/CMakeLists.txt b/code/CMakeLists.txt
 index 55538d9..f5553e5 100644
 --- a/code/CMakeLists.txt
 +++ b/code/CMakeLists.txt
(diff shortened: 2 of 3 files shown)


PhoebeHui:
@waych, thanks for your contribution! Could you please report an issue to track firstly? so others may replicate this problem?

PhoebeHui:
Thanks for the PR; we're closing this for now since there's been no response. If you'd like to continue working on it, please reopen and ping us if it's ready for review!

PR #2669, opened 2018-01-28
[libopenmpt] init

atkawa7:
#503

PR #18933, opened 2021-07-13
[libopenmpt] Upgrade to 0.5.10, include author feedback

### LINKED ISSUE #15551: [mpg123] Use CMake build system
****

The `mpg123` library supports building with CMake using the CMakeLists.txt inside `ports/cmake`

**Proposed solution**
Use the CMake build system.

ras0219-msft:
This PR addresses feedback given by @manxorist in https://github.com/microsoft/vcpkg/pull/2669 and https://github.com/microsoft/vcpkg/issues/18480, while also upgrading the library to version 0.5.10.

The new port provides both CMake and pkg-config integration.

In order to provide CMake integration on Linux, this PR also fixes https://github.com/microsoft/vcpkg/issues/15551 which causes mpg123 to provide CMake integration on Linux instead of just Windows.


ports/libopenmpt/CMakeLists.txt
```diff
@@ -44,224 +27,75 @@ include_directories(
...
-  install(FILES ${lib_headers} DESTINATION include/libopenmpt)
-endif()
-
-if(NOT DISABLE_INSTALL_TOOLS)
-  install (
-    TARGETS openmpt123
-    RUNTIME DESTINATION tools/libopenmpt
-  )
-endif()
+set(LIBOPENMPT_REQUIRES_PRIVATE "zlib vorbisfile libmpg123")
```
ras0219-msft, line 82: Maybe this needs to be 
```suggestion
set(LIBOPENMPT_REQUIRES_PRIVATE "zlib vorbis vorbisfile libmpg123")
```

ports/libopenmpt/CMakeLists.txt
```diff
@@ -10,28 +10,11 @@ if(MSVC)
...
-  libopenmpt/libopenmpt_ext_impl.hpp
-  libopenmpt/libopenmpt_impl.hpp
-  libopenmpt/libopenmpt_internal.h
-  libopenmpt/libopenmpt_stream_callbacks_buffer.h
-  libopenmpt/libopenmpt_stream_callbacks_fd.h
-  libopenmpt/libopenmpt_stream_callbacks_file.h
-  libopenmpt/libopenmpt_version.h
-)
+file(GLOB HEADERS libopenmpt/*.h libopenmpt/*.hpp)
+list(FILTER HEADERS EXCLUDE REGEX "_impl\\.hpp$|_internal\\.h$")
```
manxorist, line 17: As libopenmpt's set of public interface headers is rather stable, I think it may be simpler to just list the public headers explicitly:
```
  libopenmpt/libopenmpt.h
  libopenmpt/libopenmpt.hpp
  libopenmpt/libopenmpt_config.h
  libopenmpt/libopenmpt_ext.h
  libopenmpt/libopenmpt_ext.hpp
  libopenmpt/libopenmpt_stream_callbacks_buffer.h
  libopenmpt/libopenmpt_stream_callbacks_fd.h
  libopenmpt/libopenmpt_stream_callbacks_file.h
  libopenmpt/libopenmpt_version.h
```

ports/libopenmpt/CMakeLists.txt
```diff
@@ -44,224 +27,75 @@ include_directories(
...
   target_compile_definitions(
     libopenmpt
     PRIVATE
-    -DMPT_BUILD_MSVC_SHARED -DLIBOPENMPT_BUILD_DLL
-  )
-  target_compile_definitions(
-    openmpt123
-    PRIVATE
-    -DMPT_BUILD_MSVC_SHARED -DLIBOPENMPT_BUILD_DLL
+      -DMPT_BUILD_MSVC_SHARED -DLIBOPENMPT_BUILD_DLL
```
manxorist, line 58: `MPT_BUILD_MSVC_SHARED` should not be needed. This should only be required by our msbuild project files (it's a historic misnomer and bug in our build system). It causes `#define ZLIB_DLL` in `common/BuildSettings.h`, which might or might not be correct. As I understand it, such macros should get set/unset correctly by CMake/vcpkg.

ports/libopenmpt/CMakeLists.txt
```diff
@@ -44,224 +27,75 @@ include_directories(
...
   target_compile_definitions(
     libopenmpt
     PRIVATE
-    -DMPT_BUILD_MSVC_STATIC
-  )
-  target_compile_definitions(
-    openmpt123
-    PRIVATE
-    -DMPT_BUILD_MSVC_STATIC
+      -DMPT_BUILD_MSVC_STATIC
```
manxorist, line 64: `MPT_BUILD_MSVC_STATIC` dito.

ports/libopenmpt/CMakeLists.txt
```diff
@@ -44,224 +27,75 @@ include_directories(
...
-
-install(
-  TARGETS libopenmpt
-  RUNTIME DESTINATION bin
-  LIBRARY DESTINATION lib
-  ARCHIVE DESTINATION lib
+  PRIVATE
+    MPG123::libmpg123
+    Vorbis::vorbisfile
+    ZLIB::ZLIB
```
manxorist, line 79: Probably also needs `Vorbis::vorbis`. See below.

ports/libopenmpt/CMakeLists.txt
manxorist: Yes, kind of. libopenmpt currently builds a function that calls `vorbis_comment_query()` (which is in libvorbis), however this function is unused in libopenmpt currently. In releases builds, function-level-linking and/or link-time-code-generation is smart enough to completely remove that code, however, in debug builds, linking libvorbis is currently required. I created upstream issue <https://bugs.openmpt.org/view.php?id=1477> to clean that up.

manxorist (changes requested):
Thanks for working on this!

Some further comments inline.

In `include_directories(`, `openmpt123` can be removed.

Also, related to libvorbis, it's kind of unclear how libogg should be handled. libvorbisfile uses typedefs from libogg in its public interface header (vorbis/vorbisfile.h). libvorbisfile's public header includes the libogg public header (ogg/ogg.h) (via the libvorbis public header (vorbis/codec.h)). libopenmpt also uses these libogg typedefs directly itself, and thus also includes libogg's public interface header itself. libopenmpt however does not require any linker symbols from libogg directly. In practice, things should probably work fine in all cases when not declaring a dependency on libogg, however as libopenmpt directly includes libogg's headers I tend to prefer an explicit dependency, even if that results in slight overlinking (however, libogg will be linked anyway in all cases by libvorbisfile, so this does not result in any actual bloat in the end). I have no strong opinion here, as long as things work, either way is fine by me.

One thing I did not see handled anywhere: client code consuming libopenmpt as a shared library on Windows needs to `#define LIBOPENMPT_USE_DLL`.

ports/libopenmpt/CMakeLists.txt
ras0219-msft: We generally prefer a glob approach when possible, because it makes it easier to know that nothing is missing when you aren't intimately familiar with the library.

ras0219-msft:
We generally err on the side of overlinking to make sure things work. It seems that `--as-needed` is becoming a more-and-more common practice (and is the default with MSVC), which removes almost all of the downsides.

I think I've addressed all your feedback -- thank you very much for giving it and sorry for the very long turnaround on this!

ports/libopenmpt/portfile.cmake
```diff
@@ -21,5 +21,11 @@ vcpkg_cmake_config_fixup()
 vcpkg_fixup_pkgconfig()
 vcpkg_copy_pdbs()
 
+if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
+    vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/include/libopenmpt/libopenmpt_config.h "defined(LIBOPENMPT_USE_DLL)" "0")
+else()
+    vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/include/libopenmpt/libopenmpt_config.h "defined(LIBOPENMPT_USE_DLL)" "1")
+endif()
+
```
manxorist, line 29: Is there really no cleaner way of injecting a macro definition into consuming code than text-replacing in a public header file? This feels kind of hack-ish to me, however, I am of course not the vcpkg expert here, and it looks like flac is doing a similar thing already in vcpkg (<https://github.com/microsoft/vcpkg/blob/master/ports/libflac/portfile.cmake#L56>), so I guess it's ok? I am mainly just wondering here.

Also, `LIBOPENMPT_BUILD_DLL`/`LIBOPENMPT_USE_DLL` are really intended to be used only on platforms where `dllexport`/`dllimport` is required instead of a public symbol visibility attribute, i.e. basically only on Windows. `libopenmpt_config.h` checks for the platform itself, thus overusing these macros should do no harm though.

manxorist:
> I think I've addressed all your feedback -- thank you very much for giving it and sorry for the very long turnaround on this!

Thanks!

libopenmpt is at version 0.5.11 by now, so I guess you could also update that. See <https://lib.openmpt.org/libopenmpt/2021/08/22/security-updates-0.5.11-0.4.23-0.3.32/>. Should require no build related changes.

ports/libopenmpt/portfile.cmake
ras0219-msft: We try to minimize the number of possible moving parts of our output to ensures that users of custom buildsystems or hand-rolled makefiles still get the correct results. We do this sort of "hard-code macro values" in many ports across the tree for precisely this case of `dllimport` and it has proved very resilient so far.

manxorist:
Looks good to me now.

ras0219 (approved):

JackBoosY (approved):

PR #18942, opened 2021-07-14
[mlpack] Add feature openmp

### LINKED ISSUE #18423: How do I set the compiler to use for a package and all its dependencies?
Background to issue:
https://github.com/mlpack/mlpack/issues/2971 

In the above, the advice was to compile armadillo on its own with the cl.exe compiler and the right compile flags - with much reading, I think I can do this within armadillo's CMakeLists.txt by setting:

set(CMAKE_C_COMPILER "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64" )

(or, other ways, as https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#how-do-i-use-a-different-compiler)

However, I dont think that resolves my problem as:

1.  There may be a plethora of other dependencies for mlpack, all of which will need the cl.exe compiler to be defined - then this becomes a piece-wise, hugely manual process
2.  To my mind, there must be a more elegant solution to ensure a compiler is specified at the vcpkg top level, & 
3.  As vcpkg installs a package and all its dependencies from a single command (download, unzip, compile as per inlcuded cmakelists file instructions), I'm not sure even how to break into that process and change the compile settings and re-compile via vcpkg while ensuring my modified CMakeList.txt file remains unaltered.

What I think (in my ignorance) there should be a way to say to vcpkg at the "top level", "use _this_ compiler" in the same way as you can, say:

set(VCPKG_CXX_FLAGS "/openmp- ${VCPKG_CXX_FLAGS}")
(in the triplet)

Ie. it makes no sense to me that you can set compiler flags but not the compiler - as compiler flags are not universal; they apply to specific compilers. (as outlined in the history bug/quesiton link at the top)

Nb0. https://vcpkg.readthedocs.io/en/latest/users/triplets/ - says, quote, "Triplet is a standard term used in cross compiling as a way to completely capture the target environment (cpu, os, **compiler**, runtime, etc) in a single convenient name.")

Nb1. I have toyed with the idea of adding a "set(CMAKE_C_COMPILER ..." setting in the vcpkg x64-windows triplet, but it's just a guess and I have no way of knowing what effect this would have, if it would be passed down-the-cmake-dependency-hierarchy, or if its the right thing to do. The vcpkg flags are outlined in the above link - none appear to allow compiler specification)

Nb2. Each time I try something different, removing/deleting vcpkg, re-download/installing so it's fresh (& repeatable) and then compiling the packages takes 30 - 50 mins. Consequently, I've not tried too many things - rather, I have spent time reading & reading & reading on the topic, hoping to find the precise advice to allow me to build my packages and dependencies with a single compiler.

Nb3. I've used google to search on all possible related keywords over about a week and cannot find anything that touches on enforcing cl.exe (or any other compiler). Most destinations have been places like here, or stackoverflow or similar.

I've got to the point where I'm **_sure_** I must be missing something really obvious/fundamental. Please help!!

JackBoosY:
Fixes #18423.

### DIFF
--- a/ports/mlpack/portfile.cmake
+++ b/ports/mlpack/portfile.cmake
@@ -23,8 +23,9 @@ file(REMOVE ${SOURCE_PATH}/CMake/ARMA_FindOpenBLAS.cmake)
 file(REMOVE ${SOURCE_PATH}/CMake/FindArmadillo.cmake)
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
-    tools   BUILD_CLI_EXECUTABLES
-    openmp  USE_OPENMP
+    FEATURES
+        tools   BUILD_CLI_EXECUTABLES
+        openmp  USE_OPENMP
 )
 
 vcpkg_configure_cmake(
--- a/ports/mlpack/portfile.cmake
+++ b/ports/mlpack/portfile.cmake
@@ -8,6 +8,7 @@ vcpkg_from_github(
         cmakelists.patch
         fix-configure-error.patch
         fix-test-dependency.patch
+        fix-dependencies.patch
 )
 
 file(REMOVE ${SOURCE_PATH}/CMake/ARMA_FindACML.cmake)
@@ -22,7 +23,8 @@ file(REMOVE ${SOURCE_PATH}/CMake/ARMA_FindOpenBLAS.cmake)
 file(REMOVE ${SOURCE_PATH}/CMake/FindArmadillo.cmake)
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
-    tools     BUILD_CLI_EXECUTABLES
+    tools   BUILD_CLI_EXECUTABLES
+    openmp  USE_OPENMP
 )
 
 vcpkg_configure_cmake(
--- a/versions/m-/mlpack.json
+++ b/versions/m-/mlpack.json
@@ -1,7 +1,7 @@
 {
   "versions": [
     {
-      "git-tree": "3bd9b5e857e2aacc12e3f52e46b583fda35ce557",
+      "git-tree": "587ee2575ef7543ff9a0ed30e645154db26f0001",
       "version": "3.4.1",
       "port-version": 2
     },
--- a/versions/m-/mlpack.json
+++ b/versions/m-/mlpack.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "3bd9b5e857e2aacc12e3f52e46b583fda35ce557",
+      "version": "3.4.1",
+      "port-version": 2
+    },
+    {
       "git-tree": "4793fb5802e10e645d5d33e0ec2baf6f1a0a337b",
       "version": "3.4.1",
       "port-version": 1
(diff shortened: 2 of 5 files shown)


JackBoosY:
@DanL1234 Can you please test `openmp` feature with my changes?

Thanks.

DanL1234:
Hi @JackBoosY - absolutely, would love to help. Can you tell me what I need to do? (I guess I need to try a fix you have made in a branch?) Sorry, but you will need to give me baby-steps as I am not familiar with git (I worked in tortoise SVN many years ago, so get the underlying idea, if not the procedure).

Ps. Thank you for your help on this issue - very much appreciated. :-)

JackBoosY:
@DanL1234  You can click the `command line instructions` link in this page.

ports/mlpack/portfile.cmake
```diff
@@ -22,7 +23,8 @@ file(REMOVE ${SOURCE_PATH}/CMake/ARMA_FindOpenBLAS.cmake)
 file(REMOVE ${SOURCE_PATH}/CMake/FindArmadillo.cmake)
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
-    tools     BUILD_CLI_EXECUTABLES
+    tools   BUILD_CLI_EXECUTABLES
+    openmp  USE_OPENMP
```
JackBoosY, line 27:
```suggestion
    FEATURES
        tools   BUILD_CLI_EXECUTABLES
        openmp  USE_OPENMP
```

versions/m-/mlpack.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "3bd9b5e857e2aacc12e3f52e46b583fda35ce557",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "587ee2575ef7543ff9a0ed30e645154db26f0001",
```

BillyONeal (approved):

NancyLi1013 (approved):

DanL1234:
> @DanL1234 Can you please test `openmp` feature with my changes?
> 
> Thanks.

@JackBoosY  - hi there. I just got around to trying this and had an issue. 

**Steps I took:**

1. git clone https://github.com/JackBoosY/vcpkg.git
2. git checkout dev/jack/18423
3. rename c:\src\vcpkg\ to c:\src\vcpkg-18423\
4. entered the directory \vcpkg-18423\
5. called .\bootstrap-vcpkg.bat
6. .\vcpkg install mlpack:x64-windows

All dependencies are built fine according to powershell output, until it gets to "mlpack", package 80 of 80.

**Issue seen:**

See attached screen capture and log that is being referred to. (plus an extra one for fun ;-))

1png
[install-x64-windows-rel-out.log](https://github.com/microsoft/vcpkg/files/6831098/install-x64-windows-rel-out.log)
[install-x64-windows-dbg-out.log](https://github.com/microsoft/vcpkg/files/6831105/install-x64-windows-dbg-out.log)



Did I do something wrong do you think?

Thanks in advance.
Daniel

vicroms (approved):

vicroms:
@DanL1234 

From your logs I can see that you're running out of memory when compiling mlpack. Also in order to test that the feature works it should be `vcpkg install mlpack[openmp]`. 

```
fatal error C1060: compiler is out of heap space
```

vicroms:
@JackBoosY 

Successfully tested new feature locally.

PR #18723, opened 2021-06-29
[vcpkg_check_features] remove deprecated form from ports tree

strega-nil:
cc @BillyONeal who asked me to

PR #18945, opened 2021-07-14
[many ports] remove deprecated vcpkg_check_features call [1/5]

strega-nil:
Replaces #18723 with smaller PRs that can be rolled up.


BillyONeal (approved):

strega-nil-ms:

Issue #18533, opened 2021-06-18
Versioning fails to recognise baseline, and doesn't cope with specification of versioning for dependency


tonygould:
**Describe the bug**
Following the steps at https://github.com/microsoft/vcpkg/blob/master/docs/examples/versioning.getting-started.md, configuration fails with errors on the baseline, and, if you delete the baseline, errors on the version specification  of `fmt`.

Either the "Getting started" example should work and there are bug(s) in the underlying vcpkg, or the example should be modified so that new users can easily get started. If the latter, it would be good to have an explanation why the baseline in the example is not working (despite being present in my cloned git repo), and why the version specification doesn't work -- because both of these look reasonable and intuitive to me.

**Environment**
- OS: Windows and Linux

**To Reproduce**

vcpkg.json

    {
        "name": "versions-test",
        "version": "1.0.0",
        "dependencies": [
            {
                "name": "fmt",
                "version>=": "7.1.3"
            },
            "zlib"
        ],
        "builtin-baseline": "b60f003ccf5fe8613d029f49f835c8929a66eb61"
    }

main.cpp

    #include <fmt/core.h>
    #include <zlib.h>

    int main() {
      fmt::print("fmt version is {}\n"
                "zlib version is {}\n",
                FMT_VERSION, ZLIB_VERSION);
      return 0;
    }

CMakeLists.txt

    cmake_minimum_required(VERSION 3.18)

    project(versionstest CXX)

    add_executable(main main.cpp)

    find_package(ZLIB REQUIRED)
    find_package(fmt CONFIG REQUIRED)
    target_link_libraries(main PRIVATE ZLIB::ZLIB fmt::fmt)

**Expected behavior**
It's the "Getting started" example. It should just work, per instructions.

**Failure logs** 
Initially

    [cmake] Error: Couldn't find explicitly specified baseline `"b60f003ccf5fe8613d029f49f835c8929a66eb61"` in the baseline file, and there was no baseline at that commit or the commit didn't exist.
    [cmake] Error: while checking out baseline 'b60f003ccf5fe8613d029f49f835c8929a66eb61:versions/baseline.json':
    [cmake] fatal: path 'versions/baseline.json' exists on disk, but not in 'b60f003ccf5fe8613d029f49f835c8929a66eb61'
    [cmake] This may be fixed by updating vcpkg to the latest master via `git pull`.
    [cmake] The current commit is "6e024e744e7717c06ddacd5089401109c6298553"

[...]

PR #18970, opened 2021-07-15
[docs] More Getting Started fixes

vicroms:
This fixes an issue with `fmt` changing from a `version-string` scheme to `version` scheme. 
Using `7.1.3#1` in the Getting Started guide should prevent version scheme conflicts in the future.

Second part of fixing #18533


PhoebeHui:
See https://github.com/microsoft/vcpkg/issues/18533#issuecomment-881117019, could you help take a look and see if it's a case that need to be covered.

PhoebeHui (approved):
LGTM, thanks for updating the docs!

PR #18979, opened 2021-07-16
[shaderc] Add fix for installing shaderc_util

chrismile:
**Describe the pull request**

- #### What does your PR fix?  
  The current version of the port shaderc neither installs shaderc_combined nor shaderc_shared, which both combine the targets shaderc and shaderc_util. Currently, only shaderc is installed by the port. However, if a program uses functionality from shaderc_util, there is currently no way to link to it. Thus, this PR adds shaderc_util as an installation target.

  all, no

  Yes.

  Yes.

### DIFF
--- a/ports/shaderc/fix-install-shaderc_util.patch
+++ b/ports/shaderc/fix-install-shaderc_util.patch
@@ -1,0 +1,18 @@
+diff --git a/libshaderc_util/CMakeLists.txt b/libshaderc_util/CMakeLists.txt
+index ec0e8fb..53e6fe7 100644
+--- a/libshaderc_util/CMakeLists.txt
++++ b/libshaderc_util/CMakeLists.txt
+@@ -31,6 +31,13 @@ target_link_libraries(shaderc_util PRIVATE
+   glslang OSDependent OGLCompiler HLSL glslang SPIRV
+   SPIRV-Tools-opt ${CMAKE_THREAD_LIBS_INIT})
+ 
++if(SHADERC_ENABLE_INSTALL)
++  install(TARGETS shaderc_util
++    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
++    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
++    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
++endif(SHADERC_ENABLE_INSTALL)
++
+ shaderc_add_tests(
+   TEST_PREFIX shaderc_util
+   LINK_LIBS shaderc_util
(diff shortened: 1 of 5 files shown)


ports/shaderc/fix-install-shaderc_util.patch
```diff
@@ -0,0 +1,18 @@
...
+@@ -31,6 +31,13 @@ target_link_libraries(shaderc_util PRIVATE
+   glslang OSDependent OGLCompiler HLSL glslang SPIRV
+   SPIRV-Tools-opt ${CMAKE_THREAD_LIBS_INIT})
+ 
++if(SHADERC_ENABLE_INSTALL)
++  install(TARGETS shaderc_util
++    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
++    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
++    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
++endif(SHADERC_ENABLE_INSTALL)
```
NancyLi1013, line 14: Seems we should install this target according to build type. Since it only supports static build. Please refer to this patch.
https://github.com/microsoft/vcpkg/blob/master/ports/shaderc/fix-build-type.patch

NancyLi1013 (changes requested):

ports/shaderc/fix-install-shaderc_util.patch
chrismile: Thanks a lot for your change request! I added a check for "NOT BUILD_SHARED_LIBS", so hopefully this should be fixed now.

NancyLi1013: I wonder if we also need to add `if(NOT BUILD_SHARED_LIBS)` for https://github.com/google/shaderc/blob/main/libshaderc_util/CMakeLists.txt#L17. Since it seems like a static library. What do you think?

chrismile: I think that `if(NOT BUILD_SHARED_LIBS)` can't be added in this case. Both shaderc (static library) and shaderc_shared (shared library) seem to reference shaderc_util with target_link_libraries. Here's the excerpt from the current (patched) version of libshaderc/CMakeLists.txt:

```CMake
set(SHADERC_LIBS
  glslang OSDependent OGLCompiler glslang ${CMAKE_THREAD_LIBS_INIT}
  shaderc_util
  SPIRV # from glslang
  SPIRV-Tools
)

if(NOT BUILD_SHARED_LIBS)
target_link_libraries(shaderc PRIVATE ${SHADERC_LIBS})
else()
target_link_libraries(shaderc_shared PRIVATE ${SHADERC_LIBS})
endif()
```

NancyLi1013 (approved):

PR #18980, opened 2021-07-16
[cmake-guidelines] Minor update, for `if()`

strega-nil:
Basically, introduces rules for `if()` as well as argument quoting.

Depends on #18397


docs/maintainers/cmake-guidelines.md
```diff
@@ -45,27 +45,64 @@ We hope that they will make both forwards and backwards compatibility easier.
...
+  - Exception: arguments to `if()` that are not operators must always be quoted.
+    - operators should be unquoted.
+    - Example:
+      ```cmake
+      if("${FOO}" STREQUAL "BAR")
+      endif()
+
+      if("${BAZ}" EQUAL "0")
+      endif()
+      ```
```
Neumann-A, line 94: is there a reason to not use `if(FOO STREQUAL "BAR")` ? If FOO is not set this will raise and error.

strega-nil: no it won't, in either case.

Assuming `FOO` is not defined:

```cmake
if(FOO STREQUAL "BAR") # equivalent to `if("FOO" STREQUAL "BAR")`
endif()

if("${FOO}" STREQUAL "BAR") # equivalent to `if("" STREQUAL "BAR")`
endif()
```

strega-nil:
```cmake
if(FOO STREQUAL "FOO")
	message(STATUS "yes 1")
endif()
if("${FOO}" STREQUAL "")
	message(STATUS "yes 2")
endif()
```

prints out

```no-highlight
-- yes 1
-- yes 2
```

Neumann-A: Ok after testing 
`if(${FOO} STREQUAL "BAR")`
is throwing the error I thought about.

I wonder if 
`if(FOO AND FOO STREQUAL "BAR")`
is mensurable more performant than 
`if("${FOO}" STREQUAL "BAR")`

docs/maintainers/cmake-guidelines.md
```diff
@@ -80,16 +119,14 @@ We hope that they will make both forwards and backwards compatibility easier.
   and `<start>` _must always be_ less than or equal to `<stop>`.
   - This must be checked by something like:
   ```cmake
-  if(start LESS_EQUAL end)
-    foreach(RANGE start end)
+  if("${start}" LESS_EQUAL "${end}")
+    foreach(RANGE "${start}" "${end}")
```
Neumann-A, line 123: `vcpkg_foreach(RANGE start stop CODE "<somecode>")` ?

strega-nil:
The reason I'm interested in this change is not because `if(FOO STREQUAL "bar")` can't be used without error, it's because it's extremely confusing as to _when_ it can be used -- for example, I've had bugs in the past where I've written `if(EXISTS arg_FILENAME)` instead of the required `if(EXISTS "${arg_FILENAME}")`, or `if("${foo}" MATCHES regex_var)`.

strega-nil-ms:
Closed for rollup

PR #19002, opened 2021-07-19
[libidn2] Fixes for mingw and minor issues

dg0yt:
- #### What does your PR fix?  
  - [x] Fixes building with mingw (needed for curl). For faster build, it uses the cmake build system which is also used for MSVC.
  - [x] Installs the pc file also for cmake builds.
  - [x] Modernizes the portfile
  - [x] Use libunistring for non-windows

  all, no

  yes

  Yes, but I am still working on this PR.


dg0yt:
I can pick up libunistring in another PR. But this one is needed for updating curl.

ports/libidn2/portfile.cmake
```diff
@@ -50,21 +51,21 @@ if (VCPKG_TARGET_IS_WINDOWS)
 else()
```
NancyLi1013, line 51: file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")

NancyLi1013 (changes requested):

ports/libidn2/portfile.cmake
```diff
@@ -9,62 +9,64 @@ vcpkg_download_distfile(ARCHIVE
 
 vcpkg_extract_source_archive_ex(
     OUT_SOURCE_PATH SOURCE_PATH
-    ARCHIVE ${ARCHIVE}
+    ARCHIVE "${ARCHIVE}"
     REF ${IDN2_VERSION}
 )
 
 if (VCPKG_TARGET_IS_WINDOWS)
```
NancyLi1013, line 16:
```suggestion
if (VCPKG_TARGET_IS_WINDOWS OR VCPKG_TARGET_IS_MINGW)
```
If using cmake to build on mingw, this condition is also required. I noticed that you add
```
{
      "name": "vcpkg-cmake",
      "host": true,
      "platform": "windows | mingw"
    }
```
in vcpkg.json file.

NancyLi1013 (changes requested):

ports/libidn2/portfile.cmake
dg0yt: Note that the platform expression in the manifest is not consistent with the script variables:
https://github.com/microsoft/vcpkg/blob/cdd51899f6ae3736bfbe936f9e387f86d4a63a58/scripts/cmake/vcpkg_common_definitions.cmake#L52-L55

dg0yt: Verified again: Mingw uses cmake as expected. No change needed.

NancyLi1013 (approved):

PR #19010, opened 2021-07-19
[vcpkg] Optionally limit the number of cores during vcpkg bootstrap

### LINKED ISSUE #14534: Building vcpkg on RPi fails with out of memory. Add number of cores argument option to bootstrap script
Hey guys,

one target of our multi plattform project is a RPi 3B. Thanks to all that working with vcpkg on RPi is already possible and mostly going well.

But we are facing one problem with building vcpkg from source on RPi using bootstrap.sh. As the ninja build spawns a lot of build processes in parallel by default, the device is running out of memory very quickly. This affects all multicore system with a limited ram size (RPi 3B = 1GB).

Would it be possible to add a cmd line option to the boostrap script to adjust the number of cores used while building?

We achieved this using the follwowing additional args on cmake configuration in boostrap.sh script:
-DCMAKE_JOB_POOL_COMPILE:STRING=compile DCMAKE_JOB_POOL_LINK:STRING=link DCMAKE_JOB_POOLS:STRING=compile=1;link=1'

It would be really nice if this could get it into productive code ;-)

Thank you.

Best Regards,
Michael

jherico:
Add an option to the bash bootstrap script that allows a user to force the use of only a single CPU during both compilation and linking of the vcpkg executable.  

Building on low-memory / high-core count systems like 1-2 GB variants of Raspberry PIs will often fail because of memory exhaustion.  Each compile will typically take a few hundred MB and running 4 of them concurrently will easily exhaust even a 2GB setup.  Further, such systems often have no swap space configured.  This command line option ensures that `vcpkg` will bootstrap successfully, albeit slowly, even when only 1GB of ram is available.

This PR _does not_ attempt to address any similar issues that might arise from building packages using `vcpkg`

- #### What does your PR fix?  

Fixes #14534 

This is a modification to the bootstrap script and should have no impact on the triplets, only a potential impact on the speed of bootstrapping.

Yes.  The vast majority of the guide doesn't apply since this is a change to the boostrapping mechanism, not to any port or vcpkg functionality.  

N/A

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -10,7 +10,6 @@ done
 vcpkgDisableMetrics="OFF"
 vcpkgUseSystem=false
 vcpkgAllowAppleClang=false
-vcpkgLimitBuildMemory=false
 vcpkgBuildTests="OFF"
 for var in "$@"
 do
@@ -18,8 +17,6 @@ do
         vcpkgDisableMetrics="ON"
     elif [ "$var" = "-useSystemBinaries" -o "$var" = "--useSystemBinaries" ]; then
         vcpkgUseSystem=true
-    elif [ "$var" = "-limitBuildMemory" -o "$var" = "--limitBuildMemory" ]; then
-        vcpkgLimitBuildMemory=true
     elif [ "$var" = "-allowAppleClang" -o "$var" = "--allowAppleClang" ]; then
         vcpkgAllowAppleClang=true
     elif [ "$var" = "-buildTests" ]; then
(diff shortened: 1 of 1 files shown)


aizuon:
Could this feature also be implemented when compiling packages? Would be nice to leave half of the cores unattended when compiling big packages, i.e. Qt, to do other work to spend time.

jherico:
@aizuon It would be inconsistent.  As far as I understand, not every package uses CMake for the build wrapper, and the approach here only works with CMake.  It's likely that there may be some necessary similar feature that would be useful for package building, but this PR is just focused on the bootstrap process and unblocking people in the situation described above.

vicroms:
@aizuon You can try setting the [`VCPKG_MAX_CONCURRENCY`](https://vcpkg.io/en/docs/users/config-environment.html#vcpkg_max_concurrency) environment variable which most buildsystem helper functions support (there may be some ports with weird buildsystems that won't).

vicroms (approved):

ras0219-msft:
I feel like it would make more sense to just make bootstrap respect `VCPKG_MAX_CONCURRENCY`

jherico:
@ras0219 I've modified the build script to respect `VCPKG_MAX_CONCURRENCY` and removed the other modifications.  I feel like it would be helpful to have some text in the `-help` output but I couldn't come up with something both informative and concise, so I left it as is.

vicroms (approved):

PR #19029, opened 2021-07-20
[Yara] add spectre mitigation feature

### LINKED ISSUE #18986: How do I enable spectre mitigation
We are using BinSkim to scan our assembly for any issues, one of the issue is flagged is that our assembly does not have spectre mitigation.

Our assembly depends on the yara port which we build this way:
```
vcpkg install yara:x86-windows-static yara:x64-windows-static
```

Ultimately we depend on the following lib packages:
- jansson.lib
- libcrypto.lib
- libssl.lib
- libyara.lib

Is it possible to enable spectre mitigation when building those static libraries flag without changing the port file?

When building with msbuild this is done by either by specifying the /p:SpectreMitigation=Spectre or by adding the /QSpectre

henriblMSFT:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #18986 

  no update to supported version

  `yes`

  Waiting on PR #18977 to complete before this PR can complete.

### DIFF
--- a/ports/yara/CMakeLists.txt
+++ b/ports/yara/CMakeLists.txt
@@ -5,11 +5,13 @@ if(MSVC)
   add_compile_options(/W3 /wd4005 /wd4996 /wd4018 -D_CRT_SECURE_NO_WARNINGS)
 endif()
 
+if(${YARA_ENABLE_SPECTRE_MITIGATION})
+  add_compile_options(/Qspectre)
+endif()
 
 find_package(OpenSSL REQUIRED)
 find_path(JANSSON_INCLUDE_DIR NAMES jansson.h)
 find_library(JANSSON_LIBRARY NAMES jansson)
-
 
 include_directories(
   .
--- a/ports/yara/portfile.cmake
+++ b/ports/yara/portfile.cmake
@@ -8,12 +8,23 @@ vcpkg_from_github(
   HEAD_REF dev
 )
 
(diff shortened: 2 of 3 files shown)


Hoikas:
Maybe I'm misunderstanding something, but why not use a custom triplet with the contents:
```cmake
set(VCPKG_C_FLAGS "${VCPKG_C_FLAGS} /Qspectre")
set(VCPKG_CXX_FLAGS "${VCPKG_CXX_FLAGS} /Qspectre")
```
appended to the end?

JackBoosY:
> Maybe I'm misunderstanding something, but why not use a custom triplet with the contents:
> 
> ```cmake
> set(VCPKG_C_FLAGS "${VCPKG_C_FLAGS} /Qspectre")
> set(VCPKG_CXX_FLAGS "${VCPKG_CXX_FLAGS} /Qspectre")
> ```
> 
> appended to the end?

Oh you mentioned me.
Since spectre mitigation can be applied to any port that supports Visual Studio builds, I think stacking triples would be a good choice.

@henriblMSFT  Please see overlay-triplet [docs](https://github.com/microsoft/vcpkg/blob/master/docs/examples/overlay-triplets-linux-dynamic.md).
Add the code to the triplet file which @Hoikas mentioned. Then add parameter `--overlay-triplets=TRIPLET_FILE_PATH` to install command.

PR #15522, opened 2021-01-08
[liblas] Fix features

NancyLi1013:
**Describe the pull request**

- What does your PR fix? Fixes #15494

 Add `vcpkg_check_features()`

Note: All features have passed with the following triplets:

- x86-windows
- x64-windows-static

PR #19053, opened 2021-07-22
[liblas] Add new feature gdal

### LINKED ISSUE #16005: [libLAS] GDAL support
****

The pointclouds I work with have specific requirements for georeferencing; when using libLAS I need its optional WITH_GDAL feature enabled (it's off by default); however the vcpkg port doesn't provide this option. It would certainly be nice though.

**Proposed solution**
Please add an option for GDAL support (libLAS already provides a WITH_GDAL option, I *think* this could be exposed as a vcpkg port option)

longhuan2018:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #16005
  Merge PR #15522

Fix find_package(liblas),  libLAS_INCLUDE_DIRS path incorrect bug.
liblas-config.cmake is stored in the share\liblas directory,  So the relative path should be ../.. instead of ..
```
get_filename_component (PROJECT_ROOT_DIR "${_DIR}/.." ABSOLUTE)
set (libLAS_INCLUDE_DIRS "${PROJECT_ROOT_DIR}/include")
set (libLAS_LIBRARY_DIRS "${PROJECT_ROOT_DIR}/lib")
set (libLAS_BINARY_DIRS "${PROJECT_ROOT_DIR}/bin")
```

libLAS_LIBRARIES contains  geotiff_library,  we need to add  find_package (geotiff CONFIG REQUIRED)  to the liblas-config.cmake  file

### DIFF
--- a/ports/liblas/CONTROL
+++ b/ports/liblas/CONTROL
@@ -1,7 +1,7 @@
 Source: liblas
 Version: 1.8.1
 Port-Version: 7
-Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff, libjpeg-turbo, zlib
+Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff
 Description: A C/C++ library for reading and writing the very common LAS LiDAR format.
 Default-Features: gdal
 
--- a/ports/liblas/CONTROL
+++ b/ports/liblas/CONTROL
@@ -3,6 +3,7 @@ Version: 1.8.1
 Port-Version: 7
 Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff, libjpeg-turbo, zlib
 Description: A C/C++ library for reading and writing the very common LAS LiDAR format.
+Default-Features: gdal
 
 Feature: gdal
 Build-Depends: gdal
--- a/ports/liblas/CONTROL
+++ b/ports/liblas/CONTROL
@@ -1,13 +1,9 @@
 Source: liblas
 Version: 1.8.1
-Port-Version: 6
-Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff
+Port-Version: 7
+Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff, libjpeg-turbo, zlib
 Description: A C/C++ library for reading and writing the very common LAS LiDAR format.
 
-Feature: jpeg
-Description: Support for jpeg
-Build-Depends: libjpeg-turbo
-
-Feature: zlib
-Build-Depends: zlib
-Description: Support zlib for compression
+Feature: gdal
+Build-Depends: gdal
+Description: Support gdal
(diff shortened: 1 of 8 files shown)


ports/liblas/CONTROL
```diff
@@ -1,13 +1,10 @@
 Source: liblas
 Version: 1.8.1
-Port-Version: 6
-Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff
+Port-Version: 7
+Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff, libjpeg-turbo, zlib
```
dg0yt, line 4: The PR is named "support gdal". The new core dependencies are not required for this purpose: You might use feature dependencies (at least after converting `CONTROL` to `vcpkg.json` with `vcpkg format-manifest  ports/liblas/CONTROL`).

ports/liblas/portfile.cmake
```diff
@@ -13,20 +13,25 @@ vcpkg_extract_source_archive_ex(
...
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
-    OPTIONS 
+    OPTIONS ${FEATURE_OPTIONS}
         -DBUILD_OSGEO4W=OFF # Disable osgeo4w
         -DWITH_TESTS=OFF
         -DWITH_UTILITIES=OFF
-        -DCMAKE_DISABLE_FIND_PACKAGE_ZLIB=${CMAKE_DISABLE_FIND_PACKAGE_ZLIB}
-        -DCMAKE_DISABLE_FIND_PACKAGE_JPEG=${CMAKE_DISABLE_FIND_PACKAGE_JPEG}
```
dg0yt, line 29: If keeping jpeg and zlib features, these might be handled via `vcpkg_check_features` as inverted feature options.

ports/liblas/CONTROL
dg0yt: OTOH, libgeotiff pulls in libjpeg-turbo and zlib dependencies anyway. So there is nothing gained from optional features, and CMAKE_DISABLE_FIND_PACKAGE_... is not desirable. 
It just isn't clear from the PR structure and comments.

ports/liblas/portfile.cmake
longhuan2018: JPEG and zlib are the dependencies of libgeotif, so it is not necessary to be a feature of liblas  (and the CMAKE_DISABLE_FIND_PACKAGE_ZLIB,  CMAKE_DISABLE_FIND_PACKAGE_JPEG option is not used in liblas)

ports/liblas/CONTROL
```diff
@@ -1,13 +1,10 @@
 Source: liblas
 Version: 1.8.1
-Port-Version: 6
+Port-Version: 7
 Build-Depends: boost, boost-thread, boost-system, boost-iostreams, boost-filesystem, boost-detail, libgeotiff
 Description: A C/C++ library for reading and writing the very common LAS LiDAR format.
+Default-Features: gdal
```
dg0yt, line 6: IMO this shouldn't be a default feature, given that the combination of GDAL 3.x and liblas is not supported upstream. (In fact, upstream recommends switching to pdal.) 
I understant that this is needed to get CI coverage for this PR now. But it also adds a maintenance burden to future changes, without upstream support.

longhuan2018: When the CI compilation passes, I will delete the default value

JackBoosY:

JackBoosY:

PR #19164, opened 2021-07-27
[vcpkg_build_make] Always use msys make on win32 host

dg0yt:
- #### What does your PR fix?  
  This PR ensures that `cmake_build_make` uses the `make` program it acquires via msys, instead of allowing an instance of make which is earlier in the path, e.g. side by side with the compiler.
  Fixes issues like #19117, hopefully (mingw32-make invoked as make with mingw32-w64 toolchain from Chocolatey.)

  all (changing behaviour for windows hosts), no

  yes

  not needed


NancyLi1013 (approved):

ras0219-msft (approved):

cenit:
why not adding no_default_paths too?

dg0yt:
@cenit You are right, I meant to add this too.

PR #19194, opened 2021-07-28
[kf5archive] fix cmake.in to allow consumption by static builds

wrobelda:
**Describe the pull request**

- #### What does your PR fix?  
  Includes patch from https://invent.kde.org/frameworks/karchive/-/merge_requests/23

  No change

  Yes

  Yes


ports/kf5archive/portfile.cmake
```diff
@@ -4,17 +4,21 @@ vcpkg_from_github(
...
+        23.patch # https://invent.kde.org/frameworks/karchive/-/merge_requests/23
 )
 
 vcpkg_configure_cmake(
     DISABLE_PARALLEL_CONFIGURE
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
     OPTIONS 
+        -DBUILD_TESTING=OFF
+    MAYBE_UNUSED_VARIABLES
```
NancyLi1013, line 18: Where do you get the option `MAYBE_UNUSED_VARIABLES`? Seems it doesn't exist in source codes.

NancyLi1013 (changes requested):

ports/kf5archive/portfile.cmake
wrobelda: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_configure_cmake.md#maybe_unused_variables

ports/kf5archive/23.patch
```diff
@@ -0,0 +1,46 @@
...
++
++    if (@LIBLZMA_FOUND@)
++        find_dependency(LibLZMA)
++    endif()
++
++    if (@LibZstd_FOUND@)
++        find_package(PkgConfig)
++        pkg_check_modules(LibZstd IMPORTED_TARGET "libzstd")
++    endif()
++endif()
```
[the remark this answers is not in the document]
wrobelda: That's a good call, will fix that.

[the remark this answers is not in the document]
wrobelda: @NancyLi1013 added.

ports/kf5archive/portfile.cmake
dg0yt: IIUC this parameter should list the variables names, not act as another form of `OPTIONS`.

wrobelda: @dg0yt good call: https://github.com/microsoft/vcpkg/blob/0e1dc121856eef84bf1852134e548bf9898dabac/scripts/cmake/vcpkg_configure_cmake.cmake#L96

wrobelda: Fixed.

ports/kf5archive/portfile.cmake
```diff
@@ -4,21 +4,27 @@ vcpkg_from_github(
...
     OPTIONS 
+        -DBUILD_TESTING=OFF
         -DBUILD_HTML_DOCS=OFF
         -DBUILD_MAN_DOCS=OFF
         -DBUILD_QTHELP_DOCS=OFF
-        -DBUILD_TESTING=OFF
+    MAYBE_UNUSED_VARIABLES 
+        BUILD_HTML_DOCS
+        BUILD_MAN_DOCS
+        BUILD_QTHELP_DOCS
```
NancyLi1013, line 23:
```suggestion
    MAYBE_UNUSED_VARIABLES 
        -DBUILD_HTML_DOCS=OFF
        -DBUILD_MAN_DOCS=OFF
        -DBUILD_QTHELP_DOCS=OFF
```
Seems you added these options like this in other ports.

NancyLi1013 (changes requested):

ports/kf5archive/vcpkg.json
```diff
@@ -1,12 +1,35 @@
...
+      ]
+    },
+    "zstd": {
+      "description": "Support for zstd compressed files and data streams",
+      "dependencies": [
+        "zstd"
+      ]
+    }
+  }
 }
```
NancyLi1013, line 35: You still need to handle these two features in portfile.cmake via `vcpkg_check_features()`.

NancyLi1013 (changes requested):

ports/kf5archive/portfile.cmake
NancyLi1013, line 23: These variables seem not exist in source codes. Why do we still need them here?

```
        BUILD_HTML_DOCS
        BUILD_MAN_DOCS
        BUILD_QTHELP_DOCS
```

ports/kf5archive/vcpkg.json
wrobelda: I have a problem finding a way of disabling the `pkg_check_modules` for `zstd` without patching the CMakeLists.txt

wrobelda: Fixed.

ports/kf5archive/portfile.cmake
wrobelda: Fixed.

dg0yt:
On zstd lookup:
AFAIU there is no pkgconfig setup in `vcpkg_configure_cmake`/`vcpkg_cmake_configure`. So unfortunately, the zstd lookup based on pkg-config is fragile, i.e. depending on system configuration. Here, even twice: when building the port, and when using the config files.

(That's why I switched to the experimental `vcpkg-pkgconfig-get-modules` utility port for the WIP curl update, #18971.)

NancyLi1013:
For `zstd`, we suggest to use it like this instead of pkgconfig:

```
find_package(zstd CONFIG REQUIRED)
target_link_libraries(main PRIVATE zstd::libzstd_shared)

find_package(zstd CONFIG REQUIRED)
target_link_libraries(main PRIVATE zstd::libzstd_static)
```

wrobelda:
> For `zstd`, we suggest to use it like this instead of pkgconfig:

The patch used in my version has already been pushed upstream. Is there any real disadvantage of using `pkg_check_modules` instead of `target_link_libraries` directly?

NancyLi1013:
> > For `zstd`, we suggest to use it like this instead of pkgconfig:
> 
> The patch used in my version has already been pushed upstream. Is there any real disadvantage of using `pkg_check_modules` instead of `target_link_libraries` directly?

If this is accepted by upstream, it is fine to use `pkg_check_modules`. `pkg_check_modules` uses .pc files, `target_link_libraries` uses .lib files. To be honest, I'm not clear the advantages and disadvantages between these two ways.

ports/kf5archive/portfile.cmake
```diff
@@ -4,21 +4,29 @@ vcpkg_from_github(
...
     HEAD_REF master
+    PATCHES
+        9ab5f2bf.patch # https://invent.kde.org/frameworks/karchive/-/commit/9ab5f2bfbe59038b0d0b6ca7f1b22d1c9229c67e
+        5a79756f.patch # https://invent.kde.org/frameworks/karchive/-/commit/5a79756f381e1a1843cb2171bdc151dad53fb7db
+        23.patch # https://invent.kde.org/frameworks/karchive/-/merge_requests/23
 )
 
-vcpkg_configure_cmake(
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
```
NancyLi1013, line 14:
```suggestion
```
There is no feature in `FEATURES`. So you can remove it here.

ports/kf5archive/portfile.cmake
```diff
@@ -4,21 +4,29 @@ vcpkg_from_github(
...
+        ${FEATURE_OPTIONS}
 )
 
-vcpkg_install_cmake()
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/KF5Archive)
+vcpkg_cmake_install()
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/KF5Archive)
 vcpkg_copy_pdbs()
 
 if(VCPKG_LIBRARY_LINKAGE STREQUAL "static")
```
NancyLi1013, line 32: Could you please add double quotes for these lines? 

```
file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/etc")
file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/etc")
file(INSTALL "${SOURCE_PATH}/LICENSES/" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}/copyright")
```

NancyLi1013 (changes requested):

dg0yt:
> > Is there any real disadvantage of using `pkg_check_modules` instead of `target_link_libraries` directly?
> 
> If this is accepted by upstream, it is fine to use `pkg_check_modules`. `pkg_check_modules` uses .pc files, `target_link_libraries` uses .lib files. To be honest, I'm not clear the advantages and disadvantages between these two ways.

There is confusion here. You need to use `target_link_libraries` in any case. The question is how you provide the libraries to be linked. Ideally, you have a target which has all necessary properties like transitive dependencies and include directories.

`pkg_check_modules` with `IMPORTED_TARGET` seems to have the capacity to provide such a target. However, it is using information from the .pc file, and this information wasn't decide for cmake. For exampe, there can't be information about debug vs. release variants.

`find_package()` will look for a cmake find module or config file. Config files are normally provided by the libraries themselve, letting them provide cmake properties as needed. Find modules are often provided by cmake or by the consuming software, and quality varies with provider and cmake version.

So in general, I would prefer `find_package()` because it doesn't need to leave the cmake world. For zstd, there are cmake config  files, but IIUC they are more difficult to use because there interface is different for static vs. dynamic linkage.

wrobelda:
> So in general, I would prefer `find_package()` because it doesn't need to leave the cmake world. For zstd, there are cmake config files, but IIUC they are more difficult to use because there interface is different for static vs. dynamic linkage.

All you said makes sense, however, as explained, the `.config.cmake` patch was already accepted upstream and it is based on the library's CMakeLists.txt, which uses pkgconfig to find the zstd as well. Whether CMake could be used, that's upstream's decision.

dg0yt:
@wrobelda There is no "however" which needs to be explained. I was just targeting the confusion in the cited comment.

The essence of my first comment is still valid: "there is no pkgconfig setup in vcpkg_configure_cmake/vcpkg_cmake_configure" so I wonder how pkgconfig can work for this portfile. What did I miss?

NancyLi1013 (approved):

NancyLi1013:
Have you tested the features `lzma `and `zstd`?

wrobelda:
> Have you tested the features `lzma `and `zstd`?

Yes, they get enabled and built if the features are requested (and disabled if they are not):

```
-- The following features have been enabled:

 * LibZstd, Support for zstd compressed files and data streams

-- The following OPTIONAL packages have been found:

 * LibLZMA, Support for xz compressed files and data streams, <http://tukaani.org/xz/>
   Support for xz compressed files and data streams
 * PkgConfig
 ```

The `KF5ArchiveConfig.cmake` is also formed as expected:

```
$ cat installed/arm64-osx/share/kf5archive/KF5ArchiveConfig.cmake

(...)
if (NOT OFF)
    if (TRUE)
        find_dependency(ZLIB)
    endif()

    if (TRUE)
        find_dependency(BZip2)
    endif()

    if (TRUE)
        find_dependency(LibLZMA)
    endif()

    if (1)
        find_package(PkgConfig)
        pkg_check_modules(LibZstd IMPORTED_TARGET "libzstd")
    endif()
endif()
```

**Comment by wrobelda** [comment]
> The essence of my first comment is still valid: "there is no pkgconfig setup in vcpkg_configure_cmake/vcpkg_cmake_configure" so I wonder how pkgconfig can work for this portfile. What did I miss?

Really not sure what you're referring to. There isn't any pkgconfig setup needed for this portfile. What this PR does is that it merely includes a patch (that has since been pushed upstream) for the config.in.cmake file. This is required so that when compiled statically it correctly passes its dependencies down.

**Comment by dg0yt** [comment]
> There isn't any pkgconfig setup needed for this portfile.

I'm talking about using the exported configuration.
~~~
        find_package(PkgConfig)
        pkg_check_modules(LibZstd IMPORTED_TARGET "libzstd")
~~~
With this line in the exported config file, depending cmake projects will call pkgconfig during the configure step in order to find zstd. Even if there is a `libzstd.pc` file and a `pkg-config` executable, how will cmake find `pkg-config`, and how will `pkg-config` know where to look for vcpkg's `libzstd.pc`?` Unless I am missing something, ports with cmake build system, and user projects, may return a "NOTFOUND" in the first line line and/or in the second line.
(Upstream probably doesn't know enough about vcpkg but assumes a system wide installation in a location which already is in pkg-config search path.)

For reference, look how these environment variables are initialized in `vcpkg_build_make`, not in `vcpkg_configure_cmake`:
`PKG_CONFIG`
`PKG_CONFIG_PATH`

**Review by NancyLi1013 — CHANGES_REQUESTED** [review]

File: ports/kf5archive/portfile.cmake [hunk 7]
```diff
@@ -4,30 +4,38 @@ vcpkg_from_github(
     REF v5.84.0
     SHA512 82926f62424446df0f4fc300f57ae9bd5baf8e13da2ce4135ac56c0c52a0307bffb06f84ac7e8e658e96ace2ae3d530f27e232061284ac87271404f218e9fdd4
     HEAD_REF master
+    PATCHES
+        9ab5f2bf.patch # https://invent.kde.org/frameworks/karchive/-/commit/9ab5f2bfbe59038b0d0b6ca7f1b22d1c9229c67e
+        5a79756f.patch # https://invent.kde.org/frameworks/karchive/-/commit/5a79756f381e1a1843cb2171bdc151dad53fb7db
+        23.patch # https://invent.kde.org/frameworks/karchive/-/merge_requests/23
+        24.patch # https://invent.kde.org/frameworks/karchive/-/merge_requests/24
```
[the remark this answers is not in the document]
wrobelda: @NancyLi1013 ping

ports/kf5archive/23.patch
```diff
@@ -0,0 +1,46 @@
...
++        find_dependency(BZip2)
++    endif()
++
++    if (@LIBLZMA_FOUND@)
++        find_dependency(LibLZMA)
++    endif()
++
++    if (@LibZstd_FOUND@)
++        find_package(PkgConfig)
++        pkg_check_modules(LibZstd IMPORTED_TARGET "libzstd")
```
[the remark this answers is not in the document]
NancyLi1013: I see, but why not merge these two patches as one patch? For the current source codes, they are all patches. We don't need to consider if they are merged to upstream or not. If 24.pacth is not merged in next version, we can make a new patch based on the new sources. But in current case, I think we should merge them.

ports/kf5archive/24.patch
```diff
@@ -0,0 +1,49 @@
...
+-    pkg_check_modules(LibZstd IMPORTED_TARGET "libzstd")
+-endif()
+-add_feature_info(LibZstd LibZstd_FOUND
+-                "Support for zstd compressed files and data streams"
++find_package(ZSTD)
++set_package_properties(LibZstd PROPERTIES
++    URL "http://www.zstd.net"
++    DESCRIPTION "Support for zstd compressed files and data streams"
++    PURPOSE "Support for zstd compressed files and data streams"
+ )
```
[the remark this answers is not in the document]
NancyLi1013: `set_package_properties` is not the problem. But I don't think we need to download URL here. Since we prefer to use the packages provided by vcpkg instead of third party.

ports/kf5archive/portfile.cmake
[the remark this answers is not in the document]
NancyLi1013: Make sense. Thanks for your detailed clarification about this. Your cases are a little special. I will talk with our colleagues and update the opinions here.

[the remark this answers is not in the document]
wrobelda: For what it's worth, almost all of these patches are already upstream and will be dropped when I update the kf5 framework next.

ports/kf5archive/24.patch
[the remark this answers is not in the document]
wrobelda: This doesn't download any package, it's just for information purposes. Please refer to the upstream PR this patch was generated from: https://invent.kde.org/frameworks/karchive/-/merge_requests/24/diffs#9a2aa4db38d3115ed60da621e012c0efc0172aae_24_24

ports/kf5archive/portfile.cmake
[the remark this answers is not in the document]
NancyLi1013: The current behavior is reasonable. It will be more convenient for version update. Thanks again for your patience.

ports/kf5archive/24.patch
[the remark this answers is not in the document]
NancyLi1013: Thanks again for your information. It's me that consider too much. I'm sorry about this.

ports/kf5archive/23.patch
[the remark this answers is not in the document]
NancyLi1013: This is reasonable. it's unnecessary to update this. Thanks!

NancyLi1013 (approved):

NancyLi1013:
LGTM now, thanks for your efforts and patience. @wrobelda

wrobelda:
> LGTM now, thanks for your efforts and patience. @wrobelda

Thank you very much for yours as well!

ras0219-msft:
This LGTM, thanks everyone for the efforts!

PR #19219, opened 2021-07-29
[spirv-tools] Fix build, support dynamic build

### LINKED ISSUE #18612: spirv-tools explicitly deletes DLLs, then aborts when consumed via find_package due to missing DLLs
**Describe the bug**
The `spirv-tools` package explicitly deletes the output of the `SPIRV-Tools-shared` target here: https://github.com/barcharcraz/vcpkg/blob/master/ports/spirv-tools/portfile.cmake#L46-L51

The `SPIRV-Tools-shared` target is still created even with `-DSPIRV_TOOLS_BUILD_STATIC=ON` (which is set explicitly in the portfile) so the upstream CMake still installs it, and creates a check for its presence in the module importing script.

That means if the package is consumed via `find_package`, the check will fail, and a `FATAL_ERROR` message is logged. This aborts the build.

**Environment**
- OS: Windows 10
- Compiler: MSVC 2019 16.9.5

**To Reproduce**
1. `./vcpkg install spirv-tools`
2. Create a new CMake project including `find_package(SPIRV-Tools REQUIRED)`. A `CMakeLists.txt` containing this single line should be enough.
3. Configure the CMake project using the normal Vcpkg toolchain.
4. Observe that a fatal error is printed stating that:
   ```
    The imported target "SPIRV-Tools-shared" references the file

       "C:/Clones/vcpkg/installed/x64-windows/debug/bin/SPIRV-Tools-shared.dll"

       "C:/Clones/vcpkg/installed/x64-windows/share/SPIRV-Tools/SPIRV-ToolsTarget.cmake"

     C:/Clones/vcpkg/installed/x64-windows/share/SPIRV-Tools/SPIRV-ToolsConfig.cmake:1 (include)
     C:/Clones/vcpkg/scripts/buildsystems/vcpkg.cmake:772 (_find_package)
     CMakeLists.txt:119 (find_package)
   ```

**Expected behavior**
Once `find_package(SPIRV-Tools REQUIRED)` has been called, the configure proceeds and the `SPIRV-Tools-static` library is available (also as `SPIRV-Tools` with no suffix).

**Failure logs** 
- The relevant parts of the log are above.
- This is a log from a dummy CMake project with literally just the `find_package` call, so it's got some other warnings about not being a real CMake project: [log.log](https://github.com/microsoft/vcpkg/files/6705076/log.log)

**Additional context**
N/A

JackBoosY:
- Generate the corresponding library according to the build type.
- Support dynamic build.

### DIFF
--- a/ports/spirv-tools/fix-build-type.patch
+++ b/ports/spirv-tools/fix-build-type.patch
@@ -1,0 +1,62 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d1c9d5e..9acfe47 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -159,7 +159,7 @@ endif()
+ # Ideally, all of these will go away - see https://github.com/KhronosGroup/SPIRV-Tools/issues/3909.
+ option(SPIRV_TOOLS_BUILD_STATIC "Build ${SPIRV_TOOLS}-static target. ${SPIRV_TOOLS} will alias to ${SPIRV_TOOLS}-static or ${SPIRV_TOOLS}-shared based on BUILD_SHARED_LIBS" ON)
+ if(SPIRV_TOOLS_BUILD_STATIC)
+-  set(SPIRV_TOOLS_FULL_VISIBILITY ${SPIRV_TOOLS}-static)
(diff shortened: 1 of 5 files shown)


NancyLi1013:
Related issue #18612.

JackBoosY:
Will reopen this PR until the upstream response.

PR #19238, opened 2021-07-29
[scripts-audit] vcpkg_download_distfile

strega-nil:
depends on PR microsoft/vcpkg-tool#142


strega-nil-ms:
Closed for rollup #19469.

PR #19245, opened 2021-07-30
[vcpkg-cmake] Fix docs and usage of option MAYBE_UNUSED_VARIABLES

JackBoosY:
In the current document, our description of `MAYBE_UNUSED_VARIABLES` caused the user to misunderstand the parameter and incorporate the wrong changes.

Fix the option description and add sample code to the docs.

Related: https://github.com/microsoft/vcpkg/pull/13467

### DIFF
--- a/ports/kf5sonnet/portfile.cmake
+++ b/ports/kf5sonnet/portfile.cmake
@@ -14,13 +14,6 @@ vcpkg_cmake_configure(
         -DKDE_INSTALL_PLUGINDIR=plugins
         -DKDE_INSTALL_DATAROOTDIR=data
         -DKDE_INSTALL_QTPLUGINDIR=plugins
-        -DBUILD_HTML_DOCS=OFF
-        -DBUILD_MAN_DOCS=OFF
-        -DBUILD_QTHELP_DOCS=OFF
-    MAYBE_UNUSED_VARIABLES
-        BUILD_HTML_DOCS
-        BUILD_MAN_DOCS
-        BUILD_QTHELP_DOCS
 )
 
 vcpkg_add_to_path(PREPEND "${CURRENT_INSTALLED_DIR}/bin")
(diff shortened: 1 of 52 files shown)


JackBoosY:
cc @wrobelda

ports/kf5sonnet/portfile.cmake
```diff
@@ -14,13 +14,6 @@ vcpkg_cmake_configure(
         -DKDE_INSTALL_PLUGINDIR=plugins
         -DKDE_INSTALL_DATAROOTDIR=data
         -DKDE_INSTALL_QTPLUGINDIR=plugins
-        -DBUILD_HTML_DOCS=OFF
-        -DBUILD_MAN_DOCS=OFF
-        -DBUILD_QTHELP_DOCS=OFF
-    MAYBE_UNUSED_VARIABLES
-        BUILD_HTML_DOCS
-        BUILD_MAN_DOCS
-        BUILD_QTHELP_DOCS
```
JackBoosY, line 23: @wrobelda I checked the upstream CMakeLists.txt, but these options doesn't exist.

JackBoosY:
Need to check the related warning message after pipeline test.

ports/kf5sonnet/portfile.cmake
wrobelda: Indeed. However, these are standard KDE variables used across KDE software. If at any point a library adds a manual or HTML docs, these will apply. I thought that it makes sense for consistency with other frameworks where such variables are already present in their CMakeLists.txt.

JackBoosY: @wrobelda 
`MAYBE_UNUSED_VARIABLES` is only for the corresponding ports. If some of them have these parameters, we should only add `MAYBE_UNUSED_VARIABLES` to these ports.

JackBoosY:
Question: should we also add the following requirement in https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#build-techniques?

strega-nil-ms (approved):

PhoebeHui (approved):

PR #19256, opened 2021-07-30
[signalrclient]Update homepage URL

BrennanConroy:
https://github.com/aspnet/SignalR-Client-Cpp is for the newer microsoft-signalr package.

### DIFF
--- a/versions/s-/signalrclient.json
+++ b/versions/s-/signalrclient.json
@@ -1,7 +1,7 @@
 {
   "versions": [
     {
-      "git-tree": "dbe638f959b879da354861fb39b3d4dbb179ea7d",
+      "git-tree": "dbddf4175663355eea96e5474d904fc08bda4847",
       "version-string": "1.0.0-beta1-9",
       "port-version": 0
     },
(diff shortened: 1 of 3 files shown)


JonLiu1993 (approved):

versions/s-/signalrclient.json
```diff
@@ -1,9 +1,9 @@
 {
   "versions": [
     {
-      "git-tree": "dbe638f959b879da354861fb39b3d4dbb179ea7d",
+      "git-tree": "a1cf11834f81f78f10ff18bcec0d7f1c42309f62",
       "version-string": "1.0.0-beta1-9",
-      "port-version": 0
+      "port-version": 1
```
dan-shaw, line 6: Can we add a new entry to the version list?

JonLiu1993: > Can we add a new entry to the version list?

When we add "port-version" to the CONTROL file, we must run the command _./vcpkg x-add-version signalrclient --overwrite-version_ after submitting the modification to update "git-tree", once the command is used, "port-version" The value will automatically keep the same as in the CONTROL file

PR #19272, opened 2021-07-31
[yaml-cpp] update to 0.7.0

### LINKED ISSUE #19255: yaml-cpp update to 0.7.0
Library name: yaml-cpp

New version number: 0.7.0

https://github.com/jbeder/yaml-cpp/releases/tag/yaml-cpp-0.7.0

autoantwort:
Fixes #19255


JonLiu1993 (commented):
@autoantwort ,Why delete the opencolorio-tools port?

autoantwort:
> @autoantwort ,Why delete the opencolorio-tools port?

Because it was simply a tools feature if the `opencolorio` port, I assume that was done when features were not supported in vcpkg so that they created a new `...-tools` port for that.

ports/opencolorio/fix-pystring-path.patch
```diff
@@ -0,0 +1,178 @@
+diff --git a/share/cmake/modules/Findpystring.cmake b/share/cmake/modules/Findpystring.cmake
+index a3e396de..14374eed 100644
+--- a/share/cmake/modules/Findpystring.cmake
++++ b/share/cmake/modules/Findpystring.cmake
+@@ -33,7 +33,7 @@ if(NOT OCIO_INSTALL_EXT_PACKAGES STREQUAL ALL)
+     # Find include directory
+     find_path(pystring_INCLUDE_DIR
+         NAMES
+-            pystring/pystring.h
```
ras0219, line 9: This looks like we have a bug in `pystring` and the header should actually be at `pystring/pystring.h`. Considering `pystring` is owned by the same entity as `opencolorio`, I trust that their `Findpystring` module is correct.

Could you change `pystring` to install the header into `pystring/pystring.h` instead of adding this patch?

ports/opencolorio/fix-half-version.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/share/cmake/modules/FindHalf.cmake b/share/cmake/modules/FindHalf.cmake
+index da308c02..430dbd25 100644
+--- a/share/cmake/modules/FindHalf.cmake
++++ b/share/cmake/modules/FindHalf.cmake
+@@ -23,7 +23,7 @@
+ #
+ 
+ # IlmBase components may have the version in their name
+-set(_Half_LIB_VER "${Half_FIND_VERSION_MAJOR}_${Half_FIND_VERSION_MINOR}")
++set(_Half_LIB_VER "${Half_FIND_VERSION_MAJOR}_5")
```
ras0219, line 10: This looks an awful lot like hardcoding in a specific version of the dependency; can we change this to use `find_package(OpenEXR CONFIG)` instead?

ports/opencolorio/vcpkg.json
```diff
@@ -1,29 +1,24 @@
 {
   "name": "opencolorio",
-  "version-semver": "1.1.1",
-  "port-version": 5,
+  "version-semver": "2.0.1",
   "description": "OpenColorIO (OCIO) is a complete color management solution geared towards motion picture production with an emphasis on visual effects and computer animation. OCIO provides a straightforward and consistent user experience across all supporting applications while allowing for sophisticated back-end configuration options suitable for high-end production usage. OCIO is compatible with the Academy Color Encoding Specification (ACES) and is LUT-format agnostic, supporting many popular formats.",
   "homepage": "https://opencolorio.org/",
   "dependencies": [
+    "expat",
```
ras0219, line 7: I see a dependency on `openimageio` which should either be explicitly disabled or added as a dep: https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/src/CMakeLists.txt#L11

ports/opencolorio/vcpkg.json
```diff
@@ -1,29 +1,24 @@
...
-    {
-      "name": "yaml-cpp",
-      "default-features": false
+    "yaml-cpp"
+  ],
+  "features": {
+    "tools": {
+      "description": "Installs tools",
+      "dependencies": [
+        "lcms"
```
ras0219, line 20: It looks like building the apps also depends on `GLEW`: https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/0021372b4dba3444e21a36d846b327efad054f8d/share/cmake/utils/CheckSupportGL.cmake#L10

ras0219 (changes requested):

ports/opencolorio/portfile.cmake
```diff
@@ -1,73 +1,48 @@
...
+    FEATURES
+        tools OCIO_BUILD_APPS
+)
 
-# TODO(theblackunknown) build additional targets based on feature
 
-vcpkg_configure_cmake(
+vcpkg_cmake_configure(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
```
JonLiu1993, line 21:
```suggestion
vcpkg_cmake_configure(
    SOURCE_PATH "${SOURCE_PATH}"
```

ports/opencolorio/portfile.cmake
```diff
@@ -1,73 +1,48 @@
...
-    ${CURRENT_PACKAGES_DIR}/OpenColorIOConfig.cmake
 )
+if (OCIO_BUILD_APPS)
+    vcpkg_copy_tools(
+        TOOL_NAMES ociowrite ociomakeclf ociochecklut ociocheck ociobakelut
+        AUTO_CLEAN
+    )
+endif()
 
 file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
```
JonLiu1993, line 48:
```suggestion
file(INSTALL "${SOURCE_PATH}/LICENSE" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}" RENAME copyright)
```

ports/yaml-cpp/portfile.cmake
```diff
@@ -14,7 +12,7 @@ else()
     set(YAML_BUILD_SHARED_LIBS OFF)
 endif()
 
-vcpkg_configure_cmake(
+vcpkg_cmake_configure(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
```
JonLiu1993, line 17:
```suggestion
vcpkg_cmake_configure(
    SOURCE_PATH "${SOURCE_PATH}"
```

ports/yaml-cpp/portfile.cmake
```diff
@@ -23,17 +21,18 @@ vcpkg_configure_cmake(
...
+    vcpkg_cmake_config_fixup(CONFIG_PATH cmake)
 endif()
 if(EXISTS ${CURRENT_PACKAGES_DIR}/lib/cmake/yaml-cpp)
-    vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/yaml-cpp)
+    vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/yaml-cpp)
 endif()
 
 # Remove debug include files
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
+file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
```
JonLiu1993, line 35:
```suggestion
file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
```

ports/yaml-cpp/vcpkg.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "yaml-cpp",
+  "version": "0.7.0",
```
JonLiu1993, line 3:
```suggestion
  "version-semver": "0.7.0",
```

JonLiu1993 (changes requested):

ports/yaml-cpp/vcpkg.json
ras0219: Semver seems correct here.

ports/yaml-cpp/vcpkg.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "yaml-cpp is a YAML parser and emitter in C++ matching the YAML 1.2 spec.",
+  "homepage": "https://github.com/jbeder/yaml-cpp",
```
ras0219, line 5:
```suggestion
  "homepage": "https://github.com/jbeder/yaml-cpp",
  "documentation": "https://codedocs.xyz/jbeder/yaml-cpp/index.html",
```

ports/opencolorio/fix-pystring-path.patch
autoantwort: Hm opencolorio uses a [custom build script](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/master/share/cmake/projects/Buildpystring.cmake) to build pystring and move it to the pystring folder, but the [original makefile](https://github.com/imageworks/pystring/blob/master/Makefile) does not do that.

ports/opencolorio/vcpkg.json
autoantwort: Openimageio has an optional dependency on opencolorio, that can cause a circle.

ports/yaml-cpp/vcpkg.json
autoantwort: Have you checked that the lib follows the strict semver rules?

autoantwort:
Thank you for your detailed review @ras0219 ❤️

ports/yaml-cpp/vcpkg.json
JonLiu1993: Yes, I checked the historical version records, "port-semver" is more suitable

ports/openimageio/portfile.cmake
```diff
@@ -8,12 +8,11 @@ endif()
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO OpenImageIO/oiio
-    REF 5167b11277fffcd9fe18fe4dc35b3eb2669d8c44 # 2.2.10
-    SHA512 d5812cf93bbaf8a384e8ee9f443db95a92320b4c35959a528dff40eac405355d1dec924a975bef7f367d3a2179ded0a15b4be9737d37521719739958bb7f3123
+    REF 099c8585e3add6f58fab9aa438a491fa55d3f67e # 2.2.16.0
```
JonLiu1993, line 11:
```suggestion
    REF 099c8585e3add6f58fab9aa438a491fa55d3f67e # 2.2.17.0
```

ports/pystring/vcpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "pystring",
+  "version-string": "1.1.3",
```
JonLiu1993, line 3:
```suggestion
  "version-semver": "1.1.3",
```

JonLiu1993 (changes requested):

JonLiu1993 (approved):

PR #19274, opened 2021-08-01
[wxwidgets] make wxUSE_STL an optional triplet feature disabled by default

Be-ing:
wxUSE_STL defaults to OFF so that is how it is in Linux
distribution packages. Downstream projects developed with
wxUSE_STL=ON are not necessarily compatible with wxUSE_STL=OFF
without modification. So, by default, go with wxWidgets' default
for compatibility with downstream codebases. vcpkg users who need
wxUSE_STL=ON can still do so by simply opting into the new 'stl'
feature of this port.

https://forums.wxwidgets.org/viewtopic.php?p=165208

**Describe the pull request**

- #### What does your PR fix?  
  Fixes downstream applications not building with wxwidgets port

  all

  yes

  Yes

### DIFF
--- a/ports/wxwidgets/portfile.cmake
+++ b/ports/wxwidgets/portfile.cmake
@@ -23,6 +23,13 @@ if(VCPKG_TARGET_ARCHITECTURE STREQUAL arm64 OR VCPKG_TARGET_ARCHITECTURE STREQUA
     )
 endif()
 
+# This may be set to ON by users in a custom triplet.
+# wxUSE_STL=ON and wxUSE_STL=OFF are not API compatible which is why this must be set
+# in a custom triplet rather than a port feature.
+if(NOT DEFINED WXWIDGETS_USE_STL)
+    set(WXWIDGETS_USE_STL OFF)
+endif()
+
 vcpkg_configure_cmake(
     SOURCE_PATH ${SOURCE_PATH}
     PREFER_NINJA
@@ -33,8 +40,8 @@ vcpkg_configure_cmake(
         -DwxUSE_LIBJPEG=sys
         -DwxUSE_LIBPNG=sys
         -DwxUSE_LIBTIFF=sys
-        -DwxUSE_STL=ON
         -DwxBUILD_DISABLE_PLATFORM_LIB_DIR=ON
+        -DwxUSE_STL=${WXWIDGETS_USE_STL}
         ${OPTIONS}
 )
 
--- a/ports/wxwidgets/vcpkg.json
+++ b/ports/wxwidgets/vcpkg.json
@@ -1,8 +1,11 @@
 {
   "name": "wxwidgets",
   "version-semver": "3.1.5",
-  "port-version": 2,
-  "description": "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications.",
+  "port-version": 3,
+  "description": [
+    "Widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications. ",
+    "Set WXWIDGETS_USE_STL in a custom triplet to build with the wxUSE_STL build option."
+  ],
   "homepage": "https://github.com/wxWidgets/wxWidgets",
   "supports": "!uwp",
   "dependencies": [
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -6738,7 +6738,7 @@
     },
     "wxwidgets": {
       "baseline": "3.1.5",
-      "port-version": 2
+      "port-version": 3
     },
     "x-plane": {
       "baseline": "3.0.3",
--- a/versions/w-/wxwidgets.json
+++ b/versions/w-/wxwidgets.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "dcd0a52fd041f8c7b658a01ec1141f36d77220c6",
+      "version-semver": "3.1.5",
+      "port-version": 3
+    },
+    {
       "git-tree": "6fa230bffdee1e7d700570c31e6f08367460c0c9",
       "version-semver": "3.1.5",
       "port-version": 2


PhoebeHui (approved):
LGTM

JackBoosY (approved):

dg0yt:
I would like to ask for clarificiation whether this feature is additive and compatible with binary caching:
- How does it affect ABI?
- How does it affect the library's backward compatibility.
- Basically, it Port B requires wxwidgets[stl], will it break Port A which was built agains wxwidgets?

Be-ing:
wxUSE_STL=ON and wxUSE_STL=OFF are not source compatible so I can't imagine they would be ABI compatible.

dg0yt:
> wxUSE_STL=ON and wxUSE_STL=OFF are not source compatible so I don't imagine they would be ABI compatible.
> 
> > Basically, it Port B requires wxwidgets[stl], may it break Port A which was built agains wxwidgets[core]?
> 
> I suppose so.

This is the list of "Do not" for features: 
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#features

I get the impression that the proposed feature is in conflict with at least one item.

Be-ing:
The only alternatives are:
* Keep this port as-is and make it unusable for lots of wxWidgets applications.
* Provide no option and only use wxUSE_STL=OFF which would make the upstream port unusable if anyone was previously depending on wxUSE_STL=ON.

Be-ing:
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-control-alternatives-in-published-interfaces lists some exceptions made for backwards compatibility. I think this case should have an exception made for backwards compatibility.

dg0yt:
> I think this case should have an exception made for backwards compatibility.

Don't misunderstand backwards compatibility and the existing exceptions: These ports already had a broken feature list. The feature were *left* for backwards compatibility.

> The only alternatives are:
> 
> * Keep this port as-is which is unusable for lots of wxWidgets applications.
> * Provide no option and only use wxUSE_STL=OFF which would make the upstream port unusable if anyone was previously depending on wxUSE_STL=ON.

There is more:

 * Let the user use overlay ports to change the default.
 * Use another variable in the portfile, and let users override this variable in *custom triplets* if needed.

Be-ing:
> Let the user use overlay ports to change the default.

I don't think this is a good idea because the port currently does not work with downstream codebases that were developed with wxWidget's default of wxUSE_STL=OFF.

> Use another variable in the portfile, and let users override this variable in custom triplets if needed.

This would be quite cumbersome. I don't know why that would be better than a port feature.

dg0yt:
> > Use another variable in the portfile, and let users override this variable in custom triplets if needed.
> 
> This would be quite cumbersome. I don't know why that would be better than a port feature.

It creates an entire consistent universe of compatible binaries bound to that particular triplet. 

With features, this is not possible: The binaries created from `wxwidgets[stl]` are incompatible with ports depending on `wxwidgets[core]`  as far as we are told. This is not additive.

Be-ing:
I would be annoyed if I had to use a custom triplet just to use wxWidgets with its upstream defaults. But if that's what the vcpkg maintainers decide, it should be documented.

strega-nil-ms:
Do you have a repro? I'm trying to figure out why this is necessary, except for the `wxUSE_STL_BASED_WXSTRING` stuff (which we could fix by defining that).

Be-ing:
Unfortunately due to other bugs in the wxwidgets port (#17111) I cannot tell you how to reproduce this with the wxwidgets port in vcpkg. I am working with this branch of Tenacity: https://github.com/tenacityteam/tenacity/pull/228 This branch is still very much a work-in-progress and I am force pushing frequently.

When I tried to build that branch with the wxwidgets port in vcpkg which uses wxUSE_STL=ON, the build failed [with these errors](https://github.com/tenacityteam/tenacity/pull/228#issuecomment-880113285). When I build wxwidgets manually without vcpkg using the default wxUSE_STL=OFF, I am able to build that branch.

JackBoosY:
> I am working with this branch of Tenacity: [tenacityteam/tenacity#228](https://github.com/tenacityteam/tenacity/pull/228) This branch is still very much a work-in-progress and I am force pushing frequently.
> 
> When I tried to build that branch with the wxwidgets port in [wxWidgets/wxWidgets#2402](https://github.com/wxWidgets/wxWidgets/pull/2402) which uses wxUSE_STL=ON, the build failed [with these errors](https://github.com/tenacityteam/tenacity/pull/228#issuecomment-880113285). When I build wxwidgets manually without vcpkg using the default wxUSE_STL=OFF, I am able to build that branch. When I rebuilt wxwidgets manually with wxUSE_STL=ON, I reproduced the compile errors.

Those failures are due to the upstream doesn't handle the library path. See https://github.com/wxWidgets/wxWidgets/pull/2402.

talregev:
@JackBoosY Will you help us to continue the PR that solve the problems?

Be-ing:
> Those failures are due to the upstream doesn't handle the library path. See wxWidgets/wxWidgets#2402.

Yes the first error I posted in https://github.com/tenacityteam/tenacity/pull/228#issuecomment-880113285 was due to that upstream bug. But then I ran into the compile errors because of wxUSE_STL=ON.

Be-ing:
https://docs.wxwidgets.org/3.0/classwx_string.html has this note:

> If you built wxWidgets with wxUSE_STL set to 1, the implicit conversions to both narrow and wide C strings are disabled and replaced with implicit conversions to std::string and std::wstring.

I think this is the source of the compile error I encountered with wxUSE_STL=ON.

strega-nil-ms:
@Be-ing could you try instead with `wxUSE_STL=1`, and `wxUSE_STL_BASED_WXSTRING=0`?

Be-ing:
That's not an available CMake variable:
```
CMake Warning:

    wxUSE_STL_BASED_WXSTRING
```

strega-nil-ms:
urggh... okay, I think the best thing is to default to `-DwxUSE_STL=OFF` then, and allow for turning it back on.

Be-ing:
So merge this PR?

strega-nil-ms:
@Be-ing no, it shouldn't be a feature. Could you turn it off by default, and make turning it back on a triplet option?

Be-ing:
Okay, that sounds like a good solution. Is there documentation or an example how to do that? Just set the `wxUSE_STL` CMake option based on a CMake variable? Then users can choose to set that CMake variable in a custom triplet?

Be-ing:
Where would I document that the triplet option exists?

dg0yt:
> Where would I document that the triplet option exists?

Good question. Documentation exists for global triplet variables, but not for per-port variables. How about adding a sentence to the port's description field? In this way, it would be visible in online package search. E.g. 
~~~
To enable 'wxUSE_STL', set 'VCPKG_WXWIDGETS_USE_STL' to 'ON' in a custom triplet file.
~~~

ports/wxwidgets/portfile.cmake
```diff
@@ -23,6 +23,13 @@ if(VCPKG_TARGET_ARCHITECTURE STREQUAL arm64 OR VCPKG_TARGET_ARCHITECTURE STREQUA
     )
 endif()
 
+# This may be set to ON by users in a custom triplet.
+# wxUSE_STL=ON and wxUSE_STL=OFF are not API compatible which is why this must be set
+# in a custom triplet rather than a port feature.
+if(NOT DEFINED WXWIDGETS_USE_STL)
```
talregev, line 29: In triplet, you can set only VCPKG_ prefix vars.
I am not sure where neet to add this var. 
https://vcpkg.readthedocs.io/en/latest/users/triplets/

ports/wxwidgets/vcpkg.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "wxwidgets",
   "version-semver": "3.1.5",
-  "port-version": 2,
-  "description": "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications.",
+  "port-version": 3,
+  "description": "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications. Set WXWIDGETS_USE_STL in a custom triplet to build with the wxUSE_STL build option.",
```
talregev, line 5: Can you force it to next line? Because usually you cannot see long lines.
I think we also need to take care of it globally.

talregev (changes requested):

ports/wxwidgets/portfile.cmake
Be-ing: Are you sure? All the global variables for vcpkg use the `VCPKG_` prefix but I don't know if this convention should be followed for port-specific variables.

talregev: I know that from android, When I wanted to add some vars. It needed vcpkg_ prefix. also to add in some place.
I will let other here to comment that know this matter to explain it better.

strega-nil-ms: No, `VCPKG_` should be used for vcpkg variables; for this kinda thing, WXWIDGETS is a perfectly fine prefix for the wxwidgets port.

ports/wxwidgets/vcpkg.json
strega-nil-ms:
```suggestion
  "description": [
    "a widget toolkit and tools library for creating graphical user interfaces (GUIs) for cross-platform applications.",
    "Set WXWIDGETS_USE_STL in a custom triplet to build with the wxUSE_STL build option."
  ],
```

strega-nil-ms (approved):

ports/wxwidgets/portfile.cmake
talregev: ok. I didn't know that. thank you for the correction.

talregev (approved):

JackBoosY (approved):

Be-ing:
This has been approved by 3 reviewers. Can it be merged?

Be-ing:
Thanks for reviewing. I am glad we found a solution that works with vcpkg's policies. Now, onto #17111...

talregev:
@Be-ing @JackBoosY 
> Now, onto #17111...
I so agree ...

PR #19276, opened 2021-08-01
[qhull] fix missing symbols

### LINKED ISSUE #18854: [qhull] qset.c functions are not built
Hi. I installed PCL with vcpkg and so qhull is installed too. I'm using PCL already but I want to access qhull functionality too. When I try to use qhull's runQhull function (in orgQhull:Qhull), linker fails as "unresolved external symbol qh_memcheck" and some other qset.c related functions. I tried to build qhull myself and when I did that, it fails because my build is a different version against vcpkg's. How can I solve this issue? I'm using latest version of both vcpkg and all other libraries.

fran6co:
I need this fix for https://github.com/microsoft/vcpkg/pull/17423

It contains upstream fix https://github.com/qhull/qhull/pull/93

- #### What does your PR fix?  
  Fixes #18854

  all, Yes

  Yes

### DIFF
--- a/ports/qhull/fix-missing-symbols.patch
+++ b/ports/qhull/fix-missing-symbols.patch
@@ -1,0 +1,29 @@
+From 02ba907908ccf5ed0d40a2251272b67cb05c23ea Mon Sep 17 00:00:00 2001
+From: Francisco Facioni <fran6co@gmail.com>
+Date: Sat, 31 Jul 2021 22:30:45 +0100
+Subject: [PATCH] Missing symbols in the export
+
+---
+ src/libqhull_r/qhull_r-exports.def | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/src/libqhull_r/qhull_r-exports.def b/src/libqhull_r/qhull_r-exports.def
+index 4c5e17c..b789673 100644
+--- a/src/libqhull_r/qhull_r-exports.def
++++ b/src/libqhull_r/qhull_r-exports.def
+@@ -183,6 +183,7 @@ qh_maxouter
+ qh_maxsimplex
+ qh_maydropneighbor
+ qh_memalloc
++qh_memcheck
+ qh_memfree
+ qh_memfreeshort
+ qh_meminit
+@@ -351,6 +352,7 @@ qh_setdelnth
+ qh_setdelnthsorted
+ qh_setdelsorted
+ qh_setduplicate
++qh_setendpointer
+ qh_setequal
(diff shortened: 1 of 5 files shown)


ports/qhull/portfile.cmake
```diff
@@ -5,6 +5,7 @@ vcpkg_from_github(
     SHA512 5b8ff9665ba73621a9859a6e86717b980b67f8d79d6c78cbf5672bce66aed671f7d64fcbec457bca79eef2e17e105f136017afdf442bb430b9f4a059d7cb93c3
     HEAD_REF master
     PATCHES include-qhullcpp-shared.patch
+            fix-missing-symbols.patch # upstream https://github.com/qhull/qhull/pull/93
```
NancyLi1013, line 8:
```suggestion
    PATCHES 
        include-qhullcpp-shared.patch
        fix-missing-symbols.patch # upstream https://github.com/qhull/qhull/pull/93
```

NancyLi1013 (changes requested):

NancyLi1013 (approved):

ports/qhull/fix-missing-symbols.patch
```diff
@@ -0,0 +1,29 @@
...
+ qh_memfreeshort
+ qh_meminit
+@@ -351,6 +352,7 @@ qh_setdelnth
+ qh_setdelnthsorted
+ qh_setdelsorted
+ qh_setduplicate
++qh_setendpointer
+ qh_setequal
+ qh_setequal_except
+ qh_setequal_skip
```
NancyLi1013, line 29: For the patch, we're not sure if upstream will accept the changes. So maybe we can wait for the response from upstream, what do you think? @fran6co

fran6co: I mean, on windows right now qhull is not working if it's used as a shared library not sure what other fix could be done for this

cenit:
glad you are still working on open3d here on vcpkg!!

fran6co:
> glad you are still working on open3d here on vcpkg!!

it's slow going, but getting there

cenit:
yeah but the byproduct is many ports finally working and not just appearing here on vcpkg (in the past there was a rush to add ports but usually people didn’t even tried them or used them, with the obvious result). Just look at the sequence of PR extracted from my original one and how many ports you already helped fixing and made working.

this is another very good one (and qhull was a mess with static/dynamic cmake targets too in the past, i don’t know current status)

ras0219 (approved):

PR #19297, opened 2021-08-02
[boost-build/boost-uninstall] Update to 1.76.0

PhoebeHui:
Others boost components has been upgraded to 1.76.0, however, we missed boost-build and boost-uninstall. 

Remove patch 'fix_options.patch', since the release version 1.76.0 included the changes.
See https://github.com/boostorg/build/blob/develop/src/tools/msvc.jam


autoantwort:
I assume the problem is still https://github.com/boostorg/build/issues/724. That was also the Problem in the upgrade PR: https://github.com/microsoft/vcpkg/pull/17335#issuecomment-821872323. I have no idea how to fix that, so it currently still use boost build 1.75.0, that works. To be honest I don't think that that will be fixed, maybe we should wait until boost fully supports CMake and than switch to CMake.

xavier2k6:
@autoantwort you may want to open a "new issue" over at https://github.com/bfgroup/b2 ref.: https://github.com/boostorg/build/issues/724#issuecomment-850859562 so that you may get a quicker response & you could reference your OG issue from https://github.com/boostorg/build/issues/724

PhoebeHui:
Closing in favor of https://github.com/microsoft/vcpkg/pull/19556.

PR #19298, opened 2021-08-02
[proj4] Add usage, fixup target PROJ::proj

### LINKED ISSUE #16906: FindProj not working as expected

JackBoosY:
Since the upstream indicated in the document that it prefers `PROJ` instead of `PROJ4`, fixup the target `PROJ` and only provide the usage of target `PROJ` in the usage and keep the exported PROJ4 cmake configuration file.

Fixes #16906.


dg0yt:
Which upstream changes does this need? 
We need an update to port vtk in order to allow an update to port proj in order to follow upstream.

versions/p-/proj4.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "e6b33c3a3987f7198d8cd25fd12640fcaccc9773",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "2a79aed1616c4f305f0231da48764e93a60ac74a",
```

dg0yt:
The PROJ project exports valid file names. Which problem is actually fixed by changing them?
If there is no good reason, it seems quite intrusive to ask upstream for changes. I would reject the change.

JackBoosY:
@dg0yt According to cmake [docs](https://cmake.org/cmake/help/latest/command/export.html):
```
Store the current build directory in the CMake user package registry for package <PackageName>. 
The find_package() command may consider the directory while searching for package <PackageName>. 
This helps dependent projects find and use a package from the current project's build tree without help from the user. 
Note that the entry in the package registry that this command creates works only in conjunction with a package configuration file (<PackageName>Config.cmake) that works with the build tree. 
In some cases, for example for packaging and for system wide installations, it is not desirable to write the user package registry.
```

dg0yt:
> @dg0yt According to cmake [docs](https://cmake.org/cmake/help/latest/command/export.html):
> 
> ```
> Store the current build directory in the CMake user package registry for package <PackageName>. 
> ...
> ```

@JackBoosY Now how do we get to discussing [package registries](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-registry)?
Vcpkg builds cmake projects with package registry export and usage disabled:

~~~
$ git grep PACKAGE_REGISTRY scripts/
scripts/buildsystems/vcpkg.cmake:set(Z_VCPKG_UNUSED "${CMAKE_EXPORT_NO_PACKAGE_REGISTRY}")
scripts/buildsystems/vcpkg.cmake:set(Z_VCPKG_UNUSED "${CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY}")
scripts/buildsystems/vcpkg.cmake:set(Z_VCPKG_UNUSED "${CMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY}")
scripts/cmake/vcpkg_configure_cmake.cmake:        "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON"
scripts/cmake/vcpkg_configure_cmake.cmake:        "-DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON"
scripts/cmake/vcpkg_configure_cmake.cmake:        "-DCMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY=ON"
~~~

dg0yt:
For completeness, as posted before in PROJ, cmake's `find_package` explicitly looks also for `<lower-case-package-name>-config.cmake`. References:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-configuration-file
https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode

JackBoosY:
> 为了完整起见，正如之前在 PROJ 中发布的那样，cmake`find_package`还明确查找`<lower-case-package-name>-config.cmake`. 参考资料：
> [https](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-configuration-file) : [//cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-configuration-file ](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-configuration-file)
> [https://cmake.org/cmake/help/latest/command/find_package.html #full-signature-and-config-mode](https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode)

Reasonable.

ports/proj4/usage
```diff
@@ -0,0 +1,7 @@
+The package ${PORT}:${TARGET_TRIPLET} provides CMake targets:
+
+    find_package(PROJ CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE PROJ::proj PROJ4::proj)
```
dg0yt, line 4: No `PROJ4::proj` here. It is just one library (target).

ports/proj4/usage
```diff
@@ -0,0 +1,7 @@
...
+    # Or
+    find_package(PROJ4 CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE PROJ::proj PROJ4::proj)
```
dg0yt, line 7: And no `PROJ::proj` here. It is just one library (target). 

But I would suggest to remove all references to PROJ4 in `usage` i.e. the last three lines. It is legacy, don't promote it.
(And it just works as intented anyway. Only vcpkg is unable to automatically detect what is intented.)

dg0yt (changes requested):
Thank for your patience with my comments @JackBoosY.

ports/proj4/usage
JackBoosY: In _config.cmake_, it includes both proj-targets and proj4-targets, so we can use both of them.

ports/proj4/usage
JackBoosY: See my comment above.

ports/proj4/usage
dg0yt: You can use *either*, but it is definitely *not meant to use them both simultaneously*: PROJ4::proj is legacy. It will just work for everyone how is using it already, or cannot switch yet.

ports/proj4/usage
JackBoosY, line 7:
```suggestion
    target_link_libraries(main PRIVATE PROJ::proj)
```

versions/p-/proj4.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "786f91b5cc03cadfa2e37e69600b8683baa7c5ca",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "8fc1b6bdde08a3b1f131df5a4f042885dcc7a58d",
```

PhoebeHui (approved):

ports/proj4/usage
```diff
@@ -0,0 +1,4 @@
+The package ${PORT}:${TARGET_TRIPLET} provides CMake targets:
```
ras0219-msft, line 1: How does this work? The configure line above is `@ONLY` but this is using `${}` syntax.

+ @JackBoosY

PhoebeHui: @ras0219-msft, thanks for pointing out this issue! I have submitted PR https://github.com/microsoft/vcpkg/pull/19474 to fix it.

JackBoosY: Ops, sorry for that.

PR #19299, opened 2021-08-02
[Azure-*] Remove version constraint

### LINKED ISSUE #19267: Failure with vcpkg.json manifest file and azure-
**Describe the bug**
When trying to use a vcpkg.json manifest file, it works with every library except the azure-* ones.  The vcpkg install works correctly from the command line.  It does NOT work when using the manifest file.  Adding "version>=" entry for each one does not solve the issue.  This issue was found by using the vcpkg integration with Visual Studio Manifest file.  The Manifest file works for other libraries such as rapidjson, protobuf, openssl, boost, etc.  The Manifest file only starts to fail when the azure- libraries are included.  Installing them manually using the command line works correctly, so it is only a problem when using the manifest file.  I've trimmed down the MSBuild command line which is displayed to show minimum required commands to reproduce.  I tried uninstalling, then deleting, then rechecking out, then re-bootstrapping, then re-integrating vcpkg and get the same results.  

**Environment**
- OS: Windows 10 21H1 compiling for arm64-windows
- Compiler: Microsoft Visual Studio Professional 2019 Version 16.10.4

**To Reproduce**
1. Run command "git clone https://github.com/microsoft/vcpkg"
2. Run command ".\vcpkg\bootstrap-vcpkg.bat"
3. Drop the attached vcpkg.json [vcpkg.json.txt](https://github.com/microsoft/vcpkg/files/6911457/vcpkg.json.txt) file inside the .\vcpkg folder
4. Run command ".\vcpkg\vcpkg install --triplet=arm64-windows --x-manifest-root=.\vcpkg\"
5. Error results in:
Error: no version entry for azure-storage-common-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).
Error: no version entry for azure-core-cpp at version 1.0.0.
We are currently using the version in the ports tree (1.1.0).
Error: no version entry for azure-storage-blobs-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).
Error: no version entry for azure-storage-common-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).

**Expected behavior**
I expect it to function the same when using the normal command line command:
vcpkg install --triplet=arm64-windows azure-core-cpp azure-identity-cpp azure-storage-blobs-cpp azure-storage-common-cpp azure-storage-files-datalake-cpp azure-storage-files-shares-cpp
Without it working through the manifest file, it breaks CI/CD toolchains.

**Failure logs** 
1>------ Build started: Project: Sample, Configuration: Debug ARM64 ------
1>Installing vcpkg dependencies to D:\Sample\vcpkg_installed\
1>"D:\vcpkg\scripts\buildsystems\msbuild\..\..\..\vcpkg.exe" install  --x-wait-for-lock --triplet "arm64-windows" --vcpkg-root "D:\vcpkg\scripts\buildsystems\msbuild\..\..\..\\" "--x-manifest-root=D:\Sample\\" "--x-install-root=D:\Sample\vcpkg_installed\\"
1>EXEC : error : no version entry for azure-storage-common-cpp at version 12.0.0.
1>We are currently using the version in the ports tree (12.0.1).
1>EXEC : error : no version entry for azure-core-cpp at version 1.0.0.
1>We are currently using the version in the ports tree (1.1.0).
1>EXEC : error : no version entry for azure-storage-blobs-cpp at version 12.0.0.
1>We are currently using the version in the ports tree (12.0.1).
1>EXEC : error : no version entry for azure-storage-common-cpp at version 12.0.0.
1>We are currently using the version in the ports tree (12.0.1).
1>D:\vcpkg\scripts\buildsystems\msbuild\vcpkg.targets(104,5): error MSB3073: The command ""D:\vcpkg\scripts\buildsystems\msbuild\..\..\..\vcpkg.exe" install  --x-wait-for-lock --triplet "arm64-windows" --vcpkg-root "D:\vcpkg\scripts\buildsystems\msbuild\..\..\..\\" "--x-manifest-root=D:\Sample\\" "--x-install-root=D:\Sample\vcpkg_installed\\" " exited with code 1.
1>C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(486,5): error MSB4181: The "CompileXaml" task returned false but did not log an error.
1>Done building project "Sample.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

**Additional context**
It says "We are currently using the version in the ports tree" which is newer than a version entry for the file, so what seems to be the problem?

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/19267

This version constraint used in related ports cause the failures in manifest mode. See https://github.com/microsoft/vcpkg/pull/18779#discussion_r664850641

Failures:
```
Error: no version entry for azure-storage-common-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).
Error: no version entry for azure-core-cpp at version 1.0.0.
We are currently using the version in the ports tree (1.1.0).
Error: no version entry for azure-storage-blobs-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).
Error: no version entry for azure-storage-common-cpp at version 12.0.0.
We are currently using the version in the ports tree (12.0.1).
```

cc @ahsonkhan cc @RickWinter, @vhvb1989

### DIFF
--- a/ports/azure-security-keyvault-keys-cpp/vcpkg.json
+++ b/ports/azure-security-keyvault-keys-cpp/vcpkg.json
@@ -7,8 +7,7 @@
     "This library provides Azure Key Vault Keys SDK."
   ],
   "homepage": "https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/keyvault/azure-security-keyvault-keys",
-  "license": "MIT",
-  "builtin-baseline": "14c54c49b56a964ac7f2f701a6857adb02ae1bec",
+  "license": "MIT"
   "dependencies": [
     {
       "name": "azure-core-cpp",
--- a/ports/azure-security-keyvault-keys-cpp/vcpkg.json
+++ b/ports/azure-security-keyvault-keys-cpp/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "azure-security-keyvault-keys-cpp",
   "version-semver": "4.0.0-beta.4",
+  "port-version": 1,
   "description": [
     "Microsoft Azure Key Vault Keys SDK for C++",
     "This library provides Azure Key Vault Keys SDK."
@@ -11,8 +12,7 @@
   "dependencies": [
     {
       "name": "azure-core-cpp",
-      "default-features": false,
-      "version>=": "1.1.0"
+      "default-features": false
     },
     {
       "name": "vcpkg-cmake",
--- a/ports/azure-security-keyvault-keys-cpp/vcpkg.json
+++ b/ports/azure-security-keyvault-keys-cpp/vcpkg.json
@@ -7,7 +7,7 @@
     "This library provides Azure Key Vault Keys SDK."
   ],
   "homepage": "https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/keyvault/azure-security-keyvault-keys",
-  "license": "MIT"
+  "license": "MIT",
   "dependencies": [
     {
       "name": "azure-core-cpp",
--- a/ports/azure-storage-blobs-cpp/vcpkg.json
+++ b/ports/azure-storage-blobs-cpp/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "azure-storage-blobs-cpp",
   "version-semver": "12.0.1",
+  "port-version": 1,
   "description": [
     "Microsoft Azure Storage Blobs SDK for C++",
     "This library provides Azure Storage Blobs SDK."
@@ -11,8 +12,7 @@
   "dependencies": [
     {
       "name": "azure-storage-common-cpp",
-      "default-features": false,
-      "version>=": "12.0.0"
+      "default-features": false
     },
     {
       "name": "vcpkg-cmake",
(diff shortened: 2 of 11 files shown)


NancyLi1013 (approved):

MarkVabulas:
Thank you, this fixes the issues raised in #19267

antkmsft:
@PhoebeHui, @ras0219-msft, what should be our approach with version constraints now?
These versions should have version entries, because we released them after the feature was public.
I can see an entry for the version `1.0.0`, for example: https://github.com/microsoft/vcpkg/blob/master/versions/a-/azure-core-cpp.json#L8-L12
Is this a way the bug should have been fixed?

We release `azure-core` version `1.2.0`, and are release `azure-identity` version `1.1.0`, which will only compile with `azure-core` `1.2.0`, how should we manifest that? You removed "`version>=`", is there other way?
And if not, how do we get there to achieve what we want?

Out of curiosity, if we should not be using `version>=` in this case, what scenarios should `version>=` be used in?

PR #19307, opened 2021-08-02
[angle] Fix mingw build

dg0yt:
- #### What does your PR fix?  
  - [x] Fixes dulicate "lib" in name of import libs for libEGL and libGLESv2.
  - [x] Fixes building with mingw.

  no change, no.

  yes

  yes

---

Disclaimer: not tested with actual app, just another step towards building Qt.

### DIFF
--- a/ports/angle/CMakeLists.txt
+++ b/ports/angle/CMakeLists.txt
@@ -679,9 +679,9 @@ if(NOT BUILD_SHARED_LIBS)
     add_definitions("-DANGLE_EXPORT=" "-DANGLE_UTIL_EXPORT=" "-DEGLAPI=" "-DGL_APICALL=" "-DGL_API=")
 endif()
 
-SET_TARGET_PROPERTIES(libANGLE PROPERTIES PREFIX "")
-SET_TARGET_PROPERTIES(libGLESv2 PROPERTIES PREFIX "")
-SET_TARGET_PROPERTIES(libEGL PROPERTIES PREFIX "")
+set_target_properties(libANGLE PROPERTIES PREFIX "")
+set_target_properties(libGLESv2 PROPERTIES PREFIX "" IMPORT_PREFIX "")
+set_target_properties(libEGL PROPERTIES PREFIX "" IMPORT_PREFIX "")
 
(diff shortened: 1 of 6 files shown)


dg0yt:
qt5-base doesn't use angle on windows?!

ras0219 (approved):

ports/angle/CMakeLists.txt
```diff
@@ -679,9 +686,9 @@ if(NOT BUILD_SHARED_LIBS)
     add_definitions("-DANGLE_EXPORT=" "-DANGLE_UTIL_EXPORT=" "-DEGLAPI=" "-DGL_APICALL=" "-DGL_API=")
 endif()
 
-SET_TARGET_PROPERTIES(libANGLE PROPERTIES PREFIX "")
-SET_TARGET_PROPERTIES(libGLESv2 PROPERTIES PREFIX "")
-SET_TARGET_PROPERTIES(libEGL PROPERTIES PREFIX "")
+set_target_properties(libANGLE PROPERTIES PREFIX "")
```
ras0219, line 689: Shouldn't `libANGLE` be given a similar `IMPORT_PREFIX ""`?

dg0yt: `libANGLE` is `STATIC` so there is no import library:

https://github.com/microsoft/vcpkg/blob/a863c84812089836a3c0f2f215ab3e2579fc8acf/ports/angle/CMakeLists.txt#L537

dg0yt:
> qt5-base doesn't use angle on windows?!

@ras0219 Do you have an idea why it is not supported for windows?

https://github.com/microsoft/vcpkg/blob/a863c84812089836a3c0f2f215ab3e2579fc8acf/ports/qt5-base/vcpkg.json#L7-L11

Neumann-A:
> @ras0219 Do you have an idea why it is not supported for windows?

on windows the default is always to use `-opengl dynamic` even for static builds so `angle` is not required for windows. (see https://doc.qt.io/qt-5/windows-requirements.html.)
 `angle` also has been completely removed from Qt6. It is a completely optional runtime dependency which nowadays can commonly be replaced by the normal windows opengl driver. Alternatively software rendering via mesa would also be possible.

PhoebeHui (approved):

PR #19315, opened 2021-08-02
[windows ci] ignore rmdir errors

strega-nil:
cc @BillyONeal


strega-nil:
Closed in favor of #19314

PR #19348, opened 2021-08-04
[scripts-audit] vcpkg_fixup_pkgconfig

strega-nil:
Tested on abseil - no changes except for ordering.

### DIFF
--- a/scripts/cmake/vcpkg_fixup_pkgconfig.cmake
+++ b/scripts/cmake/vcpkg_fixup_pkgconfig.cmake
@@ -116,18 +116,15 @@ function(vcpkg_fixup_pkgconfig)
         file(GLOB_RECURSE arg_RELEASE_FILES "${CURRENT_PACKAGES_DIR}/**/*.pc")
         file(GLOB_RECURSE arg_DEBUG_FILES "${CURRENT_PACKAGES_DIR}/debug/**/*.pc")
         foreach(debug_file IN LISTS arg_DEBUG_FILES)
-            list(REMOVE_ITEM arg_RELEASE_FILES "${debug_file}")
+            vcpkg_list(REMOVE_ITEM arg_RELEASE_FILES "${debug_file}")
         endforeach()
     endif()
 
     vcpkg_find_acquire_program(PKGCONFIG)
     debug_message("Using pkg-config from: ${PKGCONFIG}")
 
-    if(VCPKG_TARGET_IS_WINDOWS)
-        # Absolute Unix like paths 
-        string(REGEX REPLACE "^([a-zA-Z]):/" [[/\1/]] unix_packages_dir "${CURRENT_PACKAGES_DIR}")
-        string(REGEX REPLACE "^([a-zA-Z]):/" [[/\1/]] unix_installed_dir "${CURRENT_INSTALLED_DIR}")
-    endif()
+    string(REGEX REPLACE "^([a-zA-Z]):/" [[/\1/]] unix_packages_dir "${CURRENT_PACKAGES_DIR}")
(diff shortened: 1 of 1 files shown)


JackBoosY:

scripts/cmake/vcpkg_fixup_pkgconfig.cmake
```diff
@@ -45,149 +45,181 @@ Still work in progress. If there are more cases which can be handled here feel f
...
-        file(GLOB_RECURSE _vfpkg_DEBUG_FILES "${CURRENT_PACKAGES_DIR}/debug/**/*.pc")
-        if(_vfpkg_DEBUG_FILES)
-            list(REMOVE_ITEM _vfpkg_RELEASE_FILES ${_vfpkg_DEBUG_FILES})
-        endif()
+    if(NOT DEFINED arg_RELEASE_FILES)
+        file(GLOB_RECURSE arg_RELEASE_FILES "${CURRENT_PACKAGES_DIR}/**/*.pc")
+        file(GLOB_RECURSE arg_DEBUG_FILES "${CURRENT_PACKAGES_DIR}/debug/**/*.pc")
+        foreach(debug_file IN LISTS arg_DEBUG_FILES)
+            list(REMOVE_ITEM arg_RELEASE_FILES "${debug_file}")
+        endforeach()
```
Neumann-A, line 120: does 
`list(REMOVE_ITEM _vfpkg_RELEASE_FILES ${_vfpkg_DEBUG_FILES})` 
have problems with spaces in paths?
Otherwise I don't know why you are switching to a loop.
(as far as i remember the cmake guidelines allow variable list expansion in this case.)

scripts/cmake/vcpkg_fixup_pkgconfig.cmake
```diff
@@ -45,149 +45,181 @@ Still work in progress. If there are more cases which can be handled here feel f
...
     debug_message("Using pkg-config from: ${PKGCONFIG}")
 
-    #Absolute Unix like paths 
-    string(REGEX REPLACE "([a-zA-Z]):/" "/\\1/" _VCPKG_PACKAGES_DIR "${CURRENT_PACKAGES_DIR}")
-    string(REGEX REPLACE "([a-zA-Z]):/" "/\\1/" _VCPKG_INSTALLED_DIR "${CURRENT_INSTALLED_DIR}")
+    if(VCPKG_TARGET_IS_WINDOWS)
+        # Absolute Unix like paths 
+        string(REGEX REPLACE "^([a-zA-Z]):/" [[/\1/]] unix_packages_dir "${CURRENT_PACKAGES_DIR}")
+        string(REGEX REPLACE "^([a-zA-Z]):/" [[/\1/]] unix_installed_dir "${CURRENT_INSTALLED_DIR}")
+    endif()
```
Neumann-A, line 130: I wonder if this depends more on the host than the target. It basically depends on what is passed via `--prefix` or how the pc file is configured/generated with meson/cmake/autotools

scripts/cmake/vcpkg_fixup_pkgconfig.cmake
strega-nil: `list(REMOVE_ITEM` has problems; this should actually be `vcpkg_list`.

scripts/cmake/vcpkg_fixup_pkgconfig.cmake
strega-nil: hm, if you think it's better to just always do it, then I'll switch it back.

strega-nil-ms:
Closed for rollup #19469.

Issue #5518, opened 2019-03-04
Cannot install libconfig:x64-osx on macos High Sierra


shranet:
**Install log:**

```
macbook:vcpkg shranet$ ./vcpkg install libconfig:x64-osx
The following packages will be built and installed:
    libconfig[core]:x64-osx
Starting package 1/1: libconfig:x64-osx
Building package libconfig[core]:x64-osx... 
-- Using cached /Users/shranet/vcpkg/downloads/hyperrealm-libconfig-v1.7.2.tar.gz
-- Using source at /Users/shranet/vcpkg/buildtrees/libconfig/src/v1.7.2-ddbd1a1da6
-- Configuring x64-osx-dbg
-- Configuring x64-osx-rel
-- Building x64-osx-dbg
CMake Error at scripts/cmake/vcpkg_build_cmake.cmake:175 (message):
    Command failed: /Users/shranet/vcpkg/downloads/tools/cmake-3.12.4-osx/cmake-3.12.4-Darwin-x86_64/CMake.app/Contents/bin/cmake;--build;.;--config;Debug;--target;install;--;-v;-j1
    Working Directory: /Users/shranet/vcpkg/buildtrees/libconfig/x64-osx-dbg
    See logs for more information:
      /Users/shranet/vcpkg/buildtrees/libconfig/install-x64-osx-dbg-out.log

Call Stack (most recent call first):
  scripts/cmake/vcpkg_install_cmake.cmake:24 (vcpkg_build_cmake)
  ports/libconfig/portfile.cmake:19 (vcpkg_install_cmake)
  scripts/ports.cmake:71 (include)


Error: Building package libconfig:x64-osx failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: libconfig:x64-osx
  Vcpkg version: 2018.11.23-unknownhash

Additionally, attach any relevant sections from the log files above.
```

**Build log:**
```
...
[...]

PR #19374, opened 2021-08-05
[libconfig] build on macos

jamesathey:
**Describe the pull request**

- #### What does your PR fix?  
Fixes compilation errors on macOS by making the compilation flags better match what the upstream libconfig CMakeLists.txt is doing. In particular, the line `find_path(STDINT_H stdint.h)` was picking up the kernel header version of `stdint.h` on macOS, altering the header search paths and preventing libconfig from compiling against the C standard library.

Also updates the library to the latest patch release (1.7.2 -> 1.7.3).

This PR addresses the build failure on macOS, so this PR also removes the "fail" line from `ci.baseline.txt`.

Yes.

Yes.

### DIFF
--- a/ports/libconfig/CMakeLists.txt
+++ b/ports/libconfig/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.5.1)
 project(libconfig C CXX)
 
 if(MSVC)
-  add_compile_options(/W3 /wd4005 /wd4996 /wd4018 -D_CRT_SECURE_NO_WARNINGS)
+  add_compile_options(/W3 /wd4005 /wd4996 /wd4018 -D_CRT_SECURE_NO_WARNINGS -DYY_NO_UNISTD_H -DYY_USE_CONST)
 endif()
 
 set(C_SOURCES
@@ -20,11 +20,9 @@ set(CPP_SOURCES
   lib/libconfigcpp.cc
 )
 
-find_path(STDINT_H stdint.h)
+set(CMAKE_C_STANDARD 99)
 
 include_directories(lib ${STDINT_H})
-
-add_definitions(-DYY_NO_UNISTD_H -DYY_USE_CONST)
 
 add_library(libconfig  ${C_SOURCES})
 add_library(libconfig++ ${CPP_SOURCES})
--- a/ports/libconfig/CONTROL
+++ b/ports/libconfig/CONTROL
@@ -1,5 +1,5 @@
 Source: libconfig
-Version: 1.7.2
-Port-Version: 4
+Version: 1.7.3
+Port-Version: 0
 Homepage: https://github.com/hyperrealm/libconfig
 Description: C/C++ library for processing configuration files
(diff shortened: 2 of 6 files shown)


ports/libconfig/CONTROL
```diff
@@ -1,5 +1,5 @@
 Source: libconfig
-Version: 1.7.2
-Port-Version: 4
+Version: 1.7.3
+Port-Version: 5
```
JackBoosY, line 3:
```suggestion
```
Since you updated the version, we should reset `port-version`.

ports/libconfig/CMakeLists.txt
```diff
@@ -20,12 +20,10 @@ set(CPP_SOURCES
   lib/libconfigcpp.cc
 )
 
-find_path(STDINT_H stdint.h)
+set(CMAKE_C_STANDARD 99)
 
 include_directories(lib ${STDINT_H})
 
-add_definitions(-DYY_NO_UNISTD_H -DYY_USE_CONST)
```
JackBoosY, line 27: Any reason to do that?

ports/libconfig/CMakeLists.txt
```diff
@@ -20,12 +20,10 @@ set(CPP_SOURCES
   lib/libconfigcpp.cc
 )
 
-find_path(STDINT_H stdint.h)
+set(CMAKE_C_STANDARD 99)
```
JackBoosY, line 23: Why remove finding stdint.h ? And is libconfig only support c99 standard?

JackBoosY (changes requested):

ports/libconfig/CMakeLists.txt
jamesathey: The [upstream libconfig CMakeLists.txt](https://github.com/hyperrealm/libconfig/blob/master/lib/CMakeLists.txt) only applies those two definitions to MSVC builds.

```cmake
if(MSVC)
    target_compile_definitions(${libname}
        PRIVATE
            _CRT_SECURE_NO_DEPRECATE
            YY_NO_UNISTD_H
            YY_USE_CONST )

    target_compile_definitions(${libname}++
        PRIVATE
            _CRT_SECURE_NO_DEPRECATE
            YY_NO_UNISTD_H
            YY_USE_CONST )
endif()
```

ports/libconfig/CMakeLists.txt
jamesathey: Compilers don't need help finding `stdint.h`. 😄 By including the `find_path() call`, it was forcing CMake to find another folder with that header in it, and on macOS / clang, it finds the kernel headers path. From there, all `#include` lines were picking up kernel header versions instead of the C standard library versions, which then broke the compile.

The `set(CMAKE_C_STANDARD 99)` line may not be necessary, although the [upstream libconfig CMakeLists.txt](https://github.com/hyperrealm/libconfig/blob/master/CMakeLists.txt#L16) sets it for all platforms. I know that's meaningless on MSVC, which doesn't support full C99, but I figured making the build match upstream was saner.

ports/libconfig/CONTROL
jamesathey: Agreed. Fixed.

JackBoosY:
I think it's good. Please click `ready for review`.

JackBoosY (approved):

jamesathey:
If #5518 were still open and not closed wont-fix, this PR would fix it.

PR #19429, opened 2021-08-07
Fix the false warning on missing pdbs

hellozee:
Couldn't find from where `unmatched_dlls_length` was adjusted and it seemingly throwing wrong warnings for missing pdbs. Might be a hack from my side though.


dg0yt:
Duplicates #19369.

hellozee:
Ahh my bad should have checked before.

PR #19430, opened 2021-08-07
[Phonon4qt5] new port

hellozee:
- #### What does your PR fix?  
  Adds phonon4qt5 port.

  all, No

  Yes

  Yes

### DIFF
--- a/ports/phonon4qt5/portfile.cmake
+++ b/ports/phonon4qt5/portfile.cmake
@@ -1,3 +1,5 @@
+vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
+
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO KDE/phonon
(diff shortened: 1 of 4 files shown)


ports/phonon4qt5/portfile.cmake
```diff
@@ -0,0 +1,29 @@
+vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
```
NancyLi1013, line 1: Could you please make a simple clarification why this port doesn't support static?

wrobelda: @hellozee you don't need to disable static build, just follow the suggestions from the CI: https://dev.azure.com/vcpkg/public/_build/results?buildId=57390&view=logs&j=821e2c31-c252-5a5c-eeb9-89df78f2f610&t=34ac2787-4652-59a0-fda0-8b82a81de66d&l=1983

hellozee: The shared library thing is [hard coded](https://github.com/KDE/phonon/blob/master/phonon/CMakeLists.txt#L87) and I think that is the way it is intended to be used. Though could be wrong, I would ask the authors whenever I can make some time, 😓.

PR #19439, opened 2021-08-07
[kf5coreaddons] fix Windows static builds & CMake Config

wrobelda:
- #### What does your PR fix?  
  https://invent.kde.org/frameworks/kcoreaddons/-/merge_requests/122

  Also:
   - replace deprecated functions
   - embrace paths in quotes

  No change

  Yes

  Yes


ports/kf5coreaddons/portfile.cmake
```diff
@@ -3,41 +3,36 @@ vcpkg_from_github(
...
 
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/bin/data)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/bin/data)
-file(REMOVE ${CURRENT_PACKAGES_DIR}/debug/bin/desktoptojson${EXECUTABLE_SUFFIX})
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/etc)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/etc)
+file(REMOVE "${CURRENT_PACKAGES_DIR}/debug/bin/desktoptojson${EXECUTABLE_SUFFIX}")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/data")
```
NancyLi1013, line 30: Why do you only remove `desktoptojson `and `data` in debug type?

wrobelda: I don't quite understand. I also remove the `debug/include` and `debug/share`? The `debug/bin/desktoptojson` line is a leftover, however, should have deleted it after adding `vcpkg_copy_tools`.

wrobelda: @NancyLi1013

NancyLi1013: Sorry, I forgot to reply this. I mean why do you only handle these files in debug type? 
I don't what files are listed in data, etc. But we should handle them in both debug and release.

For `debug/include` and `debug/share`, vcpkg only needs to keep `include/` and `share/`. So we need to remove `debug/include` and `debug/share`. But other files listed in debug and release folds should be handled at the same time.

For desktoptojson, seems like a executable file, we might use `vcpkg_copy_tools` to handle this.

wrobelda: @NancyLi1013 I am _not_ removing anything else now than:
```
```

The exec file is already being handled by `vcpkg_copy_tools`, too.

wrobelda: pinging @NancyLi1013

NancyLi1013: Oh, sorry for the late. I forgot to submit my comments. It's my mistake.

wrobelda: `/data/` contains MIME definitions and the locale files, as well some other helper files. `/debug/data` would effectively duplicate that.

NancyLi1013 (approved):

ports/kf5coreaddons/portfile.cmake
dg0yt: @wrobelda Do you have an idea why this data is installed to `/data` at all? I strongly feel like this belongs to `/share`. And maybe this could be achieved by configuration options?
In the context of vcpkg, I would even move `/etc` to `/share` until there is a filesystem layout spec which says different.

wrobelda: @dg0yt good question. I checked debian's package and all these files are stored in `/share`. I don't see anything in the code that would change the location of these files. I'll ask around.

wrobelda: OK, this is most likely happening because of what some dependency `kf5` libs do. Will get that fixed.

wrobelda: Fixed now.

NancyLi1013 (approved):

wrobelda:
@BillyONeal if you don't mind looking at this one as well, since this is a dependency of many of my other PRs. I'd greatly appreciate it!

PR #19441, opened 2021-08-08
[icu] fixes for macOS dynamic lib install

adam-bloom:
**Describe the pull request**

- #### What does your PR fix?  
  1. Fixes the install step of macOS dynamic libraries. This was previously split into separate build/fix rpath/install phases. The install phase broke since the split was introduced. This switches back to using `vcpkg_install_make` after the fix rpath phase, which appears to work just fine (only the install is executed, the existing fixed build is used). The installed libraries had the correct rpath at the end.
  2. I added two safety guards for ensuring that there are files to remove before attempting to remove the files. A guard like this was already present for `DUMMY_DLLS` at the end of the port file, it was just missing in the other two places. This prevents errors like the following from occurring. This error, of course, was in fact due to issue 1, but still seems like guards here are a good idea.
 ```
CMake Error at ports/icu/portfile.cmake:160 (file):
  file must be called with at least two arguments.
```

No changes

Yes

Yes

### CODE REVIEW

File: ports/icu/portfile.cmake [hunk 1]
```diff
@@ -143,7 +136,9 @@ if(VCPKG_TARGET_IS_MINGW)
         ${CURRENT_PACKAGES_DIR}/debug/bin/*${VCPKG_HOST_EXECUTABLE_SUFFIX}
         ${CURRENT_PACKAGES_DIR}/bin/icu-config
         ${CURRENT_PACKAGES_DIR}/debug/bin/icu-config)
-    file(REMOVE ${ICU_TOOLS})
+    if(ICU_TOOLS)
+        file(REMOVE ${ICU_TOOLS})
+    endif()
```
dg0yt, line 141: AFAICT this port won't install tools to `bin` for mingw, so it needs a different change.
I would prefer this PR wouldn't mix macos and mingw fixes. This port really needs an update for mingw support, for native and for cross builds, and there is already one PR pending.

ports/icu/portfile.cmake
```diff
@@ -125,14 +125,7 @@ if(VCPKG_TARGET_IS_OSX AND VCPKG_LIBRARY_LINKAGE STREQUAL "dynamic" AND (NOT DEF
...
     )
-
-    # make install
-    vcpkg_execute_build_process(
-        COMMAND bash --noprofile --norc -c "make install"
-        WORKING_DIRECTORY "${CURRENT_BUILDTREES_DIR}/${RELEASE_TRIPLET}"
-        LOGNAME "make-install-${RELEASE_TRIPLET}")
-    message(STATUS "Package ${RELEASE_TRIPLET} done")
-
+    vcpkg_install_make()
```
dg0yt, line 128: It looks like `vcpkg_install_make()` should be moved out of the `if` and `else` branches now.

ports/icu/portfile.cmake
adam-bloom: I've removed this.

ports/icu/portfile.cmake
adam-bloom: Yes, it indeed does! Done.

JonLiu1993 (approved):

adam-bloom:
Hi @PhoebeHui,
Any expectations for when this might be merged?

BillyONeal (approved):

ras0219-msft:
This looks like a simple, straightforward change. Thanks for the fix!

PR #19453, opened 2021-08-09
[vcpkg baseline][rbdl-orb/rbdl/ptex] file conflicts, skip rbdl-orb in CI testing

PhoebeHui:
Related to https://github.com/microsoft/vcpkg/pull/19369

Changes:
1. rbdl-orb and rbdl are conflicts, from https://github.com/ORB-HD/rbdl-orb, the two main differences to the original [rbdl](https://github.com/rbdl/rbdl) is that this version has error handling and uses polymorphism for constraints. It tends to be conflict, instead of depending on each other. 

2. ptex: Add DISABLE_PARALLEL_CONFIGURE to ptex to disable parallel.

Failures:
rbdl-orb:
```
The following files are already installed in D:/installed/x86-windows and are in conflict with rbdl-orb:x86-windows

Installed by rbdl:x86-windows
    debug/lib/pkgconfig/rbdl.pc
    debug/lib/rbdl.lib
    include/rbdl/Body.h
    include/rbdl/Constraints.h
    include/rbdl/Dynamics.h
    include/rbdl/Joint.h
    include/rbdl/Kinematics.h
    include/rbdl/Logging.h
    include/rbdl/Model.h
    include/rbdl/Quaternion.h
    include/rbdl/SpatialAlgebraOperators.h
    include/rbdl/compileassert.h
    include/rbdl/rbdl.h
    include/rbdl/rbdl_config.h
    include/rbdl/rbdl_eigenmath.h
    include/rbdl/rbdl_math.h
    include/rbdl/rbdl_mathutils.h
    include/rbdl/rbdl_utils.h
    lib/pkgconfig/rbdl.pc
    lib/rbdl.lib
```

ptex:
```
CMake Error at src/ptex/CMakeLists.txt:7 (configure_file):
  File exists
```
cc @ju6ge

### DIFF
--- a/ports/rbdl-orb/portfile.cmake
+++ b/ports/rbdl-orb/portfile.cmake
@@ -1,12 +1,16 @@
-string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" RBDL_STATIC)	
-	
+if (EXISTS ${CURRENT_INSTALLED_DIR}/share/rbdl/copyright)
+    message(FATAL_ERROR "${PORT} conflict with rbdl, please remove rbdl before install ${PORT}.")
+endif()
+
+string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" RBDL_STATIC)
+
 vcpkg_from_github(
   OUT_SOURCE_PATH SOURCE_PATH
   REPO ORB-HD/rbdl-orb
   REF b22abab856a90dbc62e6b2e79f148bd383b5ce43
   SHA512 744a60145243454a9d148971d998ae7a3cc5b9d66131b5d6f3c7be80d6c9ef8b8bf4390b9d1b90b14be6c619c2e1d14c7c6104b3ca6e606e22e3581b548e4f9d	
   HEAD_REF master
-)	
+)
 
 vcpkg_from_github(
   OUT_SOURCE_PATH PARSER_SOURCE_PATH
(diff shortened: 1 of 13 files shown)


JackBoosY (approved):

ports/rbdl-orb/portfile.cmake
```diff
@@ -1,12 +1,16 @@
-string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" RBDL_STATIC)	
-	
+if (EXISTS ${CURRENT_INSTALLED_DIR}/share/rbdl/copyright)
```
dg0yt, line 1: Double quotes are missing here.

dg0yt (commented):
:+1:

ports/rbdl-orb/portfile.cmake
PhoebeHui: Thanks for your review!

JackBoosY (approved):

ju6ge:
@PhoebeHui jup that is correct, to give a little context: Our research group has been writing patches and fixing bugs for rbdl. rbdl was originally developed by Martin Felis at our research group, after he left the original version has not changed very much. Since rbdl is heavily used in our group with time we have had the need to update rbdl with new features, removing bugs and in general making rbdl easier to use and less error prone. Sadly trying to upstream these features has lead no where. This means the two version "conflict" in the sense that they can't be installed at the same time, but anything that works with the original rbdl should also be able to just use 'rbdl-orb' without needing to change the code much. Also unlike the port of `rbdl`, `rbdl-orb` does contain some of the addons that are also present in the original version but for some reason are not included in the port. The addons enable some very important features such as reading in a model from urdf or lua files, which is important if you want to write code that can be used with multiple different models. Hopefully at some future point we will be able to have a single version of rbdl again, but at the moment this seems unlikely.

Anyway marking the two libraries as conflicting seems like the correct move here.

Kind regards,
ju6ge

PhoebeHui:
@ju6ge, thanks for confirming this!

PR #19462, opened 2021-08-09
bugfix unmatched dlls in vcpkg_copy_pdbs

silverqx:
Fix `if` condition to check unmatched `dlls`.

Currently, the `unmatched_dlls_length` variable is used, but this variable is not defined anywhere, it should be `dlls_without_matching_pdbs`.

- #### What does your PR fix?  
Currently, if you use `vcpkg_copy_pdbs()`, then following [condition](https://github.com/microsoft/vcpkg/blob/5304f826b5736eea0aa4749ce49c84539badaf4a/scripts/cmake/vcpkg_copy_pdbs.cmake#L68) is always `true` because `unmatched_dlls_length` doesn't exist and because of that the vcpkg output contains this error message everytime:
```
CMake Warning at scripts/cmake/vcpkg_copy_pdbs.cmake:70 (message):
  Could not find a matching pdb file for:


  ports/hello-world-library/portfile.cmake:57 (vcpkg_copy_pdbs)
```

### DIFF
--- a/scripts/cmake/vcpkg_copy_pdbs.cmake
+++ b/scripts/cmake/vcpkg_copy_pdbs.cmake
@@ -65,7 +65,7 @@ function(vcpkg_copy_pdbs)
 
         set(ENV{VSLANG} "${vslang_backup}")
 
-        if(NOT unmatched_dlls_length STREQUAL "")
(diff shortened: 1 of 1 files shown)


cenit:
duplicate of #19369 ?

PR #19464, opened 2021-08-09
[openal-soft] enable coreaudio backend for ios

edo9300:
**Describe the pull request**

- #### What does your PR fix?  
  Currently if openal-soft is built for an ios target, it will be built with no audio backend. Now apple's coreaudio backend will be enabled when building for both ios and mac os targets.


ports/openal-soft/portfile.cmake
```diff
@@ -27,7 +27,10 @@ if(VCPKG_TARGET_IS_WINDOWS)
     set(ALSOFT_REQUIRE_WINDOWS ON)
 endif()
 if(VCPKG_TARGET_IS_OSX)
-    set(ALSOFT_REQUIRE_OSX ON)
+    set(ALSOFT_REQUIRE_APPLE ON)
+endif()
+if(VCPKG_TARGET_IS_IOS)
+    set(ALSOFT_REQUIRE_APPLE ON)
 endif()
```
NancyLi1013, line 34: Seems `ALSOFT_REQUIRE_OSX `and `ALSOFT_REQUIRE_APPLE ` do the same thing here. Both of them is used to judge if `ALSOFT_BACKEND_COREAUDIO `is enabled or not. So we can use one of them. 

For the condition, we might put the targets together.
```
if(VCPKG_TARGET_IS_OSX OR VCPKG_TARGET_IS_IOS)
        set(ALSOFT_REQUIRE_OSX ON)
endif()
```

NancyLi1013:
Hi @edo9300 
Thanks for your PR!

```
versions/o-/openal-soft.json(2,): error : While reading versions for port openal-soft from file: version/o-/openal-soft.json
Version `1.21.1#2` was not found in versions file.
Run:

    vcpkg x-add-version openal-soft

to add the new port version.
```
You also need to run `./vcpkg x-add-version openal-soft` to update the version files.

ports/openal-soft/portfile.cmake
edo9300: I didn't notice there was still ``ALSOFT_REQUIRE_OSX`` being used, i removed it and replaced completely with ``ALSOFT_REQUIRE_APPLE`` (as it makes more sense since it applies to apple products and not only osx). Also i put both the checks together.

edo9300:
> You also need to run `./vcpkg x-add-version openal-soft` to update the version files.

I forgot to run the command, my bad

NancyLi1013 (approved):

NancyLi1013:
LGTM, thanks for your PR @edo9300.

PR #19500, opened 2021-08-11
[icu] Fix removing tests with alternative lib dir name

nicolasfella:
On some Linux distros (e.g. OpenSUSE) lib64 is used instead of lib for
libraries.

Take that into account when removing the tests, otherwise the tests are
not removed and the file(REMOVE) call fails because the variable is
empty

### DIFF
--- a/ports/icu/portfile.cmake
+++ b/ports/icu/portfile.cmake
@@ -150,11 +150,13 @@ file(REMOVE_RECURSE
     ${CURRENT_PACKAGES_DIR}/share
     ${CURRENT_PACKAGES_DIR}/debug/share
     ${CURRENT_PACKAGES_DIR}/lib/icu
+    ${CURRENT_PACKAGES_DIR}/lib64/icu
     ${CURRENT_PACKAGES_DIR}/debug/lib/icu
     ${CURRENT_PACKAGES_DIR}/debug/lib/icud)
 
 file(GLOB TEST_LIBS
     ${CURRENT_PACKAGES_DIR}/lib/*test*
+    ${CURRENT_PACKAGES_DIR}/lib64/*test*
     ${CURRENT_PACKAGES_DIR}/debug/lib/*test*)
 file(REMOVE ${TEST_LIBS})
 


nicolasfella:
the lib64 problem bites me in other places as well, so maybe it makes more sense to force icu to install to lib instead of lib64 somehow

JackBoosY (changes requested):

ports/icu/portfile.cmake
```diff
@@ -150,11 +150,13 @@ file(REMOVE_RECURSE
     ${CURRENT_PACKAGES_DIR}/share
     ${CURRENT_PACKAGES_DIR}/debug/share
     ${CURRENT_PACKAGES_DIR}/lib/icu
+    ${CURRENT_PACKAGES_DIR}/lib64/icu
```
JackBoosY, line 153: We do need to consider this because many ports on Linux will install libraries to `lib64` when their machine is x64.

dg0yt: Bot how does the `lib64` installation path get into vcpkg ports? I think it is necessary to suppress that. Portfiles are written with the assumption that `lib` is used exclusively. vcpkg even doesn't allow executables in `bin` although this *is* a standard.

JackBoosY: @dg0yt This is very common in *inx, and I think we need some discussion.

dg0yt: I'm discussing here... 
Until now, I don't have the impression that vcpkg cares much about what is common in *nix/POSIX. E.g.:
- `/bin` for executables.
- completely separate for debug vs. release.

JackBoosY: @BillyONeal what do you think about?

nicolasfella:
I think the better way to fix this is passing '--libdir=${prefix}/lib' to the configure args since the lib64 causes trouble in other places too

JackBoosY:
> I think the better way to fix this is passing '--libdir=${prefix}/lib' to the configure args since the lib64 causes trouble in other places too

If it works, we should set it into `vcpkg_configure_make`, but:
https://github.com/microsoft/vcpkg/blob/63aa65e65b9d2c08772ea15d25fb8fdb0d32e557/scripts/cmake/vcpkg_configure_make.cmake#L494

cc @Neumann-A

dg0yt:
I guess icu just use a different (or no) parameter for the test libs. In that case it needs another parameter or patch.

ports/icu/portfile.cmake
dg0yt: IIUC #19666 takes cares of the undesired `lib64` now, in a more general approach.

nicolasfella:
Closing in favor of https://github.com/microsoft/vcpkg/pull/19666

PR #19522, opened 2021-08-12
[boost-context] arm64 linux abi fix

siposcsaba89:
Fixes missing function definitions from libboost_context.a (for example make_fcontext) when crosscompiling to arm64 Linux.

I am not sure this is the best solution, but it solves the linker issues in my case.


PhoebeHui:
@siposcsaba89, thanks for the PR!

Could you add "port-version": 1 to https://github.com/microsoft/vcpkg/blob/master/ports/boost-context/vcpkg.json, submit the changes, and update the baseline revision, and then submit the changes again?

siposcsaba89:
> @siposcsaba89, thanks for the PR!
> 
> Could you add "port-version": 1 to https://github.com/microsoft/vcpkg/blob/master/ports/boost-context/vcpkg.json, submit the changes, and update the baseline revision, and then submit the changes again?

Hi @PhoebeHui ! Thanks for checking out, added "port-version" and updated the baseline revision. I hope I did it correctly :)

PhoebeHui (approved):

ras0219-msft:

PR #19557, opened 2021-08-13
[scripts-audit WIP] vcpkg_find_acquire_program

strega-nil:
PR-ing for CI results.


dg0yt:
Isn't there a more declarative way for platform alternatives in JSON then `if` `then` `else`? After all, vcpkg doesn't need these keywords for manifests, and the use case is quite similar.

strega-nil-ms:
@dg0yt it's actually fairly declarative, it's just that I used `$if` because it makes sense and is unlikely to clash with anything.

Neumann-A:
hmm what about introducing a port for everything `vcpkg-tool-<name>` and slowly migrate to that instead + make `vcpkg_find_acquire_program` search in the dirs given by `CURRENT_HOST_INSTALLED_DIR` + `vcpkg-tool-<name>`. (Instead of this JSON migration.)

strega-nil-ms:
@Neumann-A that's an even bigger migration and might be the best option given all the time in the world, but this needs to be done at a decent clip.

Neumann-A:
@strega-nil-ms: I just don't see how moving from the CMake definitions to the JSON definitions is an improvement. It just feels like making it even more difficult to see what is happening.

strega-nil-ms:
@Neumann-A it separates data from code, which I think is good. It also, at least for me, clears up any confusion about what variables are set when.

dg0yt:
> hmm what about introducing a port for everything `vcpkg-tool-<name>` and slowly migrate to that instead

That's what I thought when I looked at this PR. Ports can be dependencies in manifests. Port functions are not available to users. This is source of confusion for pkgconf, flex, bison etc.
Last not least, consumers would be decoupled from changes to unrelated tools.

Neumann-A:
> it separates data from code, which I think is good. It also, at least for me, clears up any confusion about what variables are set when.

Nothing against that. Just `include(<tooldetails>.cmake)` and prefix the variables with `<TOOLNAME>_(URL|HASH|....)`. It just feels more natural for me to stay in CMake for this instead of switching to JSON and adding the required code for parsing the JSON.

strega-nil-ms:
@Neumann-A I understand the opinion... I will try switching back to CMake before this is out of draft mode.

Hoikas:
I have to agree with the others - when I first read this JSON, I was not sure that I understood how exactly it was an improvement, considering that there are fields that are effectively cmake code instead of just declarations.

strega-nil-ms:
@Hoikas I think it's important that the data be pulled out from the actual code; whether that data is in JSON or in CMake, is not a big deal. I just wanted to be able to see what was going on.

PR #19574, opened 2021-08-14
[openxr-loader] Update to 1.0.18

### LINKED ISSUE #19573: [<port name>] update to <version>
Library name: OpenXR-loader

New version number: 1.0.18

Current version of 1.0.11 is exactly one year out of date as of today.

jherico:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #19573 

UWP and ARM are not supported.  This is unchanged from the previous version of the port. 

I believe the portfile may be using a deprecated function but I'm attempting to make this a minimum touch PR.

Yes.


ports/openxr-loader/002-fix-hpp-gen.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/scripts/hpp_genxr.py
+@@ -80,6 +80,11 @@ def genTarget(args):
+             "XR_MSFT_controller_model",
+             # Projection of static string fails
+             "XR_MSFT_spatial_graph_bridge",
++            "XR_MSFT_spatial_anchor_persistence",
++            "XR_MSFT_holographic_window_attachment",
++            # Projection of UuidMSFT fails
++            "XR_MSFT_scene_understanding",
++            "XR_MSFT_scene_understanding_serialization",
```
JackBoosY, line 13: Is this the upstream changes?

JackBoosY (changes requested):

ports/openxr-loader/002-fix-hpp-gen.patch
jherico: The OpenXR-hpp project generates C++ bindings, but the generation code isn't handling every case properly.  Excluding these extensions was required to get a version of the header that compiles.  This replaces the old patch method in the previous version where I patched the output headers directly.  That approach is significantly more difficult now because the C++ bindings have been broken up into many files.

JackBoosY: Is this caused by upstream? If so, please report to the upstream first.

jherico: I've already reported the corresponding issues here:  https://github.com/KhronosGroup/OpenXR-Hpp/issues/24 but they haven't gotten any traction.  I'm one of the two main contributors to the upstream repository and I haven't had time to try to resolve this bug yet, but I still need an updated version of the XR bindings from vcpkg.

JackBoosY (approved):

PR #19581, opened 2021-08-15
[comms,commsdsl,comms-ublox] update ports

mathisloge:
**Describe the pull request**

- #### What does your PR fix?  

* updates comms to 3.4
* adds feature tools to comms. 
* update commsdsl to 3.6.4
* update comms-ublox to 0.20.2

yes

yes

yes


mathisloge:
\ping https://github.com/commschamp/comms_champion/issues/13

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,60 @@
...
 
+vcpkg_check_features(
+    OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        tools   COMMS_BUILD_TOOLS
+)
+
+if(COMMS_BUILD_TOOLS)
+    set(VCPKG_POLICY_DLLS_WITHOUT_LIBS enabled)
+endif()
```
JackBoosY, line 17: Why do that?

JackBoosY (changes requested):

ports/comms/portfile.cmake
mathisloge: good catch. Forgot to remove that.

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,56 @@
...
+    file(COPY "${CURRENT_PACKAGES_DIR}/lib/CommsChampion/plugin" DESTINATION "${CURRENT_PACKAGES_DIR}/tools/${PORT}/lib/CommsChampion/")
+endif()
+
+vcpkg_fixup_cmake_targets(CONFIG_PATH "lib/LibComms/cmake" TARGET_PATH "share/LibComms")
+
+if(COMMS_BUILD_TOOLS)
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/CommsChampion/plugin")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/LibComms")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/CommsChampion/plugin")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/LibComms")
```
JackBoosY, line 46: Can you confirm which file was installed in those directories?

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,56 @@
...
-file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
-file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug")
+
+if(COMMS_BUILD_TOOLS)
+    vcpkg_copy_tools(
+        TOOL_NAMES cc_dump cc_view
+        SEARCH_DIR "${CURRENT_PACKAGES_DIR}/bin"
+        AUTO_CLEAN
+    )
+    file(COPY "${CURRENT_PACKAGES_DIR}/lib/CommsChampion/plugin" DESTINATION "${CURRENT_PACKAGES_DIR}/tools/${PORT}/lib/CommsChampion/")
```
JackBoosY, line 37: Maybe `RENAME` would be better.

JackBoosY (changes requested):

ports/comms/portfile.cmake
mathisloge: with the fix now, all dirs are empty (except plugins which are installed now)

ports/comms/portfile.cmake
mathisloge: Rename gives some strange errors that the target dir does not exists

JackBoosY:
Depends on https://github.com/microsoft/vcpkg/pull/19629.

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,57 @@
...
+
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/LibComms")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/CommsChampion")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/LibComms")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/CommsChampion")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+else()
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
+    file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug")
```
JackBoosY, line 52: Why did you remove that? Is this port a header only library?

JackBoosY (changes requested):

ports/comms/portfile.cmake
mathisloge: Yes

JackBoosY (approved):

JackBoosY:
When building `comms[core,tools]:x64-windows-static`:
```
-- Installing: G:/19581/vcpkg/packages/comms_x64-windows-static/share/comms/copyright

    G:/19581/vcpkg/packages/comms_x64-windows-static/bin/comms_champion.dll
    G:/19581/vcpkg/packages/comms_x64-windows-static/debug/bin/comms_champion.dll

There should be no bin\ directory in a static build, but G:\19581\vcpkg\packages\comms_x64-windows-static\bin is present.
There should be no debug\bin\ directory in a static build, but G:\19581\vcpkg\packages\comms_x64-windows-static\debug/bin is present.

    endif()

    G:\19581\vcpkg\ports\comms\portfile.cmake
Error: Building package comms:x64-windows-static failed with: POST_BUILD_CHECKS_FAILED
```

mathisloge:
@JackBoosY Is there a way to add `supports` fields to a feature property? 
The tool itself builds a shared library (hardcoded) which is needed for the build plugins. 
So the feature tools can only be build if the triplet is dynamic, but the port without the feature tools supports all triplets, since it is header only

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,62 @@
...
+# check before configure
+if(COMMS_BUILD_TOOLS)
+    vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY) 
+endif()
+
 vcpkg_configure_cmake(
-    SOURCE_PATH ${SOURCE_PATH}
+    SOURCE_PATH "${SOURCE_PATH}"
     PREFER_NINJA
     OPTIONS
```
JackBoosY, line 23:
```suggestion
    OPTIONS ${FEATURE_OPTIONS}
```

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,62 @@
...
+vcpkg_check_features(
+    OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        tools   COMMS_BUILD_TOOLS
+)
+
+# check before configure
+if(COMMS_BUILD_TOOLS)
+    vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY) 
+endif()
```
JackBoosY, line 18: I don't think we can deprive users of the power to build static, even with feature tools.
Please replace `vcpkg_check_linkage` with `message(FATAL_ERROR)` here.

JackBoosY (changes requested):

dg0yt:
Is the DLL only used by the tools (= a private implementation detail)?
If yes, can't the DLL and its import library be removed from everything but `tools/${PORT}`?

mathisloge:
> Is the DLL only used by the tools (= a private implementation detail)?
> If yes, can't the DLL and its import library be removed from everything but `tools/${PORT}`?

It's not a private implementation detail. It can be used to develop own plugins for the tool application.

ports/comms/portfile.cmake
mathisloge: not applicable in this case. There is not a `COMMS_BUILD_TOOLS` in the upstream cmake. 
The tools feature depends on multiple variables. 
```
-DCC_BUILD_TOOLS_LIBRARY=${COMMS_BUILD_TOOLS}
-DCC_INSTALL_TOOLS_LIBRARY=${COMMS_BUILD_TOOLS}
-DCC_BUILD_TOOLS=${COMMS_BUILD_TOOLS}
-DCC_INSTALL_TOOLS=${COMMS_BUILD_TOOLS}
-DCC_BUILD_DEMO_PROTOCOL=OFF
```

dg0yt: But you can set multiple variables from the same feature, even inverted ones.

mathisloge: Ah thanks for the hint. wasn't aware of it

mathisloge:
@JackBoosY please review again. I've also replaced the deprecated cmake functions with the new one

ports/comms/portfile.cmake
mathisloge: Ah thanks! wasn't aware of it

ports/comms/portfile.cmake
```diff
@@ -1,26 +1,61 @@
...
+    OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        tools   CC_BUILD_TOOLS_LIBRARY
+        tools   CC_INSTALL_TOOLS_LIBRARY
+        tools   CC_BUILD_TOOLS
+        tools   CC_INSTALL_TOOLS
+)
+
+# check before configure
+if(CC_BUILD_TOOLS)
```
JackBoosY, line 19:
```suggestion
if("tools" IN_LIST FEATURES)
```

JackBoosY (changes requested):

JackBoosY (changes requested):

JackBoosY (approved):

NancyLi1013 (changes requested):

NancyLi1013 (approved):

ras0219-msft:
This all looks clean and reasonable. Thanks for the update!

PR #19599, opened 2021-08-16
[Help needed][brotli] Remove PowerShell dependency when compile in mingw ports

talregev:
~~Update deprecated function~~
https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#avoid-deprecated-helper-functions

> 4. vcpkg_copy_tool_dependencies() should be replaced by vcpkg_copy_tools()

I want to remove PowerShell dependency when compile in mingw.
It using PowerShell to copy a file. I think It overkill. 
I will happy if someone can help me to solve this.

### DIFF
--- a/ports/brotli/CONTROL
+++ b/ports/brotli/CONTROL
@@ -1,5 +1,5 @@
 Source: brotli
 Version: 1.0.9
-Port-Version: 1
+Port-Version: 2
 Homepage: https://github.com/google/brotli
 Description: a generic-purpose lossless compression algorithm that compresses data using a combination of a modern variant of the LZ77 algorithm, Huffman coding and 2nd order context modeling.
--- a/ports/brotli/portfile.cmake
+++ b/ports/brotli/portfile.cmake
@@ -21,7 +21,7 @@ vcpkg_install_cmake()
 
 vcpkg_copy_pdbs()
 
-vcpkg_copy_tool_dependencies(${CURRENT_PACKAGES_DIR}/tools/brotli)
+vcpkg_copy_tools(${CURRENT_PACKAGES_DIR}/tools/brotli)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-brotli TARGET_PATH share/unofficial-brotli)
 vcpkg_fixup_pkgconfig()
 
--- a/ports/brotli/portfile.cmake
+++ b/ports/brotli/portfile.cmake
@@ -22,6 +22,10 @@ vcpkg_install_cmake()
 vcpkg_copy_pdbs()
 
 vcpkg_copy_tools(TOOL_NAMES ../tools/brotli/brotli)
+vcpkg_copy_tools(
+    TOOL_NAMES brotli
+    SEARCH_DIR "${CURRENT_PACKAGES_DIR}/tools/brotli/"
+)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-brotli TARGET_PATH share/unofficial-brotli)
 vcpkg_fixup_pkgconfig()
 
--- a/ports/brotli/portfile.cmake
+++ b/ports/brotli/portfile.cmake
@@ -21,7 +21,7 @@ vcpkg_install_cmake()
 
 vcpkg_copy_pdbs()
 
-vcpkg_copy_tools(TOOL_NAMES ${CURRENT_PACKAGES_DIR}/tools/brotli/brotli)
+vcpkg_copy_tools(TOOL_NAMES ../tools/brotli/brotli)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-brotli TARGET_PATH share/unofficial-brotli)
 vcpkg_fixup_pkgconfig()
 
--- a/ports/brotli/portfile.cmake
+++ b/ports/brotli/portfile.cmake
@@ -21,7 +21,7 @@ vcpkg_install_cmake()
 
 vcpkg_copy_pdbs()
 
-vcpkg_copy_tools(TOOL_NAMES ${CURRENT_PACKAGES_DIR}/tools/brotli)
+vcpkg_copy_tools(TOOL_NAMES ${CURRENT_PACKAGES_DIR}/tools/brotli/brotli)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-brotli TARGET_PATH share/unofficial-brotli)
 vcpkg_fixup_pkgconfig()
 
--- a/ports/brotli/portfile.cmake
+++ b/ports/brotli/portfile.cmake
@@ -21,7 +21,7 @@ vcpkg_install_cmake()
 
 vcpkg_copy_pdbs()
 
-vcpkg_copy_tools(${CURRENT_PACKAGES_DIR}/tools/brotli)
+vcpkg_copy_tools(TOOL_NAMES ${CURRENT_PACKAGES_DIR}/tools/brotli)
 vcpkg_fixup_cmake_targets(CONFIG_PATH share/unofficial-brotli TARGET_PATH share/unofficial-brotli)
 vcpkg_fixup_pkgconfig()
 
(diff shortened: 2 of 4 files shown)


talregev:
It looking for this file now:
/home/tal/vcpkg/packages/brotli_x64-linux/bin/../tools/brotli/brotli
The problem there isn't a bin folder there:
image

JonLiu1993:
The gdal and paraview breaks due CI basline error,just wait CI error fixed.

talregev:
Also for mingw it using powershell to copy the tool. It seems to me overkill. Can we do it without powerShell?

JackBoosY:
We currently have to use powershell in Windows to call dumpbin through the ps1 script to analyze executable dependencies, so as to copy these dependent libraries to the executable directory.

JackBoosY:
I saw that the execution of powershell is restricted by `VCPKG_TARGET_IS_WINDOWS`, this issue should not happen again:
https://github.com/microsoft/vcpkg/blob/6b24f0fa76d171c5756f01f407682f7274cfe238/scripts/cmake/vcpkg_copy_tool_dependencies.cmake#L36-L49

JackBoosY:
@longnguyen2004 @cristianadam 
https://github.com/microsoft/vcpkg/blob/6b24f0fa76d171c5756f01f407682f7274cfe238/scripts/cmake/vcpkg_common_definitions.cmake#L52-L55
It seems like our community triplets `x64-mingw-*` only support Windows, right?
How about add *inx support?

talregev:
@JackBoosY 
mingw is support also on linux. The main goal is mingw is to compile binaries on linux.
Other ports are compile in mingw on linux.
For example curl and openssl. I can show you the log how they compile.

longnguyen2004:
Seems like we just need to add `AND CMAKE_HOST_WIN32` to `vcpkg_copy_tool_dependencies`, to prevent it from running under Linux hosts.

talregev:
> Seems like we just need to add `AND CMAKE_HOST_WIN32` to `vcpkg_copy_tool_dependencies`, to prevent it from running under Linux hosts.

Thank you.

talregev:
@longnguyen2004 After this change, it still copy the tool in mingw linux to the right place?

longnguyen2004:
> @longnguyen2004 After this change, it still copy the tool in mingw linux to the right place?

It'd copy the tools (`*.exe`) but not the dependencies (`*.dll`), since we rely on the PowerShell script for that.

talregev:
> > @longnguyen2004 After this change, it still copy the tool in mingw linux to the right place?
> 
> It'd copy the tools (`*.exe`) but not the dependencies (`*.dll`), since we rely on the PowerShell script for that.

Great. on mingw on linux there still `tools/brotli/brotli`. 
So it will copy it with linux bash command?

JackBoosY:
@longnguyen2004 I think we should step further, in my [comment](https://github.com/microsoft/vcpkg/pull/19599#issuecomment-900837553) we set the vcpkg target, so it will affect more code.

talregev:
On mingw on linux, and the linux port. there still a binary tool that need to copy.
How you going to handle that after the change?

longnguyen2004:
IIRC, that would be handled by `vcpkg_copy_tools`, not `vcpkg_copy_tools_dependencies`.

talregev:
> IIRC, that would be handled by `vcpkg_copy_tools`, not `vcpkg_copy_tools_dependencies`.

Thank you for the explanation!

talregev:
> Seems like we just need to add `AND CMAKE_HOST_WIN32` to `vcpkg_copy_tool_dependencies`, to prevent it from running under Linux hosts. So that would be `if (VCPKG_TARGET_IS_WINDOWS AND CMAKE_HOST_WIN32)`

Can you make the change?

dg0yt:
> I want to remove PowerShell dependency when compile in mingw.
> It using PowerShell to copy a file. I think It overkill.
> I will happy if someone can help me to solve this.

Please don't remove this unconditionally based on a single opinion. I do understand the wish to use vpckg without powershell on Linux. But I don't want different scripts to copy DLLs for mingw dynamic triplets. 
ATM the powershell scripts for EXE/DLL deployment don't use path separators in a portable way but this can be fixed. Then it will not only copy "a single file".

However, I would happily support migration to portable cmake scripts instead of powershell.

talregev:
> > I want to remove PowerShell dependency when compile in mingw.
> > It using PowerShell to copy a file. I think It overkill.
> > I will happy if someone can help me to solve this.
> 
> Please don't remove this unconditionally based on a single opinion. I do understand the wish to use vpckg without powershell on Linux. But I don't want different scripts to copy DLLs for mingw dynamic triplets.
> ATM the powershell scripts for EXE/DLL deployment don't use path separators in a portable way but this can be fixed. Then it will not only copy "a single file".
> 
> However, I would happily support migration to portable cmake scripts instead of powershell.

That acceptable. 
When will  you start work on this migration?

dg0yt:
> When will you start work on this migration?

Perhaps when the PR review experience gets a little smoother in particular with regard to explaining CMake behaviour...

But basically migrating from PS to cmake scripts should be a conscious decision. There is nothing broken by requiring Powershell, but there is a lot to break by replacing scripts. It is not just appdeploy.ps1 but also additional scripts in other ports. And there are not unit tests, only full rebuilds.

IMO it is not necessary to do this switch (at least at the moment). Powershell is an excellent tool. The only trouble is not used very often outside Windows, and there is already a huge number of scripting languages in use in vcpkg. With CMake, vcpkg already has a cross-platform language, and it even includes parts of the features of appdeploy.ps1.

longnguyen2004:
So, should we move on with removing this functionality on Linux hosts, or just leave it there as is?

JackBoosY:
@longnguyen2004  How about change here according with HOST_WIN:
https://github.com/microsoft/vcpkg/blob/6b24f0fa76d171c5756f01f407682f7274cfe238/scripts/cmake/vcpkg_common_definitions.cmake#L52-L55

longnguyen2004:
No, that part is correct as it is.

JackBoosY:
@longnguyen2004  Oh yes it is. Please ping me in your PR so I can review it.

dg0yt:
> So, should we move on with removing this functionality on Linux hosts, or just leave it there as is?

IMO the DLL copying must be fixed, not made impossible, for mingw on Linux hosts: If a binary package was built on Linux using `x64-mingw-dynamic` triplet, it must be immediately usable on a x64 Windows host, just as if it was compiled there (or as if compiled with x64-windows).

cristianadam:
CMake starting with version 3.16 has [file(GET_RUNTIME_DEPENDENCIES)](https://cmake.org/cmake/help/latest/command/file.html?#get-runtime-dependencies) which could be used instead of relying on Powershell.

The less dependencies the better.

dg0yt:
> CMake starting with version 3.16 has [file(GET_RUNTIME_DEPENDENCIES)](https://cmake.org/cmake/help/latest/command/file.html?#get-runtime-dependencies) which could be used instead of relying on Powershell.
> 
> The less dependencies the better.

Yes, I raised this already in another discussion. And even before that, I successfully used `BundleUtilities` across all platforms. 
A common difficulty is handling plugins, e.g. for Qt.

Neumann-A:
> A common difficulty is handling plugins, e.g. for Qt.

run `<platform>deployqt` if available or just install all plugins which are available at that point.

dg0yt:
> just install all plugins which are available at that point.

I guess this what is done by qtdeploy.ps1. (?)

Neumann-A:
No it tries to be clever and look what qt libraries the target depends upon. I didn't write that script so I don't know all the details

dg0yt:
> No it tries to be clever and look what qt libraries the target depends upon. I didn't write that script so I don't know all the details.

That is why I'm concerned about porting to cmake. It is hard to understand all the details and past decisions.  And CI won't catch missing plugins, or transitive dependencies of plugins.

IIRC another aspect is that the appdeploy script must run from msbuild, and I can't even test this locally.

JackBoosY:
In a favor of https://github.com/microsoft/vcpkg/pull/19631.

PR #19600, opened 2021-08-16
[google-cloud-cpp] use features to stop package growth

coryan:
We have plans to grow `google-cloud-cpp`, but we think the build times
are large enough already. This creates "features" to at least stop the
package growth as new features are added. The "default-features"
reflects the current features in the package.

- #### What does your PR fix?  

N/A

No chage.

Yes.

Yes.


ports/google-cloud-cpp/portfile.cmake
```diff
@@ -12,11 +12,16 @@ vcpkg_from_github(
 
 vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/grpc")
 
+set(GOOGLE_CLOUD_CPP_ENABLE "${FEATURES}")
+list(REMOVE_ITEM GOOGLE_CLOUD_CPP_ENABLE "core")
+list(REMOVE_ITEM GOOGLE_CLOUD_CPP_ENABLE "googleapis")
```
JackBoosY, line 17: Please note that the feature name in vcpkg.json is not necessarily the same as the feature name in source.

JackBoosY (changes requested):

ports/google-cloud-cpp/portfile.cmake
coryan: I am one of the authors of `google-cloud-cpp` ([ref](https://github.com/googleapis/google-cloud-cpp/graphs/contributors)) and I am the primary maintainer of the `google-cloud-cpp` port.

Yes, I understand that the features in `vcpkg` and the project are not *necessarily* the same.  It is our (me and my team) *intention* to keep these in sync.  That will make both our life (as maintainers of both) easier, and will make life for users also easier (we hope).

JackBoosY (approved):

JackBoosY:
Tested all feature on `x86-windows` and `x64-windows-static` successfully.

ports/google-cloud-cpp/vcpkg.json
```diff
@@ -20,5 +21,115 @@
...
+  "features": {
+    "bigquery": {
+      "description": "The Google Cloud BigQuery C++ client library",
+      "dependencies": [
+        "abseil",
+        "grpc",
+        {
+          "name": "grpc",
+          "host": true
+        },
```
ras0219-msft, line 42: These are already listed in the core dependencies and so don't need to be duplicated. Is the intent to remove them from the core deps list?

coryan: Thanks for noticing that, and I am sorry it has taken me so many iterations to get this right.  Fixed, and rebased.  Tested each feature with x64-linux.

JackBoosY:

ras0219-msft:
Thanks!

PR #19607, opened 2021-08-17
[vcpkg_fixup_pkgconfig] HOTFIX -- revert #19469

### LINKED ISSUE #19606: [Curl] libcurl.pc in pkgconfig are broken for linux and android.
> **Describe the bug**
> A clear and concise description of what the bug is.

vcpkg install curl[core,openssl] on linux and android. create the lib with the libcurl.pc
libcurl.pc are broken
pkg-config libcurl.pc --libs -> errors:

> **Environment**

- OS: Linux, Android. (maybe more?)
- Compiler: for linux gcc 9.

> **To Reproduce**

1. vcpkg install curl[core,openssl]
2. pkg-config libcurl.pc --libs

> **Expected behavior**
> A clear and concise description of what you expected to happen.

That it give you the correct libs.

> **Failure logs** 
> -(please attached failure logs)

```
 pkg-config libcurl.pc --libs
Package " was not found in the pkg-config search path.
Perhaps you should add the directory containing `".pc'
Package '"', required by 'libcurl', not found
Package 'field', required by 'libcurl', not found
Package 'added', required by 'libcurl', not found
```

> **Additional context**
> Add any other context about the problem here.

ras0219:
PR #19469 regressed pkgconfig for libcurl. This was not detected because it also regressed the self-check mechanism (removed argument from `vcpkg_fixup_pkgconfig_check_files` without removing the argument from the callers).

Because there are a large number of changes in the helper and the internal consistency check was disabled, I'm reverting the entire change so it can go through a proper separate PR process again.

Fixes https://github.com/microsoft/vcpkg/issues/19606

### DIFF
--- a/scripts/cmake/vcpkg_fixup_pkgconfig.cmake
+++ b/scripts/cmake/vcpkg_fixup_pkgconfig.cmake
@@ -45,122 +45,115 @@ Still work in progress. If there are more cases which can be handled here feel f
 * [brotli](https://github.com/Microsoft/vcpkg/blob/master/ports/brotli/portfile.cmake)
 #]===]
 
-function(z_vcpkg_fixup_pkgconfig_check_files file config)
-    set(path_suffix_debug /debug)
-    set(path_suffix_release "")
+function(vcpkg_fixup_pkgconfig_check_files pkg_cfg_cmd _file _config)
+    set(PATH_SUFFIX_DEBUG /debug)
+    set(PATH_SUFFIX_RELEASE)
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Bypassing full CI check because this is a revert back to a known good state. I've manually reviewed the changes in https://github.com/microsoft/vcpkg/pull/19469 to ensure that this file can be independently rolled back.

dg0yt:
Thanks for the quick fix.
For the record, port gdal heavily depends on pkgconfig files, and I intentionally added configuration verification to the portfile because I thought this could be fragile. Now on the first true positive alert from the gdal alert bells, it was ignored although the CI logs provided clear hints. 😞

PR #19608, opened 2021-08-17
[vcpkg baseline] Re-fix lapack-reference to find blas on windows-static

### LINKED ISSUE #20952: [lapack-reference] Does not contain instructions for using cblas
**Describe the bug**
The lapack-reference port does not contain working instructions for use of the cblas feature.

**Environment**
- OS: Windows
- Compiler: Microsoft Visual Studio 2019 Professional (14.29.30133)

**To Reproduce**

I created a minimal CMake project for one of the official cblas examples.
https://gist.github.com/cgmb/e2ec5f91a109bad62d739e63c3c185e3

1. Download the sample project I linked above.
```
cmake -S. -Bbuild -DUSE_VCPKG_LAPACK=ON -DCMAKE_TOOLCHAIN_FILE=%VCPKG_PATH%/scripts/buildsystems/vcpkg.cmake -DVCPKG_PREFER_SYSTEM_LIBS=OFF -DVCPKG_TARGET_TRIPLET=x64-windows
cmake --build build
```

2. Observe Build Failure:
```
Microsoft (R) Build Engine version 16.11.1+3e40a09f8 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  Checking Build System
  Building Custom Rule C:/ws/vcpkg_cblas/CMakeLists.txt
  cblas_example.cpp
C:\ws\vcpkg_cblas\cblas_example.cpp(7,11): fatal error C1083: Cannot open include file: 'cblas.h': No such file or directory [C:\ws\vcpkg_cblas\build\ex.vcxproj]
```

**Expected behavior**
It should be possible to build this sample using the CMake targets that vcpkg suggests. At configure time, vcpkg states:
```
The package lapack-reference provides CMake targets:

    find_package(lapack CONFIG REQUIRED)
    target_link_libraries(main PRIVATE LAPACK::LAPACK)

    find_package(lapack-3.8.0 CONFIG REQUIRED)
    target_link_libraries(main PRIVATE blas lapack)
```

However, linking `LAPACK::LAPACK` is insufficient to find cblas. Replacing `find_package(lapack CONFIG REQUIRED)` with `find_package(lapack-3.8.0  CONFIG REQUIRED)` results in an even earlier failure (at configure time):

```
CMake Error at C:/Workspace/3rdparty/vcpkg/scripts/buildsystems/vcpkg.cmake:788 (_find_package):
  Could not find a package configuration file provided by "lapack-3.8.0" with
  any of the following names:

    lapack-3.8.0Config.cmake
    lapack-3.8.0-config.cmake

  Add the installation prefix of "lapack-3.8.0" to CMAKE_PREFIX_PATH or set
  "lapack-3.8.0_DIR" to a directory containing one of the above files.  If
  "lapack-3.8.0" provides a separate development package or SDK, be sure it
  has been installed.
```

**Failure logs** 
Using find_package(lapack REQUIRED): [lapack_failure_log.txt](https://github.com/microsoft/vcpkg/files/7403531/lapack_failure_log.txt)
Using find_package(lapack-3.8.0 CONFIG REQUIRED):  [lapack_380_failure_log.txt](https://github.com/microsoft/vcpkg/files/7403532/lapack_380_failure_log.txt)
This one is not a failure log, but it's a successful build using the upstream tarball of Netlib LAPACK manually built from source:  [lapack_success_log.txt](https://github.com/microsoft/vcpkg/files/7403535/lapack_success_log.txt)

JackBoosY:
The origin issue is:

openblas comes with lapack, lapack comes with blas, cblas, lapacke.
vcpkg only wants to build one lib of those in the ports, so lapack-reference should only build lapack, openblas should only build openblas. Due to the gfortran constrains lapack needs to build its internal blas in x64-windows-static.

**The problem:** FindBlas.cmake picks up a spurious openblas installation of vcpkg in one port since it was installed while another port was looking for it but that port did not declare the dependency on openblas -> error in colmap since openblas is not restored by CI.

This PR will
1. Add port `blas-reference` and disable `lapack-reference` to build internal `blas`.
2. Add `FindBLAS.cmake` provided by cmake and modify some codes to adapt to Windows or x64 build.
3. Add `vcpkg-cmake-wrapper` to `blas-reference` to avoid finding openblas.

Related: https://github.com/microsoft/vcpkg/pull/19628

Thanks @Neumann-A!


JackBoosY:
cc @Neumann-A @cenit

Neumann-A:
?!?!? who removed the `blas` metaport?

also what you are adding here is `blas-reference`. just like `lapack-reference`.

JackBoosY:
@Neumann-A I want to clarify the relationship between blas and lapack's reference and non-reference, can you elaborate on it?

Neumann-A:
@JackBoosY: Please keep blas a metaport. What you are adding here is blas-reference. There a many ways vcpkg could provide blas/lapack but the current idea is to have the blas port just check that blas can be found and depend on vcpkg defaults blas implementation (openblas). Using an overlay this could be changed to e.g. using mkl instead.

JackBoosY:
@Neumann-A What about the dependencies between them? This PR start with a ci regression when building colmap:x64-windows-static:
```
ninja: error: 'D:/installed/x64-windows-static/debug/lib/openblas.lib', needed by 'src/exe/colmap.exe', missing and no known rule to make it
```
Colmap's dependency chain:
colmap -> ceres/suitesparse -> lapack -> lapack-reference -> blas -> openblas
Their messy dependencies have now affected other ports such as colmap.

Neumann-A:
@JackBoosY That is just someone linking openblas but not declaring the dependency correctly:
```
neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
$ ./vcpkg.exe depend-info colmap:x64-windows-static | findstr openblas

neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
```
vs
```
neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
$ ./vcpkg.exe depend-info colmap:x64-windows | findstr openblas
openblas:
blas: openblas
```
so some target has a dependency on openblas but does not have it in its build-depends which is why ci is not restoring openblas. Find the target which links openblas and make the dependency on blas explicit.

Neumann-A:
furthermore this is related to #15571. vcpkg needs to force the internally used blas and lapack version to avoid finding spurious other stuff.

JackBoosY:
@Neumann-A But the libblas generated by lapack-reference contains the same symbols as the openblas generated by openblas.

cenit:
> And according to the upstream issue xianyi/OpenBLAS#203, the blas used in lapack-reference is not openblas.

i think that issue talks about the opposite: openblas contains lapack-reference 

the two algebra libraries are really tangled together, the best option is disabling lapack in openblas and blas in lapack-reference, and then explicitly depend on both. 

as @Neumann-A is saying, colmap error should be just a matter of fact of a port depending (wrongly) on openblas(!static)

JackBoosY:
After comparing the blas source code in lapack-reference with the source code in blas, I found that they are almost the same (except some comments). 
Although the feature noblas of lapack-reference depends on openblas, it still uses the internal blas code to generate libblas, and creates a link with the pkgconfig of openblas In the lib, this obviously violates the dependency.

Neumann-A:
The problem is lapack-reference cannot link against openblas from vcpkg if `VCPKG_CRT_LINKAGE` is static since gfortran cannot simply link against the MSVC CRT in this case. As such `lapack-reference` has to use its internal blas implementation in this case. As already said switching to ifort solves this problem.

Neumann-A:
> After comparing the blas source code in lapack-reference with the source code in blas, I found that they are almost the same (except some comments).

Yeah because it is the reference implementation in both cases ;) lapack-reference could probably do a FETCH_CONTENT of the blas stuff

JackBoosY:
> @JackBoosY That is just someone linking openblas but not declaring the dependency correctly:
> 
> ```
> neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
> $ ./vcpkg.exe depend-info colmap:x64-windows-static | findstr openblas
> 
> neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
> ```
> 
> vs
> 
> ```
> neumann@heineken MINGW64 /e/qt6 (update_qt6_6.2)
> $ ./vcpkg.exe depend-info colmap:x64-windows | findstr openblas
> openblas:
> blas: openblas
> ```
> 
> so some target has a dependency on openblas but does not have it in its build-depends which is why ci is not restoring openblas. Find the target which links openblas and make the dependency on blas explicit.

That because:
https://github.com/microsoft/vcpkg/blob/d679a1e0befa9c9e36fcda169c94f6daba2224b7/ports/lapack-reference/CONTROL#L15-L17

According to my poor experience with blas and lapack, I can't distinguish the relationship between them at all.

Neumann-A:
@JackBoosY: Just think of the CMakeLists.txt of both ports as superbuilds.

openblas comes with lapack
lapack comes with blas, cblas, lapacke

vcpkg only wants to build one lib of those in the ports: 
so lapack-reference should only build lapack
openblas should only build openblas

due to the gfortran constrains lapack needs to build its internal blas in x64-windows-static. 
The problem: `FindBlas.cmake` picks up a spurious openblas installation of vcpkg in one port since it was installed while another port was looking for it but that port did not declare the dependency on openblas -> error in colmap since openblas is not restored by CI.

JackBoosY:
> The problem: FindBlas.cmake picks up a spurious openblas installation of vcpkg in one port since it was installed while another port was looking for it but that port did not declare the dependency on openblas -> error in colmap since openblas is not restored by CI.

But why create a link to the pkgconfig file of openblas?

Neumann-A:
> If it uses internal blas, it should not depend on blas in CONTROL.

it does. since lapack is providing blas. in this case. 

> But why create a link to the pkgconfig file of openblas?

that is probably a bug.

JackBoosY:
> > If it uses internal blas, it should not depend on blas in CONTROL.
> 
> it does. since lapack is providing blas. in this case.
> 
> > But why create a link to the pkgconfig file of openblas?
> 
> that is probably a bug.

This will be fixed in https://github.com/microsoft/vcpkg/pull/19628.

JackBoosY:
Okay, I will rename the new port as `blas-reference`.

JackBoosY:
Now we encounter the following problems:
1. Since `noblas` relies on external `openblas`, we cannot add the corresponding conflicting feature `innerblas`. Because according to our policy, any conflicting features should not be added. So feature `cblas` also encountered the same problem.
2. As `openblas` conflicts with `blas-reference`, do we need to disable one of them in the baseline?

ports/blas-reference/portfile.cmake
```diff
@@ -0,0 +1,51 @@
...
+if (VCPKG_TARGET_ARCHITECTURE STREQUAL "x64")
+    set(BUILD_X64 ON)
+else()
+    set(BUILD_X64 OFF)
+endif()
+
+include(vcpkg_find_fortran)
+vcpkg_find_fortran(FORTRAN_CMAKE)
+
+vcpkg_configure_cmake(
```
JackBoosY, line 32: We can't use vcpkg-cmake since issue https://github.com/microsoft/vcpkg/issues/19818

JackBoosY:
This baseline regression blocked 5-6 PRs, @ras0219-msft please make some suggestions.

ras0219-msft (changes requested):
@Neumann-A Thanks for the summary (https://github.com/microsoft/vcpkg/pull/19608#issuecomment-900108479)

> openblas comes with lapack
> lapack comes with blas, cblas, lapacke
>
> vcpkg only wants to build one lib of those in the ports:
> so lapack-reference should only build lapack
> openblas should only build openblas


To put more words on the subject for other readers, here are our objectives:

1. If a consumer depends on `blas`, then `find_package(BLAS)` will resolve (+pkgconfig)
2. If a consumer depends on `lapack`, then `find_package(LAPACK)` will resolve (+pkgconfig)
3. There will be only one provider of BLAS or LAPACK symbols in a given build tree
4. (ideally) Official vcpkg triplets will provide a consistent default choice for lapack backend. I'm not yet certain whether the gfortran issue makes this intractable for static-crt triplets.
5. Consumers have a choice of backends to provide blas and lapack by replacing `blas` or `lapack` and their choice should be respected by the tree (including builds failing if appropriate)

Then, we need to determine what the space of backends looks like. Given @Neumann-A's summary, it looks to me like there are several possible configurations:

1. lapack-reference[noblas] & openblas[nolapack]
2. openblas[blas,lapack]
3. mpl & lapack-reference[noblas] (?)
4. lapack-reference[noblas] & blas-reference[noblas]
5. Built in MacOS option for BLAS?

Open questions:
1. How does `cblas` fit into this? Can we have it be implied by `blas` or is it useful for it to be a unique metaport?
2. What restrictions do the major packages have w.r.t. incompatibilities (such as the gfortran issue)?
3. What other major packages are relevant to this space?
4. How do we test this? This is another "forks the entire graph" situation like openssl/libressl/boringssl. Perhaps we need to just pick the default subset that we will test the world against and relegate other configurations to community registries that can provide more domain-specific testing?

cenit:
about the configurations: we have also clapack already in tree, it was our default lapack port before lapack-reference was accepted

also, macOS Accelerate Framework contains BLAS but also LAPACK.

Final note: all implementations must have compatible ABI and they have, at least as far as i know  (not headers, it is common to just call libraries knowing their interface)

ports/blas-reference/FindBLAS.cmake
```diff
@@ -0,0 +1,1191 @@
+
```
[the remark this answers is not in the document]
Neumann-A: > we just need to define the outputs listed in

Please also add the check that the libraries are actual usable.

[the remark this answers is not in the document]
JackBoosY: The reason why I manually put `FindBLAS.cmake` into the port directory is that the version provided by cmake is not suitable for the port `blas-reference`.
In windows, if `BUILD_X64` is turned on, `blas-reference` will generate `libblas64.lib`, and when searching, only `blas` will be found.

ports/blas-reference/vcpkg-cmake-wrapper.cmake
```diff
@@ -0,0 +1,12 @@
+message(STATUS "Using VCPKG FindLAPACK from package 'blas-reference'")
+set(LAPACK_PREV_MODULE_PATH ${CMAKE_MODULE_PATH})
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
+
+list(REMOVE_ITEM ARGS "NO_MODULE")
+list(REMOVE_ITEM ARGS "CONFIG")
+list(REMOVE_ITEM ARGS "MODULE")
+
+set(BLA_VENDOR "Generic")
+_find_package(${ARGS})
```
JackBoosY, line 10:
```suggestion
if (WIN32)
    set(CMAKE_FIND_LIBRARY_PREFIXES_BAK ${CMAKE_FIND_LIBRARY_PREFIXES})
    set(CMAKE_FIND_LIBRARY_PREFIXES lib)
endif()

set(BLA_VENDOR "Generic")
_find_package(${ARGS})

if (WIN32)
    set(CMAKE_FIND_LIBRARY_PREFIXES ${CMAKE_FIND_LIBRARY_PREFIXES})
endif()
```

JackBoosY: In [`FindBLAS.cmake`](https://github.com/Kitware/CMake/blob/master/Modules/FindBLAS.cmake) line 1156:
```cmake
# Generic BLAS library?
if(BLA_VENDOR STREQUAL "Generic" OR
   BLA_VENDOR STREQUAL "NVHPC" OR
   BLA_VENDOR STREQUAL "All")
  if(NOT BLAS_LIBRARIES)
    check_blas_libraries(
      BLAS_LIBRARIES
      BLAS
      sgemm
      ""
      "blas"
      ""
      ""
      ""
      )
  endif()
endif()
```
line 275:
```cmake
function(CHECK_BLAS_LIBRARIES LIBRARIES _prefix _name _flags _list _deps _addlibdir _subdirs)
...
        find_library(${_prefix}_${_lib_var}_LIBRARY
          NAMES ${_library}
          NAMES_PER_DIR
          PATHS ${_extaddlibdir}
          PATH_SUFFIXES ${_subdirs}
        )
```

Unless we forced the 64bit openblas-reference library name to `blas.lib` or `libblas.lib`, we can't search it using `FindBLAS.cmake` provided by cmake.

Neumann-A:
> I'm not yet certain whether the gfortran issue makes this intractable for static-crt triplets.

the problem lies mainly with the Fortran library. Because that is linked with the mingw stuff and requires ld as a linker due to internal symbol stuff. If We could link against a naively build windows Fortran library the problem goes away (#19413 does this by switching to ifort). One problem still is that the compiler Fortran libraries have not set the /APPCONTAINER bit so strictly speaking Fortran on UWP is currently not supported. 
The only solution I see for the UWP problem in the future is to switch to the Fortran library provided by LLVM via Flang which will be build by vcpkg itself and will set the bit correclty. 

> There will be only one provider of BLAS or LAPACK symbols in a given build tree

that would be the ideal situation but it is already broken due to intel-mkl also providing blas/lapack symbols. It is more important to force the lookup of `find_package(BLAS)`, `find_package(LAPACK)` to the selection of the metaport dependency blas/lapack. Here XOR features would be helpful to cover the cases without forcing users to use an overlay (Just like `BLA_VENDOR` in CMake).

> How does cblas fit into this?

cblas is just a C wrapper to the legacy blas library. As such it can be installed using any blas implementation. It will wrap the fortran symbols using a `cblas_` prefix

> What restrictions do the major packages have w.r.t. incompatibilities (such as the gfortran issue)?

as said the incompatibility is mainly due to the linkage problems with libgfortran.lib in static builds. If we ca get a native windows variant there is only the issue with UWP an the missing appcontainer bit. There is no ABI implication. 

> What other major packages are relevant to this space?

Just look into CMakes `FindBLAS.cmake/FindLAPACK.cmake`. We currently have intel-mkl, openblas, blas-reference, lapack-reference, macOS Accelerate Framework, clapack (outdated IMHO). There are probably also variants with openMP/MPI enabled?

ports/blas-reference/FindBLAS.cmake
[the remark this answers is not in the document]
ras0219-msft: If the `FindBLAS` from CMake is not usable and we need to replace it in this case, that's fine. However, we do not need a thousand lines of CMake to do that -- a much smaller file will do fine.

ras0219-msft:
>If We could link against a naively build windows Fortran library the problem goes away

Hmm, maybe splitting the `.lib` back into objects and then using msvc to re-create the static lib would help? I'll try this.

>... intel-mkl also providing blas/lapack ...

Isn't this handled by simply overlaying both `lapack` and `blas` to point at `mkl` then?

> It is more important to force the lookup of find_package(BLAS), find_package(LAPACK) to the selection of the metaport dependency blas/lapack.

Agreed, this seems critical. If we think "cross" configurations like `lapack-reference` LAPACK on top of `intel-mkl` BLAS are also important, then this may require features to request "find_package(X) support" -- the exclusivity would be across packages and not within a package. It still is suboptimal, but I think it should be functional enough to work for now. Specifically, `intel-mkl[lapack]` would provide `share/lapack/vcpkg-cmake-wrapper.cmake` to force `mkl` as your lapack provider, but it wouldn't provide that by default. To make the global selection, you'd overlay a `lapack -> intel-mkl[lapack]` and `blas -> intel-mkl[core]`. Same story for the various other "interesting" configurations.

> XOR features would be helpful to cover the cases without forcing users to use an overlay

Agreed, though I think that will still require the basic structure I've indicated above.

> As such it can be installed using any blas implementation

But can we assume every blas implementation will install it? Do we need to have an explicit port that just creates the wrapper?

PhoebeHui (approved):

ports/blas-reference/usage
```diff
@@ -0,0 +1,5 @@
+Use this package via the module FindBLAS that comes with CMake. To use in your CMakeLists.txt:
```
cenit, line 1: this is not true because we are checking-in our own modified version

JackBoosY:
The solution accepted, but the `FindBLAS.cmake` needs to be minimized. 
Does `FindBLAS.cmake` only provided for port blas-reference? We have other port cblas, openblas.
We now have two approaches:
1. Provide Findblas.cmake to blas-reference and reduce its content.
2. Provide Findblas.cmake to all blas-like ports and keep all the content.

JackBoosY:
In favor of #21479.

Issue #19623, opened 2021-08-17
binary cache on macOS is not deterministic 


autoantwort:
**Describe the bug**
On M1 macs you can build the vcpkg tool as arm64 or x64. But for some libs the resulting lib file has the same architecture as the vcpkg tool regardless of the selected triplet. So if you build the tool as arm64, you get arm64 libs, if you build the tool as x64, you get x64 libs. This problem is getting worse with binary caching, because the hash is the same for the arm64 and the x64 version of the library. So you can't trust the whole binary cache because it maybe contains binary caches for the wrong architecture.

**Environment**
- OS: macOS with M1
- Compiler: apple clang

**Suggested Solution**
Fail when the resulting lib has the wrong architecture (post build check). You can check the architecture of a libs with `lipo -archs ...`.

**Alternative Solution**
Always build the tool as x64. Then you maybe get x64 libs when you use the `arm64-osx` triplet, but then this is deterministic.

Be-ing:
> same architecture as the vcpkg tool

Do you mean that the architecture the vcpkg executable was built for with `./bootstrap-vcpkg.sh`?

autoantwort:
Yes. On arm macs the default is arm64, but you can change it to x64 for example by `arch -x86_64 ./bootstrap-vcpkg.sh`. Then you get an x64 vcpkg tool

JackBoosY:
As more and more users use M1 mac, this matter has become more urgent. 
For bootstrap, we may need to make improvements immediately. 
For pipeline machines, can we use half of the work for arm64-osx?

Neumann-A:
Just like the dumpbin check on windows vcpkg should integrate a lipo check on osx an probably some kind of arch check on normal linux. It is probably way easier to implement on the other platforms since vcpkg can simply assume that the required tools are available on the path.

PR #19624, opened 2021-08-17
Make macOS binary cache deterministic. Workaround for #19623

autoantwort:
Workaround for #19623

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -299,7 +299,7 @@ echo "Building vcpkg-tool..."
 rm -rf "$baseBuildDir"
 mkdir -p "$buildDir"
 vcpkgExtractArchive "$tarballPath" "$srcBaseDir"
-cmakeConfigOptions="-DCMAKE_BUILD_TYPE=Release -G 'Ninja' -DCMAKE_MAKE_PROGRAM='$ninjaExe'"
+cmakeConfigOptions="-DCMAKE_BUILD_TYPE=Release -G 'Ninja' -DCMAKE_MAKE_PROGRAM='$ninjaExe' -DCMAKE_OSX_ARCHITECTURES='x86_64'"
 
 if [ "${VCPKG_MAX_CONCURRENCY}" != "" ] ; then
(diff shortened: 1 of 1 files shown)


autoantwort:
paraview is currently generally broken, this is not caused by this PR

JackBoosY:
I think we can't simply do that.

cc @strega-nil-ms

cenit:
in any case, valid or not, please add a comment with a reference to the issue next to the code, otherwise in a month or two someone will reverse it without any knowledge, "to improve m1 compatibility"

autoantwort:
Not needed anymore when https://github.com/microsoft/vcpkg-tool/pull/163 gets merged

scripts/bootstrap.sh
```diff
@@ -299,7 +299,7 @@ echo "Building vcpkg-tool..."
 rm -rf "$baseBuildDir"
 mkdir -p "$buildDir"
 vcpkgExtractArchive "$tarballPath" "$srcBaseDir"
-cmakeConfigOptions="-DCMAKE_BUILD_TYPE=Release -G 'Ninja' -DCMAKE_MAKE_PROGRAM='$ninjaExe'"
+cmakeConfigOptions="-DCMAKE_BUILD_TYPE=Release -G 'Ninja' -DCMAKE_MAKE_PROGRAM='$ninjaExe' -DCMAKE_OSX_ARCHITECTURES='x86_64'"
```
strega-nil-ms, line 302: I'd prefer that we choose the correct architecture here rather than always x86_64; I want an arm64 binary on my M1.

autoantwort: Yeah but that maybe breaks the binary cache. This PR is unnecessary when https://github.com/microsoft/vcpkg-tool/pull/163 gets merged soon

strega-nil-ms: Could you do `'-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64'`?

strega-nil-ms: Also, I don't understand how setting it to `arm64` breaks the binary cache any less or more than `x64`?

autoantwort: > Also, I don't understand how setting it to arm64 breaks the binary cache any less or more than x64?

Yes that is true, setting it to arm64 also works. But some ports don't build when you build the tool as arm64, so x64 is more reliable. 

> Could you do '-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64'?

I could. But we should merge microsoft/vcpkg-tool#163

autoantwort:
Close in favor of microsoft/vcpkg-tool#163

Issue #16259, opened 2021-02-17
[macOS] Some ports reference hardcoded library paths


davidebeatrici:
After relocating a vcpkg build environment to another machine, I encountered the following error when building my application:
```bash
ninja: error: '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/CoreFoundation.framework', needed by 'murmurd', missing and no known rule to make it
```

Searching for `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk` in the environment's directory yields the following results:

In `installed/x64-osx/share/absl/abslTargets.cmake`:
```cmake
set_target_properties(absl::time_zone PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<\$<PLATFORM_ID:Darwin>:/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/CoreFoundation.framework>"
)
```

In `installed/x64-osx/share/unofficial-angle/unofficial-angle-config.cmake`:
```cmake
set_target_properties(unofficial::angle::angle_gpu_info_util PROPERTIES
  INTERFACE_LINK_LIBRARIES "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/IOKit.framework;/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/CoreFoundation.framework;/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/CoreGraphics.framework;\$<LINK_ONLY:unofficial::angle::angle_common>"
)
```

Removing `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/` from the path (leaving only the library's filename) fixes the issue.

JackBoosY:
[...]

PR #19629, opened 2021-08-18
[vcpkg-fixup-cmake-targets] Rewrite the traversal code with `-framework NAME` on OSX

JackBoosY:
https://github.com/microsoft/vcpkg/blob/ab60f27567ec24ab97289fb7a1d8455c74bcad04/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L230-L232
Because `foreach (IN LISTS)` separates each item according to **a space**, it incorrectly replaces the contents of non-`INTERFACE_LINK_LIBRARIES`. Such as:
```
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_INCLUDE_DIRECTORIES "-framework OpenGL"
  INTERFACE_LINK_LIBRARIES "-framework OpenGL;-framework OpenGL"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "-framework OpenGL"
)
```

Fix that.

Now, this will fix
```
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_LINK_LIBRARIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
)
```
to
```
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_LINK_LIBRARIES "-framework OpenGL;-framework OpenGL"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
)
```


JackBoosY:
cc @strega-nil-ms for review this PR.

ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.cmake
```diff
@@ -250,7 +250,7 @@ get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)]]
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "\"-framework ${name}\"" fixed_line "${fixed_line}")
```
JackBoosY, line 253:
```suggestion
                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```

scripts/cmake/vcpkg_fixup_cmake_targets.cmake
```diff
@@ -249,7 +249,7 @@ function(vcpkg_fixup_cmake_targets)
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "\"-framework ${name}\"" fixed_line "${fixed_line}")
```
JackBoosY, line 252:
```suggestion
                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```

Neumann-A:
So this is what broke paraview

dg0yt:
> ```cmake
> target_link_libraries(main PRIVATE -framework OpenGL)
> ```

How about 
~~~cmake
target_link_libraries(main PRIVATE OpenGL.framework)
~~~
instead of evil spaces?

scripts/cmake/vcpkg_fixup_cmake_targets.cmake
```diff
@@ -249,7 +249,7 @@ function(vcpkg_fixup_cmake_targets)
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```
JackBoosY, line 252:
```suggestion
                        string(REPLACE "${framework}" "${name}.framework" fixed_line "${fixed_line}")
```

ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.cmake
```diff
@@ -250,7 +250,7 @@ get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)]]
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```
JackBoosY, line 253:
```suggestion
                        string(REPLACE "${framework}" "${name}.framework" fixed_line "${fixed_line}")
```

JackBoosY:
> > ```cmake
> > target_link_libraries(main PRIVATE -framework OpenGL)
> > ```
> 
> How about
> 
> ```cmake
> target_link_libraries(main PRIVATE OpenGL.framework)
> ```
> 
> instead of evil spaces?

No, it didn't work.

ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.cmake
```diff
@@ -250,7 +250,7 @@ get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)]]
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "${name}.framework" fixed_line "${fixed_line}")
```
JackBoosY, line 253:
```suggestion
                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```

scripts/cmake/vcpkg_fixup_cmake_targets.cmake
```diff
@@ -249,7 +249,7 @@ function(vcpkg_fixup_cmake_targets)
                     endif()
                     list(FIND VCPKG_DETECTED_CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES "${path}" index)
                     if(NOT index EQUAL -1)
-                        string(REPLACE "${framework}" "-framework ${name}" fixed_line "${fixed_line}")
+                        string(REPLACE "${framework}" "${name}.framework" fixed_line "${fixed_line}")
```
JackBoosY, line 252:
```suggestion
                        string(REPLACE "${framework}" "\\\"-framework ${name}\\\"" fixed_line "${fixed_line}")
```

JackBoosY:
No, even add the double quotes and `\` didn't work, will revert the changes in https://github.com/microsoft/vcpkg/pull/19469 now to temporary avoid this issue.

Neumann-A:
`SHELL:-framework <Name>` ?

Neumann-A:
https://cmake.org/cmake/help/latest/command/add_compile_options.html

JackBoosY:
> `SHELL:-framework <Name>` ?

Or use cmake expression.

Neumann-A:
> Or use cmake expression.

It could also be that it needs to be a simple list instead. 
So instead of `-framework <Name>` ->  `-framework;<Name>`

Do you have an example config where this problem occurs and the replacement is necessary?

dg0yt:
> > > ```cmake
> > > target_link_libraries(main PRIVATE -framework OpenGL)
> > > ```
> > 
> > 
> > How about
> > ```cmake
> > target_link_libraries(main PRIVATE OpenGL.framework)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of evil spaces?
> 
> No, it didn't work.

Hm, probably it is more 
~~~cmake
find_library(SOME_FRAMEWORK x.framework)
target_link_libraries(x ${SOME_FRAMEWORK})
~~~

dg0yt:
I'm not finding documentation right now. But curl uses what was proposed initally: `"-framework CoreFoundation"`.

JackBoosY:
Ther results are:
1. we can use `-framework=NAME`.
2. currently we wrongly replaced the value of `INTERFACE_INCLUDE_DIRECTORIES` and `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`.

dg0yt:
That's all what I found evidence of:
- There is `target_link_libraries(tgt PRIVATE "-framework A")`
- CMake documentation on `find_library` indicates special behaviour for 
  `target_link_libraries(tgt PRIVATE "<fullPath>/A.framework")` 
  > If the library found is a framework, then `<VAR>` will be set to the full path to the framework `<fullPath>/A.framework`. When a full path to a framework is used as a library, CMake will use a `-framework A`, and a `-F<fullPath>` to link the framework to the target.
- For includes, there seems to be similar behaviour for `<fullPath>/A.framework`, documented for `find_path`:
  > When searching for frameworks, if the file is specified as `A/b.h`, then the framework search will look for `A.framework/Headers/b.h`. If that is found the path will be set to the path to the framework. CMake will convert this to the correct -F option to include the file.

Given the lack of documentation, it might make sense to look at CMake unit tests, or other Kitware sources.

JackBoosY:
https://github.com/microsoft/vcpkg/blob/bd7fc2c75b87b815c96abff1990a0d2333b20d6d/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L231
Here we found all the string begin with `INTERFACE_LINK_LIBRARIES` and drop `INTERFACE_LINK_LIBRARIES`.
And here
https://github.com/microsoft/vcpkg/blob/bd7fc2c75b87b815c96abff1990a0d2333b20d6d/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L255
We replaced all the values of framework with `-framework ${VALUE}` except `INTERFACE_LINK_LIBRARIES`, so it would replace all the string with framework.

JackBoosY:
> That's all what I found evidence of:
> 
> * There is `target_link_libraries(tgt PRIVATE "-framework A")`
> * CMake documentation on `find_library` indicates special behaviour for
>   `target_link_libraries(tgt PRIVATE "<fullPath>/A.framework")`
>   > If the library found is a framework, then `<VAR>` will be set to the full path to the framework `<fullPath>/A.framework`. When a full path to a framework is used as a library, CMake will use a `-framework A`, and a `-F<fullPath>` to link the framework to the target.
> * For includes, there seems to be similar behaviour for `<fullPath>/A.framework`, documented for `find_path`:
>   > When searching for frameworks, if the file is specified as `A/b.h`, then the framework search will look for `A.framework/Headers/b.h`. If that is found the path will be set to the path to the framework. CMake will convert this to the correct -F option to include the file.
> 
> Given the lack of documentation, it might make sense to look at CMake unit tests, or other Kitware sources.

We can't simply add double quotes to the string because this value exists in `INTERFACE_LINK_LIBRARIES`:
```cmake
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_LINK_LIBRARIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
)
```
Add double quotes or `\\\"` is a disaster, and I confirm that does work.

dg0yt:
Thanks for sharing the data. I have poor IPv4 connectivity for a few more days, so contributing is harder than usual.

> We can't simply add double quotes to the string because this value exists in `INTERFACE_LINK_LIBRARIES` which contains double quotes already:

We don't want to add quotes. We need one of two things:
- a space between `-framework` and the framework name, as part of a quoted string.
- stay with absolute paths. but replace the prefix with a variable which is initialized from the system which is using the framework (vcpkg toolchain), not from the system which is building the port.

The following line does the right thing for `target_link_libraries` only iff the repacement takes place within a quoted string. (I hope this is also true for `INTERFACE_LINK_LIBRARIES` but I can't verify ATM). It is not sufficient if there aren't quotes:
https://github.com/microsoft/vcpkg/blob/bd7fc2c75b87b815c96abff1990a0d2333b20d6d/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L252 

I can try some experiments with curl. VTK is too many ports away for me.

JackBoosY:
> Thanks for sharing the data. I have poor IPv4 connectivity for a few more days, so contributing is harder than usual.
> 
> > We can't simply add double quotes to the string because this value exists in `INTERFACE_LINK_LIBRARIES` which contains double quotes already:
> 
> We don't want to add quotes. We need one of two things:
> 
> * a space between `-framework` and the framework name, as part of a quoted string.
> * stay with absolute paths. but replace the prefix with a variable which is initialized from the system which is using the framework (vcpkg toolchain), not from the system which is building the port.
> 
> The following line does the right thing for `target_link_libraries` only iff the repacement takes place within a quoted string. (I hope this is also true for `INTERFACE_LINK_LIBRARIES` but I can't verify ATM). It is not sufficient if there aren't quotes:
> https://github.com/microsoft/vcpkg/blob/bd7fc2c75b87b815c96abff1990a0d2333b20d6d/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L252
> 
> I can try some experiments with curl. VTK is too many ports away for me.

Yes, for now, I can confirm that the origin changes is good, and the only problem is we wrongly replace `INTERFACE_INCLUDE_DIRECTORIES` and `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` content since [here](https://github.com/microsoft/vcpkg/pull/19629#issuecomment-901721015).

ports/vcpkg-cmake-config/vcpkg_cmake_config_fixup.cmake
```diff
@@ -227,12 +227,17 @@ get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)]]
...
         foreach(targets_file IN LISTS targets_files)
             file(READ "${targets_file}" targets_content)
-            string(REGEX MATCHALL "INTERFACE_LINK_LIBRARIES[^\n]*\n" library_contents "${targets_content}")
-            foreach(line IN LISTS library_contents)
-                set(fixed_line "${line}")
-                string(REGEX MATCHALL [[/[^ ;"]+/[^ ;"/]+\.framework]] frameworks "${line}")
+            string(REGEX MATCHALL "INTERFACE_LINK_LIBRARIES[^\n]*\n" matched_lines "${targets_content}")
+            # For every line begin with `INTERFACE_LINK_LIBRARIES`
+            string(REGEX MATCH "INTERFACE_LINK_LIBRARIES[^\n]*\n" current_line "${matched_lines}")
+            while (current_line)
```
JackBoosY, line 236: Now it will intercept each line in the list according to the keyword `INTERFACE_LINK_LIBRARIES`, and replace the fixed content in the original content line by line to prevent incorrect replacement of other content (such as `INTERFACE_INCLUDE_DIRECTORIES`).

versions/v-/vcpkg-cmake-config.json
```diff
@@ -1,5 +1,10 @@
 {
   "versions": [
+    {
+      "git-tree": "dd7d67c81c2442c831a298d136b019a2bd138565",
```
JackBoosY, line 4:
```suggestion
      "git-tree": "d4b5ea86aea3416a957fbe2384610feb6106f730",
```

Neumann-A:
> INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"

Isn't there some kind of env variable the full path could be replaced with ? Because this changes doesn't change the fact that there are still full paths in the configs.

dg0yt:
>> INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
>
> Isn't there some kind of env variable the full path could be replaced with ? Because this changes doesn't change the fact that there are still full paths in the configs.

There is existing code that removes the path part for *link libraries* if it is in `CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES`

https://github.com/microsoft/vcpkg/blob/ab60f27567ec24ab97289fb7a1d8455c74bcad04/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L250

Should this be applied to include directories, too?

JackBoosY:
> > INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
> 
> Isn't there some kind of env variable the full path could be replaced with ? Because this changes doesn't change the fact that there are still full paths in the configs.

I wish too, but I didn't find that.

JackBoosY:
> > > INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
> > 
> > 
> > Isn't there some kind of env variable the full path could be replaced with ? Because this changes doesn't change the fact that there are still full paths in the configs.
> 
> There is existing code that removes the path part for _link libraries_ if it is in `CMAKE_CXX_IMPLICIT_LINK_FRAMEWORK_DIRECTORIES`
> 
> https://github.com/microsoft/vcpkg/blob/ab60f27567ec24ab97289fb7a1d8455c74bcad04/scripts/cmake/vcpkg_fixup_cmake_targets.cmake#L250
> 
> Should this be applied to include directories, too?

No, the original error was caused by changing the value of `*_INCLUDE_DIRECTORIES` to `-framework VALUE`.

dg0yt:
> No, the original error was caused by changing the value of *_INCLUDE_DIRECTORIES to -framework VALUE.

Sure. 

I am concerned that `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"` contains an absolute path from the port build environment, and it might limit the usability of the binary cache items if the consuming environment uses another path for the SDK.

cenit:
i agree with @dg0yt and @Neumann-A 

this is a half-backed fix, because there are still absolute paths. Why fixing only some? We still have the problems of having absolute paths in the configs! Better to fix none in the end, instead of a half-solution?

JackBoosY:
@cenit See https://github.com/microsoft/vcpkg/issues/16259, the origin PR wants to replace those absolute path in cmake files.

cenit:
> @cenit See #16259, the origin PR wants to replace those absolute system paths in cmake files.

yeah I know.
But why leaving absolute paths in INTERFACE_INCLUDE_DIRECTORIES and INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ? Since you added -framework properly, those lines might even be removed if they do not contain anything important.
Leaving them with an absolute path is just asking for troubles in future!

dg0yt:
We know how to the substitution now, and we know the ports to look at for verification. Do it right and completely.
(We might immediately add a hard check which just blindly searches for known build system path strings in ...config.cmake files, .pc files and ...-config scripts.)

JackBoosY:
@dg0yt  But we don't know which expression should be replaced.

Neumann-A:
@JackBoosY:

`(/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/[^;"]+)` ? + maybe a `;;` and `;"` replacement

cenit:
> @dg0yt But we don't know which expression should be replaced.

nothing, it's not necessary if you have a proper "-framework VALUE" in the interface_link_library slot.
You can remove them from INTERFACE_INCLUDE_DIRECTORIES and INTERFACE_SYSTEM_INCLUDE_DIRECTORIES and it should work, without having to fix them! Did you try this way?

cenit:
Of course, before removing everything from those fields, you have to be sure there are not other libraries, which are not part of the "fixed" frameworks included in the fixed "interface_link_libraries" line ;)

cenit:
so to say

```
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_LINK_LIBRARIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/System/Library/Frameworks/OpenGL.framework"
)
```

should become

```
set_target_properties(VTK::opengl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GL_SILENCE_DEPRECATION"
  INTERFACE_LINK_LIBRARIES "-framework OpenGL"
)```

JackBoosY:
Temporary close this issue because I have no good idea about that.

cenit:
@autoantwort how did you manage this issue in your “remove absolute paths” heroic effort?

autoantwort:
The absolute path check only checks for absolute paths to the package and the installed dir paths.

PR #19631, opened 2021-08-18
[vcpkg_copy_tool_dependencies] Only run on Windows hosts

longnguyen2004:
**Describe the pull request**

- #### What does your PR fix?  
  Only copy dependencies on Windows hosts, since PowerShell isn't present on other platforms by default.

  mingw triplets, not needed

  Yes

### DIFF
--- a/scripts/cmake/vcpkg_copy_tool_dependencies.cmake
+++ b/scripts/cmake/vcpkg_copy_tool_dependencies.cmake
@@ -38,7 +38,7 @@ function(vcpkg_copy_tool_dependencies tool_dir)
         message(WARNING "${CMAKE_CURRENT_FUNCTION} was passed extra arguments: ${ARGN}")
     endif()
 
-    if(VCPKG_TARGET_IS_WINDOWS)
+    if(VCPKG_TARGET_IS_WINDOWS AND CMAKE_HOST_WIN32)
         find_program(Z_VCPKG_POWERSHELL_CORE pwsh)
         if (NOT Z_VCPKG_POWERSHELL_CORE)
             message(FATAL_ERROR "Could not find PowerShell Core; please open an issue to report this.")


longnguyen2004:
@talregev Please test this PR

JackBoosY (approved):

dg0yt:
Repeating my comment from #19599:

The DLL copying must be fixed, not made impossible, for mingw on Linux hosts: If a binary package was built on Linux using x64-mingw-dynamic triplet, it must be immediately usable on a x64 Windows host, just as if it was compiled there (or as if compiled with x64-windows).

talregev:
I Will test it.

talregev:
> Repeating my comment from #19599:
> 
> The DLL copying must be fixed, not made impossible, for mingw on Linux hosts: If a binary package was built on Linux using x64-mingw-dynamic triplet, it must be immediately usable on a x64 Windows host, just as if it was compiled there (or as if compiled with x64-windows).

Until we have the cmake implementation for copy dll of mingw dynamic. 
Can we not use the `vcpkg_copy_tool_dependencies` only for mingw static on linux?

dg0yt:
> Until we have the cmake implementation for copy dll of mingw dynamic.
> Can we not use the `vcpkg_copy_tool_dependencies` only for mingw static on linux?
> Is it acceptable solution?

You mean "only for mingw dynamic`, don't you?
In any case, it would probably also need to be aware that some ports may use non-standard linkage.

talregev:
> mean "only for

I mean do the dll copy with `vcpkg_copy_tool_dependencies` for mingw dynamic and windows.
What I mean that the only port that don't fit with the PowerShell dependency is mingw static on linux.

If this solution is blocking copy of tool (binaries and dlls) in mingw dynamic on linux, then it not good solution, and should not approve. I agree. 

Any other suggestion? maybe to do the cmake solution as you suggested? Is it hard work?

dg0yt:
> do the cmake solution as you suggested? Is it hard work?

It is much harder than just installing powershell.
It is cross-platform work, for contributor and reviewers. mingw is not even tested in CI. And it is not a single central script but also per-port scripts for thing such as Qt plugins.

talregev:
> > do the cmake solution as you suggested? Is it hard work?
> 
> It is much harder than just installing powershell.
> It is cross-platform work, for contributor and reviewers. mingw is not even tested in CI. And it is not a single central script but also per-port scripts for thing such as Qt plugins.

You say you will support such an option. What do you mean by that?

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -38,7 +38,7 @@ function(vcpkg_copy_tool_dependencies tool_dir)
         message(WARNING "${CMAKE_CURRENT_FUNCTION} was passed extra arguments: ${ARGN}")
     endif()
 
-    if(VCPKG_TARGET_IS_WINDOWS)
+    if(VCPKG_TARGET_IS_WINDOWS AND CMAKE_HOST_WIN32)
```
talregev, line 41: Can you change it to 
VCPKG_TARGET_IS_WINDOWS AND dynamic)?

JackBoosY: I think we always need that because some ports will use `ONLY_DYNAMIC_LIBRARY`.

JackBoosY:

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
talregev: So if port use only dynamic they should enter here. if port is dynamic it should enter here.
What about windows and static. if it will not enter here, it will be ok?

talregev: For now mingw dynamic on linux should enter here, and it not. That why is not good solution.

JackBoosY: @talregev mingW on Linux shouldn't enter here because `CMAKE_HOST_WIN32`. Can you double confirm that?

talregev: As @dg0yt say, mingw dynamic should enter here, because it create dlls.

talregev: > mingW on Linux shouldn't enter here because CMAKE_HOST_WIN32

And because of that flag, mingw dynamic on linux do not enter here, and because of that, is not good solution.

talregev: I propose different check instead CMAKE_HOST_WIN32. Can we check if we compile dynamic or static?

longnguyen2004:
Whatever it is, if we want it to work on Linux, we have to move away from PowerShell, and use something like `file(GET_RUNTIME_DEPENDENCIES)`

dg0yt:
> if we want it to work on Linux, we have to move away from PowerShell

It is not a strict *must*. The current script just fails due to poor path separator handling. This can be fixed easily. (CI does use PowerShell for all platforms.)

But it is certainly desired.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
dg0yt: At the time of calling `vcpkg_copy_tool_dependencies`, isn't it enough for most (or all) ports to look at `VCPKG_LIBRARY_LINKAGE`? The remaining ports might use a parameter to indicate whether some tools are known to have non-default linkage.

JackBoosY:
Any progress?

JackBoosY:
Please reopen this PR when you continue it.

dg0yt:
For the record, I already fixed mingw DLL copying with powershell on Linux hosts (plugins excluded).

PR #19633, opened 2021-08-18
[vcpkg baseline][paraview:x64-osx][vcpkg_cmake_config_fixup] Revert #19469

PhoebeHui:
Revert OSX related changes in vcpkg_cmake_config_fixup and vcpkg-cmake-config in https://github.com/microsoft/vcpkg/pull/19469.

This PR expects to resolve the CI baseline regression 'paraview:x64-osx', since it blocks many PRs, we'd better to revert it now.

Confrimed with Jack, the changes will be re-test and fixed via PR https://github.com/microsoft/vcpkg/pull/19629.


versions/baseline.json
```diff
@@ -6618,7 +6618,7 @@
     },
     "vcpkg-cmake-config": {
       "baseline": "2021-08-11",
-      "port-version": 0
+      "port-version": 1
```
Neumann-A, line 6621: Just edit the date instead?

ports/vcpkg-cmake-config/vcpkg.json
```diff
@@ -1,4 +1,5 @@
 {
   "name": "vcpkg-cmake-config",
-  "version-date": "2021-08-11"
+  "version-date": "2021-08-11",
+  "port-version": 1
```
Neumann-A, line 4: also edit the date instead of pushing the port-version

versions/baseline.json
PhoebeHui: Updated! thanks for your review!

ports/vcpkg-cmake-config/vcpkg.json
PhoebeHui: Updated! thanks for your review!

JackBoosY (approved):

JackBoosY:
Ping @strega-nil-ms for merge this first.

PhoebeHui:
@dan-shaw @strega-nil, could you help merge this PR?

PhoebeHui:
Thank you Nicole! ☺

versions/baseline.json
```diff
@@ -6617,8 +6617,8 @@
       "port-version": 0
     },
     "vcpkg-cmake-config": {
-      "baseline": "2021-08-11",
-      "port-version": 0
+      "baseline": "2021-05-22",
+      "port-version": 1
```
JackBoosY, line 6621: @strega-nil-ms We couldn't do that because the version shoud only be added, not be reverted.

JackBoosY (changes requested):

versions/baseline.json
strega-nil: I think that it's fine, but would like confirmation from @ras0219-msft

dg0yt: It may look fine now, but with the next change, we will have two variants of port version 2...

dg0yt: Cut the "port version 2", it is different "version-date". It is really easy to start confusion ;-)

PhoebeHui:
@strega-nil, the CI pipeline failed with following failures:
```
versions/v-/vcpkg-cmake-config.json(2,): error : While reading versions for port vcpkg-cmake-config from file: version/v-/vcpkg-cmake-config.json
Local port version `2021-05-22#1` exists in version file but it's not the first entry in the "versions" array.
```

JackBoosY:
I think we can drop this PR and take https://github.com/microsoft/vcpkg/pull/19629 instead.

PR #19641, opened 2021-08-18
Fix freebsd vcpkg-tool compilation issue

ghost:
Adds an xml entry to cmake 3.20.4 for freebsd 13


JackBoosY:
Can you please attach the failure log? Did `cmake-3.21.1-linux-x86_64.tar.gz` not work?

InfRandomness:
> Can you please attach the failure log? Did `cmake-3.21.1-linux-x86_64.tar.gz` not work?

Well, originally, when running the vcpkg bootstrap.sh script on freebsd, I would get : 

```
Building vcpkg-tool...
eval: /usr/home/build/sllang/vcpkg/downloads/tools/cmake-3.9.6-freebsd/python.exe: Permission denied
```

but I'm not sure I get your point; we sure could use the freebsd linux compatibility and then just download the linux cmake archive but I don't think that's fair for all of the peeps that do not want to install the linux compatibility packages.
This would most likely also have lots of version mismatching.

JackBoosY (approved):

strega-nil-ms:
Merging since this is good and paraview was my fault.

Issue #11262, opened 2020-05-09
[eastl] Compiling solution fails with "EABase/eabase.h not found"


Nelarius:
**Describe the bug**
Installing `eastl` also installs the `eabase` dependency. But including any `eastl` headers leads to a compilation failure, because `eastl`'s internal includes are are of the form `#include <EABase/eabase.h>`, and the `eabase` package places the headers into `include/Common/EABase/**.h`.

**Environment**
- OS: Windows 10
- Compiler: Visual Studio Community 2019 Version 16.5.4

**To Reproduce**
1. ./vcpkg install eastl
2. Create a new Console App project in Visual Studio.
3. Add `#include <EASTL/vector.h>` and compile
4. The build will fail with
```
1>eastl-repro.cpp
1>C:\Users\nelarius\vcpkg\installed\x86-windows\include\EASTL\internal\config.h(61,11): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
```

**Expected behavior**
I would expect that the solution should work without manually specifying the inclusion path, just like the other packages I installed. EABase's CMake install seems to export the correct include path `include/Common`, but I was under the impression that you don't have to use CMake with `vcpkg` to generate your solution files 🤔 

**Additional context**
EABase's folder structure seems to be a recurring theme with the project, see for instance https://github.com/electronicarts/EASTL/issues/320.

PhoebeHui:
@Nelarius, thanks for reporting this issue!

I think the changes should be updated in Upstream, eabase install its headers to ./include/Common, so the port(eg: eastl) which depends on eabase would find the headers under ./include/Common, if we changes the path, all these ports affect.

PhoebeHui:
Please report to upstream instead, reopen this issue if you still expect vcpkg do something here.

xgalaxy:
I think this ticket should be reopened. Its the ports patch file that is causing the EABase headers to be installed into the incorrect location. This doesn't seem like an upstream issue to me.

PR #19643, opened 2021-08-18
[eastl] fix dependency eabase

### LINKED ISSUE #12734: [eastl] Incorrect EABase include
**Describe the bug**
EABase is not included correctly, and not in sync with the EASTL port either. EABase is stored in a different file directory level than what EASTL expects.

**Environment**
- OS: Windows
- Compiler: MSVC

**To Reproduce**
1. ./vcpkg install eastl 
2. See error

**Expected behavior**
Should just work out of the box (compile and link with no issues).

**Failure logs** 
-(please attached failure logs)
image
image
image

FabienPean:
Trying to use EASTL fails hard due to some include problem with EABase (see #12734). It is still not usable to this day and it seems highly unlikely solution will come from upstream.

All EASTL files rely on having the problematic `include/Common/` folder on the include path so it is never mentioned in the source files. Currently, the folder somehow exists on the path for the user (not in the eyes of EASTL though, don't know why, the cmake files seem correct), so the user could already use simply `#include<EABase/xxxx>`. This PR is a small change to existing patch to install EABase at the root of vcpkg include folder, which solved the problem. 

- #### What does your PR fix?  
  Fixes #12734

  Same as in previous version of the port

  Yes

  Yes

### DIFF
--- a/ports/eabase/CONTROL
+++ b/ports/eabase/CONTROL
@@ -1,4 +1,0 @@
-Source: eabase
-Version: 2.09.12-1
-Homepage: https://github.com/electronicarts/EABase
-Description: Electronic Arts Base. EABase is a small set of header files that define platform-independent data types and macros.
--- a/ports/eabase/fix_cmake_install.patch
+++ b/ports/eabase/fix_cmake_install.patch
@@ -36,7 +36,7 @@ index 89c6703..ab8e553 100644
 +    #-------------------------------------------------------------------------------------------
 +    target_include_directories(EABase INTERFACE
 +        $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include/Common>
-+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/Common>
++        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 +    )
 +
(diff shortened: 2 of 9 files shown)


PhoebeHui:
LGTM. @FabienPean, thanks for the PR!

From https://github.com/electronicarts/EABase, vcpkg installed the include directory correctly, so the changes shoud be updated by EASTL. However the Upstream refused to change their reference, see https://github.com/electronicarts/EASTL/issues/390.

We don't like block users, it solves the problem and it looks not cause the regressions. The only risk may from projects who use the eabase, I searched #include<Common/EABase/eabase.h> in all github projects, I didn't find any reference, many projects used '#include<EABase/eabase.h>', so it looks safe for now.

cc @ras0219 @ras0219-msft @c0rp3n 

Related:
https://github.com/microsoft/vcpkg/issues/11262
https://github.com/microsoft/vcpkg/issues/12734

c0rp3n (approved):
This looks good to me, stops people having to use workarounds.

Won't have a chance to test compiling a project with this change though.

PhoebeHui:
@c0rp3n, thanks for your comments!

The baseline issue would be fixed by PR https://github.com/microsoft/vcpkg/pull/19633. I will rerun the CI after the PR merged.

PhoebeHui (approved):

ras0219-msft:

PR #19677, opened 2021-08-20
[imgui] Update to 1.84.1

RT2Code:
Update imgui from 1.83 to 1.84.1. I also merged all the different opengl3 bindings into one because imgui now provides its own lightweight loader to make the opengl3 backend usage easier : https://github.com/ocornut/imgui/issues/4445

Changelogs:
https://github.com/ocornut/imgui/releases/tag/v1.84
https://github.com/ocornut/imgui/releases/tag/v1.84.1

### DIFF
--- a/ports/imgui/CMakeLists.txt
+++ b/ports/imgui/CMakeLists.txt
@@ -65,27 +65,7 @@ if(IMGUI_BUILD_OPENGL2_BINDING)
     target_sources(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/backends/imgui_impl_opengl2.cpp)
 endif()
 
-if(IMGUI_BUILD_OPENGL3_GLEW_BINDING)
-    find_package(GLEW REQUIRED)
-    target_link_libraries(${PROJECT_NAME} PUBLIC GLEW::GLEW)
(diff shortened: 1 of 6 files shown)


JonLiu1993 (approved):

JonLiu1993:
@RT222, thanks for your PR, have you tested these features locally?

ras0219-msft:
Sweet, this is an awesome improvement! Thank you so much!

cenit:
what about changing features and upgrade procedure? also, is it ok to break compatibility with "older" manifests? just asking to understand procedures to implement better how to handle these situations in future on my side

RT222:
Sorry, I forgot to mention it, but I indeed tested the features locally on the following platforms :

- x86-windows
- x64-windows
- x64-windows-static
- x64-uwp
- arm64-windows
- arm-uwp

Here's the details :

Before the update to 1.84 :
- dx12-binding fails on x86-windows.
- glut-binding fails on x64-uwp, arm64-windows and arm-uwp because freeglut dependency doesn't build on these platforms.
- glfw-binding fails on x64-uwp and arm-uwp because glfw3 dependency doesn't build on these platforms.
- dx9-binding, dx10-binding, dx12-binding, opengl2-binding, sdl2-binding and win32-binding fails on x64-uwp and arm-uwp.

With the update to 1.84 :
- opengl3-binding fails on uwp platforms because of the new loader.

PR #19692, opened 2021-08-22
[vcpkg] Document vcpkg cmake wrappers

dg0yt:
- #### What does your PR fix?  
  This PR fixes the lack of documentation for vcpg cmake wrappers. It is a start - some topics remain to be added.

  not relevant

  yes

  not relevant

### DIFF
--- a/docs/maintainers/wrapper-guidelines.md
+++ b/docs/maintainers/wrapper-guidelines.md
@@ -1,0 +1,156 @@
+# Wrappers for CMake find modules
+
+The vcpkg CMake toolchain modifies CMake's `find_package` behaviour by allowing
+ports to provide wrappers for genuine CMake find modules. Wrappers do not
+replace the original behaviour of find modules but help initializing variables
+and targets with vcpkg port configuration details such as installation paths,
+file names and transitive usage requirements related to active port features.
+
+The original parameters passed to the `find_package` called are made available
+to the wrapper in variable `ARGS`. The genuine find module can be called via
+`_find_package`. The typical structure of a wrapper is the following:
+
+```cmake
+# Early setup: cache variables for library directories etc.
+# (Goal: Make genuine module succeed with consistent configuration.)
+...
+# Call the genuine CMake find module.
+_find_package(${ARGS})
+# Late setup: complementary configuration for transitive usage requirements etc.
+# (Goal: Improve usability and maintainability for vcpkg.)
+...
+```
+
+
+## Wrapper naming and location
+
+CMake's find module names (aka package names) are case sensitive. However, the
+install location of a wrapper uses lower-case spelling of the package name for
+the install location. On calling `find_package(<Pkg>)`, vcpkg will look for
+`share/<pkg>/vcpkg-cmake-wrapper.cmake`.
+
+
+## CMake version and policies
+
+Wrappers cannot rely on a recent version of CMake. Unlike portfiles, wrappers
+can run in the context of a user project where the version of CMake is determined
+by the user. In order to not limit the usability of vcpkg, wrappers must not use
+language features of CMake newer than version 3.4 unless really necessary.
+
+Note that some language features depend not only the actual CMake version but
+also on the activated policies. The default configuration is defined by the user
+project's `cmake_minimum_required` statement. To use some features of CMake 3.4,
+such as `if ("word" IN_LIST <list>)`, some policies must be activate locally:
+
+```cmake
+cmake_policy(PUSH)
+cmake_policy(SET CMP0057 NEW)
+...
+cmake_policy(POP)
+```
+
+
+## Scoping of variables and targets
+
+Find modules don't establish an own scope but operate in the scope of the
+subdirectory where they are used. Some care is needed in order to avoid
+interference with user projects and CMake behaviour.
+
+- If a wrapper needs to create variables, these variables shall be prefixed
+  with `Z_VCPKG`. They must not be used without initialization.
+- If a wrapper creates CMake targets, these targets shall be namespaced by
+  using prefix `unofficial::`.
+- CMake macro `find_dependency` from module `CMakeFindDependencyMacro` cannot
+  be used.
+
+
+## Library locations and selection
+
+Regular CMake find modules use `find_library(<Pkg>_LIBRARY_RELEASE ...)` and
+`find_library(<Pkg>_LIBRARY_DEBUG ...)` to locate the link libraries for a
+given `<Pkg>`. If the module fails to locate the proper release and debug
+variants on its own, the usual wrapper pattern for early setup is:
+
+```cmake
+find_library(<Pkg>_LIBRARY_DEBUG NAMES name1d name2_d NAMES_PER_DIR PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/debug/lib" NO_DEFAULT_PATH)
+find_library(<Pkg>_LIBRARY_RELEASE NAMES name1 name2 NAMES_PER_DIR PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib" NO_DEFAULT_PATH)
+```
+
+If not implemented in the find module, the wrapper must take care of selecting
+the right configuration for the `<Pkg>_LIBRARIES` variable. Remember that this
+depends on the behaviour of the find module in the lowest supported version of
+CMake.
+
+```cmake
+include(SelectLibraryConfigurations)
+select_library_configurations(<Pkg>)
+unset(<Pkg>_FOUND)  # https://gitlab.kitware.com/cmake/cmake/-/issues/22509
+```
+
+
+## Setting up targets
+
+Wrappers may add extra targets which are not provided by a particular version
+of a genuine find module. But find modules may be called multiple times, from
+different subdirectories, in a user project. There are some pitfalls which must
+be avoided:
+
+- Before adding an imported target, always check if it already exists.
+- Calling `target_link_libraries` on an imported target is only allowed in the
+  subdirectory which created the target, for CMake versions lower than 3.21.
+  (CMP0079/CMake 3.13 lifts the restriction for normal targets only.) 
+  For simplicity, set/append the properties directly, e.g.
+  
+  ```cmake
+  set_properties(TARGET <Pkg>::Tgt APPEND PROPERTIES INTERFACE_LINK_LIBRARIES some::lib)
+  ```
+
+
+## Handling port features
+
+Vcpkg port features often result in transitive usage requirements which must
+be added to variables and targets in late setup. The transfer of the list of
+active features from port build time (portfile variable `FEATURES`) to wrapper
+usage time must be done by a configuration step in `portfile.cmake`:
+
+```cmake
+configure_file(
+    "${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake.in"
+    "${CURRENT_PACKAGES_DIR}/share/<pkg>/vcpkg-cmake-wrapper.cmake"
+    @ONLY
+)
+```
+
+To capture the features in the installed wrapper, `vcpkg-cmake-wrapper.cmake.in`
+must contain a line like:
+
+```cmake
+set(Z_VCPKG_FEATURES "@FEATURES@")
+```
+
+After this (and with policy CMP0057 activated) the port file may now inspect
+the features and adjust the `_find_package` results accordingly:
+
+```cmake
+if("zlib" IN_LIST Z_VCPKG_FEATURES)
+    find_package(ZLIB)
+    list(APPEND <Pkg>_LIBRARIES ${ZLIB_LIBRARIES})
+    if(TARGET <Pkg>::Tgt)
+        set_property(TARGET <Pkg>::Tgt APPEND PROPERTY INTERFACE_LINK_LIBRARIES ZLIB::ZLIB)
+    endif()
+endif()
+```
+
+
+## Enforce loading of exported config files
+
+If a port exports config files which provide accurate information as expected
+by users of the find module, wrappers may enforce the loading of the config
+files by adjusting the argument list:
+
+```cmake
+list(REMOVE_ITEM ARGS "NO_MODULE")
+list(REMOVE_ITEM ARGS "CONFIG")
+list(REMOVE_ITEM ARGS "MODULE")
+_find_package(${ARGS} CONFIG)
+```
--- a/docs/maintainers/wrapper-guidelines.md
+++ b/docs/maintainers/wrapper-guidelines.md
@@ -154,3 +154,20 @@ list(REMOVE_ITEM ARGS "CONFIG")
 list(REMOVE_ITEM ARGS "MODULE")
 _find_package(${ARGS} CONFIG)
 ```
+
+
+## Handling the `REQUIRED` keyword
+
+When `find_package` is called with the `REQUIRED` keyword (i.e. `REQUIRED`
+occurs in list `ARGS`), any error in setting up the configuration and transitive
+usage requirements may immediately raise a fatal error.
+
+However, a normal call to `find_package` without passing the `REQUIRED` keyword
+must not cause fatal CMake errors. It can indicate failure only by setting
+`<Pkg>_FOUND` to `FALSE`. This implies that a wrapper must not use the
+`REQUIRED` keyword when looking for transitive usage requirements via additional
+calls to `find_package`.
+
+(Normally, failures to find transitive usage requirements indicate serious
+issues with the vcpkg setup. However, users may explicitly disable finding and
+using some modules by setting `CMAKE_DISABLE_FIND_PACKAGE_<Pkg>` to `ON`.)
--- a/docs/README.md
+++ b/docs/README.md
@@ -32,6 +32,7 @@ Vcpkg helps you manage C and C++ libraries on Windows, Linux and MacOS. This too
 - [Control files](maintainers/control-files.md)
 - [Portfile functions](maintainers/portfile-functions.md)
 - [Common CMake definitions](maintainers/vcpkg_common_definitions.md)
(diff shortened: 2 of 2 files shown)


docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Enforce loading of exported config files
+
+If a port exports config files which provide accurate information as expected
+by users of the find module, wrappers may enforce the loading of the config
+files by adjusting the argument list:
+
+```cmake
+list(REMOVE_ITEM ARGS "NO_MODULE")
+list(REMOVE_ITEM ARGS "CONFIG")
```
Hoikas, line 153: Shouldn't this be added instead of removed?

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+Find modules don't establish an own scope but operate in the scope of the
+subdirectory where they are used. Some care is needed in order to avoid
+interference with user projects and CMake behaviour.
+
+- If a wrapper needs to create variables, these variables shall be prefixed
+  with `Z_VCPKG`. They must not be used without initialization.
+- If a wrapper creates CMake targets, these targets shall be namespaced by
+  using prefix `unofficial::`.
+- CMake macro `find_dependency` from module `CMakeFindDependencyMacro` cannot
+  be used.
```
Hoikas, line 64: I think it would be good to explain why this module is inappropriate. My understanding was the opposite--that it was the preferred approach.

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+the install location. On calling `find_package(<Pkg>)`, vcpkg will look for
+`share/<pkg>/vcpkg-cmake-wrapper.cmake`.
+
+
+## CMake version and policies
+
+Wrappers cannot rely on a recent version of CMake. Unlike portfiles, wrappers
+can run in the context of a user project where the version of CMake is determined
+by the user. In order to not limit the usability of vcpkg, wrappers must not use
+language features of CMake newer than version 3.4 unless really necessary.
```
Neumann-A, line 38: Should be dependent on the CMake version the port internally uses itself. If the ports `CMakeLists.txt` is requiring 3.17 I think we should also require that in the wrapper since the `-Config.cmake` will probably not work with an earlier version

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Scoping of variables and targets
+
+Find modules don't establish an own scope but operate in the scope of the
+subdirectory where they are used. Some care is needed in order to avoid
+interference with user projects and CMake behaviour.
+
+- If a wrapper needs to create variables, these variables shall be prefixed
+  with `Z_VCPKG`. They must not be used without initialization.
+- If a wrapper creates CMake targets, these targets shall be namespaced by
```
Neumann-A, line 61:
```suggestion
- If a wrapper creates additional/new CMake targets, these targets shall be namespaced by
```

docs/maintainers/wrapper-guidelines.md
Neumann-A: It only correctly works in the original implemented find_package function and the wrapper is outside of it.

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+    @ONLY
+)
+```
+
+To capture the features in the installed wrapper, `vcpkg-cmake-wrapper.cmake.in`
+must contain a line like:
+
+```cmake
+set(Z_VCPKG_FEATURES "@FEATURES@")
+```
```
Neumann-A, line 129: why declare an extra variable? theoretically it is often used as `if(<somefeature>)`.

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+If a port exports config files which provide accurate information as expected
+by users of the find module, wrappers may enforce the loading of the config
+files by adjusting the argument list:
+
+```cmake
+list(REMOVE_ITEM ARGS "NO_MODULE")
+list(REMOVE_ITEM ARGS "CONFIG")
+list(REMOVE_ITEM ARGS "MODULE")
+_find_package(${ARGS} CONFIG)
+```
```
Neumann-A, line 156:
```suggestion
```

This is not allowed in general. Make normal `Find<Modules>.cmake` work with the wrapper instead!

docs/maintainers/wrapper-guidelines.md
dg0yt: In fact, it is added.
One cannot add it blindly, because it may already be in the list. The given code removes it blindly (which is always allowed), and then explicitly adds `CONFIG` to the actual `_find_package` call.

docs/maintainers/wrapper-guidelines.md
dg0yt: It is intended for *config files*.
Cf. CMake's documentation: 

> **Note** The call to return() makes this macro unsuitable to call from Find Modules.

https://cmake.org/cmake/help/v3.21/module/CMakeFindDependencyMacro.html

(Reminder to self: CMake spelling is "Find Modules".)

Hoikas: Right, I know what the intention is. But because the wrappers are `include()`ed, I don't understand why:
> Note The call to return() makes this macro unsuitable to call from Find Modules.

Is still relevant. That line implies that calling `return()` in a find module is a bad thing(TM), but the wrappers are included, so I must be missing something.

docs/maintainers/wrapper-guidelines.md
dg0yt: Not allowed, or simply not possible? I tried to express the limitations in the first part of the sentence. 
I agree with the recommendation to make regular Find Modules just work. Loading config files is a road into the CMake language features jungle, limiting usability.

docs/maintainers/wrapper-guidelines.md
dg0yt: @Neumann-A got it right, but it probably needs an extra section in this guide because I already fell over it on another occassion.

> It only correctly works in the original implemented `find_package` function 
> and the wrapper is outside of it.

The point is that `find_dependency` relies on [Package File Interface Variables](https://cmake.org/cmake/help/v3.21/command/find_package.html#package-file-interface-variables) which are defined by CMake's original implementation of `find_package`: `CMAKE_FIND_PACKAGE_NAME`, `<PackageName>_FIND_QUIETLY` etc. 

But the wrapper is included from a macro in `vcpkg.cmake`, and the Package File Interface Variables are not available. `find_dependency` simply fails to achieve the side effects which are its key feature.

Hoikas: Thanks for the explanation. I think a synopsis of that would be great in the documentation.

docs/maintainers/wrapper-guidelines.md
Neumann-A: Yeah that was the intention. we are free to add the same targets before/as a Find<Module> does to fix the target before the module messes it ups. This only works because all modules do a `if(NOT TARGET <sometarget>)` check

docs/maintainers/wrapper-guidelines.md
Neumann-A: > There aren't variables like <somefeature> which reflect the individual

`vcpkg_check_features` does a create variables. So we could either use those or just directly pass `-DZ_VCPKG_FEATURES=${FEATURES}` via `vcpkg_configure_make`. No need to create `Z_VCPKG_FEATURES` in the wrapper

docs/maintainers/wrapper-guidelines.md
Neumann-A: > Not allowed, or simply not possible?

not possible. Since you do not know if a custom Find<Module> with a same name is used which sets completely different variables/targets. by rerouting to config it is not possible to patch just the Find<Module> but the whole build-system needs to be patched.

docs/maintainers/wrapper-guidelines.md
dg0yt: > vcpkg_check_features does a create variables. 

Just one variable, and not in a way which has any added value here: Basically, it transforms variable `FEATURES` to some `<OUT_VAR>` which will e.g. contain `-DENABLE_ZLIB=ON` where `FEATURES` has an item `zlib`. 

When `<OUT_VAR>` is passed to cmake, cmake will create a (cache) variables - in project mode while building the port. But these variables are not available in the wrapper which runs in consuming projects, and also not in the portfile which calls `configure_file`.

dg0yt: > So we could either use those or just directly pass `-DZ_VCPKG_FEATURES=${FEATURES}` via `vcpkg_configure_make`.

This goes to the ported software, but this software knows nothing about wrappers without patching).

Neumann-A: ah yeah forgot that the wrapper needs to be hardcoded. I still prefer `if(NOT "@<somefeature>@" STREQUAL "")`

dg0yt: Okay. Well, I don't see a `<somefeature>` most of the time.
To use `FEATURES` without another variable, I can offer once more:

```cmake
if(";@FEATURES@;" MATCHES ";my-feature;")
```

Last not least we may add a vcpkg function if this leads to a clean approach.

dg0yt: Okay. I now realized that "`vcpkg_check_features` does create variables", and how. It is documented.

However, I'm really surprised:
- There are three examples in the documentation - and none uses anything but `FEATURE_OPTIONS`.
- There is no need to make `OUT_FEATURE_OPTIONS` a mandatory parameter when the function has useful effects even without it.

But I don't want to start a `vcpkg_check_features` here. I will update this PR to avoid `@FEATURES@`.

dg0yt: I tested this for port tiff. There is one aspect which I dislike about using the individual variables from `vcpkg_check_features` directly: The `if` statements *in the generated wrapper* don't give any hint about the actual condition:
 
```diff
-if("zip" IN_LIST _tiff_features)
+if(ON)
```

The upside is that typos in variable names become quickly visible, because empty replacements lead to illegal syntax (`if()`).

ras0219-msft (approved):
This is _awesome_, thank you so much for writing all this tribal knowledge down!

I think a lot of the suggestions above are really great, so I'm marking this PR as a draft while you have a chance to look through them and apply the ones you'd like. Once you're ready for this to be merged, just mark it ready for review :)

dg0yt:
On `<PKG>_LIBRARY`:

Some find modules and wrappers use `select_library_configuration` to merge `<PKG>_LIBRARY_RELEASE` and `<PKG>_LIBRARY_DEBUG` into `<PKG>_LIBRARY`. However, some find modules (`FindJPEG.cmake` before CMake 3.12), and possibly user projects, rely on `<PKG>_LIBRARY` being a singe path, e.g. using `get_filename_component` on this variable.

Apart from matching the *particular find module* behaviour, I wonder if it would be preferred to *generally* avoid `select_library_configuration` in wrappers, and rely on a single `find_library(<PKG>_LIBRARY NAMES ... NAMES_PER_DIR)` instead, with expected debug and release names, leveraging the toolchain's debug/release order and the known binary names?
This could still be limited to vcpkg provided binaries by a `PATHS` argument containing vcpkg's lib paths.

However, I have no idea if this works with the multi-config generators. 
- The binary would be selected at cmake configuration time, not at cmake generation time.
- This would be based the basic release/debug build type configuration, not the mapping of additional build types.

(I keep coming back to the idea of encapsulating this stuff in a helper function. Such a function might provide a user variable to change behaviour when needed in a user project.)

Neumann-A:
> However, some find modules (FindJPEG.cmake before CMake 3.12), and possibly user projects, rely on <PKG>_LIBRARY being a singe path, e.g. using get_filename_component on this variable.

In general I never found a valid case of `get_filename_component` of something `find_library` returned. Most of those are just hacks and generally not necessary. 
In general I expect downstream to fix those cases and switch to use proper targets instead. 

>  and rely on a single find_library(<PKG>_LIBRARY NAMES ... NAMES_PER_DIR) instead,

This breaks multi configuration generators. 
Multi config generators need both variables to be set at configure time. It even breaks the generated `*Targets.cmake` if the knowledge of both variants are not supplied at configure time. (e.g. missing genex in interface_link_libraries)

docs/maintainers/wrapper-guidelines.md
```diff
@@ -0,0 +1,173 @@
...
+If not implemented in the find module, the wrapper must take care of selecting
+the right configuration for the `<Pkg>_LIBRARIES` variable. Remember that this
+depends on the behaviour of the find module in the lowest supported version of
+CMake.
+
+```cmake
+include(SelectLibraryConfigurations)
+select_library_configurations(<Pkg>)
+unset(<Pkg>_FOUND)  # https://gitlab.kitware.com/cmake/cmake/-/issues/22509
+```
```
Neumann-A, line 88: `select_library_configurations` should only be called after `find_package()`. Either the module itself already called it then it is unnecessary or the module used `<PKG>_LIBRARY` to maybe set `IMPORTED_LOCATION` of a target. In this case we cannot modify `<PKG>_LIBRARY` before `find_package()`

dg0yt: This section is about (older) modules (versions) that do not call `select_library_configurations` but do use `find_library(<PKG>_LIBRARY ...)`. So the "Either" part is out of scope.

I try to reflect the argument against calling `select_library_configurations` and setting cache variable `<PKG>_LIBRARY` before `find_package()` in this case. `IMPORTED_LOCATION` is documented to be a `Full path`, not a list. This is in line with the problem for `get_filename_component(... <PKG>_LIBRARY ...)`. However, `<PKG>_LIBRARY` has no multi-config alternative, but `IMPORTED_LOCATION` has.

I don't think this can generally be resolved. But maybe the scope of support for older versions of CMake in user projects should be explicitly limited to
- that CMake version being at least 3.4,
- the config files used by the project supporting that CMake version,
- the user project, and the CMake Find modules used by the project,
  - either not using config-unaware CMake features (e.g.
    target property `IMPORTED_LOCATION`, `get_filename_component(... <PKG>_LIBRARY ...)`)
  - or using single-config generators.

(This is not for the wrapper documentation but for the CMake requirements.)

dg0yt:
> This breaks multi configuration generators.
> Multi config generators need both variables to be set at configure time. It even breaks the generated `*Targets.cmake` if the knowledge of both variants are not supplied at configure time. (e.g. missing genex in interface_link_libraries)

That's why it could be only an opt-in alternative. But to implement a global choice, it needs a single function implementing that choice.

JackBoosY:
Any progress here?

dg0yt:
Well, I could do an update, based on the experiences collected.

JackBoosY:
Consider to finish this PR now?

JackBoosY:
Ping for response.

JackBoosY:

JackBoosY:

dg0yt:

ras0219-msft:
This is incredibly useful information and it's a shame it has been sitting for so long! Thank you so much for writing it, @dg0yt.

Since our new docs are in Microsoft/vcpkg-docs, I've taken liberty to migrate the work over to a branch https://github.com/microsoft/vcpkg-docs/compare/main...ras0219-msft:vcpkg-docs:pr-19692. This branch currently contains two commits:
1. https://github.com/ras0219-msft/vcpkg-docs/commit/18098d333bce50f9ee68d0a662a2b45eefb25fc3, which is a (near) direct squash of this PR so far
2. https://github.com/microsoft/vcpkg-docs/commit/3a31546abdc1732a605d49f671af12669823c696 is a bunch of editorializing I made on top that I believe makes it ready to merge.

I've attempted to apply all the critical comments in this PR, though I've probably missed some things. Most importantly, I've rewritten the recommendation on how to bake in features:

```cmake
# portfile.cmake
    PREFIX feature
    FEATURES
        ZLIB WITH_ZLIB
        TIFF WITH_TIFF
)
configure_file(
    "${CMAKE_CURRENT_LIST_DIR}/vcpkg-cmake-wrapper.cmake.in"
    "${CURRENT_PACKAGES_DIR}/share/<pkg>/vcpkg-cmake-wrapper.cmake"
    @ONLY
)
```
```cmake
# vcpkg-cmake-wrapper.cmake.in
if("@feature_WITH_ZLIB@" STREQUAL "ON") # feature_WITH_ZLIB
    find_package(ZLIB)
    list(APPEND <Pkg>_LIBRARIES ${ZLIB_LIBRARIES})
    if(TARGET <Pkg>::Tgt)
        set_property(TARGET <Pkg>::Tgt APPEND PROPERTY INTERFACE_LINK_LIBRARIES ZLIB::ZLIB)
    endif()
endif()
```

This resolves both the problem of `IN_LIST` requiring a policy and needing to reset `Z_VCPKG_FEATURES` after each subordinate call to `find_package()` (which could trigger a different wrapper that itself sets `Z_VCPKG_FEATURES`!).

@dg0yt How would you like to proceed with the work you've done here? My suggestions are:
1. You can grab the first or both commits from my branch and open a PR against vcpkg-docs.
2. Tell me to handle it and I'll open a PR in the new repo and get it merged.

dg0yt:
I'm hesitating to put time into this at the moment. 

TBH I would prefer to discourage the current form of wrappers and get to an implementation based on vcpkg find modules which call into external find modules.
- The wrappers can't see the package interface variables.
- They can't tell module mode from config mode. (Some look for `CONFIG`, `NO_MODULE` or `MODULE`, but each keyword of the full `find_package` signature activates config mode.)
- They depend on undocumented CMake behaviour (overriding built-in commands).
- The `find_package` macro has problems with policy scopes: #28960.

dan-shaw:
Closing this PR for now, as our docs have moved. Please open a new PR there if you are still working on this. Thanks!

PR #19697, opened 2021-08-22
[vcpkg] Update cmake 3.20.4 hash

ghost:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes freebsd cmake hash mismatch issue (?, looks like it was updated for some reason)

yes


autoantwort (approved):

Hoikas:
Doesn't vcpkg require 3.21 for ports.cmake?

InfRandomness:
> Doesn't vcpkg require 3.21 for ports.cmake?

yes, it does, but that version isn't available as of right now on freebsd, unfortunately

PhoebeHui (approved):
@InfRandomness, thanks for your fix!

ras0219-msft:
LGTM, though if this URL is not bit-stable over time we will need to find another source for the package.

PR #19738, opened 2021-08-25
[folly] also support arm64

### LINKED ISSUE #19733: [folly] Add support for arm64
Currently, the portfile for Folly checks if the underlying architecture string contains `x64`. This fails on `arm64` for example, where Folly can be built properly.

A potential solution is to just add the `arm64` string check at the beginning of the portfile.

As a note, I've currently tested this on an Raspberry Pi 4 w/ 2GB of RAM (lowest end RPi4), using Ubuntu Server 20.04 aarch64 and latest `vcpkg` from master. The build succeeds, along with all dependencies. I can contribute the changes back while referencing this issue.

autoantwort:
Fixes #19733  
Tested on my arm64 mac

### DIFF
--- a/ports/folly/portfile.cmake
+++ b/ports/folly/portfile.cmake
@@ -1,5 +1,7 @@
-if(NOT VCPKG_TARGET_ARCHITECTURE STREQUAL x64)
-  message(FATAL_ERROR "Folly only supports the x64 architecture.")
+if (VCPKG_TARGET_IS_WINDOWS)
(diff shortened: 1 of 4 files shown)


JackBoosY (changes requested):

JackBoosY (changes requested):

JackBoosY (approved):

ras0219-msft:
LGTM, thanks!

PR #19753, opened 2021-08-26
[scripts-audit] Make buildsystem (port)

JackBoosY:
Comply with the regulations in https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/cmake-guidelines.md
- [x] Convert functions and variables
- [x] remove/deprecate ENABLE_INSTALL (depending on whether it's used) (see also INSTALL_TARGET)
No because some port's makefile doesn't contains install command.
- [x] look and see if MAKEFILE is used usefully vs SUBPATH; if not, deprecate/remove depending on whether it's used
Yes, will do this job after https://github.com/microsoft/vcpkg/pull/20090 merge.
- [x] Modify docs
- [x] Improve code
- [x] Test
- [x] Depends on PR https://github.com/microsoft/vcpkg/pull/20165

Related: #16188.


ports/vcpkg-make/vcpkg_make_configure.cmake
```diff
@@ -0,0 +1,911 @@
...
+    include("${cmake_vars_file}")
+    if(DEFINED VCPKG_MAKE_BUILD_TRIPLET)
+        set(arg_BUILD_TRIPLET ${VCPKG_MAKE_BUILD_TRIPLET}) # Triplet overwrite for crosscompiling
+    endif()
+
+    set(src_dir "${arg_SOURCE_PATH}/${arg_PROJECT_SUBPATH}")
+
+    set(requires_autogen FALSE) # use autogen.sh
+    set(requires_autoconfig FALSE) # use autotools and configure.ac
+    if(EXISTS "${src_dir}/configure" AND "${src_dir}/configure.ac") # remove configure; rerun autoconf
```
dg0yt, line 279: This line misses an `EXISTS` after `AND`. This bug is present already in the legacy function, but the script always had too many PRs pending...
IMO this whole `if ... endif` block should be reviewed. For example, I don't feel like I would want to see autoconfig being run when `SKIP_CONFIGURE` was passed as an argument.

JackBoosY: We should always generate configure when `configure.ac` exists, and we need to provide a new option to skip this step.

@Neumann-A Do you agreed?

ports/vcpkg-make/vcpkg_make_configure.cmake
```diff
@@ -0,0 +1,863 @@
...
+        # Don't bother with whitespace. The tools will probably fail and I tried very hard trying to make it work (no success so far)!
+        message(WARNING "Detected whitespace in root directory. Please move the path to one without whitespaces! The required tools do not handle whitespaces correctly and the build will most likely fail")
+        set(vcm_paths_with_spaces TRUE)
+    endif()
+
+    # Pre-processing windows configure requirements
+    if (VCPKG_TARGET_IS_WINDOWS)
+        if(CMAKE_HOST_WIN32)
+            list(APPEND msys_require_packages binutils libtool autoconf automake-wrapper automake1.16 m4)
+            vcpkg_acquire_msys(MSYS_ROOT PACKAGES ${msys_require_packages} ${arg_ADDITIONAL_MSYS_PACKAGES})
```
JackBoosY, line 295: This violates the documentation https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/cmake-guidelines.md:
```
Scripts in the scripts tree should not be expected to need observable changes as part of normal operation.
Example violation: vcpkg_acquire_msys() has hard-coded packages and versions that need updating over time due to the MSYS project dropping old packages.
```
but this is required for cross-build.
@strega-nil-ms what do you think about this?

strega-nil-ms: Calling `vcpkg_acquire_msys` is not a violation, but `vcpkg_acquire_msys` has violations of this.

strega-nil-ms:
Could you please do the script audit of the `vcpkg_*_make` commands first? The `vcpkg_make_*` commands are also necessary, but we should do the scripts in the `scripts/cmake` folder first.

JackBoosY:
@strega-nil-ms  Yeah going to do that.

JackBoosY:
Will continue this after https://github.com/microsoft/vcpkg/pull/20165 merge.

dg0yt:
This PR hasn't seen team feedback in three weeks. What is the time frame for getting this finished? How to deal with improvements to the existing maintainer functions which still have dead code, redundant code, underused parallelization, and other deficencies? 
Either this PR is finished quickly (with existing maintainer function behaviour as a starting point) so that changes can focus on the new port alone, or we will have to waste extra time on duplicating and merging changes.

JackBoosY:
@dg0yt  Sorry, I don't have much time, and I will handle this PR this week.

dg0yt:
Okay, thanks.

For future maintainability: 
`vcpkg_configure_make` and `vcpkg_build_make` cooperate by side effects and convention. Example: the name of the build directories. 
Does it make sense to turn everything into explicit parameter passing?
Are any (environment) variables which are modified in the configure function allowed to have an effect on the build function?

JackBoosY:
@dg0yt 

> Does it make sense to turn everything into explicit parameter passing?

For example?

> Are any (environment) variables which are modified in the configure function allowed to have an effect on the build function?

For this one, I think we should follow `vcpkg_*_make`.

dg0yt:
> > Does it make sense to turn everything into explicit parameter passing?
> 
> For example?

In the first place: the name of the build directory?

> > Are any (environment) variables which are modified in the configure function allowed to have an effect on the build function?
> 
> For this one, I think we should follow `vcpkg_*_make`.

ATM it is not obvious if vcpkg_configure_make resets all environment variables that it modifies, due to "code diversity". I need to know if not resetting some variables is intentional or a bug.

To combine both points, the make function duplicates some of the code of the configure function because things are not passed from one to another.

JackBoosY:
Close this PR temporary.

PR #16321, opened 2021-02-20
[tbb] Update to 2021.1.1 to use official cmake build method

JackBoosY:
Since the upstream supported the official cmake build method, always use that to build tbb.

Fixes #16308.

PR #19909, opened 2021-09-01
[tbb] Update to 2021.3.0

### LINKED ISSUE #18074: [tbb] update to <2021.2.0>
Library name: tbb

New version number: 2021.2.0

Please update it.

ras0219-msft:
This PR updates Intel's TBB to the latest release version.

As part of this update, Intel made several breaking changes to their API which have downstream effects:
- `embree2` no longer builds. This version was not supported any longer in favor of `embree3` anyway, so I've removed it from the current baseline.
- `openvdb` hasn't had a release supporting the new version, so I've backported their patch
- Updated `pagmo2`
- `usd` has stated upstream that they will not be fixing compatibility with the new version of TBB for the near term; I've therefore added the port as skip for now with a message for users about how to pin the older version of `tbb` if they need it. I've left the port in the catalog because the implication is that they will eventually fix compatibility -- in comparison to embree 2 which is completely unsupported.


ras0219-msft:
The request changes by the bot is incorrect because it does not know how to handle package removals from the baseline (`embree2`).

JackBoosY:
I think that a port that is still provided by the upstream and can be used normally should not be deleted anyway.

JackBoosY:
Any progress here?

JackBoosY:
Oh I remember I also have a tbb update PR: https://github.com/microsoft/vcpkg/pull/16321.
That is blocked on the downstrem.

JackBoosY:
Draft this PR until it has any progress.

JonLiu1993:
Duplicate of #23567

JonLiu1993:
@ras0219-msft, I temporarily close this pr, if you have time to come back to deal with it please reopen

PR #19985, opened 2021-09-04
Fix libgit2 ssh feature

JafarAbdi:
**Describe the pull request**

I'm trying to use libgit2 with ssh feature but I wasn't able to get it to work, the reason is libssh2 from vcpkg doesn't have a `libssh2.pc` which is how libgit2 is trying to find it, so this package fixes it by introducing a patch file to fix it

I created a repo to reproduce the behavior https://github.com/JafarAbdi/test_libgit2, if you build without my fix it will print

```
GIT_FEATURE_THREADS: 1
GIT_FEATURE_HTTPS: 1
GIT_FEATURE_SSH: 0
GIT_FEATURE_NSEC: 1
```
After applying this fix I was able to get the ssh feature, and running the example prints
```
GIT_FEATURE_THREADS: 1
GIT_FEATURE_HTTPS: 1
GIT_FEATURE_SSH: 1
GIT_FEATURE_NSEC: 1
```

  linux, No

  Yes

### DIFF
--- a/ports/libgit2/0001-Fix-libssh2.patch
+++ b/ports/libgit2/0001-Fix-libssh2.patch
@@ -1,0 +1,25 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index 45dec2796..313af3af6 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -231,15 +231,14 @@ ENDIF()
+ 
+ # Optional external dependency: libssh2
+ IF (USE_SSH)
+-       FIND_PKGLIBRARIES(LIBSSH2 libssh2)
++       FIND_PACKAGE(Libssh2 REQUIRED)
+ ENDIF()
+-IF (LIBSSH2_FOUND)
++IF (Libssh2_FOUND)
+        SET(GIT_SSH 1)
+-       LIST(APPEND LIBGIT2_SYSTEM_INCLUDES ${LIBSSH2_INCLUDE_DIRS})
+-       LIST(APPEND LIBGIT2_LIBS ${LIBSSH2_LIBRARIES})
+-       LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
++       LIST(APPEND LIBGIT2_LIBS Libssh2::libssh2)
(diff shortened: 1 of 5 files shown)


ports/libgit2/0001-Fix-libssh2.patch
```diff
@@ -0,0 +1,25 @@
...
++       FIND_PACKAGE(Libssh2 REQUIRED)
+ ENDIF()
+-IF (LIBSSH2_FOUND)
++IF (Libssh2_FOUND)
+        SET(GIT_SSH 1)
+-       LIST(APPEND LIBGIT2_SYSTEM_INCLUDES ${LIBSSH2_INCLUDE_DIRS})
+-       LIST(APPEND LIBGIT2_LIBS ${LIBSSH2_LIBRARIES})
+-       LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
++       LIST(APPEND LIBGIT2_LIBS Libssh2::libssh2)
++       LIST(APPEND LIBGIT2_PC_LIBS -lssh2)
```
JackBoosY, line 19: Does libssh provides pkgconfig file? I think we should add `requires: ssh2` instead of this.

JackBoosY (changes requested):

ports/libgit2/0001-Fix-libssh2.patch
JafarAbdi: I think it does, but we're removing it [here](https://github.com/microsoft/vcpkg/blob/master/ports/libssh2/portfile.cmake#L25)

cenit: better to restore the pkgconfig there and use @JackBoosY suggestion!

JackBoosY: > I think it does, but we're removing it [here](https://github.com/microsoft/vcpkg/blob/master/ports/libssh2/portfile.cmake#L25)

No, please don't. Vcpkg now support to use the pkgconfig file.

JafarAbdi: I didn't remove it :), sorry I meant it's have been removed from the link I shared, I'll push a new commit to restore it

JackBoosY:

Issue #2109, opened 2017-11-02
Installing OpenSSL not working [Hash Issue]


xtreameprogram:
I was trying to install OpenSSL on my device, however, I got an issue as shown below. I don't know where any logs are, but I tried looking. 

    /Downloads/vcpkg/scripts/cmake/vcpkg_find_acquire_program.cmake:167 (file):
    file DOWNLOAD HASH mismatch

    for file: [C:/Users/Ford/Downloads/vcpkg/downloads/strawberry-perl-5.24.1.1-32bit-portable.zip]
      expected hash: [a6e685ea24376f50db5f06c5b46075f1d3be25168fa1f27fa9b02e2ac017826cee62a2b43562f9b6c989337a231ba914416c110075457764de2d11f99d5e0f26]
        actual hash: [f3355d2b16232e34d28713216e7f49fae02c6b8e957548fd0eda19a3894e43c4ad613acc0685c86f450de36338cf9681d8619c47bd11c147a9079f300d83e40f]
             status: [0;"No error"]

    Call Stack (most recent call first):    
    C:/Users/Ford/Downloads/vcpkg/ports/openssl/portfile.cmake:10 (vcpkg_find_acquire_program)
    C:/Users/Ford/Downloads/vcpkg/scripts/ports.cmake:72 (include)

alexkaratarakis:
Seems like the file was partially downloaded. 
Please delete `C:/Users/Ford/Downloads/vcpkg/downloads/strawberry-perl-5.24.1.1-32bit-portable.zip` and try again.

xtreameprogram:
Worker perfectly. Thanks!

alexkaratarakis:
@xtreameprogram 
"Incomplete downloads" is an issue that has occurred a number of times. In response, the downloading functionality in `vcpkg` has been improved in 66ad158be46e09b2db4fe8be90841fc01d271c68...0fb0191f3cb9ba4ea0f0e0675f929e1e599e1461. 
The downloads are now performed in a temp file; if the download is complete successfully, the file is moved to its actual target destination.

This avoids issues with incomplete downloads.
Thanks for reporting. :)

corymickelson:
Also having this issue, but removing and re-trying just results in the same error message.
λ .\vcpkg.exe install openssl
The following packages will be built and installed:
    openssl:x86-windows
Starting package 1/1: openssl:x86-windows
Building package openssl:x86-windows...
[...]

PR #20012, opened 2021-09-06
[imgui] update to 1.84.2

mathisloge:
**Describe the pull request**

- #### What does your PR fix?  
update imgui to 1.84.2

 y
y

y


RT222:
The docking-experimental feature will build but projects depending on it won't, because you used a commit from the master branch where the docking feature doesn't exist. They did not merge the master to the docking branch for the 1.84.2, so the only way to get it would be from [this merge commit](https://github.com/ocornut/imgui/commit/dedb381c510cc0b87164e16b9e7ef6bf50ffccec), which would also bring some wip changes from the 1.85, which is not desirable.

mathisloge:
Thanks for the hint. Thought I used the correct one. Will fix later today

mathisloge:
>(https://github.com/ocornut/imgui/commit/dedb381c510cc0b87164e16b9e7ef6bf50ffccec) happened after the 1.84.2

I would use this commit. 

> and using it would also bring some changes from the 1.85 wip which is not desirable. 

In my opinion it is not desirable to not update a port because a _experimental_ feature is some steps ahead. Have build a example application with docking and it works and build. 

So IMHO the focus should be the stable release version and not the docking branch. I know that a lot of people using it, but if you experience some malfunctioning inside the experimental-docking feature, you could also pinpoint the version to 1.84.1. 
The regression fix in 1.84.2 is important.

RT222:
I actually agree that it is better to update it in this case, but I would like to clarify that the docking branch is not experimental, it is in beta, which I think is very different. I don't know with the port feature it was named like this and it is an error in my opinion.

From the Imgui changelog :
> The docking (#2109) and multi-viewports (#1542) features are available in the docking branch, they are in beta but actively maintained and being used by many teams already. Your continuous feedback is always appreciated.

mathisloge:
@RT222 yeah. agree with thats beta. I think the wording _experimental_ is indeed a bit ambiguous and open to interpretation. But in the end we have to agree to some policy that the docking is not the offical release. Therefore it should be the version guidance. However there are then two possible ways to handle docking if the commit is newer or older than the release version. 
1. Use the latest minor version since merge 
2. Use the last merge after release

In my opinion the second point is the way to go. (If there aren't to many differences between release and latest merge)

JonLiu1993:
@mathisloge , This pr is ready for review?

mathisloge:
From my point of view - yes

JonLiu1993:
Have you tested the feature locally?

JonLiu1993 (approved):

mathisloge:
> Have you tested the feature locally?

docking-experimental with sdl2 and opengl

PR #20068, opened 2021-09-09
[cimg/libgd/libfabric] Update to latest release version

### LINKED ISSUE #19451: [libgd] update to 2.3.x and later
Library name: libgd

New version number: 2.3.x

Good morning :)

Thanks for the great work here!

We are the libGd teams and finally cleaning up the building scripts to only rely on CMake and autoconf. 

I see that libgd is already in the list of available packages but it seems stuck to 2.2.x. 

What would be the process to add our own registry, anything else than https://devblogs.microsoft.com/cppblog/registries-bring-your-own-libraries-to-vcpkg/ ? Given it is in there already.

We have a task for it https://github.com/libgd/libgd/issues/724

Mainly about missing deps. To see where it fits best (libxpm and libimagequant are not yet on vcpkg).

Thanks,
Pierre

### LINKED ISSUE #19870: [cimg] update to 2.9.9
Library name: cimg

New version number: 2.9.9

https://github.com/dtschump/CImg/releases/latest

### LINKED ISSUE #19927: [libfabric] update to 1.13.1
Library name: libfabric

New version number: 1.13.1

https://github.com/ofiwg/libfabric/releases

### LINKED ISSUE #19928: [libgd] update to 2.3.2
Library name: libgd

New version number: 2.3.2

https://github.com/libgd/libgd/releases

PhoebeHui:
Fixes https://github.com/microsoft/vcpkg/issues/19870
Fixes https://github.com/microsoft/vcpkg/issues/19451
Fixes https://github.com/microsoft/vcpkg/issues/19927

libfabric: Update patch to match the source changes.
cimg: Update to latest release.
libgd: 
1. Update exist 3 patches to match the source changes.
2. Add new patch fix_msvc_build.patch to fix source issue https://github.com/libgd/libgd/issues/679 and find dependency port webp. This issue has been fixed in Upstream, will be included in future release 2.3.4.

### DIFF
--- a/ports/libgd/portfile.cmake
+++ b/ports/libgd/portfile.cmake
@@ -19,7 +19,6 @@ file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindPackageHandleStandardArgs.
 file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindPNG.cmake")
 file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindWEBP.cmake")
 
-
 vcpkg_check_features(
     OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     FEATURES
--- a/ports/libgd/portfile.cmake
+++ b/ports/libgd/portfile.cmake
@@ -1,51 +1,35 @@
 vcpkg_from_github(
     OUT_SOURCE_PATH SOURCE_PATH
     REPO libgd/libgd
-    REF gd-2.2.5
-    SHA512 e4ee4c0d1064c93640c29b5741f710872297f42bcc883026a63124807b6ff23bd79ae66bb9148a30811907756c4566ba8f1c0560673ccafc20fee38d82ca838f
+    REF 2e40f55bfb460fc9d8cbcd290a0c9eb908d5af7e # gd-2.3.2
+    SHA512 c3f2db40f774b44e3fd3fbc743efe70916a71ecd948bf8cb4aeb8a9b9fefd9f17e02d82a9481bac6fcc3624f057b5a308925b4196fb612b65bb7304747d33ffa
     HEAD_REF master
     PATCHES
         0001-fix-cmake.patch
         no-write-source-dir.patch
         intrin.patch
+        fix_msvc_build.patch
 )
 
 #delete CMake builtins modules
-file(REMOVE_RECURSE ${SOURCE_PATH}/cmake/modules/CMakeParseArguments.cmake)
-file(REMOVE_RECURSE ${SOURCE_PATH}/cmake/modules/FindFreetype.cmake)
-file(REMOVE_RECURSE ${SOURCE_PATH}/cmake/modules/FindJPEG.cmake)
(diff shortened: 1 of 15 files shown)


ports/libgd/portfile.cmake
```diff
@@ -1,51 +1,35 @@
...
-    set(ENABLE_FONTCONFIG ON)
-endif()
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/CMakeParseArguments.cmake")
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindFreetype.cmake")
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindJPEG.cmake")
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindPackageHandleStandardArgs.cmake")
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindPNG.cmake")
+file(REMOVE_RECURSE "${SOURCE_PATH}/cmake/modules/FindWEBP.cmake")
+
+
```
PhoebeHui, line 22:
```suggestion
```

pierrejoye:
Good afternoon,

thanks for taking care of libgd here :)

ssize_t is fixed in libgd/libgd and will be in the upcoming 2.3.4. By the way, we do support windows (now with Windows CI up and running so that kind of build issues should not happen anymore), would it be possible to have such patches upstream as well? It would be easier than maintaining async trees on each side? 

For the record here, I added the intrinsics header fix (will be in 2.3.4).

I am not sure what are the goals of the Cmake changes tho', happy to apply a PR in libgd/libgd if that helps and I know what they are for :)

PhoebeHui:
@pierrejoye, thanks for your comments, I will submit a PR to Upstream.

pierrejoye:
thanks @phoebe :)

I think the last patch is fixed already (unistd.h), I see now the cmake
change and it makes sense, I am applying it now too.

PhoebeHui:
@pierrejoye, yes, the patch would be removed when upgrade it in next release.

NancyLi1013 (approved):

ports/libgd/vcpkg.json
```diff
@@ -1,9 +1,14 @@
 {
   "name": "libgd",
-  "version-string": "2.2.5",
-  "port-version": 5,
+  "version-string": "2.3.2",
```
BillyONeal, line 3: Did upstream indicate that they actually do semver or should this just be `version`?

BillyONeal (changes requested):
Looking for upstream confirmation that they actually do semver.

ports/libgd/vcpkg.json
PhoebeHui: Thanks for pointing out this! From https://github.com/libgd/libgd/blob/master/CHANGELOG.md, libgd adheres to semver.

BillyONeal: Awesome, thanks!

BillyONeal:
Thanks for the update!

PR #20145, opened 2021-09-14
[qt5] Refactor to avoid shared ownership in tools/qt5

ras0219:


ports/qt5-base/cmake/qt_build_submodule.cmake
```diff
@@ -51,77 +73,70 @@ function(qt_build_submodule SOURCE_PATH)
...
+            continue()
+        endif()
+        if(EXISTS "${CURRENT_INSTALLED_DIR}/tools/qt5/bin/${bin}")
+            file(REMOVE "${CURRENT_PACKAGES_DIR}/tools/qt5/bin/${bin}")
+        else()
+            # I expected a dependency that doesn't currently exist to have already been deployed
+            message(FATAL_ERROR "Port ${PORT} expected ${CURRENT_INSTALLED_DIR}/tools/qt5/bin/${bin} to be deployed by another port")
         endif()
+    endforeach()
+    file(GLOB programs "${CURRENT_PACKAGES_DIR}/bin/*.exe" "${CURRENT_PACKAGES_DIR}/debug/bin/*.exe")
```
ras0219, line 134: This needs to be generalized to other platforms -- each qt port should supply the basename of the executable

ports/qt5-webengine/portfile.cmake
```diff
@@ -1,7 +1,7 @@
 vcpkg_check_linkage(ONLY_DYNAMIC_LIBRARY)
 string(LENGTH "${CURRENT_BUILDTREES_DIR}" buildtrees_path_length)
-if(buildtrees_path_length GREATER 35 AND CMAKE_HOST_WIN32)
-    vcpkg_buildpath_length_warning(35)
+if(buildtrees_path_length GREATER 37 AND CMAKE_HOST_WIN32)
+    vcpkg_buildpath_length_warning(37)
```
Neumann-A, line 4: Why did you change this?

ports/qt5-base/cmake/qt_build_submodule.cmake
Neumann-A: On other platforms you can simply move everything in `bin` to `tools/qt5/bin` since bin should contain only scripts and binaries and no `.so` or `.a`. Only windows has dlls there.

Neumann-A:
After studying #19922 a bit more there might by a hidden optional dependency on `qt5-virtualkeyboard` somewhere which pulls the rest in because the conflict with `tools/qt5/bin/Qt5VirtualKeyboard.dll` is the only one I cannot explain after looking at `qtdeploy.ps1`.

ports/qt5-webengine/portfile.cmake
ras0219: The warning message directly suggests "move to a path like C:\src\vcpkg" -- however that example path results in a 37 character path and still warned.

Neumann-A: yeah but it will fail with 36 and 37 as far as i remember

Neumann-A: please check the git blame/history for it and the associated issues. I think I remember that it will fail with that which is why it is set to 35 even though the suggestion is wrong.

ras0219-msft: It does succeed on my local machine with the path C:\src\vcpkg

Neumann-A:
@wrobelda :

Do you now why the qt ports try to install `KF5Archive.dll`? According to the file list I also see it in:
```
	Zeile 40725: openmvg:/tools/openmvg/KF5Archive.dll
	Zeile 44488: paraview:/tools/paraview/KF5Archive.dll
```

wrobelda:
> Do you now why the qt ports try to install `KF5Archive.dll`? According to the file list I also see it in:

Nothing off the bat. This is a recently added port, nothing special in it. Very odd!

ras0219:
I do observe that Kf5 installs `mkspecs/modules/qt_KArchive.pri` with the contents:

```
QT.KArchive.VERSION = 5.84.0
QT.KArchive.MAJOR_VERSION = 5
QT.KArchive.MINOR_VERSION = 84
QT.KArchive.PATCH_VERSION = 0
QT.KArchive.name = KF5Archive
QT.KArchive.module = KF5Archive
QT.KArchive.defines = 
QT.KArchive.includes = $$PWD/../../include/KF5/KArchive
QT.KArchive.private_includes =
QT.KArchive.libs = $$PWD/../../lib
QT.KArchive.depends = core
QT.KArchive.module_config = 
```

I would hope this doesn't automatically get picked up by qt's buildsystem though

ras0219-msft:
Ok, an update on this situation: I believe I've tracked down where the dependency is coming from.

Because QML is a plugin model, we deploy the entire `qml` directory as part of applocal.ps1 if we detect the program to be using `Qt5Qml.dll`[1]. Note though that dependencies of those qml dlls must be placed in the root folder. To illustrate:

```
Given:
    $proj/my.exe -> Qt5Qml.dll
    $vcpkg/qml/qt5/qmlwidget1.dll -> ssl.dll
Then we need to deploy:
    $proj/qt.conf
    $proj/Qt5Qml.dll
    $proj/qml/qt5/qmlwidget1.dll
    $proj/ssl.dll
```

This is _fine_ within the core qt ecosystem because every qt port coordinates to own its own deployment of that slice of qml. Additionally, I have not observed any diamond issues with independent submodules of qt5 that would need to deploy a shared dependency like `ssl.dll`.

However, the ecosystem is no longer closed with the introduction of the kf5* ports, which want to provide their own qml plugins that then depend on other DLLs -- such as (hypothetically; this isn't a real example I think) `qml/org/kf5/kholiday.dll -> KF5Archive.dll`.

I see two paths forward:

1. We brute force the kf5 ports into following the same requirements and restrictions as the qt5 ports; they'll need to all get sorted out about who owns what in the `tools/qt5` subdirectory. Ideally, with some helpers to make this less painful.
2. We find a way to restrict the qml plugin deployment into `tools/qt5` to only be ones from the `qt5-*` set of ports.

I don't know how to evaluate the impact of these two -- how essential is it for the kf5 qml plugins to be present in `tools/qt5/qml`?

[1] https://github.com/microsoft/vcpkg/blob/59a3f35e7e5a3281e59863a4a0123c325899f8f9/ports/qt5-base/qtdeploy.ps1#L65

Neumann-A:
I think we could completely avoid deploying qml stuff. Only the main .dlls need to be deployed. The qml plugins could be discovered by a [qt.conf](https://doc.qt.io/qt-5/qt-conf.html). The only reason the dlls are virtually needed is to get the linked dlls in a recursive way.

JackBoosY:
Any progress here?

JackBoosY:
Closing this PR since it seems that no progress is being made. Please ping me to reopen if work is still being done.

PR #20183, opened 2021-09-15
[hdf5] Update to 1.12.1

Thieum:
Update hdf5 to 1.12.1 - There were some changes in the cmake build, I tried to address them as much as I can.

- #### What does your PR fix?  
N/A

  same as previous version - tested only on x86 windows so far.

Yes

Yes


ports/hdf5/mingw-import-libs.patch
```diff
@@ -1,16 +1,17 @@
...
++++ b/config/cmake_ext_mod/HDFMacros.cmake
+@@ -142,7 +142,7 @@ macro (HDF_SET_LIB_OPTIONS libtarget libname libtype)
    endif ()
  
-   #----- Use MSVC Naming conventions for Shared Libraries
--  if (MINGW AND ${libtype} MATCHES "SHARED")
-+  if (FALSE AND MINGW AND ${libtype} MATCHES "SHARED")
+   option (HDF5_MSVC_NAMING_CONVENTION "Use MSVC Naming conventions for Shared Libraries" OFF)
+-  if (HDF5_MSVC_NAMING_CONVENTION AND MINGW AND ${libtype} MATCHES "SHARED")
++  if (FALSE AND HDF5_MSVC_NAMING_CONVENTION AND MINGW AND ${libtype} MATCHES "SHARED")
```
dg0yt, line 10: Did you actually test with mingw?

ports/hdf5/szip.patch
```diff
@@ -1,16 +1,17 @@
...
+--- a/CMakeFilters.cmake
++++ b/CMakeFilters.cmake
+@@ -113,7 +113,11 @@ if (HDF5_ENABLE_SZIP_SUPPORT)
      find_package (SZIP NAMES ${SZIP_PACKAGE_NAME}${HDF_PACKAGE_EXT} COMPONENTS static shared)
      if (NOT SZIP_FOUND)
        find_package (SZIP) # Legacy find
 -      if (SZIP_FOUND)
 +      if (DEFINED SZIP_SHARED_LIBRARY AND NOT DEFINED SZIP_STATIC_LIBRARY)
-+          set (SZIP_STATIC_LIBRARY "${SZIP_SHARED_LIBRARY}")
++        set (SZIP_STATIC_LIBRARY "${SZIP_SHARED_LIBRARY}")
```
dg0yt, line 11: Does this line have to change?

dg0yt (commented):
You revert some of the pkgconfig fixes I added. You will need to verify with gdal on unix or osx if this broke things. 
And I also made this usable with mingw.

ports/hdf5/mingw-import-libs.patch
Thieum: it installs on mingw64: 
image

ports/hdf5/szip.patch
Thieum: it did not, fixed.

Thieum:
> You revert some of the pkgconfig fixes I added. You will need to verify with gdal on unix or osx if this broke things.
> And I also made this usable with mingw.

Are those verifications part of the azure pipelines, or is it something I need to run on my side?

ports/hdf5/mingw-import-libs.patch
dg0yt: It is not just about installing hdf5. It must be found by consuming ports. This is a question of the library names and pc files etc., depending how other software looks for hdf5. You may have notices that hdf5 now added a switch for MSVC naming conventions on mingw. We should not use MSVC naming conventions for mingw.

dg0yt:
> > You revert some of the pkgconfig fixes I added. You will need to verify with gdal on unix or osx if this broke things.
> > And I also made this usable with mingw.
> 
> Are those verifications part of the azure pipelines, or is it something I need to run on my side?

Normally yes, by building depending ports. However, it is better to check the CI logs that it actually did build or install e.g. gdal, not skip it due to some other missing dependency.

ports/hdf5/mingw-import-libs.patch
Thieum: @dg0yt gdal install on windows mingw, with some fixes that were also needed for msvc (not sure about the pc files duplication, but it works this way, having one version or the other only lead to errors in the hdf5 package, or its dependencies)

dg0yt: Good to know.
I still think that adding `FALSE AND` is not needed. IIRC hdf5 just added the `HDF5_MSVC_NAMING_CONVENTION`. So what I needed to patch in the previous version, should now be an option that can be controlled by a parameter.

ports/hdf5/portfile.cmake
```diff
@@ -114,15 +113,29 @@ set(PKG_FILES hdf5 hdf5_hl hdf5_cpp hdf5_hl_cpp)
...
+    file(REMOVE "${CURRENT_PACKAGES_DIR}/bin/mirror_server${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
+endif()
+file(REMOVE "${CURRENT_PACKAGES_DIR}/debug/bin/mirror_server${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
+
+if(EXISTS "${CURRENT_PACKAGES_DIR}/bin/mirror_server_stop${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
+    file(INSTALL "${CURRENT_PACKAGES_DIR}/bin/mirror_server_stop${VCPKG_TARGET_EXECUTABLE_SUFFIX}"
+            DESTINATION "${CURRENT_PACKAGES_DIR}/tools/${PORT}")
+    file(REMOVE "${CURRENT_PACKAGES_DIR}/bin/mirror_server_stop${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
+endif()
+file(REMOVE "${CURRENT_PACKAGES_DIR}/debug/bin/mirror_server_stop${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
```
dg0yt, line 137: This looks like a use case for `vcpkg_copy_tools`.

dg0yt (commented):
I will test this locally as soon as time permits.

ports/hdf5/portfile.cmake
Thieum: @dg0yt fixed, vcpkg_copy_tools work fine for this.

PR #20200, opened 2021-09-16
[neon2sse] create a new port

luncliff:
### What does your PR fix?  

Add a library `neon2sse`.  https://github.com/intel/ARM_NEON_2_x86_SSE
This project provides 1 header file and supports `find_package` for CMake projects.

```cmake
find_package(NEON_2_SSE CONFIG REQUIRED) # NEON_2_SSE::NEON_2_SSE
```

This is a header only. So every triplet must be available.

I think so


PhoebeHui (approved):
LGTM, thanks for the PR!

ports/neon2sse/fix-cmakelists.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index e4b8980..231b75e 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -2,7 +2,7 @@
+ # Modification: Copyright (C) 2017 David Hirvonen
+ 
+ cmake_minimum_required(VERSION 3.0)
+-project(NEON_2_SSE VERSION 1.0.0)
++project(neon2sse VERSION 1.0.0)
```
BillyONeal, line 10: Can you add a comment somewhere in here describing why we are changing the project title?

BillyONeal: (If you remove the patch, the `vcpkg_cmake_config_fixup` call must be changed to `vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/NEON_2_SSE)`)

luncliff:
Indeed that would be great.

The project uses its name for config file generation.
So config fix up of `NEON_2_SSE` to `neon2sse` will break find_package with
neon2sse.

I thought matching port name and project looks better.

BillyONeal:
Yes, I think it looks better too, but it also forces people who get this from vcpkg to use a different `find_package` name than the normal one upstream uses; ideally we match upstream exactly unless we have some reason we can't.

luncliff:
Hmm... I will use package name 'NEON_2_SSE' instead.
 No patch file, no complexity. I agree.

BillyONeal:
Thanks for the new port!

PR #20210, opened 2021-09-16
[vcpkg_copy_tools] support copying .app bundles

### LINKED ISSUE #15164: [vcpkg] vcpkg_copy_tools does not handle app packages for macos
Just as the title says `vcpkg_copy_tools` should handle app packages on macos.

### LINKED ISSUE #20209: vcpkg_copy_tools() doesn't support macOS .app bundles
**Describe the bug**
Some of the ports I work with generate a complete .app bundle for some of their tools on macOS. `vcpkg_copy_tools()` unfortunately fails to detect it and copy to a location.

**Environment**
- OS: macOS

**To Reproduce**
1. Checkout #20207 
2. `./vcpkg install libalkimia[tools]` on macOS

```
CMake Error at scripts/cmake/vcpkg_copy_tools.cmake:62 (message):
  Couldn't find tool "onlinequoteseditor5":

      "/Users/cromo/Documents/Sourcecode/vcpkg/packages/libalkimia_arm64-osx/bin/onlinequoteseditor5" does not exist
```

**Expected behavior**
Bundle directory should be treated as an executable and copied as if it was a regular ELF or Windows .exe.

wrobelda:
**Describe the pull request**
    - Adds `VCPKG_HOST_BUNDLE_SUFFIX` and `VCPKG_TARGET_BUNDLE_SUFFIX` variables, currently having a non-empty value on macOS only.
    - modifies `vcpkg_copy_tools()` to attempt to copy tools using `VCPKG_TARGET_BUNDLE_SUFFIX` first. If target is not found, continues with the previous logic.

- #### What does your PR fix?  
  Fixes #20209 
  Fixes #15164

  All. 

  Yes.

  Not applicable.


scripts/cmake/vcpkg_common_definitions.cmake
```diff
@@ -88,6 +90,19 @@ else()
     set(VCPKG_TARGET_EXECUTABLE_SUFFIX "")
 endif()
 
+#Helper variables to identify bundles on host/target
+if(VCPKG_HOST_IS_OSX)
+    set(VCPKG_HOST_BUNDLE_SUFFIX ".app")
+else()
+    set(VCPKG_HOST_BUNDLE_SUFFIX "")
```
JackBoosY, line 97: Should we not to set `VCPKG_HOST_BUNDLE_SUFFIX` if the triplet is not OSX?

JackBoosY (changes requested):

scripts/cmake/vcpkg_common_definitions.cmake
wrobelda: `set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_BUNDLE_SUFFIX}")` would then potentially warn about an uninitialized variable, wouldn't it?

JackBoosY: Please note that this value is empty under other triplet, which may cause undefined issues.

scripts/cmake/vcpkg_common_definitions.cmake
```diff
@@ -88,6 +90,19 @@ else()
...
+endif()
+
+if(VCPKG_HOST_IS_OSX)
+    set(VCPKG_TARGET_BUNDLE_SUFFIX ".app")
```
dg0yt, line 101: This uses the same `if(VCPKG_HOST_IS_OSX)` condition as above, but sets `VCPKG_TARGET_BUNDLE_SUFFIX`. This looks like a wrong condition.

scripts/cmake/vcpkg_common_definitions.cmake
dg0yt: An explicit empty value `""` is not the same as undefined.

JackBoosY:
I need my colleague to review this PR further.

scripts/cmake/vcpkg_common_definitions.cmake
wrobelda: Still, the second condition should be `if(VCPKG_TARGET_IS_OSX)`. Good catch!

Neumann-A:
closes  #15164

JackBoosY:
Please download [this](https://dev.azure.com/vcpkg/public/_build/results?buildId=60019&view=artifacts&pathAsName=false&type=publishedArtifacts), apply it and commit changes.

scripts/cmake/vcpkg_common_definitions.cmake
wrobelda: Fixed it.

wrobelda:
> Please download [this](https://dev.azure.com/vcpkg/public/_build/results?buildId=60019&view=artifacts&pathAsName=false&type=publishedArtifacts), apply it and commit changes.

Fixed, thanks.

scripts/cmake/vcpkg_copy_tools.cmake
```diff
@@ -54,8 +54,11 @@ function(vcpkg_copy_tools)
     endif()
 
     foreach(tool_name IN LISTS arg_TOOL_NAMES)
-        set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
+        set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_BUNDLE_SUFFIX}")
         set(tool_pdb "${arg_SEARCH_DIR}/${tool_name}.pdb")
+        if(NOT EXISTS "${tool_path}")
+            set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
```
JackBoosY, line 60: I always think this is incorrect.
We can directly generate executables in osx, but only bundles are found here.

wrobelda: It first tries to find a bundle (a directory with `.app` extension). If it does not exist, it ill then try to locate a binary file without any extension at all. Don't understand where the problem is?

JackBoosY: For macOS, what if the bundle and the directly generated executable exist at the same time?

wrobelda: It's rather unlikely this would happen, because the .app bundle always contains the actual executable inside of it (in `Contents/MacOS` subfolder), so it would mean the executable was duplicated. And in that case the bundle should take the precedent, because the assumption then is that the bundle provides a more complete functionality – i.e. the binary itself and any additional resources (in `Contents/Resources` subfolder)

JackBoosY: Can we check these two files at the same time?

JackBoosY:
cc @strega-nil-ms for review this PR.

scripts/cmake/vcpkg_copy_tools.cmake
wrobelda: If I check for both of them, then I have to deal with the fact that it will get copied over twice on Linux (once because VCPKG_TARGET_BUNDLE_SUFFIX is empty and once because VCPKG_TARGET_EXECUTABLE_SUFFIX is empty). If I want to deal with that, I have to add another layer of logic.

wrobelda: I can change it to something like:
```
    foreach(tool_name IN LISTS arg_TOOL_NAMES)
        set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
        set(tool_pdb "${arg_SEARCH_DIR}/${tool_name}.pdb")
        if(EXISTS "${tool_path}")
            file(COPY "${tool_path}" DESTINATION "${arg_DESTINATION}")
        else()
            message(FATAL_ERROR "Couldn't find tool \"${tool_name}\":
    \"${tool_path}\" does not exist")
        endif()
        if(NOT ${VCPKG_TARGET_EXECUTABLE_SUFFIX} STREQUAL ${VCPKG_TARGET_BUNDLE_SUFFIX})
            set(tool_path "${arg_SEARCH_DIR}/${tool_name}${VCPKG_TARGET_BUNDLE_SUFFIX}")
            if(EXISTS "${tool_path}")
                file(COPY "${tool_path}" DESTINATION "${arg_DESTINATION}")
            else()
                message(FATAL_ERROR "Couldn't find tool \"${tool_name}\":
    \"${tool_path}\" does not exist")
            endif()
        endif()
        if(EXISTS "${tool_pdb}")
            file(COPY "${tool_pdb}" DESTINATION "${arg_DESTINATION}")
        endif()
    endforeach()
```

wrobelda: @JackBoosY updated as suggested.

JackBoosY:
```
CMake Error at scripts/cmake/vcpkg_copy_tools.cmake:65 (if):
  if given arguments:

    "NOT" ".app" "STREQUAL"

  Unknown arguments specified
```

strega-nil-ms:
Are there bundles on non-macOS operating systems?

wrobelda:
> Are there bundles on non-macOS operating systems?

Not that I am aware of, but also not a specialist of Android nor iOS.

JackBoosY (approved):

scripts/cmake/vcpkg_copy_tools.cmake
```diff
@@ -58,6 +58,14 @@ function(vcpkg_copy_tools)
         set(tool_pdb "${arg_SEARCH_DIR}/${tool_name}.pdb")
         if(EXISTS "${tool_path}")
             file(COPY "${tool_path}" DESTINATION "${arg_DESTINATION}")
+        elseif(VCPKG_TARGET_BUNDLE_SUFFIX AND NOT "${VCPKG_TARGET_BUNDLE_SUFFIX}" STREQUAL "${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
```
strega-nil-ms, line 61:
```suggestion
        elseif(NOT "${VCPKG_TARGET_BUNDLE_SUFFIX}" STREQUAL "" AND NOT "${VCPKG_TARGET_BUNDLE_SUFFIX}" STREQUAL "${VCPKG_TARGET_EXECUTABLE_SUFFIX}")
```

strega-nil-ms (approved):

PR #20277, opened 2021-09-22
[s2n] update to 1.1.0

camarois:
Update port to latest version 1.1.0.
Also Debug target was pointing to release lib location. Cleaned the portfile and used vcpkg_cmake_config_fixup() instead to generate good debug/release targets.


PhoebeHui (approved):

strega-nil-ms (approved):

strega-nil-ms:
Thanks @camarois!

PR #20309, opened 2021-09-23
[qt5] Fix mingw builds

dg0yt:
- #### What does your PR fix?  
  Fixes building qt5-base ... qt5-tools with a mingw toolchain on Windows hosts.

  as before + mingw, no

  yes

  I am still working on this PR

---
Depends on #20253 (icu mingw fix).

TODO: Remove absolute `compiler paths from tools/qt5/mkspecs/qmodule.pri`.


ports/qt5-base/cmake/find_qt_mkspec.cmake
```diff
@@ -65,8 +67,15 @@ function(find_qt_mkspec TARGET_PLATFORM_MKSPEC_OUT HOST_PLATFORM_MKSPEC_OUT EXT_
...
                     list(APPEND _test_triplets x86-windows x86-windows-static)
+                    if(VCPKG_TARGET_IS_MINGW)
+                        list(APPEND _test_triplets x64-mingw-dynamic x64-mingw-static)
+                        list(APPEND _test_triplets x86-mingw-dynamic x86-mingw-static)
+                    endif()
                 elseif($ENV{PROCESSOR_ARCHITECTURE} MATCHES "x86")
                     list(APPEND _test_triplets x86-windows x86-windows-static)
+                    if(VCPKG_TARGET_IS_MINGW)
+                        list(APPEND _test_triplets x86-mingw-dynamic x86-mingw-static)
+                    endif()
```
Neumann-A, line 78: you can probably change the whole code to use host triplets instead. When the code was written host triplets did not exist yet.

ports/qt5-base/portfile.cmake
```diff
@@ -84,6 +84,7 @@ qt_download_submodule(  OUT_SOURCE_PATH SOURCE_PATH
                             #patches/static_opengl.patch       #Use this patch if you really want to statically link angle on windows (e.g. using -opengl es2 and -static).
                                                                #Be carefull since it requires definining _GDI32_ for all dependent projects due to redefinition errors in the
                                                                #the windows supplied gl.h header and the angle gl.h otherwise.
+                            patches/fix-mingw-create_pc.patch  #Fix qmake mingw generator to write pkgconfig files also for "aux" pro files (qttools/src/designer/src/uiplugin)
```
Neumann-A, line 87: why write pc files for plugins?

ports/qt5-base/portfile.cmake
```diff
@@ -127,7 +128,6 @@ list(APPEND CORE_OPTIONS
     -system-sqlite
     -system-harfbuzz
     -icu
-    -no-vulkan
```
Neumann-A, line 130: why is this removed?

ports/qt5-base/portfile.cmake
```diff
@@ -258,6 +258,20 @@ set(DEBUG_OPTIONS
...
+        include("${cmake_vars_file}")
+        list(APPEND CORE_OPTIONS
+            QMAKE_CC="${VCPKG_DETECTED_CMAKE_C_COMPILER}"
+            QMAKE_CXX="${VCPKG_DETECTED_CMAKE_CXX_COMPILER}"
+            QMAKE_LINK="${VCPKG_DETECTED_CMAKE_CXX_COMPILER}"
+            QMAKE_LINK_C="${VCPKG_DETECTED_CMAKE_C_COMPILER}"
+        )
+        if(CMAKE_HOST_WIN32)
+            set(ENV{CC} "${VCPKG_DETECTED_CMAKE_C_COMPILER}")
+            set(ENV{CXX} "${VCPKG_DETECTED_CMAKE_CXX_COMPILER}")
```
Neumann-A, line 273: please apply to all platforms not just mingw.
for compiler flags use a approach similar to 
https://github.com/Neumann-A/vcpkg/commit/d882040de1ec668e6407ef61bf36db37fa0863d9 

also this needs to be added to `vcpkg_configure_qmake`

Setting CC and CXX should be avoided. As far as I understand you do it to make the host tool work. Please try to setup qt5 properly so that it has a host dependency on itself and use the artifacts from there instead.

ports/qt5-base/cmake/qt_port_functions.cmake
```diff
@@ -1,5 +1,13 @@
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
 
+# For mingw builds on Windows, establish a single msys root with perl.
+if(CMAKE_HOST_WIN32 AND VCPKG_TARGET_IS_MINGW)
+    vcpkg_acquire_msys(QT_MSYS_ROOT PACKAGES perl)
+    vcpkg_add_to_path(PREPEND "${QT_MSYS_ROOT}/usr/bin")
+    # Override vcpkg_find_acquire_program.
+    find_program(PERL "perl" PATHS "${QT_MSYS_ROOT}/usr/bin" NO_DEFAULT_PATH REQUIRED)
+endif()
```
Neumann-A, line 9: Why acquire perl? I would handle mingw just as the linux and osx triplets and assume a setup environment.

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -13,7 +13,13 @@ function(vcpkg_build_qmake)
...
 
     if(CMAKE_HOST_WIN32)
-        if (VCPKG_QMAKE_USE_NMAKE)
+        if(VCPKG_TARGET_IS_MINGW)
+            if(NOT DEFINED QT_MSYS_ROOT) # from qt_port_functions.cmake
+                vcpkg_acquire_msys(QT_MSYS_ROOT)
+            endif()
+            find_program(MAKE "make" PATHS "${QT_MSYS_ROOT}/usr/bin" NO_DEFAULT_PATH REQUIRED)
+            set(INVOKE "${MAKE}")
+        elseif (VCPKG_QMAKE_USE_NMAKE)
```
Neumann-A, line 22: Same as perl. I would assume make being available by the system in mingw.

ports/qt5-base/portfile.cmake
dg0yt: I would like to avoid changes to platform which I don't know enough about (in particular MSVC), or I don't know well enough how vpckg's flags interfer with Qt's flags.

I agree on `vcpkg_configure_qmake`. This is what overrides the absolute paths in the installed `qmodule.pri`.

Refactoring port qt5-base for host dependencies is also out of scope for this PR. AFAIU Qt5 wasn't prepared for this. Such a change should be started and reviewed separately.

scripts/cmake/vcpkg_build_qmake.cmake
dg0yt: Same answer as for perl, minus the breakage.

dg0yt (commented):
Thanks for the quick review!

ports/qt5-base/portfile.cmake
Neumann-A: > Because the plugin module name is refered to in other pc files.

that is strange for a plugin. Seems more like a bug in the makefile generation of qmake. Is this only happening for static builds?

> In Qt, definitions (cflags) carry information about feature availability (that is, at build time).

Normally this is done by the .prl files for qmake.

ports/qt5-base/cmake/find_qt_mkspec.cmake
Neumann-A: >  And I keep stumbling about the host triplet not being knowm in manifest mode.

that is only within your CMake code ;). Also you can just workaround that by referencing `${VCPKG_INSTALLED_DIR}/${VCPKG_HOST_TRIPLET}` and explicitly setting `VCPKG_HOST_TRIPLET` and preprend the tool paths to `CMAKE_PROGRAM_PATH`. But that is a different topic

ports/qt5-base/portfile.cmake
Neumann-A: >  I don't know well enough how vpckg's flags interfer with Qt's flags.

It doesn't matter. Qt needs to use vcpkg flags if it wants them or not ;) (the fact that qt6 is fine seems to indicate that it will just work.)

> AFAIU Qt5 wasn't prepared for this.

Qt5 itself supports cross-compiling. The scripts just needs adjustment for that. (e.g. no need to run qt5-base build via the bootstrap script). The rest ist just having the correct `qt.conf` for the job.

ports/qt5-base/cmake/qt_port_functions.cmake
Neumann-A: > I would have to add a condition arround vcpkg_find_acquire(PERL) in another spot, because that perl breaks the build,

The first thing `vcpkg_find_acquire_program` does is a `find_program` call (first within vcpkg downloads then outside.). So if your path is correct it should find the correct version (unless you already got the vcpkg one by running a normal windows triplet using a shared download folder).

ports/qt5-base/portfile.cmake
dg0yt: > > Because the plugin module name is refered to in other pc files.
> 
> that is strange for a plugin. Seems more like a bug in the makefile generation of qmake. Is this only happening for static builds?

Tested with native x64-mingw-dynamic. And the issue is raired by `vcpkg_fixup_pkgconfig`.
The point is that this is not a real run-time plugin (TEMPLATE = lib) but just some headers (TEMPLATE = aux). And the patch fixes "makefile generation of qmake", for mingw makefiles. For Unix makefiles, writing pc files for aux is the regular behaviour.

> > In Qt, definitions (cflags) carry information about feature availability (that is, at build time).
> 
> Normally this is done by the .prl files for qmake.

.prl is .pc is config.cmake, each in a world of its own.

ports/qt5-base/cmake/qt_port_functions.cmake
dg0yt: The first thing `vcpkg_find_acquire_program` does is test if the named variable is already set...
Anyway, there is a `find_program`. And for PERL, it first looks into the downloaded path. However if perl was downloaded and unpacked two months ago, it will be used now :boom:

ports/qt5-base/portfile.cmake
dg0yt: Anyway, good to mention `.prl` here. There is a new port conflict with unnamed `.prl` files on windows, and I assume it is caused by the patch. I will take another look on the Unix makefile generator.

ports/qt5-base/cmake/qt_port_functions.cmake
dg0yt: I don't even think there is something wrong with `vcpkg_find_acquire_program`. It is just a different use case here: For building qt5, I want to use a particular perl bnary which normally isn't found (when the msys root is not prepended to the path).

The current change applies this behaviour to ports using the special functions for building qt5 modules, and excludes normal apps. If this scope is considered too wide, it can be limited to qt5-base. (But I prefer to use the same environment for all qt5 modules.)

ports/qt5-base/portfile.cmake
dg0yt: For reference:
~~~
$ cat packages/qt5-tools_x64-linux/lib/pkgconfig/Qt5UiPlugin.pc 
prefix=${pcfiledir}/../..

exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include/qt5


Name: Qt5 UiPlugin
Description: Qt UiPlugin module
Version: 5.15.2
Cflags: -DQT_UIPLUGIN_LIB -I"${includedir}/QtUiPlugin" -I"${includedir}"
Requires: Qt5Core Qt5Gui Qt5Widgets
~~~
On x64-linux, there are also `lib/libQt5UiPlugin.prl` files for debug and release.

Neumann-A: I don't see how this pc file is helpful to consumers? In general the header path could be useful but I doubt that anything from there is even required for direct consumers. It is probably indirectly consumed by other Qt modules.

ports/qt5-base/cmake/qt_port_functions.cmake
Neumann-A: > And for PERL, it first looks into the downloaded path. However if perl was downloaded and unpacked two months ago, it will be used now

yeah that should be fixed. Probably by introducing a host subpath or something alike.

>  I want to use a particular perl bnary which normally isn't found (when the msys root is not prepended to the path).

personally I think you are responsible to setup the path correctly in this case. vcpkg really doesn't want to take control of the environment of other platforms than pure windows.

dg0yt: > personally I think you are responsible to setup the path correctly in this case. vcpkg really doesn't want to take control of the environment of other platforms than pure windows.

I do my best to control the environment. However, between msys root setup and perl acquisition, there is so much CMake code which is modified by other contributors, and often not with mingw in mind. That's why I want to set PERL early, just after acquiring msys root, before passing control to other code.
What I can do to improve consistency is to call `vcpkg_find_acquire_program(PERL)` instead of calling `find_program(PERL)`.

Neumann-A: Why do you even need perl in all Qt ports? Currently only qt5-base uses perl. Running syncqt is not done for the binary packages of the other qt modules. 

> What I can do to improve consistency is to call 

Do that if it is actually required.

ports/qt5-base/cmake/find_qt_mkspec.cmake
dg0yt: This whole section is somewhat strange: it is entered only when `DEFINED VCPKG_QT_HOST_PLATFORM` (a variable not used anywhere else) and `NOT "${_tmp_host_out}" MATCHES "${_tmp_host_out}"` (what?).

Neumann-A: `VCPKG_QT_HOST_PLATFORM` is a undocumented triplet hook. Just rip the whole thing apart and use the host triplet stuff instead.

Neumann-A: Also the check should probably check against `_tmp_targ_out` instead.

ports/qt5-base/portfile.cmake
dg0yt: > please apply to all platforms not just mingw.
> for compiler flags use a approach similar to

I noticed "deployment target" linker issues on osx now. Probably another case for passing compiler flags to Qt.

JackBoosY:
Any progress here?

dg0yt:
I didn't work on it recently but it is still desired. I want to continue with this PR.

ports/qt5-base/portfile.cmake
dg0yt: > > AFAIU Qt5 wasn't prepared for this.
> 
> Qt5 itself supports cross-compiling. The scripts just needs adjustment for that. (e.g. no need to run qt5-base build via the bootstrap script). The rest ist just having the correct `qt.conf` for the job.

Building mingw on Linux, I can now indeed start with the host `qmake`, skipping `configure`. However, the mingw triplet will still build all tools for the host. I cannot disable it.

I also need to keep the "host data" in the target triplet. Modules out data into the the mkspecs. But I could redirect the host lib and bin paths in the mingw build so that these binaries could be removed between before the end of the portfile.

JackBoosY:

dg0yt:
It is meant to be visible as a draft until I decide to close it. Would be glad to attract other people.

dg0yt:
> Ping @dg0yt for response. Is work still being done for this PR?

Yes. But it is waiting for #23668.

dg0yt:
Resolved by #29505, including contributions based on this PR.

PR #20165, opened 2021-09-15
[scripts-audit] Make buildsystem

JackBoosY:
Audit the Makefile build system according to the [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/cmake-guidelines.md).

Related: https://github.com/microsoft/vcpkg/issues/17691

- [x] Depends on PR https://github.com/microsoft/vcpkg/pull/19658

PR #20322, opened 2021-09-24
[scripts-audit] QMake buildsystem

JackBoosY:
Audit the QMake build system according to the [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/cmake-guidelines.md).

Related: https://github.com/microsoft/vcpkg/issues/17691


scripts/cmake/vcpkg_configure_qmake.cmake
```diff
@@ -21,36 +21,37 @@ The options passed to qmake.
...
-    cmake_parse_arguments(PARSE_ARGV 0 _csc "" "SOURCE_PATH" "OPTIONS;OPTIONS_RELEASE;OPTIONS_DEBUG;BUILD_OPTIONS;BUILD_OPTIONS_RELEASE;BUILD_OPTIONS_DEBUG")
+    cmake_parse_arguments(PARSE_ARGV 0 arg
+        ""
+        "SOURCE_PATH"
+        "OPTIONS;OPTIONS_RELEASE;OPTIONS_DEBUG;BUILD_OPTIONS;BUILD_OPTIONS_RELEASE;BUILD_OPTIONS_DEBUG"
+    )
      
     # Find qmake executable
-    set(_triplet_hostbindir ${CURRENT_INSTALLED_DIR}/tools/qt5/bin)
+    set(triplet_hostbindir "${CURRENT_INSTALLED_DIR}/tools/qt5/bin")
```
Neumann-A, line 31:
```suggestion
    set(triplet_hostbindir "${CURRENT_HOST_INSTALLED_DIR}/tools/qt5/bin")
```

scripts/cmake/vcpkg_configure_qmake.cmake
```diff
@@ -21,36 +21,37 @@ The options passed to qmake.
...
     if(DEFINED VCPKG_QT_HOST_TOOLS_ROOT_DIR)
-        find_program(QMAKE_COMMAND NAMES qmake PATHS ${VCPKG_QT_HOST_TOOLS_ROOT_DIR}/bin ${_triplet_hostbindir} NO_DEFAULT_PATH)
+        find_program(qmake_executable NAMES qmake PATHS "${VCPKG_QT_HOST_TOOLS_ROOT_DIR}/bin" "${triplet_hostbindir}" NO_DEFAULT_PATH)
     else()
-        find_program(QMAKE_COMMAND NAMES qmake PATHS ${_triplet_hostbindir} NO_DEFAULT_PATH)
+        find_program(qmake_executable NAMES qmake PATHS "${triplet_hostbindir}" NO_DEFAULT_PATH)
     endif()
```
Neumann-A, line 36: Remove `if()` since host tools are a thing now this is no longer necessary.

scripts/cmake/vcpkg_configure_qmake.cmake
```diff
@@ -21,36 +21,37 @@ The options passed to qmake.
 
 function(vcpkg_configure_qmake)
     # parse parameters such that semicolons in options arguments to COMMAND don't get erased
-    cmake_parse_arguments(PARSE_ARGV 0 _csc "" "SOURCE_PATH" "OPTIONS;OPTIONS_RELEASE;OPTIONS_DEBUG;BUILD_OPTIONS;BUILD_OPTIONS_RELEASE;BUILD_OPTIONS_DEBUG")
+    cmake_parse_arguments(PARSE_ARGV 0 arg
+        ""
+        "SOURCE_PATH"
+        "OPTIONS;OPTIONS_RELEASE;OPTIONS_DEBUG;BUILD_OPTIONS;BUILD_OPTIONS_RELEASE;BUILD_OPTIONS_DEBUG"
```
dg0yt, line 27: `BUILD_OPTIONS`, `BUILD_OPTIONS_RELEASE` and `BUILD_OPTIONS_DEBUG` don't have documenation. It is not obvious how they differ from `OPTIONS` etc.

dg0yt (commented):
Just when I submitted a PR touching one of these files...

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
     # parse parameters such that semicolons in options arguments to COMMAND don't get erased
-    cmake_parse_arguments(PARSE_ARGV 0 _csc "SKIP_MAKEFILES" "BUILD_LOGNAME" "TARGETS;RELEASE_TARGETS;DEBUG_TARGETS")
+    cmake_parse_arguments(PARSE_ARGV 0 arg
+        "SKIP_MAKEFILES"
+        "BUILD_LOGNAME"
+        "TARGETS;RELEASE_TARGETS;DEBUG_TARGETS"
+    )
 
     if(CMAKE_HOST_WIN32)
         if (VCPKG_QMAKE_USE_NMAKE)
```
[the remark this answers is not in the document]
JackBoosY: I also didn't find this, @Neumann-A should I remove the related code?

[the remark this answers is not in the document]
Neumann-A: > Is this variable documented?

No. It is only meant to be set by people for which jom acts up.

> @Neumann-A should I remove the related code?

No. It was introduced for people who have problems with jom. This happens rarely but it happens.

scripts/cmake/vcpkg_configure_qmake.cmake
JackBoosY: @Neumann-A The pipeline test results shows that this changes will requires all the qt component ports depend on host `qt5-base[core]`, should we do that?

Neumann-A: qt5-base needs a host dependency on itself. Just like qtbase

scripts/cmake/vcpkg_install_qmake.cmake
```diff
@@ -24,43 +24,43 @@ staging directories.
...
-    if(NOT RELEASE_LIBS AND NOT DEBUG_LIBS)
+    if(NOT release_libs AND NOT debug_libs)
         message(FATAL_ERROR "Build did not appear to produce any libraries. If this is intended, use `vcpkg_build_qmake()` directly.")
     endif()
-    if(RELEASE_LIBS)
-        file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/lib)
-        file(COPY ${RELEASE_LIBS} DESTINATION ${CURRENT_PACKAGES_DIR}/lib)
+    if(release_libs)
+        file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/lib")
+        file(COPY ${release_libs} DESTINATION "${CURRENT_PACKAGES_DIR}/lib")
```
JackBoosY, line 52: `release_libs` cannot use double quotes because it may be a list.

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
-            find_program(NMAKE nmake)
-            set(INVOKE "${NMAKE}")
-            get_filename_component(NMAKE_EXE_PATH ${NMAKE} DIRECTORY)
-            set(PATH_GLOBAL "$ENV{PATH}")
+            find_program(nmake_executable nmake)
+            set(invoke_command "${nmake_executable}")
+            get_filename_component(NMAKE_EXE_PATH "${nmake_executable}" DIRECTORY)
             set(ENV{PATH} "$ENV{PATH};${NMAKE_EXE_PATH}")
```
strega-nil-ms, line 34: This should be switched to use `vcpkg_host_path_list`.

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
 
     if(CMAKE_HOST_WIN32)
         if (VCPKG_QMAKE_USE_NMAKE)
-            find_program(NMAKE nmake)
-            set(INVOKE "${NMAKE}")
-            get_filename_component(NMAKE_EXE_PATH ${NMAKE} DIRECTORY)
-            set(PATH_GLOBAL "$ENV{PATH}")
+            find_program(nmake_executable nmake)
+            set(invoke_command "${nmake_executable}")
+            get_filename_component(NMAKE_EXE_PATH "${nmake_executable}" DIRECTORY)
```
strega-nil-ms, line 33:
```suggestion
            get_filename_component(nmake_exe_path "${nmake_executable}" DIRECTORY)
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
             vcpkg_find_acquire_program(JOM)
-            set(INVOKE "${JOM}")
+            set(invoke_command "${JOM}")
         endif()
     else()
-        find_program(MAKE make)
-        set(INVOKE "${MAKE}")
+        find_program(make_executable make)
+        set(invoke_command "${make_executable}")
     endif()
```
strega-nil-ms, line 43: This if-else block should be moved below the `env_path_backup` and `env_cl_backup`.

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
     if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "debug")
-        set(_buildname "DEBUG")
-        list(APPEND BUILDTYPES ${_buildname})
-        set(_short_name_${_buildname} "dbg")
-        set(_path_suffix_${_buildname} "/debug")
+        set(build_type_name "DEBUG")
+        vcpkg_list(APPEND buildtypes "${build_type_name}")
+        set(short_name_${build_type_name} "dbg")
+        set(path_suffix_${build_type_name} "/debug")
     endif()
```
strega-nil-ms, line 65:
```suggestion
    set(short_name_debug "dbg")
    set(path_suffix_debug "/debug")
    set(targets_debug "${arg_DEBUG_TARGETS}")
```

We can simplify this quite a bit (and same below)

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
+        vcpkg_list(APPEND buildtypes "${build_type_name}")
+        set(short_name_${build_type_name} "rel")
+        set(path_suffix_${build_type_name} "")
     endif()
-    unset(_buildname)
 
-    foreach(_buildname ${BUILDTYPES})
-        set(_installed_prefix_ "${CURRENT_INSTALLED_DIR}${_path_suffix_${_buildname}}")
-        set(_installed_libpath_ "${_installed_prefix_}/lib/${VCPKG_HOST_PATH_SEPARATOR}${_installed_prefix_}/lib/manual-link/")
+    foreach(build_type IN LISTS buildtypes)
```
strega-nil-ms, line 73:
```suggestion
    foreach(build_type IN ITEMS debug release)
        if(DEFINED VCPKG_BUILD_TYPE AND NOT "${VCPKG_BUILD_TYPE}" STREQUAL "${build_type}")
            continue()
        endif()
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
+        set(current_installed_prefix "${CURRENT_INSTALLED_DIR}${path_suffix_${build_type}}")
+        set(current_installed_libpath "${current_installed_prefix}/lib/${VCPKG_HOST_PATH_SEPARATOR}${current_installed_prefix}/lib/manual-link/")
 
-        vcpkg_add_to_path(PREPEND "${_installed_prefix_}/bin")
-        vcpkg_add_to_path(PREPEND "${_installed_prefix_}/lib")
+        vcpkg_add_to_path(PREPEND "${current_installed_prefix}/bin")
+        vcpkg_add_to_path(PREPEND "${current_installed_prefix}/lib")
```
strega-nil-ms, line 78:
```suggestion
        vcpkg_add_to_path(PREPEND "${current_installed_prefix}/lib" "${current_installed_prefix}/bin")
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
 
         # We set LD_LIBRARY_PATH ENV variable to allow executing Qt tools (rcc,...) even with dynamic linking
         if(CMAKE_HOST_UNIX)
             if(DEFINED ENV{LD_LIBRARY_PATH})
-                set(_ld_library_path_defined_ TRUE)
-                set(_ld_library_path_backup_ $ENV{LD_LIBRARY_PATH})
-                set(ENV{LD_LIBRARY_PATH} "${_installed_libpath_}${VCPKG_HOST_PATH_SEPARATOR}${_ld_library_path_backup_}")
+                set(is_ld_library_path_defined TRUE)
+                set(ld_library_path_backup $ENV{LD_LIBRARY_PATH})
```
strega-nil-ms, line 84: it's easy to just write something more like:

```cmake
    if(DEFINED ENV{LD_LIBRARY_PATH})
        set(ld_library_path_backup "$ENV{LD_LIBRARY_PATH}")
        ...
    else()
        unset(ld_library_path_backup)
    endif()
```

and then check for `if(DEFINED ld_library_path_backup)`

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
-                set(ENV{LD_LIBRARY_PATH} "${_installed_libpath_}")
+                set(is_ld_library_path_defined FALSE)
+                set(ENV{LD_LIBRARY_PATH} "${current_installed_libpath}")
             endif()
         endif()
         
-        list(APPEND _csc_${_buildname}_TARGETS ${_csc_TARGETS})
-        if(NOT _csc_SKIP_MAKEFILES)
-            run_jom(qmake_all makefiles ${_short_name_${_buildname}})
+        vcpkg_list(APPEND arg_${build_type}_TARGETS ${arg_TARGETS})
```
strega-nil-ms, line 92:
```suggestion
        vcpkg_list(SET targets ${targets_${build_type}} ${arg_TARGETS})
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
+        if(NOT arg_SKIP_MAKEFILES)
+            z_run_jom_build(qmake_all makefiles ${short_name_${build_type}})
```
strega-nil-ms, line 94:
```suggestion
            z_run_jom_build(qmake_all makefiles "${short_name_${build_type}}")
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
         endif()
-        run_jom("${_csc_${_buildname}_TARGETS}" ${_csc_BUILD_LOGNAME} ${_short_name_${_buildname}})
+        z_run_jom_build("${arg_${build_type}_TARGETS}" ${arg_BUILD_LOGNAME} ${short_name_${build_type}})
```
strega-nil-ms, line 96:
```suggestion
        z_run_jom_build("${targets}" "${arg_BUILD_LOGNAME}" "${short_name_${build_type}}")
```

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,104 @@ vcpkg_build_qmake()
...
 
         # Restore backup
         if(CMAKE_HOST_UNIX)
-            if(_ld_library_path_defined_)
-                set(ENV{LD_LIBRARY_PATH} "${_ld_library_path_backup_}")                
+            if(is_ld_library_path_defined)
```
strega-nil-ms, line 100:
```suggestion
            if(DEFINED ld_library_path_backup)
```

strega-nil-ms (changes requested):
started reviewing :)

docs/maintainers/vcpkg_configure_qmake.md
```diff
@@ -10,14 +10,20 @@ vcpkg_configure_qmake(
...
 
 ### SOURCE_PATH
 The path to the *.pro qmake project file.
 
 ### OPTIONS, OPTIONS\_RELEASE, OPTIONS\_DEBUG
-The options passed to qmake.
+The options passed to qmake to the configure step.
+
+### BUILD\_OPTIONS, BUILD\_OPTIONS\_RELEASE, BUILD\_OPTIONS\_DEBUG
+The options passed to qmake to the build step.
```
dg0yt, line 26: This may have been the intention by name, but does it make sense for qmake? There is no "build step" in `vcpkg_configure_qmake` and also no documented effect of `--` in https://doc.qt.io/qt-5/qmake-manual.html.
CC @Neumann-A

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,105 @@ vcpkg_build_qmake()
...
-                set(_ld_library_path_backup_ $ENV{LD_LIBRARY_PATH})
-                set(ENV{LD_LIBRARY_PATH} "${_installed_libpath_}${VCPKG_HOST_PATH_SEPARATOR}${_ld_library_path_backup_}")
+                set(ld_library_path_backup "$ENV{LD_LIBRARY_PATH}")
+                unset(ENV{LD_LIBRARY_PATH})
+                foreach(current_path IN ITEMS ${current_installed_libpath})
+                    vcpkg_host_path_list(APPEND ENV{LD_LIBRARY_PATH} "${current_path}")
+                endforeach()
+                foreach(current_path IN ITEMS ${ld_library_path_backup})
+                    vcpkg_host_path_list(APPEND ENV{LD_LIBRARY_PATH} "${current_path}")
+                endforeach()
```
JackBoosY, line 83: Same with https://github.com/microsoft/vcpkg/pull/20165#discussion_r718325877:
Since `current_installed_libpath` and `ld_library_path_backup` already contain host separator, we can't pass it into environment variables using `vcpkg_host_path_list `directly.

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,105 @@ vcpkg_build_qmake()
...
             else()
-                set(_ld_library_path_defined_ FALSE)
-                set(ENV{LD_LIBRARY_PATH} "${_installed_libpath_}")
+                unset(ld_library_path_backup)
+                unset(ENV{LD_LIBRARY_PATH})
+                foreach(current_path IN ITEMS ${current_installed_libpath})
+                    vcpkg_host_path_list(APPEND ENV{LD_LIBRARY_PATH} "${current_path}")
+                endforeach()
```
JackBoosY, line 89: Same as above.

docs/maintainers/vcpkg_configure_qmake.md
Neumann-A: It just means non parametric options after the `--` . `BUILD_OPTIONS` might be the wrong name here and it was a bit influenced by qt5-base bootstrap script where it is needed to pass other variables to the next build/configure step. `BUILD_OPTIONS` is used in in e.g. qt5-webengine.

strega-nil-ms (changes requested):
some stuff to change :)

Neumann-A:
@Osyotr neither respect the cmake toolchain. Qt5-base runs a bootstrap script and the flags are not passed and the rest uses the script here

strega-nil-ms (commented):
Continue marking changes to make

scripts/cmake/vcpkg_build_qmake.cmake
```diff
@@ -8,102 +8,79 @@ vcpkg_build_qmake()
 ```
 #]===]
 
+function(z_run_jom_build targets log_prefix log_suffix)
```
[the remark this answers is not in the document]
JackBoosY: Since this function has the prefix `z_`, it will only be used in this file, so I think this is also acceptable.
`invoke_command` is not a tool, so I think it should be lowercase.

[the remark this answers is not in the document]
dg0yt: `z_run_jom_build` is exposed in the stack trace on port build failures. In this situation, it is inviting to be misunderstood as "Oh no, `jom` is wrong on POSIX, vcpkg [!] broke my port again."

I know `invoke_command` is a variable. I complain that it is passed to `z_run_jom_build` by context, not as a proper formal parameter. Ideally, a function depends only on its input parameters, and has no side effects.

strega-nil-ms (approved):

PR #20393, opened 2021-09-27
[vcpkg_acquire_msys] Add gzip to default packages

wrobelda:
**Describe the pull request**
Adds `gzip` to the list of default msys environment. `gzip` is used by `autopoint`, provided by `gettext` tools. 

- #### What does your PR fix?  
  Fixes #20392

  No change.

  Yes.

  N/A

### DIFF
--- a/docs/maintainers/vcpkg_acquire_msys.md
+++ b/docs/maintainers/vcpkg_acquire_msys.md
@@ -25,7 +25,7 @@ To ensure a package is available: `vcpkg_acquire_msys(MSYS_ROOT PACKAGES make au
 ### NO_DEFAULT_PACKAGES
 Exclude the normal base packages.
 
-The list of base packages includes: bash, coreutils, sed, grep, gawk, diffutils, make, and pkg-config
+The list of base packages includes: bash, coreutils, sed, grep, gawk, gzip, diffutils, make, and pkg-config
 
 ### DIRECT_PACKAGES
 A list of URL/SHA512 pairs to acquire in msys.
--- a/scripts/cmake/vcpkg_acquire_msys.cmake
+++ b/scripts/cmake/vcpkg_acquire_msys.cmake
@@ -24,7 +24,7 @@ To ensure a package is available: `vcpkg_acquire_msys(MSYS_ROOT PACKAGES make au
 ### NO_DEFAULT_PACKAGES
 Exclude the normal base packages.
 
-The list of base packages includes: bash, coreutils, sed, grep, gawk, diffutils, make, and pkg-config
+The list of base packages includes: bash, coreutils, sed, grep, gawk, gzip, diffutils, make, and pkg-config
 
 ### DIRECT_PACKAGES
 A list of URL/SHA512 pairs to acquire in msys.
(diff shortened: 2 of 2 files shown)


JackBoosY:
Gzip is not required by most ports, I think you should pass `ADDITIONAL_MSYS_PACKAGES=gzip` into `vcpkg_configure_make`.

JackBoosY:
> > Gzip is not required by most ports, I think you should pass `ADDITIONAL_MSYS_PACKAGES=gzip` into `vcpkg_configure_make`.
> 
> The thing is this is not a port requirement, but the `autopoint` provided by `gettext[tools]` needs it (in some cases?). Adding it to a port that depends on gettext's autopoint is a workaround, not a solution.

Ahh, yes. However, you should add `gzip` to `vcpkg_configure_make` here:
https://github.com/microsoft/vcpkg/blob/8e0a801c6253a2c8e81f59427a4a5d66f10c0ca4/scripts/cmake/vcpkg_configure_make.cmake#L309
Add the following code below:
```cmake
    if (_csc_AUTOCONFIG)
        list(APPEND _csc_ADDITIONAL_MSYS_PACKAGES gzip)
    endif()
```

wrobelda:
> Ahh, yes. However, you should add `gzip` to `vcpkg_configure_make` here:

Thanks, that makes sense. Will fix it and let you know.

JackBoosY:
Sorry, just add `gzip` to this list:
https://github.com/microsoft/vcpkg/blob/8e0a801c6253a2c8e81f59427a4a5d66f10c0ca4/scripts/cmake/vcpkg_configure_make.cmake#L308

wrobelda:
> Thanks, that makes sense. Will fix it and let you know.

@JackBoosY fixed now.

JackBoosY:
The `libosip2:x86-windows` regression is not related to this changes.

JackBoosY (approved):

dg0yt:
There is too much "in the future" in the last comments. For this PR *now*, my recommendation is to remove the condition. Say yes and no, so that the author knows how to proceed.

JackBoosY (approved):

strega-nil-ms (approved):
LGTM!

dg0yt:
> Perhaps this would be a `vcpkg-port-config.cmake` for `gettext[tools]` that runs `vcpkg_acquire_msys2(GZIP_ROOT PACKAGES gzip NO_DEFAULT_PACKAGES)` and then adds it to the path.
> 
> Looking at the source of `autopoint.in`[1](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=gettext-tools/misc/autopoint.in;h=dfecd5089f804b27f10bd3ed73ee19605c578039;hb=HEAD), I see three hits for `gzip`, all of which are easily matched by the regex `" gzip "`. Therefore, it would also be viable to do something along the lines of:
> 
> ```cmake
> vcpkg_acquire_msys2(GZIP_ROOT PACKAGES gzip NO_DEFAULT_PACKAGES)
> file(INSTALL "${GZIP_ROOT}" DESTINATION ${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin RENAME gzip)
> vcpkg_replace_string(${CURRENT_PACKAGES_DIR}/tools/${PORT}/bin " gzip " " \"$bindir/gzip/bin/gzip\" ")
> ```
> 
> This ensures that the package remains portable and fully encapsulated.

Note that this may need more mingw/msys runtime environment setup than just copying of the contents of `gzip`. It seems to me that this leads into another source of fragility with regard to mingw/msys runtime.

If you want an alternative to `gzip`, there is `cmake -E tar`.

JackBoosY:
@dg0yt How to extend an msys root? What's the problem with this?

dg0yt:
> Just switch everything to vcpkg-msys-* ports. Then there is one root for msys

@Neumann-A As discussed before, that is a long-term perspective. (Who will do it, and when?) ATM we can do only a minimal change. Not to mention the unfinished script audit.

Neumann-A:
the reason different msys roots have been introduced is to find hidden dependency issues with msys which constantly broke CI.

ras0219-msft (approved):
While I do still think my suggested approach above of encapsulating the gzip dependency would be the best option, gzip is small enough and likely enough to be commonly needed that it seems acceptable to add to the base installation. This is a unique case and is not intended to set a precedent.

PR #20422, opened 2021-09-29
[OpenMVS] restore deprecated cmake scripts for configure, build & fixup targets

cenit:
During boost 1.77 upgrade, openmvs was updated to new vcpkg-cmake port for configure, build & fixup.
Unfortunately, it does not work correctly and consuming the library since then is broken.

Inside `OpenMVSTargets.cmake` the fixup deprecated script works correctly and produces this line

```
set(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_DIR}/../..")
```

while the new one produces this broken line

```
set(_IMPORT_PREFIX "${_IMPORT_PREFIX}")
```

We should fix the fixup script, but since i am not such an expert in dealing with regex-es, i suggest for now to revert the upgrade to a working script, while we try to understand what went wrong there (usually the new one works, but not with this port)

@strega-nil @strega-nil-ms

### DIFF
--- a/ports/openmvs/portfile.cmake
+++ b/ports/openmvs/portfile.cmake
@@ -21,9 +21,8 @@ file(REMOVE "${SOURCE_PATH}/build/Modules/FindCERES.cmake")
 file(REMOVE "${SOURCE_PATH}/build/Modules/FindCGAL.cmake")
 file(REMOVE "${SOURCE_PATH}/build/Modules/FindEIGEN.cmake")
 
-vcpkg_configure_cmake(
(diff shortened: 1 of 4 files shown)


strega-nil-ms:
Can you just switch back to `vcpkg_fixup_cmake_targets`? There's no reason to switch back to `vcpkg_*_cmake` aiui.

cenit:
cmake errors due to including vcpkg-cmake port and using old scripts

cenit:
any news? openmvs is broken on master since 28th Sep...

PR #20428, opened 2021-09-29
[openssl] port update to OpenSSL 3.0.0 

### LINKED ISSUE #20031: [openssl] add new port openssl 3.0.0
Library name: OpenSSL v3.0.0

Library description:
OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols. It is also a general-purpose cryptography library.

Source repository URL:
https://github.com/openssl/openssl/releases

https://www.openssl.org/

https://www.openssl.org/blog/blog/2021/09/07/OpenSSL3.Final/
https://github.com/openssl/openssl/releases
https://www.openssl.org/news/changelog.md
https://www.openssl.org/news/changelog.html#openssl-30
https://github.com/openssl/openssl/blob/master/doc/man7/migration_guide.pod

luncliff:
### What does your PR fix?  

Resolve #20031 

Most of the part is from https://github.com/microsoft/vcpkg/pull/17949.

Currently, no `feature` is defined. It was too many for me to check.
Hope (future) actual users can contribute to it.

Every non-community triplet in the mainstream. 

Checked.

* Build Techniques
   * [x] Do not use vendored dependencies --> `no-zlib`, `no-makedepend`
* Versioning
   * [x] `version-semver`
   * [x] Update the version files in versions/ of any modified ports
* Enable existing users of the library to switch to vcpkg
   * [x] provide `usage` for CMake [`FindOpenSSL`](https://cmake.org/cmake/help/latest/module/FindOpenSSL.html), [`FindPkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)


ras0219-msft:
We will be seeking to replace our `openssl` port with this; could you move this PR to name this `openssl` instead?

luncliff:
@dg0yt If so, I'd like to hear about the vcpkg team opinion. I didn't check there was an API change.

dg0yt:
@luncliff I don't know if it is that bad after re-reading https://www.openssl.org/blog/blog/2021/09/07/OpenSSL3.Final/:
"OpenSSL 3.0 is a major release and not fully backwards compatible with the previous release. Most applications that worked with OpenSSL 1.1.1 will still work unchanged and will simply need to be recompiled (although you may see numerous compilation warnings about using deprecated APIs). Some applications may need to make changes to compile and work correctly, and many applications will need to be changed to avoid the deprecations warnings. We have put together a migration guide to describe the major differences in OpenSSL 3.0 compared to previous releases."

luncliff:
Thanks for the link, @dg0yt. I will check it with the guides in the source. (I have to find where, but anyway)

vcpkg CI can check whether ports break or not. So I will try port rename then review the failure logs...
(If things are bad than expectation, I can revert them and use name `openssl3`).

dg0yt:
> vcpkg CI can check whether ports break or not.

Yes, but be aware of cascaded skips, in particular for linux, and non-default features. On windows and osx, the default features may select OS crypto functions instead of openssl.
There is also a license change. It is probably only an issue for GPL2-only software but IANAL.
(Anyway, I hope it is possible to just use the openssl port name.)

wrobelda:
> Yes, but be aware of cascaded skips, in particular for linux, and non-default features. On windows and osx, the default features may select OS crypto functions instead of openssl. There is also a license change. It is probably only an issue for GPL2-only software but IANAL. (Anyway, I hope it is possible to just use the openssl port name.)

All is true, but this wouldn't really be a problem if specifying a minimum/maximum/particular version of a dependency was actually possible in the manifest file.

dg0yt (commented):
:+1: for this work.

I don't see the installation of pkg-config files. Ideally, there should be *three* `.pc` files, on *every* platform, cf. https://github.com/microsoft/vcpkg/issues/20563.

luncliff:
Ah, so there is a removed API... It doesn't look like a simple problem if there are more.

luncliff:
Sorry for no updates... I'm lack of time to keep up this work.

I don't know how to deal with those removed API now. I will resume this work in next week

JackBoosY:
So can you continue this PR?

luncliff:
Honestly, no. I will close this. 😑

talregev:
@luncliff I think we should have do it in a different approach. 
Do it as a separate port. opensssl3 like we have for opencv.  and in the new port, we will not support backward computability or other packages that depend on the old openssl. that how slowly other packages will switch to new openssl.

luncliff:
Resurrecting https://github.com/microsoft/vcpkg/pull/20428/commits/e3924572abedeff528c94874d79d32e71b4d4481 is easy, but there was another releases after that. Hope there were not much changes.

Sadly I can't manage my schedule right now. 😭
I think I won't be able to do this until Feb week 2. I will try then if there is no volunteer for this.

PR #20458, opened 2021-10-01
[proj4] Install pc file

dg0yt:
- #### What does your PR fix?  
  Installs the pkgconfig file for proj.
  This is missing only in the cmake build system of PROJ 7.2.1.
  This is a prerequisite for improving the port's feature interface without breaking depending ports such as libspatialite (cf. #20443, #20480).

  all (unchanged), no

  yes

  yes


dg0yt:
@strega-nil-ms This change uses `VCPKG_SYSTEM_LIBRARIES` from `vcpkg_common_definitions.cmake` to determine the C++ standard library.
However, this variable doesn't have the literal name (`stdc++` etc.) but escapes `+` (`stdc\+\+`). Is this intentional?
(I also don't see how a triplet could override the C++ library e.g. for clang toolchains instead of g++ toolchains.)

ports/proj4/portfile.cmake
```diff
@@ -63,5 +64,44 @@ file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
...
+set(PACKAGE "${PORT}")
+set(requires_private "sqlite3")
+if(ENABLE_CURL)
+    string(APPEND requires_private " libcurl")
+endif()
+if(ENABLE_TIFF)
+    string(APPEND requires_private " libtiff-4")
+endif()
+set(libs_private "")
+if("c\\+\\+" IN_LIST VCPKG_SYSTEM_LIBRARIES)
```
strega-nil-ms, line 76:
```suggestion
if("c++" IN_LIST VCPKG_SYSTEM_LIBRARIES)
```

Since `IN_LIST` doesn't do a regex match, this is the correct way to check. (same for `stdc++` below)

strega-nil-ms (commented):
I need time to review, so assigning to myself.

ports/proj4/portfile.cmake
dg0yt: @strega-nil-ms Once again, it seems like part of what I wrote went unnoticed. Let me repeat:

> However, this variable doesn't have the literal name (`stdc++` etc.) but escapes + (`stdc\+\+`). Is this intentional?

And the source is:
https://github.com/microsoft/vcpkg/blob/b6b6a8b63007df2ff37167a1974652a8e948f045/scripts/cmake/vcpkg_common_definitions.cmake#L173-L181

strega-nil-ms: Okay, it seems like `VCPKG_SYSTEM_LIBRARIES` isn't used anymore anywhere; marking requires:discussion so we can discuss how we want to do this.

dg0yt: In case you are going to make changes:
It is not enough that triplets may add libraries. They need to override. E.g. for android, both c++ and stdc++ are possible alternatives.

dg0yt:
> After discussion, we've agreed that `VCPKG_SYSTEM_LIBRARIES` is not used anywhere and therefore we should just remove it entirely; the correct way to check for these things is to ask CMake which libraries it needs to link to privately. @ras0219 is currently trying to figure out _how_ to query CMake for these things.

Thanks for feedback. Well, removal was the most likely outcome that I expected when I started using it.
I wonder if you should deprecate the variable in one "release" before removing it before removing it in another, if you care for user-side overlay ports and registries. It is not used anywhere *in vcpkg*, but it is not an internal feature.

JackBoosY:

dg0yt:
FTR I did the experimentation in #20480. This is the result from the lesson-learnt.

JackBoosY (approved):

PR #20480, opened 2021-10-02
[libspatialite]  Use pkgconfig for nmake and autotools

dg0yt:
- #### What does your PR fix?  
  - [x] Use pkg-config for determining libraries to pass even to nmake, instead of hardcoding.
         Needed to allow features in dependencies.
  - [x] Fix building with mingw
  - [x] Install pc file for windows
  - [x] Add feature interface

  all, yes

  yes

  yes


Neumann-A:
I have the feeling you want to pass '--msvc-syntax' to pkg-config and avoid the transformation?

dg0yt:
> I have the feeling you want to pass '--msvc-syntax' to pkg-config and avoid the transformation?

Possibly, yes. TBH I know the option but I don't know enough about nmake to know for sure. I only see explicit `FOO_LIBS=D:/some_path/z.lib`, while the output from pkg-config is:
~~~
$ pkg-config.exe --libs --msvc-syntax zlib
/libpath:D:/msys64/mingw64/lib z.lib
~~~

If it is so easy, why all these explicit paths e.g. in https://github.com/microsoft/vcpkg/blob/master/ports/gdal/dependency_win.cmake? I'm really curious, because this nmake dependency stuff is a blocker for improvements.

Neumann-A:
It simply predates the pkg-config stuff and nobody tried to change it yet because it could be  a headache to get it to work again. I only introduced the pkg-config stuff in qt5-base because it fixed a CI build order/dependency regression.

dg0yt:
Okay, I will give it a try as long as CI is idle.

In the geographic ports, there is a lot of nmake+autotools meeting a lot of features, with significant effect on dependencies and license options. This really needs to be improved.

For `x_vcpkg_pkgconfig_get_modules` I have a draft change to catch and report errors (via `RESULT_VARIABLE`). I would suggest to add a `REQUIRED` keyword to enable this behaviour.

dg0yt:
> Okay, I will give it a try as long as CI is idle.

Works for `libspatialite`. This makes it much easier. :+1:

IIUC `vcpkg_build_nmake` doesn't do anyhing to pass `/LIBPATH` or adjust `ENV{LIB}` for vcpkg's installed dir.

dg0yt:
@PhoebeHui Ping for review. Needed for proceeding with #20443.

dg0yt:
Also needed for #21261.

dg0yt:
Ping again @PhoebeHui. This and #20443 was meant to be finished before #16494.

PhoebeHui:
@dg0yt, I'm sorry for my late response, I will test these features in my local machine firstly. And invite other members help review at the same time.

PhoebeHui:
All features test passed with x64-windows and x64-linux.

PhoebeHui (approved):
LGTM.

vicroms:
Thanks for the PR!

dg0yt:
Thank you @PhoebeHui @vicroms.

PR #20575, opened 2021-10-07
[boost-gil] remove dependency boost-filesystem

Osyotr:
This dependency is optional and only used to build and run tests and examples.
See https://github.com/boostorg/gil#requirements
However, there is also BOOST_GIL_IO_ADD_FS_PATH_SUPPORT which enables
boost::filesystem::path support for IO functions, but since users have to define
it manually, they must install the dependency manually, too.

A little benefit is that there is no need to build anything anymore: GIL itself and all its dependencies are header-only.


autoantwort:
> is vcpkg.json for autogenerated?

Yes the boost `vcpkg.json` files are autogenerated by [scripts/boost/generate-ports.ps1](https://github.com/microsoft/vcpkg/blob/master/scripts/boost/generate-ports.ps1)

strega-nil-ms:
I've made the required changes to generate-ports.ps1

PhoebeHui (approved):

yurybura (approved):

PR #20605, opened 2021-10-08
Remove tmxlite from ci.baseline.txt

chausner:
This removes tmxlite from ci.baseline.txt because the previously failing UWP builds should have been fixed in #20492.


ras0219 (approved):

PhoebeHui (approved):
Thanks for the PR!

PR #18242, opened 2021-06-02
[FreeRDP] update to 2.3.2

zhongzf:
Update FreeRDP to the latest version 2.3.2

Note : no feature need to test

PR #8060, opened 2019-09-05
[osg]Fix windows-dynamic/linux build.

JackBoosY:
This PR will fix:
1. Dynamic build.
2. Static build in Linux.
3. Dependency port missing.
4. Other errors.

Related: #7996 #8184.

Fix curl cannot link with openssl
~~Fix osg cannot link with libxml2~~

The component `present3D` has a dependency conflict with the component `osgstaticview`.
Reason:
They all require port `expat`, but for `osgstaticview`'s dependent port `fontconfig`, the library's link order must be **fontconfig->freetype->expat**. Since `osg`'s public dependency libraries is implemented using a list, the order of `expat` is **promoted** to before the `fontconfig` by `present3D`, so the `osgstaticview` build fails.

After a variety of attempts, fixing this issue still failed. So I disabled these two components in linux temporary.

PR #20620, opened 2021-10-09
[Many ports] Remove/fix unused patch files

chausner:
This removes patch files that are not referenced in the portfiles.

- 3fd: fix casing in filename (should not matter on Windows due to case-insensitive file system)
- freerdp: the patch got added in https://github.com/microsoft/vcpkg/pull/18242 but was never part of the portfile. It still seems relevant so I added it to the portfile now
- osg: the patch got added in https://github.com/microsoft/vcpkg/pull/8060 but was never part of the portfile. I am assuming it is no longer relevant because there have been many other updates to osg since then, so I removed it

For all other ports, the unused patches were part of the portfile in the past and got removed at some point but the patch files were forgotten to be removed.

  Not yet. Does port-version still need to be increased if files are removed that are not influencing the behaviour of the port?


NancyLi1013:
Hi @chausner Thanks for your PR.

Could you please also update the versions for these ports in vcpkg.json?

cenit:
looks like the forgotten freerdp patch was better left out 😃

NancyLi1013 (changes requested):

NancyLi1013 (approved):

PR #20681, opened 2021-10-12
[boost] fix boost-mpi dependency

yurybura:
Fixes #20676


PhoebeHui (approved):
Thanks for the fix again! I have double confirmed that boost-mpi has been removed from boost dependency.

PR #20717, opened 2021-10-13
[vcpkg docs] You can use the msbuild interation with multiple triplets since #19767

autoantwort:


PhoebeHui (approved):
Thanks for the PR!

PR #20778, opened 2021-10-16
[geogram] Switch to a GitHub mirror since upstream download is broken

vicroms:
Since a few days back the upstream download link for `geogram` has been down. 

This PR switches the port to use a mirror of the source in GitHub.


vicroms:
@saedrna @yann-lty @cenit @ChaoJia @dsmtE @xandox @FabienPean 

Since you have committed updates to this port at some point in time, does anyone have any opinions on using this alternative?

cenit:
ok for me. Not the ideal solution, but better than nothing.
I was thinking about writing to the authors to make a repository on github themselves, but it might not be so quick, so this is absolutely a good approach in the meantime

PR #20780, opened 2021-10-16
[xaudio2redist, directxtk] updated XAudio2 port to 1.2.8 and added cmake find

walbourn:
There's a new 1.2.8 release of [XAudio2Redist](https://www.nuget.org/packages/Microsoft.XAudio2.Redist/1.2.8) on NuGet which this updates to the port to use.

Also added a ``Findxaudio2redist.cmake`` to make it easier to consume this port.

Minor upstream change made to directxtk to use ``find_package`` instead of a bunch of CMake goo. https://github.com/microsoft/DirectXTK/commit/a0f8fcd89bd2468432f41888e6e7986de0d85180


walbourn:
@JonLiu1993 @BillyONeal The failures look like infrastructure things, not related to my port.

walbourn:
There is no 'upstream' for XAudio2Redist. It's a close-sourced DLL shipped by Microsoft as a NuGet package.  This port *is* the official way to get XAudio2Redist from vcpkg.

NancyLi1013 (changes requested):
Could you please add double quotes for the paths used in `xaudio2redist`?

NancyLi1013 (changes requested):

walbourn:
Re-verified

```
vcpkg install xaudio2redist:x86-windows
vcpkg install xaudio2redist:x64-windows

vcpkg install directxtk[xaudio2redist]:x86-windows --recurse
vcpkg install directxtk[xaudio2redist]:x64-windows --recurse
```

NancyLi1013 (approved):

NancyLi1013:
Thanks for your updates @walbourn.

walbourn:
@dg0yt  I'm happy to use the 'proper' Cmake pattern here, but I could use a pointer to an example that does it for this kind of scenario.

I'll need to do something similar for **dxsdk-d3dx** at some point, so I if I have the instructions I can use config files for that and then bump the version of the xaudio2redist port.

I largely based my find*.cmake on the half-dozen ones already in the vcpkg project.

dg0yt:
> I largely based my find*.cmake on the half-dozen ones already in the vcpkg project.

Config is mostly exported when building from source. But installed configs can be takes as an example, even the "unofficial" ones added by patches in vcpkg, and there is cmake's documentation, most relevant part:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-layout
Basically it takes a file named foo-config.cmake, and optionally a file dealing with the version, foo-config-version.cmake. Version, `REQUIRED` and `QUIET` are handled by `find_package`.

vicroms:
@walbourn I'm not opposed to merging a `FindXYZ.cmake` module but agree with @dg0yt that the config file is more desirable.  If you're willing to give it a try and have any problems let us know and we'll provide assistance.

walbourn:
@vicroms  Can we merge this one and I'll revisit it along with adding ``find_package`` support to ``dxsdk-d3dx`` which has the same issue?

walbourn:
> @vicroms @NancyLi1013 I'd like to get this in "as is" so I'm not blocking people using my latest CMakeLists that rely on using `find_package` for xaudio2redist.
> 
> I'll take a look at reworking the port as part of adding `find_package` support for dxsdk-d3dx which is on my backlog to get rid of the find.cmake approach.

Please see [this new PR](https://github.com/microsoft/vcpkg/pull/24054) for the config file revisit.

PR #20782, opened 2021-10-16
[libcanberra] Linux static .pc fix & other improvements

wrobelda:
**Describe the pull request**

- #### What does your PR fix?  
  - Fixes an issue with a missing lltdl dependency in .pc on Linux
  - incorporates Debian patches
  - featurizes all of the backends and establishes sane defaults for both platforms supported
    - macOS has "null" backend only since pulseaudio port is not available in vcpkg, the gstreamer currently doesn't build on macOS, while all other backends are irrelevant

  No change

  Yes

 Yes


PhoebeHui:
@wrobelda, thanks for testing these features! I don't want to block this PR, since it's in draft and 'gstreamer' is not a default feature, we can retest this feature after the PR merged.

PhoebeHui (approved):

PR #20855, opened 2021-10-20
[opencascade] Add feature rapidjson

### LINKED ISSUE #20821: [opencascade] Missing RapidJSON support so one cannot export to glTF
We want to be able to use opencascade to convert STEP files to glTF files. However, in order to do that, we need an opencascade with a `USE_RAPIDJSON` CMake cache var. 

**Proposed solution**
Include `rapidjson` feature here https://github.com/microsoft/vcpkg/blob/master/ports/opencascade/portfile.cmake#L23 and add it to the port dependencies.

The only alternative was to build the library from the scratch with a `USE_RAPIDJSON` set to true.

**Additional context**
We are using opencascade port on Windows 10 with MSBuild toolchain.

JackBoosY:
Fixes #20821.

Already tested this feature successfully on `x86-windows` and `x64-windows-static`.

### DIFF
--- a/ports/opencascade/portfile.cmake
+++ b/ports/opencascade/portfile.cmake
@@ -21,6 +21,7 @@ endif()
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     "freeimage"  USE_FREEIMAGE
     "tbb"        USE_TBB
+    "rapidjson"  USE_RAPIDJSON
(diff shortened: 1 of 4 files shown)


NancyLi1013 (approved):

haptiq-darko:
@JackBoosY I've just tested the prev commit and I confirm that it works.

PR #20860, opened 2021-10-20
[mbedtls] update to 3.1.0

### LINKED ISSUE #18356: [mbedtls] find_package not available
I've tried `find_package(mbedtls CONFIG REQUIRED)` but it throws me an error
```
CMake Error at /Users/kotori0/vcpkg/scripts/buildsystems/vcpkg.cmake:772 (_find_package):
  Could not find a package configuration file provided by "mbedtls" with any
  of the following names:

    mbedtlsConfig.cmake
    mbedtls-config.cmake

  Add the installation prefix of "mbedtls" to CMAKE_PREFIX_PATH or set
  "mbedtls_DIR" to a directory containing one of the above files.  If
  "mbedtls" provides a separate development package or SDK, be sure it has
  been installed.
```

I also tried to ignore that error and link directly:
```
target_link_libraries(project mbedtls)
```
But it will fail: 
```
ld: library not found for -lmbedtls
```
So what should I do to use mbedtls with vcpkg?

### LINKED ISSUE #25703: [mbedtls] update to 3.2.1
Library name: mbedtls

New version number: 3.2.1

https://github.com/Mbed-TLS/mbedtls/releases
https://github.com/Mbed-TLS/mbedtls-docs/blob/main/security-advisories/advisories/mbedtls-security-advisory-2022-07.md

JonLiu1993:
Fix #25703

Update mbedtls to the latest version 3.1.0 to support find_package 

Feature  pthreads tested successfully in the following triplet:
- x86-windows
- x64-windows
- x64-windows-static


dg0yt:
What is the state of this PR?

Is it feasible at all to update port mbedtls from 2 to 3? AFAIU there are API changes, and so it depends on downstream ports to adapt to these changes before it can be upgraded in vcpkg. 
This problem is mitigated by hiding incompatible combinations from CI by features, e.g. SSL alternatives.
In the meantime, it might be better to do a (security!) update within the version 2 family.

Port mebdtls also has issues with mingw, and it relies on invalid CMake Find module support from port pthreads.

JonLiu1993:
> What is the state of this PR?
> 
> Is it feasible at all to update port mbedtls from 2 to 3? AFAIU there are API changes, and so it depends on downstream ports to adapt to these changes before it can be upgraded in vcpkg. This problem is mitigated by hiding incompatible combinations from CI by features, e.g. SSL alternatives. In the meantime, it might be better to do a (security!) update within the version 2 family.
> 
> Port mebdtls also has issues with mingw, and it relies on invalid CMake Find module support from port pthreads.

Thanks for your suggestion, this pr is successfully tested locally but there are still some problems on CI, I will re-fix it when I have time

dg0yt:
> this pr is successfully tested locally but there are still some problems on CI

IMO the CI results confirm that the mbedtls API changes are not compatible with downstream ports.
I.e. mbedtls may require changes to downstream ports which are "feature development", not qualifying for patching in vcpkg.

JonLiu1993:
> > this pr is successfully tested locally but there are still some problems on CI
> 
> IMO the CI results confirm that the mbedtls API changes are not compatible with downstream ports. I.e. mbedtls may require changes to downstream ports which are "feature development", not qualifying for patching in vcpkg.

I'll go check my pr again to see if my changes introduced these issues

JonLiu1993:
Close this pr temporarily, I will reopen when I have time to fix it

ekilmer:
@JonLiu1993  The build failure of mbedtls on Linux can be fixed by this patch (tested locally) 
```diff
diff --git a/ports/mbedtls/portfile.cmake b/ports/mbedtls/portfile.cmake
index 5c1c84f41..9e8a16c9a 100644
--- a/ports/mbedtls/portfile.cmake
+++ b/ports/mbedtls/portfile.cmake
@@ -18,6 +18,8 @@ vcpkg_check_features(
         pthreads ENABLE_PTHREAD
 )
 
+vcpkg_find_acquire_program(PYTHON3)
+
 vcpkg_cmake_configure(
     SOURCE_PATH "${SOURCE_PATH}"
     OPTIONS
@@ -29,7 +31,7 @@ vcpkg_cmake_configure(
 
 vcpkg_cmake_install()
 
-vcpkg_cmake_config_fixup(CONFIG_PATH "CMake")
+vcpkg_cmake_config_fixup(CONFIG_PATH "cmake")
 
 vcpkg_copy_pdbs()
```
But as @dg0yt mentioned, there are multiple ports in vcpkg that either need to be updated or require non-trivial changes in order to support mbedtls v3.

Is there some vcpkg policy regarding how to update packages that aren't compatible with new major versions of their dependencies other than bugging the maintainers to fix it?

The above question is also related to the process of adding new ports that may rely on mbedtls v3+.

JonLiu1993:
@ekilmer ,Thanks for your investigation and help. At present, when we encounter this kind of problem, we need to fix the error that other ports cannot be built due to the update of this port version together.

BullyWiiPlaza:
[It would be nice to be able to use `mbedtls` with `CMake`](https://stackoverflow.com/questions/74902930) since as of right now only v3 supplies a `CMake` configuration file. Any workarounds for that or will the v3 migration happen any time soon?

EvanBalster:
It would be lovely if there were a feature or a separate `mbedtls3` port for installing newer versions of the library.  As it stands I'm building and installing it by hand.

PR #20940, opened 2021-10-22
[Docs] vcpkg_copy_tools cleans only copied tools

wrobelda:
Clarify that AUTO_CLEAN only removes the executables previously copied, not all of them.


PhoebeHui (approved):
Thanks for your updates!

PR #20947, opened 2021-10-23
[libmysql] Added limits include patch

silverqx:
Fixes numeric_limits is not a member of std build failure on GCC 11.2 in sql-common/sql_string.cc file. fixes #20946


NancyLi1013 (commented):
Would you like to make all paths quoted in this PR? Please see https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/cmake-guidelines.md

NancyLi1013 (changes requested):

NancyLi1013 (approved):

silverqx (commented):
I have tested it and it works good, thx @NancyLi1013 for help 🌹

PR #20953, opened 2021-10-23
[many ports] call vcpkg_fixup_pkgconfig()

autoantwort:
The ports generate pc files, but don't call vcpkg_fixup_pkgconfig() so that there are absolute paths in the pc files

The port list was generated like described [here](https://github.com/microsoft/vcpkg-tool/pull/172#discussion_r720096556)


PhoebeHui (approved):
LGTM. thanks for the PR!

BillyONeal (approved):

PR #20963, opened 2021-10-24
add vcpkg-qmake for qt6

Neumann-A:
QMake for Qt6 (~~untested since no dependent port so far.~~)


MehdiChinoune:
Any news?

ports/vcpkg-qmake/copyright
```diff
@@ -0,0 +1,23 @@
+Copyright (c) Microsoft Corporation
```
ras0219-msft, line 1: We may want to just provide a file that says "This helper port is part of vcpkg and is under the vcpkg license terms" instead of a duplicate license.

ports/vcpkg-qmake/vcpkg.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "vcpkg-qmake",
+  "version-date": "2021-10-23",
+  "dependencies": [
+    {
+      "name": "pkgconf",
+      "host": true,
```
ras0219-msft, line 7: It might make sense to make all these dependencies non-host and require this port to always be natively compiled. Then, you know you can use relative paths for host dependencies and `${CURRENT_INSTALLED_DIR}` for target dependencies.

ports/vcpkg-qmake/vcpkg_qmake_configure.cmake
```diff
@@ -0,0 +1,167 @@
...
+options directly passed to qmake with the form QMAKE_X=something or CONFIG=something 
+
+### OPTIONS, OPTIONS\_RELEASE, OPTIONS\_DEBUG
+The options passed after -- to qmake.
+
+#]===]
+include_guard(GLOBAL)
+function(vcpkg_qmake_configure)
+    cmake_parse_arguments(PARSE_ARGV 0 arg "" "SOURCE_PATH" "QMAKE_OPTIONS;QMAKE_OPTIONS_RELEASE;QMAKE_OPTIONS_DEBUG;OPTIONS;OPTIONS_RELEASE;OPTIONS_DEBUG")
+
```
ras0219-msft, line 37: Assuming the dependencies are converted to native dependencies, it might make sense to calculate a "QMAKE_HOST_DIR" from `${CMAKE_CURRENT_LIST_DIR}` and use that for all host dependencies in this function.

ports/vcpkg-qmake/copyright
Neumann-A: this is just copied over from vcpkg-cmake. I don't even now why you require this for the vcpkg scripts. (just touch the file vcpkg-tool wants to see)

ports/vcpkg-qmake/vcpkg_qmake_configure.cmake
Neumann-A: I don't get the added benefit of that. `CURRENT_HOST_INSTALLED_DIR` and `QMAKE_HOST_DIR` will point to the same directory however it is calculated

ports/vcpkg-qmake/vcpkg.json
Neumann-A: the command paths still need to be absolute. (Or i need to calculate the relative path from the working dir to the executable wherever that might be.)

JackBoosY:
Ping @ras0219-msft for response.

ras0219-msft:
LGTM, thanks for making the changes!

PR #20971, opened 2021-10-25
[vcpkg] Add aria2 as vcpkg required tool (step 1)

JackBoosY:
Related: https://github.com/microsoft/vcpkg/pull/20585.


PhoebeHui (approved):

BillyONeal (changes requested):

BillyONeal (approved):
@strega-nil / @strega-nil-ms confirmed this gets through the hash check but gets stopped by the unknown .bz2 extension. This is good to me as long as there's a tool update and we check that `vcpkg fetch aria2` on a Mac machine works.

JackBoosY:
@BillyONeal Ready to merge?

BillyONeal (approved):
Confirmed that this works as expected now on a MacOS box!

PR #20977, opened 2021-10-25
[qca] Fix usage

JackBoosY:
- Use `vcpkg_cmake_config_fixup` instead of hard code in `portfile.cmake`.
- Fix other macro in the config file.

Fixes #20970.


dg0yt (commented):
@JackBoosY Maybe it is time to get rid of the cmake script downloads after `# Remove these files on the cmake next update`?

m-kuhn (approved):
Fix confirmed, thanks for the quick reaction

NancyLi1013 (approved):

PR #21041, opened 2021-10-28
[alembic] Update to 1.8.3

chausner:
**Describe the pull request**

- #### What does your PR fix?  
  Updates alembic to 1.8.3. HDF5 support is now an optional feature and disabled by default.

  All except x86-windows, Yes

  Yes

  Yes


ports/alembic/vcpkg.json
```diff
@@ -1,12 +1,19 @@
...
   "name": "alembic",
-  "version": "1.7.16",
-  "port-version": 1,
+  "version": "1.8.3",
   "description": "Alembic is an open framework for storing and sharing scene data that includes a C++ library, a file format, and client plugins and applications.",
   "homepage": "https://alembic.io/",
-  "supports": "windows & !arm & !uwp & !static",
+  "supports": "!x86",
   "dependencies": [
     "hdf5",
```
simogasp, line 8: would you mind to add `hdf5` as an optional feature?
Nowadays it is disabled by default because it is less efficient (@lamiller0 can confirm or not).
You can replace it with
```json
 "features": 
  {
    "hdf5": 
    {
      "description": "Build with hdf5 support",
      "dependencies": ["hdf5"]
    }
  }
```
and then in the portfile before the call to `vcpkg_cmake_configure` you need to add

```cmake
        FEATURES
            hdf5    USE_HDF5 
)

vcpkg_cmake_configure(
    SOURCE_PATH "${SOURCE_PATH}"
    OPTIONS
        -DALEMBIC_SHARED_LIBS=${ALEMBIC_SHARED_LIBS}
        -DUSE_TESTS=OFF
        ${FEATURE_OPTIONS}
)
```

Like that you can build with hdf5 by asking it explicitly with  `alembic[hdf5]`.

I know it breaks the compatibility with the current behaviour but it seems more logic to stick with the default options of the library.
If we really want to maintain compatibility and providing the ability to build without it you can also add it to the default features like this
```json
  "dependencies": [
    "imath",
    {
      "name": "vcpkg-cmake",
      "host": true
    },
    {
      "name": "vcpkg-cmake-config",
      "host": true
    },
   "features": 
  {
    "hdf5": 
    {
      "description": "Build with hdf5 support",
      "dependencies": ["hdf5"]
    }
  },
  "default-features": [
    "hdf5"
  ]
```
like that one can still build without asking for `alembic[core]`.
But IMO it's better the first option.

What do you think?

chausner: Done in 031d1a6d8411e3834a5c84ae653d9c44e8636682. I don't have an opinion on whether it should be on or off by default, so I followed your suggestion.

simogasp: great, thanks!

dg0yt:
>     * #### Which triplets are supported/not supported? Have you updated the [CI baseline](https://github.com/microsoft/vcpkg/blob/master/scripts/ci.baseline.txt)?
>       All except x86-windows, Yes

Which triplets are really supported? (Thinking beyond CI.)
Before this PR, non-windows triplets were implicitly excluded by `!static`, and x86-windows was included.
Now, by `!x86`, x86 windows is unsupported, as is x86-linux, but arm-linux and x64-linux are supported?

chausner:
> Which triplets are really supported? (Thinking beyond CI.) Before this PR, non-windows triplets were implicitly excluded by `!static`, and x86-windows was included. Now, by `!x86`, x86 windows is unsupported, as is x86-linux, but arm-linux and x64-linux are supported?

x86 is only broken on Windows. UWP also seems to be not supported after we disabled HDF5 support. I updated the`supports` here: 423bc41ca3b039207bc4550d8d106f13e2fc0a06

JonLiu1993:
- x64-windows
- x64-windows-static
- x64-linux

PR #21042, opened 2021-10-28
[libLAS] Remove dependency on all of boost

calumr:
I suspect this was a shortcut for someone, but it's bloating our build. There's no need to pull in all of boost & all it's dependencies.


PhoebeHui (changes requested):
@calumr, thanks for your PR! Could update the version database by executing './vcpkg x-add-version liblas' and submit the changes?

PhoebeHui (approved):
Thanks for your updates!

PR #20988, opened 2021-10-25
[vcpkg_download_distfile] Check SHA512 for already downloaded files

vicroms:
Fixes #20874 

@JackBoosY  I noticed you have an open PR for `vcpkg_download_distfile`, you may want to incorporate these changes there instead.

PR #21054, opened 2021-10-29
[snappy] Fix hash and switch to vcpkg_cmake_*

Osyotr:


cenit:
another break due to #20988 
why PR check didn’t catch all of these regressions, so that we could fix it and made routine case insensitive as it should be from the beginning?

PhoebeHui:
After https://github.com/microsoft/vcpkg/pull/20988, vcpkg could detect the case issue, however, I checked https://dev.azure.com/vcpkg/public/_build/results?buildId=61942&view=logs&j=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&t=491b9f02-7edc-5990-cda1-511e95a3768e, snappy used cached binary instead when test the changes in https://github.com/microsoft/vcpkg/pull/20988. So the we missed the regressions.

PhoebeHui (approved):
@Osyotr, thanks for the PR!

Issue #20469, opened 2021-10-01
Remove absolute paths from installed files 


autoantwort:
While developing https://github.com/microsoft/vcpkg-tool/pull/172 it was discovered that many ports (160) have absolute paths the the packages dir in their result. These paths must be updated to relative paths, for example by `vcpkg_fixup_pkgconfig`. 
<details><summary>List of ports</summary>

- [x] ace
- [x] bcg729
- [x] cgns
- [x] cminpack
- [x] dcmtk
- [x] fontconfig
- [x] ftgl
- [x] gdcm
- [x] gdk-pixbuf
- [x] glib
- [ ] glibmm
- [x] graphite2
- [x] hdf5
- [x] hpx
- [ ] ignition-msgs1
- [x] ismrmrd
- [x] lager
- [x] libftdi1
- [x] libgit2
- [ ] libheif
- [ ] libpmemobj-cpp
- [x] libpng
- [x] libwebsockets
- [x] minitrace
- [x] mongo-c-driver
- [x] mongo-cxx-driver
- [x] netcdf-c
- [x] ompl
- [x] plplot
- [x] protopuf
- [x] qpid-proton
- [x] sdformat6
- [x] shapelib
- [x] simage
- [x] symengine
- [x] woff2
- [x] xtensor-blas
- [x] xtensor-io
</details>

<details><summary>List of wrong files</summary>

```
ace:x64-osx: x64-osx/share/ace/ace-devel.sh
ace:x64-osx: x64-osx/share/ace/bin/MakeProjectCreator/config/MPC.cfg
ace:x64-osx: x64-osx/share/ace/include/makeinclude/wrapper_macros.GNU
bcg729:x64-osx: x64-osx/share/Bcg729/cmake/Bcg729Config.cmake
cgns:x64-osx: x64-osx/share/cgns/cgnsBuild.defs
dcmtk:x64-osx: x64-osx/include/dcmtk/config/osconfig.h
fontconfig:x64-osx: x64-osx/etc/fonts/conf.d/README
fontconfig:x64-osx: x64-osx/etc/fonts/fonts.conf
ftgl:x64-osx: x64-osx/share/ftgl/FTGL/FTGLConfig.cmake
gdcm:x64-osx: x64-osx/include/gdcmConfigure.h
gdk-pixbuf:x64-osx: x64-osx/debug/gdk-pixbuf-2.0/2.10.0/loaders.cache
[...]

PR #21063, opened 2021-10-30
[many ports] Call vcpkg_fixup_pkgconfig

autoantwort:
Fix *some* ports of #20469


autoantwort:
Closed in favor of #21199

PR #21068, opened 2021-10-30
[quantlib] update to 1.24

### LINKED ISSUE #21067: [quantlib] update to 1.24
Library name: QuantLib

New version number: 1.24

https://github.com/lballabio/QuantLib/releases/tag/QuantLib-v1.24

sweemer:
Update quantlib to v1.24.

- #### What does your PR fix?  
  Fixes #21067

static/dynamic on Linux and static on Windows, Yes

 Yes

Yes


ports/quantlib/vcpkg.json
```diff
@@ -1,40 +1,9 @@
...
-    "boost-serialization",
-    "boost-signals2",
-    "boost-smart-ptr",
-    "boost-thread",
-    "boost-tuple",
-    "boost-type-traits",
-    "boost-ublas",
-    "boost-unordered",
-    "boost-utility"
+    "boost"
```
NancyLi1013, line 7: We prefer to use the specific components of `boost `as the dependencies instead of `boost`. So could you please update this?

scripts/ci.baseline.txt
```diff
@@ -1256,6 +1256,10 @@ qt5-x11extras:x86-windows=fail
 qt5-x11extras:x64-windows=fail
 qt5-x11extras:x64-windows-static=fail
 qt5-x11extras:x64-windows-static-md=fail
+quantlib:x86-windows=fail
+quantlib:x64-windows=fail
+quantlib:x64-windows-static-md=fail
+quantlib:arm64-windows=fail
```
NancyLi1013, line 1262: We should add `supports `field in vcpkg.json to describe this. Please also make a clarify why you add `quantlib:arm64-windows=fail` here.

NancyLi1013 (changes requested):

ports/quantlib/vcpkg.json
sweemer: I've changed back to specific components of `boost`, and also added `boost-test`, a new dependency as of version 1.24.

scripts/ci.baseline.txt
sweemer: I added a `supports` field to reflect the fact that dynamic builds on Windows are not supported. This explains why `arm64-windows` fails - it is dynamic too.

ports/quantlib/vcpkg.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "quantlib",
-  "version": "1.22",
+  "version": "1.24",
   "description": "The QuantLib C++ library",
   "homepage": "https://www.quantlib.org/",
+  "supports": "!(windows & dynamic)",
```
NancyLi1013, line 6: This is not correct. Dynamic is not the keyword for supports filed.

scripts/ci.baseline.txt
```diff
@@ -1217,6 +1217,9 @@ qpid-proton:arm-uwp=fail
 qpid-proton:x64-uwp=fail
 qpid-proton:x64-windows-static=fail
 qt5-base:arm64-windows=fail
+quantlib:arm64-windows=fail
+quantlib:x64-windows=fail
+quantlib:x86-windows=fail
```
NancyLi1013, line 1222:
```suggestion
```
If it dpesn't support arm64, please add `supports: !(arm64 | windows)`. For the dynamic build, since you have added `vcpkg_check_linkage(ONLY_STATIC_LIBRARY)`, it will be built as static library even if you build dynamic. So it's unnecessary to add this here.

NancyLi1013 (changes requested):

ports/quantlib/vcpkg.json
sweemer: Yes sorry I have changed it to `!static` instead

scripts/ci.baseline.txt
sweemer: After fixing the `supports` declaration, these builds are working again, so no change required to `ci.baseline.txt`.

NancyLi1013 (approved):

PR #21079, opened 2021-10-30
[field3d] Adapt port to use imath dependency

chausner:
**Describe the pull request**

- #### What does your PR fix?  
  This updates port `field3d` to use the newer `imath` package instead of `openexr`, in preparation to update `openexr` later.

  As before, Yes

  Yes

  Yes

### DIFF
--- a/ports/field3d/0004_fix_imath_includes.patch
+++ b/ports/field3d/0004_fix_imath_includes.patch
@@ -1,0 +1,136 @@
+--- a/export/StdMathLib.h
++++ b/export/StdMathLib.h
+@@ -38,17 +38,17 @@
+ #ifndef _INCLUDED_Field3D_StdMathLib_H_
+ #define _INCLUDED_Field3D_StdMathLib_H_
+ 
+-#include <OpenEXR/ImathBox.h> 
+-#include <OpenEXR/ImathBoxAlgo.h>
+-#include <OpenEXR/ImathColor.h>
+-#include <OpenEXR/ImathHalfLimits.h>
+-#include <OpenEXR/ImathMatrix.h>
+-#include <OpenEXR/ImathMatrixAlgo.h>
+-#include <OpenEXR/ImathPlane.h>
+-#include <OpenEXR/ImathRandom.h> 
+-#include <OpenEXR/ImathRoots.h>
+-#include <OpenEXR/ImathVec.h>
+-#include <OpenEXR/half.h> 
++#include <Imath/ImathBox.h>
++#include <Imath/ImathBoxAlgo.h>
++#include <Imath/ImathColor.h>
++#include <Imath/ImathMatrix.h>
++#include <Imath/ImathMatrixAlgo.h>
++#include <Imath/ImathPlane.h>
++#include <Imath/ImathRandom.h>
++#include <Imath/ImathRoots.h>
++#include <Imath/ImathVec.h>
++#include <Imath/half.h>
++#include <Imath/halfLimits.h>
+ 
+ //----------------------------------------------------------------------------//
+ 
+--- a/export/Curve.h
++++ b/export/Curve.h
+@@ -53,8 +53,8 @@
+ 
+ #include <boost/lexical_cast.hpp>
+ 
+-#include <OpenEXR/ImathFun.h>
+-#include <OpenEXR/ImathMatrix.h>
++#include <Imath/ImathFun.h>
++#include <Imath/ImathMatrix.h>
+ 
+ //----------------------------------------------------------------------------//
+ 
+--- a/export/FieldWrapper.h
++++ b/export/FieldWrapper.h
+@@ -6,7 +6,7 @@
+ //------------------------------------------------------------------------------
+ 
+ // Library includes
+-#include <OpenEXR/ImathMatrixAlgo.h>
++#include <Imath/ImathMatrixAlgo.h>
+ 
+ // Project includes
+ #include "DenseField.h"
+--- a/export/SpiMathLib.h
++++ b/export/SpiMathLib.h
+@@ -38,20 +38,20 @@
+ #ifndef _INCLUDED_Field3D_SpiMathLib_H_
+ #define _INCLUDED_Field3D_SpiMathLib_H_
+ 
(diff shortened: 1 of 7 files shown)


NancyLi1013:
Hi @chausner 

Thanks for your PR.

Could you please help describe why do you use `imath `as the dependency?

chausner:
> Could you please help describe why do you use imath as the dependency?

In newer openexr versions, the developers have moved out some functionality into a separate library `imath` which most of the packages that currently depend on openexr will want to use instead. I am updating these dependent ports so that we can then update openexr as a second step (https://github.com/microsoft/vcpkg/pull/20957).

NancyLi1013:
Since you use `imath `as its dependency, it might be better to report an issue or submit a PR to upstream. What's your opinion about this?

chausner:
> Since you use `imath `as its dependency, it might be better to report an issue or submit a PR to upstream. What's your opinion about this?

This has already been suggested 4 months ago at https://github.com/imageworks/Field3D/issues/101 but it seems the library is not actively maintained anymore. See also https://github.com/imageworks/Field3D/issues/102.

chausner:
We'll still want to update this port because it currently blocks the update of the openexr library.

chausner:
@NancyLi1013 Please let me know if you need anything else from me to get this to a mergeable state.

ports/field3d/0004_fix_imath_includes.patch
```diff
@@ -0,0 +1,136 @@
...
++#include <Imath/ImathBoxAlgo.h>
++#include <Imath/ImathColor.h>
++#include <Imath/ImathMatrix.h>
++#include <Imath/ImathMatrixAlgo.h>
++#include <Imath/ImathPlane.h>
++#include <Imath/ImathRandom.h>
++#include <Imath/ImathRoots.h>
++#include <Imath/ImathVec.h>
++#include <Imath/half.h>
++#include <Imath/halfLimits.h>
```
JackBoosY, line 28: Can you double confirm what is the difference between the imath and imath files in OpenEXR?

JackBoosY (changes requested):

ports/field3d/0004_fix_imath_includes.patch
chausner: This is how I understand it: the Imath parts originally shipped as part of OpenEXR but at some point have been refactored out into a separate library. This PR basically switches to a newer Imath library version which happens to now live in a separate repository/port instead of being part of OpenEXR.

Since the port still compiles fine, I assume it is compatible to the newer Imath version but I have not used the library to confirm that assumption.

The reason why I would like to update this port is to be able to update OpenEXR to the latest version. As long as field3d depends on OpenEXR, we cannot update it because the latest OpenEXR versions no longer come with Imath included.

chausner: Here is some more information about migrating from OpenEXR 2.x to Imath 3.x: https://github.com/AcademySoftwareFoundation/Imath/blob/master/docs/PortingGuide2-3.md

JackBoosY: Sadly, since your changes contains the source code, I think this should wait for the upstream approval.

chausner: It seems the upstream project is no longer actively maintained, see my earlier comment: https://github.com/microsoft/vcpkg/pull/21079#issuecomment-958047195

I still think it would be better to update this port with the patch than not updating OpenEXR and dependent ports at all. Or maybe stale ports should be removed if they prevent other, actively developed, ports from being updated.

JackBoosY: Yeah I know that, but we should follow the license.

fabiencastan: As explained here: https://github.com/microsoft/vcpkg/pull/20957
Vcpkg is now broken with 2 versions of imath installed at the same time. So a quick change is needed.

meshula:
As an OpenEXR maintainer, I can confirm that we've been trying to move Field3d forward for some time, but have been unable to verify so far that it is a maintained project. 

c.f. https://github.com/imageworks/Field3D/issues/102

The maintainer of OpenImageIO has removed field3d as a dependency, so we've done what we can identify so far, to remove field3d from dependency chains that also involve OpenEXR.

I did a small survey asking studios if they are using Field3d, and found out that some of them are indeed using it, but they build locally customized forks, using their own build systems, and do not integrate it with OpenImageIO, nor do they build it using vcpkg.

I don't know how to query the vcpkg graph for other upstream dependencies on Field3D, but in absence of any motion in the main repo for some time, I concur that the right thing to do is to elide field3d from vcpkg, at least temporarily.

chausner:
> I don't know how to query the vcpkg graph for other upstream dependencies on Field3D so I don't know the impact on vcpkg of removing Field3d,

A quick search shows that `openimageio` is currently the only package that depends on field3d. The [release notes](https://github.com/OpenImageIO/oiio/releases/tag/v2.3.9.1) of openimageio 2.3.9.1 say:
> Field3D, which is no longer actively supported, now has support disabled in the default build. To enable Field3D support, you must build with -DENABLE_FIELD3D=1. Note that we expect it to be no longer supported at all, beginning with OIIO 2.4.

So field3d is still an optional dependency that could theoretically be used with openimageio.

lgritz:
👋 Hi, maintainer of OpenImageIO here. All support for field3d has been removed from what will in the future be OpenImageIO 2.4. In the currently supported release, 2.3, we've changed the default to disable field3d, but our policy is not to completely remove support for a previously-supported format in the middle of the lifetime of an active release branch. 

That said, my recommendation is to just remove field3d from vcpkg, move imath and openexr to their current 3.1 releases, and let OpenImageIO build without the field3d support, which as far as I can tell, nobody is depending on anyway.

JackBoosY:
Ping @chausner for response, please follow the oiio maintainer's suggestion.

chausner:
I have opened https://github.com/microsoft/vcpkg/pull/22463 which removes the `field3d` port and the corresponding feature of `openimageio` and will close this PR.

PR #21087, opened 2021-10-31
[highway] Add new port

chausner:
**Describe the pull request**

- #### What does your PR fix?  
  This adds a new port for [highway](https://github.com/google/highway) (aka libhwy) 0.14.2

  Not sure yet, Yes

  Yes

  Yes

### DIFF
--- a/ports/highway/portfile.cmake
+++ b/ports/highway/portfile.cmake
@@ -16,6 +16,10 @@ vcpkg_cmake_configure(
 
 vcpkg_cmake_install()
 
+# remove test-related pkg-config files that break vcpkg_fixup_pkgconfig
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libhwy-test.pc")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig/libhwy-test.pc")
+
 vcpkg_fixup_pkgconfig()
 
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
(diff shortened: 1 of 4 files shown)


ports/highway/vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "highway",
+  "version-semver": "0.14.2",
```
chausner, line 3: Semver according to https://github.com/google/highway/blob/acc21a670582d791c6790e1f995029f57a930784/hwy/highway.h#L28.

ports/highway/portfile.cmake
```diff
@@ -0,0 +1,27 @@
...
+    SOURCE_PATH "${SOURCE_PATH}"
+    OPTIONS
+        -DBUILD_TESTING=OFF
+)
+
+vcpkg_cmake_install()
+
+# remove test-related pkg-config files that break vcpkg_fixup_pkgconfig
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libhwy-test.pc")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig/libhwy-test.pc")
```
JackBoosY, line 21: I think we should fix that instead of remove them.

JackBoosY (changes requested):

ports/highway/portfile.cmake
chausner: libhwy-test are only tests. I am not sure why a pkg-config file is generated for it.

The problem is that those pkg-config files list "gtest" as a pkg-config dependency.

JackBoosY: Does `libhwy.pc` exist in _lib/pkgconfig_?

chausner: Yes, there are `libhwy.pc`, `libhwy-contrib.pc` and `libhwy-test.pc` getting installed into  `lib/pkgconfig`.

JackBoosY (approved):

PR #21092, opened 2021-10-31
[vcpkg] Revise appdeploy and copy_tool_dependencies

dg0yt:
- #### What does your PR fix?  
  - [appdeploy] Fixes deployment for mingw on linux by 
    - making path construction portable,
    - ~catching failing~ fixing mutex creation.
  - [copy_tool_dependencies] Enables deployment for debug config.
  - [copy_tool_dependencies] Fixes overwriting of log files within a single port build.
  - [copy_tool_dependencies] Enables actual logging ~at least when `DEBUG_PORT` is set (cf. `debug_message` in `ports.cmake`)~.

  all, no

  yes

  --

- ### For discussion
  - ~Logging (`-verbose`) might always be enabled. It doesn't clutter output.~
  - ~If there is a better idea how to deal with the mutex creation issue...~
  - To correctly handle some port being set to a single config (via triplet file), it would be necessary to also search the alternative `bin` directory.
  - It is not efficient to call `appdeploy.ps1` twice. It should better take a list of directories as search path instead of a single directory.
    And it should also take a list of executables (or just a single directory).

### DIFF
--- a/scripts/buildsystems/msbuild/applocal.ps1
+++ b/scripts/buildsystems/msbuild/applocal.ps1
@@ -19,14 +19,19 @@ function computeHash([System.Security.Cryptography.HashAlgorithm]$alg, [string]$
 }
 
 function getMutex([string]$targetDir) {
-    $sha512Hash = [System.Security.Cryptography.SHA512]::Create()
-    if ($sha512Hash) {
-        $hash = computeHash $sha512Hash $targetDir
-        $mtxName = "VcpkgAppLocalDeployBinary-" + $hash
-        return New-Object System.Threading.Mutex($false, $mtxName)
-    }
+    try {
+        $sha512Hash = [System.Security.Cryptography.SHA512]::Create()
+        if ($sha512Hash) {
+            $hash = computeHash $sha512Hash $targetDir
+            $mtxName = "VcpkgAppLocalDeployBinary-" + $hash
+            return New-Object System.Threading.Mutex($false, $mtxName)
+        }
 
-    return New-Object System.Threading.Mutex($false, "VcpkgAppLocalDeployBinary")
+        return New-Object System.Threading.Mutex($false, "VcpkgAppLocalDeployBinary")
+    }
+    catch {
+        return 0
+    }
 }
 
 # Note: this function signature is depended upon by the qtdeploy.ps1 script introduced in 5.7.1-7
--- a/scripts/buildsystems/msbuild/applocal.ps1
+++ b/scripts/buildsystems/msbuild/applocal.ps1
@@ -4,7 +4,7 @@ param([string]$targetBinary, [string]$installedDir, [string]$tlogFile, [string]$
 $g_searched = @{}
 # Note: installedDir is actually the bin\ directory.
 $g_install_root = Split-Path $installedDir -parent
-$g_is_debug = $g_install_root -match '(.*\\)?debug(\\)?$'
+$g_is_debug = (Split-Path $g_install_root -leaf) -eq 'debug'
 
 # Ensure we create the copied files log, even if we don't end up copying any files
 if ($copiedFilesLog)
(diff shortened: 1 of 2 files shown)


dg0yt:
The original mutex error on linux:
~~~
New-Object: /home/kpa/Projekte/vcpkg/vcpkg/scripts/buildsystems/msbuild/applocal.ps1:26:16
Line |
  26 |          return New-Object System.Threading.Mutex($false, $mtxName)
     |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling ".ctor" with "2" argument(s): "The filename, directory name, or volume label syntax is incorrect. :
     | 'VcpkgAppLocalDeployBinary-/67pRTB1Ups2rRbHaFf9MmG5+3JTeHXhgLDIy6xHo3Wsrcl2RLAMD0XgKTwuKrwUcHqwWB5xF+iuO1frfY0xZQ=='"
~~~

dg0yt:
> The original mutex error on linux

... is caused by the `/` in the mutex name.

JackBoosY:
@BillyONeal Can you please review this PR?
Thanks.

scripts/buildsystems/msbuild/applocal.ps1
```diff
@@ -37,22 +42,24 @@ function deployBinary([string]$targetBinaryDir, [string]$SourceDir, [string]$tar
...
-            $destModTime = (Get-Item $targetBinaryDir\$targetBinaryName).LastWriteTime
+        $sourceBinaryFilePath = Join-Path $SourceDir $targetBinaryName
+        $targetBinaryFilePath = Join-Path $targetBinaryDir $targetBinaryName
+        if (Test-Path $targetBinaryFilePath) {
+            $sourceModTime = (Get-Item $sourceBinaryFilePath).LastWriteTime
+            $destModTime = (Get-Item $targetBinaryFilePath).LastWriteTime
             if ($destModTime -lt $sourceModTime) {
-                Write-Verbose "  ${targetBinaryName}: Updating $SourceDir\$targetBinaryName"
-                Copy-Item "$SourceDir\$targetBinaryName" $targetBinaryDir
+                Write-Verbose "  ${targetBinaryName}: Updating from $sourceBinaryFilePath"
```
dg0yt, line 51: Minor output change here.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
```diff
@@ -20,17 +20,25 @@ This command should always be called by portfiles after they have finished rearr
...
                 -targetBinary "${tool}"
                 -installedDir "${path_to_search}"
+                -verbose
             WORKING_DIRECTORY "${VCPKG_ROOT_DIR}"
-            LOGNAME copy-tool-dependencies
+            LOGNAME copy-tool-dependencies-${count}
         )
+        math(EXPR count "${count} + 1")
     endforeach()
+    set(Z_VCPKG_COPY_TOOL_DEPENDENCIES_COUNT ${count} CACHE INTERNAL "")
```
ras0219-msft, line 41: This probably should be `FORCE`'d to ensure it updates when called multiple times.

ras0219-msft (approved):

ras0219-msft:
>It is not efficient to call appdeploy.ps1 twice. It should better take a list of directories as search path instead of a single directory.
And it should also take a list of executables (or just a single directory).

110% agreed. Additionally, using powershell for this has shown to be no longer sufficient. Both @BillyONeal and I have medium term desires to nuke this in favor of a small portable C++ utility (maybe separate, maybe built-in to vcpkg.exe), however the need for extensibility for plugin producers (Qt, magnum, etc) has made that non-trivial. I hope to eventually find a simple text/json format that can capture all needed information for those cases in a declarative way (and we can add a small amount of hardcoded logic for backwards compatibility to older port versions).

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
dg0yt: `INTERNAL` is always forced.

dg0yt:
> > It is not efficient to call appdeploy.ps1 twice. It should better take a list of directories as search path instead of a single directory.
> > And it should also take a list of executables (or just a single directory).
> 
> 110% agreed.

I just can't put too much in a single PR.

scripts/cmake/vcpkg_copy_tool_dependencies.cmake
ras0219-msft: TIL, thanks! I guess I just have never really used `INTERNAL`.

ras0219-msft:
Thanks!

PR #21106, opened 2021-11-01
[fcl] Enable dynamic support

### LINKED ISSUE #21083: [omplapp] build failure on “fcl[core] is only supported on 'static'“
**Host Environment**
- OS: Windows10
- Compiler: VS2019 

**To Reproduce**
```
.\vcpkg install omplapp:x64-windows
Error: fcl[core] is only supported on 'static'
```

**Failure logs** 

**Additional context**

JackBoosY:
In PR https://github.com/microsoft/vcpkg/pull/20917, fcl was be decleared as a staitc port only, that caused the downstream can't be built on dynamic triplet.
Since fcl is already defined as `ONLY_STATIC_LIBRARY`, remove the `support` keyword.
https://github.com/microsoft/vcpkg/blob/5bdb9d60123df20c247a1654e6a1def51d6c5140/ports/fcl/portfile.cmake#L1

Fixes #21083.

Please note that the current `supports` field will prevent triplet build outside of the allowed rules!

### DIFF
--- a/ports/fcl/vcpkg.json
+++ b/ports/fcl/vcpkg.json
@@ -1,10 +1,9 @@
 {
   "name": "fcl",
   "version": "0.7.0",
-  "port-version": 1,
+  "port-version": 2,
   "description": "a library for performing three types of proximity queries on a pair of geometric models composed of triangles",
   "homepage": "https://github.com/flexible-collision-library/fcl",
-  "supports": "static",
   "dependencies": [
     "ccd",
     "eigen3",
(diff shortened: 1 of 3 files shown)


NancyLi1013 (approved):

autoantwort:
So this does not really enable dynamic support but it will build a static lib for an dynamic triplet.

JackBoosY:
@autoantwort  Yes, but we allow those ports build static on dynamic triplets.

PR #21114, opened 2021-11-01
[gn]Update to latest version

dave-juicelabs:
Updating gn to a more recent version

- #### What does your PR fix?  
  Many new features have been added to gn that some repositories are taking advantage of. By pulling a more recent version of gn, newer repositories will be able to build with vcpkg.

  chromium-base and skia need to be updated to work with the newer version of gn.
  CI baseline is updated.

  Yes

  N/A


dave-juicelabs:
A run of the entire port list may be required to ensure this does not break any other ports.

scripts/ci.baseline.txt
```diff
@@ -120,8 +120,17 @@ chartdir:x64-windows-static-md=fail
 chartdir:x64-osx=fail
 chmlib:arm-uwp=fail
 chmlib:x64-uwp=fail
+# gn update will require a more recent version of chromium-base
```
JackBoosY, line 123: I think we should update chromium-base and v8 first or togather.

JackBoosY (changes requested):

scripts/ci.baseline.txt
dave-juicelabs: Agreed however this PR is mainly to determine what ports are broken solely by updating the gn. I have a separate PR for skia and will make separate PRs for chromium-base and V8.

dave-juicelabs:
For reference, skia #19788.

dave-juicelabs:
@JackBoosY Would it be possible to do a full port build without pulling from the cache to make sure all ports still build with the new gn, beyond skia, chromium-base and v8?

dave-juicelabs:
For reference, chromium-base #21139, v8 #21140

JackBoosY:
> For reference, chromium-base #21139, v8 #21140

I tend to merge the contents of these two PRs into this branch, because we should not disrupt the construction of these two ports at a certain time.

dave-juicelabs:
Sure, once I have chromium-base and V8 working, I will merge them in to this branch.

ras0219-msft (changes requested):
Thanks for the PR!

Because this breaks all previous versions of the consumers (and therefore indicating GN does not preserve backwards compatibility), we must move GN into a port that can be versioned alongside the consumers.

See https://github.com/microsoft/vcpkg/tree/master/ports/vcpkg-cmake for an example of how this sort of "helper port" can be constructed.

The `vcpkg-port-config.cmake` (properly installed) will be included before the `portfile.cmake` of consumers and can provide a new set of functions like `vcpkg_gn_install()` and `vcpkg_gn_configure()`. We can then transition consumers over to these new functions to access the newer versions of gn. `vcpkg_find_acquire_program()` will forever acquire the older version to ensure previous versions still build and we can update it to print a deprecation warning pointing at the new `vcpkg-gn`.

PR #21122, opened 2021-11-02
[drogon] Update to 1.7.3

chausner:
**Describe the pull request**

- #### What does your PR fix?  
  Updates drogon to 1.7.3

  As before, Yes. All features have been tested on all Windows triplets.

  Yes

  Yes


ports/drogon/portfile.cmake
```diff
@@ -1,12 +1,15 @@
+if("ctl" IN_LIST FEATURES)
+    vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+endif()
```
chausner, line 3: This issue has been there before but it did not surface because BUILD_DROGON_SHARED was not getting set, so it was always building static builds. Now, at least as long as the "ctl" feature is not enabled, dynamic builds are created.

The problem here is that when "ctl" is enabled, it tries to run the "drogon_ctl" target as part of the build but it cannot run as it is missing DLLs from dependencies. I am not sure if there is an easy solution, so for now, I would just force static builds if "ctl" is enabled.

Hoikas: I think you would want to add the argument `ADD_BIN_TO_PATH` to `vcpkg_cmake_install()`.

Cheney-W:
All features passed with following triplets:
x86-windows 
x64-windows 
x64-windows-static

ports/drogon/portfile.cmake
chausner: > I think you would want to add the argument `ADD_BIN_TO_PATH` to `vcpkg_cmake_install()`.

Wow, that was easy. Thanks for the hint! I added `ADD_BIN_TO_PATH` and removed `vcpkg_check_linkage(ONLY_STATIC_LIBRARY)`.

PR #21130, opened 2021-11-02
[bddisasm] Update to 1.34.7

ianichitei:
This updates the port of bddisasm to version 1.34.7.

I used 0 as the `port-version` for 1.34.7, but I'm not entirely sure if this is correct.


ports/bddisasm/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bddisasm",
-  "version": "1.34.2",
+  "version": "1.34.7",
   "port-version": 1,
```
JonLiu1993, line 4:
```suggestion
```

JonLiu1993: @ianichitei ,Thanks for your pr, when we update port, we should set "port-version" to 0 or remove "port-version"

JonLiu1993 (changes requested):

ianichitei:
So the bot makes things a bit confusing. I first had `port-version` set to 0, but it suggested a change to `1`. Was that something like a false positive from the bot?

JonLiu1993:
> So the bot makes things a bit confusing. I first had `port-version` set to 0, but it suggested a change to `1`. Was that something like a false positive from the bot?

Maybe it misled you. when we update the port, we need to reset the port-version to 0 or delete it. this is correct.

JonLiu1993 (approved):

PR #21175, opened 2021-11-03
[libpq] Make sure it is possible to run configure.

horodrigues:
Sets the execution mode of the "configure" file used for configuring the compilation of the "libpq" library.

- #### What does your PR fix?  
Fixes the compilation of the "libpq" library, which can not continue on Linux due to a permission denied error caused by the "configure" file having the wrong execution flag settings.

All triplets are supported, but the only triplets affected are the Linux related ones.

Yes, it follows the guidelines. Also, it is a trivial change.

A port file was updated [libpq] and the version was incremented, as necessary.


ports/libpq/portfile.cmake
```diff
@@ -51,6 +51,7 @@ if(VCPKG_TARGET_IS_MINGW)
 endif()
 if(VCPKG_TARGET_IS_LINUX)
     list(APPEND PATCHES patches/linux/configure.patch)
+    list(APPEND PATCHES patches/linux/fixes-configure-mode.patch)
```
JackBoosY, line 54: If yes, we can use cmake code instead of patching the source:
```cmake
file(CHMOD "${SOURCE_PATH}/configure"
         PERMISSIONS
            OWNER_READ OWNER_WRITE
            GROUP_READ GROUP_WRITE
            WORLD_READ WORLD_WRITE
)
```

JackBoosY (changes requested):

ports/libpq/portfile.cmake
dg0yt: `..._EXECUTE`

dg0yt:
> Fixes the compilation of the "libpq" library, which can not continue on Linux due to a permission denied error caused by the "configure" file having the wrong execution flag settings.

Which version is broken, and on which Linux distribution? 
Neither in CI, nor on my computers, I see a wrong permission.

PR #21194, opened 2021-11-04
[gdal] Make available on arm

m-kuhn:
**Describe the pull request**

- #### What does your PR fix?  
Removes exclusion of gdal on arm.
This has been around forever, and it's hard to tell why it has been put there originally.
Builds fine here for android-arm, so I assume the cpu is not an issue.
If there are good reasons not to support arm, this PR shall be closed and a descriptive comment added instead, so once the issue has been fixed the port can be adapted accordingly.

all

yes

yes

### DIFF
--- a/ports/gdal/portfile.cmake
+++ b/ports/gdal/portfile.cmake
@@ -1,5 +1,3 @@
-vcpkg_fail_port_install(ON_ARCH "arm")
-
 set(GDAL_PATCHES
     0001-Fix-debug-crt-flags.patch
     0002-Fix-build.patch
--- a/ports/gdal/vcpkg.json
+++ b/ports/gdal/vcpkg.json
@@ -1,10 +1,9 @@
 {
   "name": "gdal",
   "version-semver": "3.3.2",
-  "port-version": 3,
+  "port-version": 4,
   "description": "The Geographic Data Abstraction Library for reading and writing geospatial raster and vector data",
   "homepage": "https://gdal.org",
-  "supports": "!arm",
   "dependencies": [
     "curl",
     "expat",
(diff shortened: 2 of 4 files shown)


dg0yt:
Unfortunately, the change is not tested in vcpkg CI: Some dependency is marked as failing or not supported on arm, and this cascades to gdal.

m-kuhn:
> Some dependency is marked as failing or not supported on arm, and this cascades to gdal.

Sorry, I'm failing to follow the implications of this.

Where I'm coming from: it builds fine for arm-android
What I'd expect is: if a dependency is failing, gdal should be excluded implicitly, not explicitly

dg0yt:
@m-kuhn I just wanted to point out that we cannot feel safe with all CI checks green because gdal:arm still *is excluded implicitly*.

m-kuhn:
> @m-kuhn I just wanted to point out that we cannot feel safe with all CI checks green because gdal:arm still _is excluded implicitly_.

_is_ or _may be_?
Whatever, even if there still is a problem somewhere else, this PR should help shifting the attention to the right spot.

autoantwort:
Works at least on arm64-osx

NancyLi1013 (approved):

dg0yt:
We know that gdal works on arm from using gdal without vcpkg.
FTR it is hdf5 which is marked as failing on arm. Hdf5 may be hard to fix unless you can test it. But that problem should go away soon, when we turn hdf5 into a non-default feature.

dg0yt:
Building for arm doesn't work yet for windows targets (building with nmake). I adopted a modified version for the gdal 3.4.0 update in #21261. This should make this PR obsolete.

m-kuhn:
Very good news, I'm more than happy to let go

PR #21222, opened 2021-11-06
Fix android ndk clang

holmesconan:
This is a fix of Android triplets that using clang as the default compiler.

- #### What does your PR fix?  
  Fix vcpkg_configure_make based port cannot find gcc (aka clang) compiler during config stage
  Fix ANDROID_NATIVE_API_LEVEL only detected from CMAKE_SYSTEM_VERSION, add ENV{ANDROID_API_LEVEL} for customizing

  android, No

  `Yes`

  No


holmescn:
The only notification is that, please add Android NDK toolchain executable path:

    export PATH=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/<host-tag>/bin:$PATH

`host-tag` is `linux-x86_64` on linux and 'darwin-x86_64' on osx. Windows is not tested yet.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -465,6 +465,50 @@ function(vcpkg_configure_make)
         endif()
     endif()
 
+    # Android - cross-compiling support
+    if(VCPKG_TARGET_IS_ANDROID)
+        if (VCPKG_DETECTED_CMAKE_CXX_COMPILER_ID MATCHES "^Clang$")
+            string(REPLACE "-static-libstdc++" "-lc++_static" VCPKG_DETECTED_CMAKE_CXX_STANDARD_LIBRARIES ${VCPKG_DETECTED_CMAKE_CXX_STANDARD_LIBRARIES})
```
dg0yt, line 471: I wonder why this should be wrong in the detected libraries. IIRC, the NDK's toolchain has variables to configure the C++ runtime. Can't we hook into chainloading and configuring this toolchain, and then leave the detected values untouched?

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -465,6 +465,50 @@ function(vcpkg_configure_make)
...
+            elseif(VCPKG_HOST_IS_LINUX)
+                z_vcpkg_determine_autotools_host_cpu(BUILD_ARCH) # machine you are building on => --build=
+                set(arg_BUILD_TRIPLET "--build=${BUILD_ARCH}-pc-linux-gnu")
+            endif()
+
+            if (DEFINED ENV{ANDROID_API_LEVEL})
+                set(ANDROID_API_LEVEL $ENV{ANDROID_API_LEVEL} CACHE STRING "")
+            else()
+                set(ANDROID_API_LEVEL ${VCPKG_DETECTED_CMAKE_SYSTEM_VERSION} CACHE STRING "")
+            endif()
```
dg0yt, line 490: I wonder if respecting `ENV{ANDROID_API_LEVEL}` at this point may break binary caching.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -518,7 +562,7 @@ function(vcpkg_configure_make)
 
     set(base_cmd)
     if(CMAKE_HOST_WIN32)
-        set(base_cmd ${bash_executable} --noprofile --norc --debug)
+        set(base_cmd ${BASH} --noprofile --norc --debug)
```
dg0yt, line 565: There is no `BASH` here. I think this was renamed to  `bash_executable` as part of the script audit, parallel to your work.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -763,7 +807,7 @@ function(vcpkg_configure_make)
         # https://www.gnu.org/software/libtool/manual/html_node/Link-mode.html
         # -avoid-version is handled specially by libtool link mode, this flag is not forwarded to linker,
         # and libtool tries to avoid versioning for shared libraries and no symbolic links are created.
-        if(VCPKG_TARGET_IS_ANDROID)
+        if(VCPKG_TARGET_IS_ANDROID AND VCPKG_DETECTED_CMAKE_C_COMPILER_ID MATCHES "^GNU$")
```
dg0yt, line 810: Isn't it needed, or does it break, with clang?

dg0yt (commented):
Thanks for the PR. This is just a quick review in Github.

scripts/cmake/vcpkg_configure_make.cmake
dg0yt: PS: `STREQUAL "GNU"` would suffice.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: I cannot answer this. I have tried to hook the `<VCPKG_ROOT>/scripts/toolchains/android.cmake` for adding some util variables, but it seems that `vcpkg_configure_make.cmake` do not reference that file. So I have to specify the ANDROID_API_LEVEL again. So I do not think it is easy to hook NDK's toolchain file to obtain the detected. Or if I could fully understand how to do it.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: I think yes, it would be. Since different API level may have different ABI,  especially for the libc++. If one library is compiled with ANDROID_API_LEVEL equals 21, but others with 23, for example. They could not be linked into the final .so file.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: Yes, I checked the upstream version, I will fix it. It was caused by merging without check. shame.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: actually, it just generated a warning for `argument unused during compilation`.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: This caused the CI to fail. fixed.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: This is removed. God bless no weird problems happen.

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -465,6 +465,50 @@ function(vcpkg_configure_make)
...
+            string(REPLACE "-static-libstdc++" "-lc++_static" VCPKG_DETECTED_CMAKE_CXX_STANDARD_LIBRARIES ${VCPKG_DETECTED_CMAKE_CXX_STANDARD_LIBRARIES})
+
+            if(VCPKG_HOST_IS_WINDOWS)
+                z_vcpkg_determine_autotools_host_cpu(BUILD_ARCH) # machine you are building on => --build=
+                z_vcpkg_determine_host_mingw(BUILD_MINGW)
+                set(arg_BUILD_TRIPLET "--build=${BUILD_ARCH}-pc-${BUILD_MINGW}")  # This is required since we are running in a msys
+                                                                                           # shell which will be otherwise identified as ${BUILD_ARCH}-pc-msys
+            elseif(VCPKG_HOST_IS_OSX)
+                z_vcpkg_determine_autotools_host_arch_mac(BUILD_ARCH) # machine you are building on => --build=
+                set(arg_BUILD_TRIPLET "--build=${BUILD_ARCH}-apple-darwin${VCPKG_CMAKE_HOST_SYSTEM_VERSION}")
```
dg0yt, line 480: I guess the last variable is named `VCPKG_DETECTED_CMAKE_HOST_SYSTEM_VERSION`.

dg0yt (commented):
We should not unconditionally override (set) `arg_BUILD_TRIPLET` for Android. It is a valid argument to be supplied by the calling portfile, and it may be overridden by the triplet file. Even if not used in vcpg at the moment, this is public interface possibly used in user's ports or triplets.
The pattern for windows and macos is:
~~~
if(VCPKG_TARGET_IS_ANDROID)
    if(arg_DETERMINE_BUILD_TRIPLET OR NOT arg_BUILD_TRIPLET)
        ...
        set(arg_BUILD_TRIPLET ...
~~~

JackBoosY:
Failed to build `python3:x64-osx`:
```
configure: WARNING: unrecognized options: --disable-silent-rules, --enable-static
./../src/v3.10.0-dcf0f0f52f.clean/configure: line 10530: PKG_PROG_PKG_CONFIG: command not found
./../src/v3.10.0-dcf0f0f52f.clean/configure: line 10673: /usr/local/bin/pkg-config --static: No such file or directory
configure: WARNING: unrecognized options: --disable-silent-rules, --enable-static
```

dg0yt:
> Failed to build `python3:x64-osx`

I wonder how this PR can have a side effect on `/usr/local/bin/pkg-config --static` being considered a single value ("No such file or directory") instead of a list for python3:x64-osx.

dg0yt:
> > Failed to build `python3:x64-osx`
> 
> I wonder how this PR can have a side effect on `/usr/local/bin/pkg-config --static` being considered a single value ("No such file or directory") instead of a list for python3:x64-osx.

Well, there is one changed line which is unrelated to android, but very much related to x64-osx:
~~~
if(host_arch MATCHES "(amd|AMD|x86_|X86_)64")
~~~
This change is reasonable, but maybe it needs investigation in a separate PR. By changing host detection, the cross-build property will change to, and this can uncover bugs in other ports.

holmescn:
made a little changes for the z_vcpkg_determine_autotools_host_cpu. if it doesn't resolve the problem, I cannot fix it now. My Macbook is dead.....

holmescn:
More tests on Android libraries are needed. Anyone who have problems are welcome.

JackBoosY:

Also, cc @Neumann-A for review the configure_make part.

holmescn:
@JackBoosY Yes, it is ready for review. @Neumann-A

scripts/cmake/vcpkg_configure_make.cmake
Neumann-A: This export should be done in the `CMakeLists.txt` of `vcpkg_get_cmake_vars` so that meson and other buildsystems are also informed about the required flags.

scripts/cmake/vcpkg_configure_make.cmake
Neumann-A: I don't see why we should consider `ENV{ANDROID_API_LEVEL}` here?

scripts/cmake/vcpkg_configure_make.cmake
```diff
@@ -465,6 +467,55 @@ function(vcpkg_configure_make)
...
+                string(APPEND configure_env " CXX=armv7a-linux-androideabi${ANDROID_API_LEVEL}-clang++")
+            else()
+                string(APPEND configure_env " CC=${TARGET_ARCH}-linux-android${ANDROID_API_LEVEL}-clang")
+                string(APPEND configure_env " CXX=${TARGET_ARCH}-linux-android${ANDROID_API_LEVEL}-clang++")
+            endif()
+
+            string(APPEND configure_env " AR=llvm-ar")
+            string(APPEND configure_env " RANLIB=llvm-ranlib")
+            string(APPEND configure_env " READELF=llvm-readelf")
+            string(APPEND configure_env " STRIP=llvm-strip")
```
Neumann-A, line 515: Simply no. Setup the toolchain correctly..... CC and CXX and the rest of the programs should be correctly setup by cmake.

scripts/get_cmake_vars/CMakeLists.txt
```diff
@@ -8,7 +8,9 @@ list(APPEND VCPKG_DEFAULT_VARS_TO_CHECK CMAKE_CROSSCOMPILING
                                         CMAKE_SYSTEM_NAME
                                         CMAKE_HOST_SYSTEM_NAME
                                         CMAKE_SYSTEM_PROCESSOR
-                                        CMAKE_HOST_SYSTEM_PROCESSOR)
+                                        CMAKE_HOST_SYSTEM_PROCESSOR
+                                        CMAKE_SYSTEM_VERSION
+                                        CMAKE_HOST_SYSTEM_VERSION)
```
Neumann-A, line 13:
```suggestion
```

don't see why we should care about `CMAKE_HOST_SYSTEM_VERSION`. vcpkg should only focus on building for the target.

scripts/toolchains/android.cmake
```diff
@@ -1,9 +1,16 @@
...
 set(CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION clang CACHE STRING "")
 
+if (DEFINED ENV{ANDROID_API_LEVEL})
+    set(ANDROID_NATIVE_API_LEVEL $ENV{ANDROID_API_LEVEL} CACHE STRING "")
+    set(ANDROID_PLATFORM $ENV{ANDROID_API_LEVEL} CACHE STRING "")
+else()
+    set(ANDROID_NATIVE_API_LEVEL ${CMAKE_SYSTEM_VERSION} CACHE STRING "")
+    set(ANDROID_PLATFORM ${CMAKE_SYSTEM_VERSION} CACHE STRING "")
+endif()
+
```
Neumann-A, line 13: without env passthrough in the triplet `ENV{ANDROID_API_LEVEL}` will be always empty on windows. So a custom triplet will be necessary anyway and if that is already necessary having a custom specialized toolchain is also not too far away. (I really hate this all purpose android toolchain which includes all possible arches .... )

scripts/cmake/vcpkg_configure_make.cmake
holmescn: I use the ENV{ANDROID_API_LEVEL} since I don't want to add another argument for vcpkg_configure_make. It needs more consideration about the general purpose. It seems that an argument is reasonable? One need to customize the ANDROID API LEVEL instead of the detected version.

holmescn: Actually, the CC and CXX environment variables are the core patch of this PR. Without that, the autotools `configure` script could not find the compiler properly.

scripts/get_cmake_vars/CMakeLists.txt
holmescn: When cross-compile python3, one needs to specify the --build option and on osx, it will be x86_64-apple-darwin20.6.0, where 20.6.0 is the host system version in cmake. In other words, when cross-compile some libraries on osx, one may need that variable. Yes I will shrink it to CMAKE_HOST_SYSTEM_VERSION only.

scripts/cmake/vcpkg_configure_make.cmake
Neumann-A: > Actually, the CC and CXX environment variables are the core patch of this PR

but not in this way because it is too specific. If necessary there needs to be a simple branch which setups all progs correctly to the `VCPKG_DETECTED_` variables (like in the windows branch)

scripts/toolchains/android.cmake
holmescn: Yes, that is why I add the ENV here. The first time I use it, the previous version only set the ANDROID_NATIVE_API_LEVEL with an empty value. I don't know how to fix this.

scripts/cmake/vcpkg_configure_make.cmake
holmescn: The toolchain did not correctly setup by cmake since it was out of control. When reading config-x86-android-dbg-out.txt in buildtree/<lib>/, I found that most cross-compile toolchains are grabbed from the system, not in the NDK. It won't work. So I need to manually specify the toolchain through configure_env.

Neumann-A: I would say that is not a problem for `vcpkg_configure_make` to solve.

scripts/get_cmake_vars/CMakeLists.txt
Neumann-A: > When cross-compile python3

that seems oddly specific to python3. The typically configure scripts don't care about anything after `*-darwin*`

Neumann-A:
Again I feel like the default android toolchain vcpkg provides should be removed and replaced with specialized toolchains for each architecture and case. There seems to be a bigger misunderstanding of vcpkg is supposed to work and what work should be done by the configure scripts.

dg0yt:
@holmescn You just closed this PR while I was writing a comment. I actively invited this PR, and I hoped we could help to form an acceptable result, no matter how hard it is to get Android right in vcpkg. I don't see another effort to move things for Android in vcpkg.

holmescn:
@dg0yt Thank you for your invite. Since @Neumann-A made a difficult goal that needs to solve the problem genetically. I don't think I have the time and skill to accomplish that target. This is a temporary solution that has many problems unknown. I hope any successor could find more elegant ways to touch the goal, maybe inspired by this effort or not. I would like to give any help if I could.

holmescn:
As a generic solution, we need to consider from scratch that how vcpkg passthrough each detected variable down to the build system and why they are not catched by the Android NDK toolchain. It needs more knowledge about how vcpkg works.

m-kuhn:
Thank you for opening this issue and the work in the PR. It would be nice to have a solution for android in vcpkg.
For reference, I'd also like to reference previous work by @xandox https://github.com/microsoft/vcpkg/pull/15605

PR #21263, opened 2021-11-08
[asio-grpc] Update to 1.3.1

Tradias:
Update [asio-grpc ](https://github.com/Tradias/asio-grpc) to 1.3.1

  all, yes

  yes

  yes

### DIFF
--- a/ports/asio-grpc/vcpkg.json
+++ b/ports/asio-grpc/vcpkg.json
@@ -1,13 +1,9 @@
 {
   "name": "asio-grpc",
-  "version": "1.1.2",
-  "description": "Asynchronous gRPC with Boost.Asio executors",
+  "version": "1.3.0",
+  "description": "Asynchronous gRPC with Asio/unified executors",
   "homepage": "https://github.com/Tradias/asio-grpc",
   "dependencies": [
-    "boost-asio",
-    "boost-core",
-    "boost-intrusive",
-    "boost-lockfree",
     "grpc",
     {
       "name": "vcpkg-cmake",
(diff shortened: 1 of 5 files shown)


ports/asio-grpc/vcpkg.json
```diff
@@ -1,13 +1,9 @@
 {
   "name": "asio-grpc",
-  "version": "1.1.2",
-  "description": "Asynchronous gRPC with Boost.Asio executors",
+  "version": "1.3.0",
+  "description": "Asynchronous gRPC with Asio/unified executors",
   "homepage": "https://github.com/Tradias/asio-grpc",
   "dependencies": [
-    "boost-asio",
```
JonLiu1993, line 7: @Tradias ,Thanks for your update ,have all these dependencies of the old version become optional in the new version?

Tradias: The dependencies `boost-core`, `boost-intrusive` and `boost-lockfree` are no longer needed.

Users are intended to link with one CMake target: ` asio-grpc::asio-grpc`, `asio-grpc::asio-grpc-standalone-asio` or `asio-grpc::asio-grpc-unifex` in which case they need exactly one of `boost-asio`, `asio` or `libunifex`. The features are therefore meant to avoid installating e.g. Boost.Asio when the user is only interested in the libunifex target. I hope that makes sense, if you have any concerns or suggestions then let me know. I am happy to adjust

ports/asio-grpc/vcpkg.json
```diff
@@ -18,12 +14,33 @@
       "host": true
     }
   ],
+  "default-features": [
+    "boost-asio"
+  ],
   "features": {
+    "boost-asio": {
```
JonLiu1993, line 21: @Tradias, if boost-asio libunifex and standalone-asio are added to features, we should add them to vcpkg_check_features

JonLiu1993 (changes requested):

ports/asio-grpc/vcpkg.json
Tradias: Ok, I can do that, although the resulting <feature-var>s won't be used for now since the library is header only. It does not need to know what dependencies are available at installation time.

Tradias: It might be best to remove these dependencies and features (boost-asio, standalone-asio and libunifex) and instead instruct the user in a `usage` file to install them, something like:

```cmake
The package asio-grpc can be used with different backends:

    find_package(asio-grpc CONFIG REQUIRED)

The standalone Boost.Asio backend. Install "boost-asio" and link with:

    target_link_libraries(main PRIVATE asio-grpc::asio-grpc)

The standalone Asio backend. Install "asio" and link with:

    target_link_libraries(main PRIVATE asio-grpc::asio-grpc-standalone-asio)

The libunifex backend. Install "libunifex" and link with:

    target_link_libraries(main PRIVATE asio-grpc::asio-grpc-unifex)
```

What do you think?

JonLiu1993: I think it's good

ras0219-msft (approved):
LGTM, thanks for the update and awesome library!

PR #21277, opened 2021-11-09
[dcmtk] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/dcmtk/portfile.cmake
+++ b/ports/dcmtk/portfile.cmake
@@ -9,9 +9,8 @@ vcpkg_from_github(
     PATCHES ${CMAKE_CURRENT_LIST_DIR}/dcmtk.patch
 )
 
-vcpkg_configure_cmake(
-    SOURCE_PATH ${SOURCE_PATH}
-    PREFER_NINJA
+vcpkg_cmake_configure(
+    SOURCE_PATH "${SOURCE_PATH}"
     OPTIONS
         -DDCMTK_WITH_DOXYGEN=OFF
         -DDCMTK_WITH_ZLIB=OFF
@@ -34,13 +33,18 @@ vcpkg_configure_cmake(
         -DINSTALL_OTHER=OFF
 )
 
-vcpkg_install_cmake()
+vcpkg_cmake_install()
 vcpkg_copy_pdbs()
 
-vcpkg_fixup_cmake_targets()
+vcpkg_cmake_config_fixup()
 
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DCMTK_PREFIX \"${CURRENT_PACKAGES_DIR}\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DCM_DICT_DEFAULT_PATH \"${CURRENT_PACKAGES_DIR}/share/dcmtk/dicom.dic:${CURRENT_PACKAGES_DIR}/share/dcmtk/private.dic\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DEFAULT_CONFIGURATION_DIR \"${CURRENT_PACKAGES_DIR}/etc/dcmtk/\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DEFAULT_SUPPORT_DATA_DIR \"${CURRENT_PACKAGES_DIR}/share/dcmtk/\"" "")
 
 # Handle copyright
-file(INSTALL ${SOURCE_PATH}/COPYRIGHT DESTINATION ${CURRENT_PACKAGES_DIR}/share/dcmtk RENAME copyright)
+file(INSTALL "${SOURCE_PATH}/COPYRIGHT" DESTINATION "${CURRENT_PACKAGES_DIR}/share/dcmtk" RENAME copyright)
--- a/ports/dcmtk/usage
+++ b/ports/dcmtk/usage
@@ -1,0 +1,18 @@
+The package dcmtk provides CMake targets:
+
+    find_package(DCMTK CONFIG REQUIRED)
+    # Note: 23 target(s) were omitted.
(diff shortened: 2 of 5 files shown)


ports/dcmtk/portfile.cmake
```diff
@@ -34,13 +33,18 @@ vcpkg_configure_cmake(
...
 
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DCMTK_PREFIX \"${CURRENT_PACKAGES_DIR}\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DCM_DICT_DEFAULT_PATH \"${CURRENT_PACKAGES_DIR}/share/dcmtk/dicom.dic:${CURRENT_PACKAGES_DIR}/share/dcmtk/private.dic\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DEFAULT_CONFIGURATION_DIR \"${CURRENT_PACKAGES_DIR}/etc/dcmtk/\"" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/dcmtk/config/osconfig.h" "#define DEFAULT_SUPPORT_DATA_DIR \"${CURRENT_PACKAGES_DIR}/share/dcmtk/\"" "")
```
JackBoosY, line 47: Are they correct? You removed those definations, and I think you should use the relative path instead.

JackBoosY (changes requested):

ports/dcmtk/portfile.cmake
autoantwort: The problem is that there is no way to get a relative path inside of header

autoantwort: So this is a define. The path must be the same regardless where the define is used, so a completely relative path don't work. It can be relative to the folder of the dir of the header, but there is not way to get the dir of the header

JackBoosY: Maybe we should change `"#define DCMTK_PREFIX \"${CURRENT_PACKAGES_DIR}\""` to some code related to get the current path, such as `getcwd`.
We should see what the upstream says.

autoantwort: You can't use `getcwd` because it changes

It is better to remove this definitions instead of providing wrong ones.

dg0yt: Either use standard prefixes, e.g. `/etc`, `/usr`, maybe guarded with `#ifndef`, and/or list the required definitions in `usage`?

autoantwort: Then I think the best solution is to provide a usage file with the content
```cmake
  target_compile_definitions(main PRIVATE -DDCMTK_PREFIX=${IS_THERE_A_CMAKE_VARIABLE_FOR_THE_INSTALLED_DIR?})
  ...
```

JackBoosY: @autoantwort That's the best way I think.

autoantwort: @JackBoosY Is there a variable for the installed dir provided by the cmake integration?

JackBoosY: This path can't be fixed because the user may export this port and use it in other place.
So I think we should notice the user to fill this value manually.

dg0yt: > Is there a variable for the installed dir provided by the cmake integration?

`${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}`? (not verified, cf. `vcpkg.cmake`)

JackBoosY: > ${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}? (not verified, cf. vcpkg.cmake)

As I said, in manifest mode the path is not that one.

dg0yt: > > `${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}`? (not verified, cf. vcpkg.cmake)
> 
> As I said, in manifest mode the path is not that one.

Did you verify the path?

IIUC the problem with manifest mode is that there are two modes of operation:
- When building the user project, ports are properly installed to `${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}` in the project's `${CMAKE_CURRENT_BINARY_DIR}`.
- Documentation indicates that users may run `vpckg install` in the directory of the manifest file. 
  However, this doesn't know neither actual `${CMAKE_CURRENT_BINARY_DIR}` nor actual target. So it is pointless and only generates issues on GH.

Correct me if I'm wrong.

JackBoosY: @dg0yt Yeah that's correct, but we can't guarantee the export scenes.

autoantwort: @JackBoosY So what should I print as usage in your opinion?

JackBoosY: We should consider all scenes, maybe print a usage like:
```
# Pass the installation path such as ${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}
target_compile_definitions(main PRIVATE -DDCMTK_PREFIX=${INSTALLATION_DIR})
```

ports/dcmtk/portfile.cmake
```diff
@@ -34,13 +33,33 @@ vcpkg_configure_cmake(
...
+  else()
+      target_compile_definitions(main
+          PUBLIC
+              "DCMTK_PREFIX=${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}"
+              "DCM_DICT_DEFAULT_PATH=${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/dcmtk/dicom.dic:${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/dcmtk/private.dic"
+              "DEFAULT_CONFIGURATION_DIR=${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/etc/dcmtk/"
+              "DEFAULT_SUPPORT_DATA_DIR=${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/dcmtk/"
+      )
+  endif()
+]])
```
JackBoosY, line 62: Can we create file `usage` to print that?

JackBoosY (changes requested):

ports/dcmtk/portfile.cmake
autoantwort: Ok I have added a usage file

JackBoosY (approved):

PR #21290, opened 2021-11-09
[libxslt] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/libxslt/portfile.cmake
+++ b/ports/libxslt/portfile.cmake
@@ -139,6 +139,12 @@ else()
         file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/lib/libxslt-plugins ${CURRENT_PACKAGES_DIR}/debug/lib/libxslt-plugins)
     endif()
     file(REMOVE ${CURRENT_PACKAGES_DIR}/lib/libxslt.so)
+
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/libxslt/bin/xslt-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../../")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/libxslt/debug/bin/xslt-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../../../")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/libxslt/xsltconfig.h" "#define LIBXSLT_DEFAULT_PLUGINS_PATH() \"${CURRENT_INSTALLED_DIR}/lib/libxslt-plugins\"" "")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/lib/xsltConf.sh" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../")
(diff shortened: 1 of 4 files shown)


ports/libxslt/portfile.cmake
```diff
@@ -139,6 +139,12 @@ else()
         file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/lib/libxslt-plugins ${CURRENT_PACKAGES_DIR}/debug/lib/libxslt-plugins)
     endif()
     file(REMOVE ${CURRENT_PACKAGES_DIR}/lib/libxslt.so)
```
Osyotr, line 141: This port is cursed...

autoantwort: I thought the same ...

autoantwort: I thought the same

JackBoosY: Why it generate the dynamic library on non-Windows?

Osyotr: @JackBoosY right now x64-linux-dynamic is broken. The code above is just incorrect.
I'll push a fix later

ports/libxslt/portfile.cmake
```diff
@@ -139,6 +139,12 @@ else()
...
+
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/libxslt/bin/xslt-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../../")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/libxslt/debug/bin/xslt-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../../../")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/libxslt/xsltconfig.h" "#define LIBXSLT_DEFAULT_PLUGINS_PATH() \"${CURRENT_INSTALLED_DIR}/lib/libxslt-plugins\"" "")
```
ras0219-msft, line 145: Is `LIBXSLT_DEFAULT_PLUGINS_PATH()` not used anywhere?

Osyotr: By the way, this folder is empty in linux build, both static and dynamic. There is no such folder in windows builds.
PS: upstream has CMake support since may 2020, but they haven't made a release since 2019.

ras0219-msft: Ahh ok, so this file probably isn't even included in the public headers.

ports/libxslt/portfile.cmake
autoantwort: @Osyotr So you will create a PR for that?

JackBoosY: Ping @Osyotr for response.

ports/libxslt/portfile.cmake
Osyotr: @ras0219-msft this particular line is guarded by `#if 0` in both static and dynamic builds.

ports/libxslt/portfile.cmake
Osyotr: I've included these changes in #21441

ras0219-msft:
I'm looking to merge https://github.com/microsoft/vcpkg/pull/21441 since it includes these fixes and also addresses dynamic linkage on Linux -- if that PR gets held up for some reason then we can revisit this smaller change.

PR #21303, opened 2021-11-10
[plplot] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/plplot/portfile.cmake
+++ b/ports/plplot/portfile.cmake
@@ -15,14 +15,14 @@ vcpkg_from_sourceforge(
       fix-pkg-config.patch
 )
 
-set(BUILD_with_wxwidgets OFF)
-if("wxwidgets" IN_LIST FEATURES)
-  set(BUILD_with_wxwidgets ON)
-endif()
+vcpkg_check_features(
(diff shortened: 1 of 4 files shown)


ports/plplot/vcpkg.json
```diff
@@ -1,12 +1,20 @@
 {
   "name": "plplot",
   "version-string": "5.13.0",
```
NancyLi1013, line 3:
```suggestion
  "version-semver": "5.13.0",
```
This should be version-semver. https://github.com/PLplot/PLplot/blob/6a12ffe24060803e1d66ae457fa7663dc4ded5bf/cmake/modules/plplot_version.cmake#L22-L23

ports/plplot/portfile.cmake
```diff
@@ -20,9 +20,8 @@ if("wxwidgets" IN_LIST FEATURES)
   set(BUILD_with_wxwidgets ON)
 endif()
```
NancyLi1013, line 21: Would you like to use `vcpkg_check_features()` to handle this feature?

NancyLi1013 (changes requested):

NancyLi1013 (approved):

PR #21307, opened 2021-11-10
[magnum-plugins] Update head features and bundled dependency

pezcode:
This PR updates magnum-plugins with new features available when compiling with `--head`, along with an update to a bundled dependency required to compile.

Note: the `spirvtoolsshaderconverter` feature is currently broken due to https://github.com/microsoft/vcpkg/issues/18612

  There shouldn't be any changes compared to the current port version

  To the best of my knowledge, yes :eyes:

  Yessss

### DIFF
--- a/ports/magnum-plugins/portfile.cmake
+++ b/ports/magnum-plugins/portfile.cmake
@@ -34,8 +34,8 @@ if("basisimporter" IN_LIST FEATURES OR "basisimageconverter" IN_LIST FEATURES)
         FILENAME "basis-universal-${_BASIS_VERSION}.tar.gz"
         SHA512 ${_BASIS_SHA512}
     )
-    vcpkg_extract_source_archive_ex(
-        OUT_SOURCE_PATH _BASIS_UNIVERSAL_SOURCE
+    vcpkg_extract_source_archive(
+        _BASIS_UNIVERSAL_SOURCE
         ARCHIVE ${_BASIS_UNIVERSAL_ARCHIVE}
         WORKING_DIRECTORY "${SOURCE_PATH}/src/external"
         ${_BASIS_UNIVERSAL_PATCHES})
--- a/ports/magnum-plugins/portfile.cmake
+++ b/ports/magnum-plugins/portfile.cmake
@@ -74,9 +74,8 @@ endforeach()
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS FEATURES ${_COMPONENTS})
 
-vcpkg_configure_cmake(
+vcpkg_cmake_configure(
     SOURCE_PATH ${SOURCE_PATH}
-    PREFER_NINJA # Disable this option if project cannot be built with Ninja
     OPTIONS
         ${FEATURE_OPTIONS}
         -DBUILD_STATIC=${BUILD_PLUGINS_STATIC}
@@ -85,7 +84,7 @@ vcpkg_configure_cmake(
         -DMAGNUM_PLUGINS_RELEASE_DIR=${CURRENT_INSTALLED_DIR}/bin/magnum
 )
 
-vcpkg_install_cmake()
+vcpkg_cmake_install()
 
 # Debug includes and share are the same as release
 file(REMOVE_RECURSE
--- a/ports/magnum-plugins/portfile.cmake
+++ b/ports/magnum-plugins/portfile.cmake
@@ -9,22 +9,20 @@ vcpkg_from_github(
 )
 
 if("basisimporter" IN_LIST FEATURES OR "basisimageconverter" IN_LIST FEATURES)
-    # Bundle Basis Universal
+    # Bundle Basis Universal. The repo has big unrequired files in its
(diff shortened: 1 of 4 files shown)


NancyLi1013 (changes requested):

ports/magnum-plugins/portfile.cmake
```diff
@@ -9,28 +9,36 @@ vcpkg_from_github(
...
-        SHA512 65062ab3ba675c46760f56475a7528189ed4097fb9bab8316e25d9e23ffec2a9560eb9a6897468baf2a6ab2bd698b5907283e96deaeaef178085a47f9d371bb2
+        SHA512 ${_BASIS_SHA512}
     )
     vcpkg_extract_source_archive_ex(
         OUT_SOURCE_PATH _BASIS_UNIVERSAL_SOURCE
         ARCHIVE ${_BASIS_UNIVERSAL_ARCHIVE}
         WORKING_DIRECTORY "${SOURCE_PATH}/src/external"
-        PATCHES
-            ${_BASIS_UNIVERSAL_PATCHES})
+        ${_BASIS_UNIVERSAL_PATCHES})
```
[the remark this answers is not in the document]
pezcode: I reverted it back to the way it was

ports/magnum-plugins/portfile.cmake
```diff
@@ -65,18 +74,19 @@ endforeach()
...
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS FEATURES ${_COMPONENTS})
 
-vcpkg_configure_cmake(
+vcpkg_cmake_configure(
     SOURCE_PATH ${SOURCE_PATH}
-    PREFER_NINJA # Disable this option if project cannot be built with Ninja
     OPTIONS
         ${FEATURE_OPTIONS}
+        # Make Basis Universal use the bundled zstd
+        -DCMAKE_DISABLE_FIND_PACKAGE_Zstd=ON
```
NancyLi1013, line 82: Why do we use the bundled `zstd`?

pezcode: This was done because otherwise a dynamic zstd wouldn't take part in the vcpkg dependency copy step (`VCPKG_APPLOCAL_DEPS`). I considered adding zstd as a feature dependency but zstd is only needed for --head so it'd needlessly build it for non-head version.

Is the feature dependency preferrable here?

pezcode:
Turned into a draft PR temporarily to coordinate with @mosra. If @NancyLi1013 confirms that head-only feature dependencies are acceptable, the bundled zstd handling can be removed upstream.

ports/magnum-plugins/portfile.cmake
JackBoosY: For `VCPKG_APPLOCAL_DEPS`, if the dependent library is linked (statically loaded) when the current library is built instead of dynamically loaded using `load_library` in the code, it will be copied automatically.

I don't see that `magnum-plugins` has an item that depends on zstd, so why do I need to modify the way to find zstd here?

pezcode: I misunderstood how `VCPKG_APPLOCAL_DEPS` then. Here's the long version of the problem:

`magnum-plugins` features `basisimageconverter` and `basisimporter` depend on [basis_universal](https://github.com/BinomialLLC/basis_universal). The vcpkg port (`basisu`) is hopelessly outdated so a current version is fetched and bundled in the portfile. The new version of `basis_universal` that `magnum-plugins --head` now uses has its own bundled copy of zstd. `magnum-plugins` tries to find zstd with `find_package`, and if it can't find it, it uses the bundled zstd sources.

So the question is: Do I add zstd as a dependency to the `basisimageconverter` and `basisimporter` features? That seemed questionable to me, since the zstd dependency is only really necessary when compiling for `--head` (where magnum uses the new `basis_universal` that requires zstd).

Note: I'll remove the `DCMAKE_DISABLE_FIND_PACKAGE_Zstd` either way, that's wrong

JackBoosY: Please do not try to fix the bugs generated by `--head`, we should fix them after the version is updated.
If this feature needs to find zstd, we must add zstd as a dependency of this feature and always ensure that it can be found.

pezcode: I removed the `DISABLE_FIND_PACKAGE` and added the zstd dependency 👍

NancyLi1013 (approved):

NancyLi1013:
Have you tested this features? @pezcode

pezcode:
> 
> 
> Have you tested this features? @pezcode

I tested all the new and updated features with `x64-windows` and `x64-windows-static`.

vicroms (changes requested):
This port should be updated to the latest commit instead. It removes the necessity for the extra logic around `--head` which was never intended to be used this way.

vicroms (approved):

vicroms:
Discussed with the rest of the team, this is fine since it acts like a no-op when a user requests unsupported features and doesn't add new dependencies for non-head users.

PR #21381, opened 2021-11-13
[libmodplug] Patch CMakeLists.txt to fix install paths

itsmattkc:
**Describe the pull request**

- #### What does your PR fix?  
  Implements Konstanty/libmodplug#61 which fixes a bug in the CMakeLists.txt where DLL files are erroneously installed to the `lib` folder rather than `bin`. Currently the `portfile.cmake` attempts to workaround this by manually renaming the files post-installation, but this is clunky (considering the filename has to be hardcoded), and will also fail if the user is installing only a release OR debug build (since both rename commands are hardcoded and CMake will throw a "file not found" error if they don't both exist).

  By patching the CMakeLists.txt, this is solved at the root of the problem. DLLs are correctly installed to `bin` in the first place (while other library/linking files are still correctly installed to `lib`). The CMake documentation recommends this solution towards the end of the "Installing Targets" section here: https://cmake.org/cmake/help/latest/command/install.html#installing-targets

  All triplets are supported. This patch corrects Win32 issues without breaking other platforms.

  I believe it does. In my opinion, this is a very small and readable patch that I believe will be easy to review. I also believe it makes no extraneous/trivial changes and is the code I am using is fairly well-documented.

  Yes.


cenit:
whenever you touch a ports/<port name> folder content, you are updating a port in vcpkg terminology.
So you need to bump port-version in vcpkg.json (or control if old port) and run the command above after committing the first steps

itsmattkc:
Understood, I believe I have done this now.

cenit:
you did everything in one commit and it’s not valid. You should have first committed your patches with the port version bump, then another commit with the modifications after running vcpkg x-add-versions

itsmattkc:
Alright. How is it now?

ports/libmodplug/portfile.cmake
```diff
@@ -23,15 +24,6 @@ vcpkg_install_cmake()
 file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
 
 if(VCPKG_LIBRARY_LINKAGE STREQUAL dynamic AND VCPKG_TARGET_IS_WINDOWS)
```
JackBoosY, line 26:
```suggestion
```

ports/libmodplug/portfile.cmake
```diff
@@ -23,15 +24,6 @@ vcpkg_install_cmake()
...
-        set(BIN_NAME libmodplug.dll)
-    else()
-        set(BIN_NAME modplug.dll)
-    endif()
-    file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/bin)
-    file(RENAME ${CURRENT_PACKAGES_DIR}/lib/${BIN_NAME} ${CURRENT_PACKAGES_DIR}/bin/${BIN_NAME})
-    file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/debug/bin)
-    file(RENAME ${CURRENT_PACKAGES_DIR}/debug/lib/${BIN_NAME} ${CURRENT_PACKAGES_DIR}/debug/bin/${BIN_NAME})
     vcpkg_copy_pdbs()
 endif()
```
JackBoosY, line 28:
```suggestion
vcpkg_copy_pdbs()
```
We don't need the judgment now.

JackBoosY (changes requested):

itsmattkc:
Made requested changes.

JackBoosY (approved):

JackBoosY:
Normally we make those changes with the port update / port feature / other port bug.
But this LGTM.

PR #21410, opened 2021-11-14
[kissfft] update, add missing file

edmBernard:
Update kiss fft and add kiss_fftr.h, kiss_fftnd.h and kiss_fftndr.h in the lib

- #### What does your PR fix?  
There is missing header in the lib

 all triplet are supported

yes

 Yes


ports/kissfft/vcpkg.json
```diff
@@ -1,7 +1,17 @@
 {
   "name": "kissfft",
-  "version-string": "2020-03-30",
+  "version-string": "2021-11-14",
```
NancyLi1013, line 3:
```suggestion
  "version-date": "2021-11-14",
```
This should be served by `version-date`.

ports/kissfft/vcpkg.json
```diff
@@ -1,7 +1,17 @@
...
   "port-version": 1,
```
NancyLi1013, line 4:
```suggestion
```
For a new version update, this field is not needed.

NancyLi1013 (changes requested):

ports/kissfft/vcpkg.json
edmBernard: Doesn't this change will break as previous version use "version-string" ?

edmBernard: Doesn't this change will break version management as previous iteration use version-string

NancyLi1013 (approved):

NancyLi1013 (approved):

PR #21419, opened 2021-11-14
[libsecret] add new port with version 0.20.4

ras0219-msft:
Cherry-picked from https://github.com/microsoft/vcpkg/pull/20185. I added a host dependency on `glib` due to use of `${CURRENT_HOST_INSTALLED_DIR}/...`.

### DIFF
--- a/ports/libsecret/vcpkg.json
+++ b/ports/libsecret/vcpkg.json
@@ -7,6 +7,10 @@
   "supports": "!(windows | uwp | osx)",
   "dependencies": [
     "glib",
+    {
+      "name": "glib",
+      "host": true
+    },
     "libgcrypt"
   ]
 }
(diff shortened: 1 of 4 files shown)


ports/libsecret/vcpkg.json
```diff
@@ -7,6 +7,10 @@
   "supports": "!(windows | uwp | osx)",
   "dependencies": [
     "glib",
+    {
+      "name": "glib",
+      "host": true
+    },
```
Be-ing: Isn't it redundant to have both a non-host dependency and a host dependency of the same package?

Be-ing: Maybe this is the trick to get arm64-windows builds to work with translations for the qtkeychain Qt6 port?

ras0219-msft: Nope, it's not redundant. Consider the case of cross compiling to ios -- you'll need both the runtime libraries for ios as well as the glib tools for osx.

PR #21438, opened 2021-11-15
[faad2] only support windows

autoantwort:
The port only supports windows


ports/faad2/vcpkg.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "faad2",
   "version-string": "2.9.1",
```
JonLiu1993, line 3:
```suggestion
  "version": "2.9.1",
```

JonLiu1993 (approved):

PR #21441, opened 2021-11-15
[libxslt] Fix linux dynamic build and replace absolute paths with relative

Osyotr:
Also remove docs, python and plugins since they all do nothing
Related: #21290

Tested on windows and linux with both static and dynamic linkage.

cc @autoantwort @JackBoosY

### DIFF
--- a/ports/libxslt/portfile.cmake
+++ b/ports/libxslt/portfile.cmake
@@ -21,9 +21,9 @@ vcpkg_from_github(
 
 if (VCPKG_TARGET_IS_WINDOWS)
     # Create some directories ourselves, because the makefile doesn't
-    file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/bin)
+    file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/bin")
     if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "debug")
-        file(MAKE_DIRECTORY ${CURRENT_PACKAGES_DIR}/debug/bin)
+        file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/debug/bin")
     endif()
     set(CONFIGURE_COMMAND_TEMPLATE
         cruntime=@CRUNTIME@
@@ -37,12 +37,14 @@ if (VCPKG_TARGET_IS_WINDOWS)
         zlib=yes
         lzma=yes
     )
+
     # Common
     if(VCPKG_LIBRARY_LINKAGE STREQUAL static)
         set(BUILDSTATIC yes)
     else()
         set(BUILDSTATIC no)
     endif()
+
     # Release params
     if(VCPKG_CRT_LINKAGE STREQUAL dynamic)
         set(CRUNTIME /MD)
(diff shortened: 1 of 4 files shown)


ports/libxslt/portfile.cmake
```diff
@@ -5,10 +5,13 @@ vcpkg_download_distfile(
     SHA512 d08a06616d732993f2131826ca06fafc2e9f561cb1edb17eaf2adaf78e276bb03cba92a773143eb939da04781f5b5e0a09b351d8e4622a941de3cb3d11da731c
 )
 
+# Get this value from configure.ac:21
+set(LIBEXSLT_VERSION 0.8.20)
+set(VERSION 1.1.34)
```
autoantwort, line 10: Imo there is no need for these variables :)

ports/libxslt/portfile.cmake
```diff
@@ -37,137 +40,142 @@ if (VCPKG_TARGET_IS_WINDOWS)
...
+    set(EXSLT_LIBS "-lexslt")
+    set(EXSLT_PRIVATE_LIBS "")
+    file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/lib/pkgconfig")
+    configure_file("${SOURCE_PATH}/libxslt.pc.in" "${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libxslt.pc" @ONLY)
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libxslt.pc" "\nRequires: " "\nRequires: liblzma ")
+    configure_file("${SOURCE_PATH}/libexslt.pc.in" "${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libexslt.pc" @ONLY)
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libexslt.pc" "\nRequires: " "\nRequires: libxslt ")
+    if(NOT VCPKG_BUILD_TYPE)
+        file(COPY "${CURRENT_PACKAGES_DIR}/lib/pkgconfig" DESTINATION "${CURRENT_PACKAGES_DIR}/debug/lib")
     endif()
```
JackBoosY, line 124: That's ... ugly, I wonder when we can provide a vcpkg function to generate the pkgconfig file.

ports/libxslt/portfile.cmake
ras0219-msft: The need for these is to substitute into the .pc file below.

ras0219-msft:
Thanks for the PR!

ports/libxslt/portfile.cmake
ras0219-msft: This way reuses as much from the original project as possible; because we're generating two files, most other methods would take the same amount of code.

dg0yt: I added similar changes recently, and I also wondered if a function on top of `configure_file` would help. The argument wouldn't be much more than a source pc file, and a listing of variables and values, possibly with `_RELEASE` and `_DEBUG`. Still, it would have some advantages:
- It moves the variables out of the main scope.
- It shows all variable usages, instead of having them hidden in some `module.pc.in`.
- It might understand enough pc files syntax to support append to fields and adding new fields.

ports/libxslt/vcpkg.json
```diff
@@ -1,12 +1,17 @@
...
+  "port-version": 2,
   "description": "Libxslt is a XSLT library implemented in C for XSLT 1.0 and most of EXSLT",
   "homepage": "https://github.com/GNOME/libxslt",
   "supports": "!uwp",
   "dependencies": [
     "liblzma",
-    "libxml2"
+    "libxml2",
+    {
+      "name": "vcpkg-pkgconfig-get-modules",
```
Osyotr, line 12: Is it used somewhere? Can't see.

ras0219-msft: Ah! Good catch -- I initially used it to get the values for libxml2, but failed to remove it when I realized that the project's configure would automatically use pkgconfig for me.

ports/libxslt/portfile.cmake
dg0yt: How about a `vcpkg_configure_file`/`vcpkg_configure_pc_file`? It might be a helper port, of course.

ras0219-msft: It seems to me like it would be a lot of code that in the end doesn't really change much. The use sites will be as verbose (or more) and we would be moving further away from standard cmake functionality (`configure_file()`). On the upside, yes, it would allow being more specific about substitutions.

PR #21444, opened 2021-11-15
extend platform expressions to support media query syntax - terminals…

markle11m:
Update documentation to support media queries syntax in platform expressions.

In media queries syntax ([Media Queries](https://www.w3.org/TR/css3-mediaqueries/#syntax)):
- "not" represents logical-not
- "and" represents logical-and
- "," represents a logical-or, but with lower precedence than logical-and
- media queries syntax expresses precedence by treating "," as a list-separator

This change extends platform expressions by using the same list-separator model for "," and by adding the other terminals ("and", "not") as synonyms for the corresponding operators that are already supported ("&", "!"). This change also adds "or" as a synonym for "|" as a potential future extension. 

Open question: does the EBNF need to disambiguate "and", "not", "or" as keywords and not identifiers?

### DIFF
--- a/docs/maintainers/manifest-files.md
+++ b/docs/maintainers/manifest-files.md
@@ -386,7 +386,7 @@ platform-expression-simple =
 
 platform-expression-unary-keyword-operand =
 | required-whitespace, platform-expression-simple
-| platform-expression-grouped ;
+| optional-whitespace, platform-expression-grouped ;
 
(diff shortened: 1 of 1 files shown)


JackBoosY:
I'm not sure that, needs other guys to review this.

ras0219-msft:
LGTM, thanks!

PR #21448, opened 2021-11-16
[libudns] use supports expression

autoantwort:


JackBoosY (approved):

cenit:
isn’t it better to write supports: linux?

JackBoosY:
> isn’t it better to write supports: linux?

We'd better use black list instead of white list. I think.

PR #21450, opened 2021-11-16
[memorymodule] use supports expression

autoantwort:


ports/memorymodule/vcpkg.json
```diff
@@ -1,9 +1,16 @@
 {
   "name": "memorymodule",
   "version-string": "2019-12-31",
-  "port-version": 1,
+  "port-version": 2,
   "description": "Library to load a DLL from memory",
   "homepage": "https://github.com/fancycode/MemoryModule",
+  "supports": "!linux & !osx & !arm",
```
JackBoosY, line 7: It seems that this should be `!linux & !osx & !uwp`?

JackBoosY (changes requested):

JackBoosY (approved):

ports/memorymodule/vcpkg.json
autoantwort: Yeah right

PR #21451, opened 2021-11-16
[gdal] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/gdal/portfile.cmake
+++ b/ports/gdal/portfile.cmake
@@ -366,6 +366,11 @@ else()
         vcpkg_replace_string("${pc_file_debug}" "${exec_prefix}/include" "${prefix}/../include")
     endif()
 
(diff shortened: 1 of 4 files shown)


PhoebeHui (approved):

ports/gdal/portfile.cmake
```diff
@@ -366,6 +366,9 @@ else()
         vcpkg_replace_string("${pc_file_debug}" "${exec_prefix}/include" "${prefix}/../include")
     endif()
 
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/gdal/bin/gdal-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../..")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/tools/gdal/debug/bin/gdal-config" "${CURRENT_INSTALLED_DIR}" "`dirname $0`/../../../..")
```
dg0yt, line 370: This breaks release-only builds. Please add a `if`.

dg0yt (changes requested):

PR #21469, opened 2021-11-16
[docs] Document `VCPKG_BUILD_TYPE`

### LINKED ISSUE #21360: How do I compile only release version?
Because I usually only use the Release version, but vcpkg will compile Debug and Release by default when installing, which consumes more installation time and harddisk space. Is there a command to compile only the Release version library? Thanks!

autoantwort:
Fixes #21360

### DIFF
--- a/docs/users/triplets.md
+++ b/docs/users/triplets.md
@@ -44,7 +44,7 @@ Specifies the preferred library linkage.
 Valid options are `dynamic` and `static`. Note that libraries can ignore this setting if they do not support the preferred linkage type.
 
 ### VCPKG_BUILD_TYPE
-You can set this value to `release` to only build release versions of the ports. By default this value is empty and release and debug versions of a port are build.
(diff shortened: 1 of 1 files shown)


docs/users/triplets.md
```diff
@@ -43,6 +43,9 @@ Specifies the preferred library linkage.
 
 Valid options are `dynamic` and `static`. Note that libraries can ignore this setting if they do not support the preferred linkage type.
 
+### VCPKG_BUILD_TYPE
+You can set this value to `release` to only build release versions of the ports. By default this value is empty and release and debug versions of a port are build.
```
NancyLi1013, line 47:
```suggestion
You can set this value to `release` to only build release versions of the ports. By default this value is empty and release and debug versions of a port are built.
```

This might be used as passive verb here. What do you think?

NancyLi1013 (approved):

PR #21481, opened 2021-11-17
[cpp-ipc] Update to 1.0.1

winsoft666:
**Describe the pull request**

- #### What does your PR fix?  
  Update version to 1.0.1

  <linux, windows / osx>, <Yes>

  `Yes`

  <Yes>

### DIFF
--- a/scripts/ci.baseline.txt
+++ b/scripts/ci.baseline.txt
@@ -160,7 +160,6 @@ cppmicroservices:arm-uwp=fail
 cppmicroservices:x64-uwp=fail
 cpp-netlib:arm-uwp=fail
 cpp-netlib:x64-uwp=fail
-cpp-ipc:x64-osx=skip
 cppcoro:x64-linux=fail
 cppcoro:arm-uwp=fail
 cppcoro:x64-uwp=fail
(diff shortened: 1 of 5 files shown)


scripts/ci.baseline.txt
```diff
@@ -160,6 +160,7 @@ cppmicroservices:arm-uwp=fail
 cppmicroservices:x64-uwp=fail
 cpp-netlib:arm-uwp=fail
 cpp-netlib:x64-uwp=fail
+cpp-ipc:x64-osx=skip
```
JonLiu1993, line 163: @winsoft666 ,Thanks for your pr, could you clarify why skip osx triplet？

winsoft666: cpp-ipc is not support macOS

JonLiu1993: We have used supports and vcpkg_fail_port_install(ON_TARGET "OSX") in the vcpkg.json and portfile.cmake files respectively, and in ci.baseline.txt it is not appropriate to indicate that the triplet that Upstream does not support itself, here is more suitable for some Upstream It supports triplet that we skipped because of other failures

JonLiu1993: So I suggest to keep the same as the previous version and delete this line

winsoft666: OK, has deleted.

JonLiu1993 (approved):

PR #21504, opened 2021-11-18
[hwloc] update to 2.6.0

### LINKED ISSUE #20899: hwloc update to 2.6 (it has native CMake support)
Library name: hwloc
New version number: 2.6

Hello
I am the upstream maintainer of hwloc. I was told about this project and we had some users requesting native CMake support for building hwloc on Windows. So I took your CMakeLists.txt as a base, updated it to hwloc 2.6 and added some additional features. See https://github.com/open-mpi/hwloc/pull/484

Differences between your CMakeLists.txt and mine:
* I commented-out your snprintf/strtoull reenabling because we had some issues with them in the past, and I don't know how to make sure we're not running on such old/broken build environment. It shoudn't hurt much anyway.
* I didn't apply any change that is currently in portfile.cmake because I am not why they are needed and why they are not in CMakeLists.txt
* all tools and lstopo are now built but may be disabled with HWLOC_SKIP_LSTOPO and HWLOC_SKIP_TOOLS
* I install these tools in bin/ instead tools/hwloc/ (if you need a way to customize that, let me know).
* dolib.c isn't needed *inside* the library, it's an external tool that helps linking the library on MinGW

If you have some feedback before I release hwloc 2.6 officially next week, I'll be happy to update the CMakeLists. Otherwise we'll update it in future releases.

In the future, I'd like to be able to rename libhwloc.dll into libhwloc-15.dll to match other builds. If there's a need to disable that, let me know.

FrankXie05:
Fix #20899 

Update hwloc version to 2.6.0

Note: no feature need to test

### DIFF
--- a/ports/hwloc/vcpkg.json
+++ b/ports/hwloc/vcpkg.json
@@ -1,12 +1,22 @@
 {
   "name": "hwloc",
-  "version": "2.5.0",
+  "version": "2.6.0",
   "port-version": 1,
   "maintainers": "bgoglin<Brice.Goglin@inria.fr>",
   "description": [
     "Portable Hardware Locality (hwloc)",
     "The Portable Hardware Locality (hwloc) software package provides a portable abstraction (across OS, versions, architectures, ...) of the hierarchical topology of modern architectures, including NUMA memory nodes, sockets, shared caches, cores and simultaneous multithreading. It also gathers various system attributes such as cache and memory information as well as the locality of I/O devices such as network interfaces, InfiniBand HCAs or GPUs."
   ],
+  "dependencies": [
+      {
+        "name": "vcpkg-cmake",
+        "host": true
+    },
+    {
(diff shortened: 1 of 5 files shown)


ports/hwloc/vcpkg.json
```diff
@@ -1,12 +1,22 @@
...
+  "dependencies": [
+    {
+      "name": "vcpkg-cmake",
+      "host": true
+    },
+    {
+      "name": "vcpkg-cmake-config",
+      "host": true
+    }
+  ]
```
NancyLi1013, line 21:
```suggestion
```
Seems like this is not used in this port, why do you add the dependencies?

NancyLi1013 (changes requested):

PhoebeHui:
Confirm with Frank, the original repo was removed by mistake, he will summit a new PR for this.

PR #21508, opened 2021-11-18
[freetype] Switch build system to meson

### LINKED ISSUE #20246: [freetype] freetype 2.11.0 dll produces crash when built with some MSVC compilers
**Host Environment**
- OS: Windows
- Compiler: Visual Studio 16 2019 (MSVC 19.29.30037.0) also github actions `windows-latest` build image

**To Reproduce**
Do this in a visual studio cmd
```
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
.\bootstrap-vcpkg.bat
.\vcpkg.exe install freetype:x64-windows
cd ..
git clone https://github.com/itszn/typetest.git
cd typetest
mkdir build
cd build
cmake -G "Visual Studio 16 2019" -DCMAKE_TOOLCHAIN_FILE=..\..\vcpkg\scripts\buildsystems\vcpkg.cmake -DCMAKE_BUILD_TYPE=Debug ..
msbuild typetest.sln
.\Debug\typetest.exe
```
Expected output is `Survived FT_Load_Glyph`

# Bug
Freetype 2.11.0 built with vcpkg will compile with an empty `af_writing_system_classes`. This causes a null pointer dereference when trying to create glyphs with autofit.

freetype uses an [include file](https://gitlab.freedesktop.org/freetype/freetype/-/blob/801cd842e27c85cb1d5000f6397f382ffe295daa/src/autofit/afwrtsys.h) with macros to initialize `af_writing_system_classes`. However when building with vcpkg, this fails to correctly populate leading to the crash.

The following patch to freetype seems to fix the issue:

```patch
diff --git a/src/autofit/afwrtsys.h b/src/autofit/afwrtsys.h
index 39aa121..81e6108 100644
--- a/src/autofit/afwrtsys.h
+++ b/src/autofit/afwrtsys.h
@@ -27,6 +27,8 @@
 #include "afcjk.h"
 #include "afindic.h"
 
+#else
+
 #endif /* AFWRTSYS_H_ */
```

The patch adds an `#else` statement to the header before the `#endif`. This theoretically should be a no-op, but vcpkg produces different objects:

Without patch:
image

With patch:
image

JackBoosY:
Since the cmake build system in freetype is unofficial and has seriously fallen behind the meson build system, switch to meson.

Fixes #20246.


dg0yt:
Will the meson build generate cmake config files or pkgconfig files?
If yes, we should verify that they don't contain invalid release/debug mixes, as current cmake config does on linux (reported from #21415).

JackBoosY:
@dg0yt meson will also generate and install pkgconfig files, and I checked the debug lib name, it's same with the release lib name.
Should we add suffix `d` to the debug lib?

dg0yt:
> Should we add suffix `d` to the debug lib?

NO, no deviation from defaults!

I'm concerned about the transitive dependencies (for static linking), not about freetype itself. As in:
~~~
freetype/stdout-x64-linux.log:  Unguarded release path /lib/ path at [>>>]:
freetype/stdout-x64-linux.log:       INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:ZLIB::ZLIB>;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/vcpkg-ci/installed/x64-linux/lib/libbz2.a>;\$<\$<CONFIG:DEBUG>:/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libbz2d.a>;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/vcpkg-ci/installed/x64-linux/lib/libpng.a>;\$<\$<CONFIG:DEBUG>:/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libpng16d.a>;\$<\$<NOT:\$<CONFIG:DEBUG>>:/mnt/vcpkg-ci/installed/x64-linux/lib/libz.a>;\$<\$<CONFIG:DEBUG>:/mnt/vcpkg-ci/installed/x64-linux/debug/lib/libz.a>;\$<LINK_ONLY:m>;[>>>]/mnt/vcpkg-ci/installed/x64-linux/lib/libbrotlidec-static.a;/mnt/vcpkg-ci/installed/x64-linux/lib/libbrotlicommon-static.a"
~~~

JackBoosY:
@dg0yt Agreed, the cmake provided module `FindFreetype.cmake` may find the mismatched library, this should be noticed in this PR.
For the pkgconfig, it should be correct because the debug and release files are separated.

dg0yt:
You are right, there is a CMake find module. So we need to get the wrapper right.

ports/imgui/CMakeLists.txt
```diff
@@ -102,8 +102,8 @@ if(IMGUI_BUILD_WIN32_BINDING)
 endif()
 
 if(IMGUI_FREETYPE)
-    find_package(freetype CONFIG REQUIRED)
-    target_link_libraries(${PROJECT_NAME} PUBLIC freetype)
+    find_package(freetype REQUIRED)
```
dg0yt, line 105: The official CMake package is spelled `Freetype`, the find module is `FindFreetype.cmake`.

ports/freetype/portfile.cmake
```diff
@@ -41,6 +41,19 @@ file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
...
+        "\${includedir}/freetype2"
+        "\${includedir}"
+    )
+endif()
+if(EXISTS "${CURRENT_PACKAGES_DIR}/lib/pkgconfig/freetype2.pc")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/lib/pkgconfig/freetype2.pc"
+        "\${includedir}/freetype2"
+        "\${includedir}"
+    )
+endif()
```
dg0yt, line 55: Please don't use `EXISTS` for build type variants. It hides configuration errors such as misspelled file names or installed paths.

JackBoosY:
@Neumann-A Do you know why the freetype library suffix generated by meson is `.so`?

JackBoosY:
Depends on #21537.

JackBoosY:
Temporary close this issue.

PR #21522, opened 2021-11-18
[botan] pass -fPIC for static libraries

### LINKED ISSUE #21461: botan2 build static botan-2.a with enable -fPIC
****

```
/usr/bin/ld: /usr/local/share/vcpkg/installed/x64-linux/debug/lib/libbotan-2.a(src_lib_ffi.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
```

Full log: https://github.com/rnpgp/rnp/runs/4222343272?check_suite_focus=true#step:7:724

**Proposed solution**

Always compile with `--extra-cxxflags=-fPIC`

IDK about possible alternatives

**Additional context**

More logs https://github.com/rnpgp/rnp/runs/4222343068

CAMOBAP:
**Describe the pull request**

- #### What does your PR fix?  
  Fixes #21461

  all, No

  This is my first PR, but looks like it follow the rules

  Done

### DIFF
--- a/ports/botan/portfile.cmake
+++ b/ports/botan/portfile.cmake
@@ -80,7 +80,8 @@ function(BOTAN_BUILD BOTAN_BUILD_TYPE)
                             "--distribution-info=vcpkg ${TARGET_TRIPLET}"
                             --prefix=${BOTAN_FLAG_PREFIX}
                             --with-pkg-config
-                            --link-method=copy)
+                            --link-method=copy
+                            --extra-cxxflags="-fPIC")
     if(CMAKE_HOST_WIN32)
         list(APPEND configure_arguments ${BOTAN_MSVC_RUNTIME}${BOTAN_MSVC_RUNTIME_SUFFIX})
     endif()
--- a/ports/botan/vcpkg.json
+++ b/ports/botan/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "botan",
   "version": "2.18.1",
-  "port-version": 4,
+  "port-version": 5,
   "description": "A cryptography library written in C++11",
   "homepage": "https://botan.randombit.net",
   "supports": "!(windows & arm)",
--- a/versions/b-/botan.json
+++ b/versions/b-/botan.json
@@ -1,9 +1,9 @@
 {
   "versions": [
     {
-      "git-tree": "477e259691b81a016da459532f616b5fca230a91",
+      "git-tree": "1bcf63938b24ecc9684b1941bb908ec052a28eac",
       "version": "2.18.1",
-      "port-version": 4
+      "port-version": 5
     },
     {
       "git-tree": "045c95bf0e7f1352d91398bc0dc41ce5ec5a288a",
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -1130,7 +1130,7 @@
     },
     "botan": {
       "baseline": "2.18.1",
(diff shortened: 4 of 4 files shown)


LilyWangLL:
@CAMOBAP Thanks for your PR. Please bump the port version. See [documentation](https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#update-the-version-field-in-the-control-file-of-any-modified-ports).

LilyWangLL (changes requested):

CAMOBAP:
@LilyWangLL I think I already did this, here is result in this PR https://github.com/microsoft/vcpkg/pull/21522/files#diff-d3ae18b4fe81bb994249a364fa35b1d04110254662de8793da49ffb8f33a44a5L4-R4

CAMOBAP:
@LilyWangLL can I update `port-version` manually?

CAMOBAP:
Updated manually @LilyWangLL could you please have a look

dg0yt:
I do understand that linking `libbotan-2.a` into a a shared object may need `-fpic` (#21461). But I doubt that adding the parameter in vcpkg is the right solution, because this forces the disadvantages of PIC on all users.
Alternatives:
- Use a dynamic triplet when you want shared objects.
- Use a custom triplet which modifies the flags only for building port `botan`.
- Use an overlay port for `botan` which sets the flag.

CAMOBAP:
@dg0yt thanks for the feedback

> Use a dynamic triplet when you want shared objects.

Not an option unfortunatelly

> Use a custom triplet which modifies the flags only for building port botan.

AFAIK custom triplets means creation for dedicated triplet cmake file, which cannot affect boatn port. Initally I searched for some way to 'override' `BOTAN_FLAG_STATIC` but there is no such option

> Use an overlay port for botan which sets the flag.

Never heard about it, let me check

dg0yt:
> > Use a custom triplet which modifies the flags only for building port botan.
>
> AFAIK custom triplets means creation for dedicated triplet cmake file, which cannot affect boatn port. Initally I searched for some way to 'override' BOTAN_FLAG_STATIC but there is no such option

The triplet file is loaded before `portfile.cmake`. So in your custom triplet file, you could do something like:

~~~cmake
if(PORT STREQUAL "botan")
    # modify vcpkg flags as you like
endif()
~~~

CAMOBAP:
@dg0yt thanks for the comment overlay ports solve my issues

> The triplet file is loaded before portfile.cmake. So in your custom triplet file, you could do something like...

I understand but because botan build system is configure-based, I don't see an option how custom triplet can help here (probably because I has lack of experience with vcpkg). I will be glad if you can share some examples of custom tripets just for me to understand their limits.

I'm totally ok to close this PR

Thanks again

dg0yt:
> I understand but because botan build system is configure-based, I don't see an option how custom triplet can help here.

`vcpkg_configure_make` tries to use the same flags that would be used for cmake. So you can use `VCPKG_C_FLAGS` etc., 
https://github.com/microsoft/vcpkg/blob/master/docs/users/triplets.md

> I will be glad if you can share some examples of custom tripets just for me to understand their limits.

There is nothing magical. The main downside is that changing the triplet changes the hashes which control binary artifact caching.

PR #21540, opened 2021-11-19
[OpenSSL] Allow customization of openssldir

lstipakov:
OpenSSL searches for config file in openssldir at runtime. By default
openssldir is set to current build directory.

When an app, which uses vcpkg for building dependencies, is
distributed to another machine, searching for openssl config
in build directory on build machine doesn't make much sense.

This enables OpenSSL port users to customize this directory
with custom triplet, for example:

    set(OPENSSL_OPENSSLDIR "c:\\Program Files\\SSL")

In this case config is searched in directory which is not writable
by default.

Since vcpkg utilizes install_ssldirs target to copy config etc
to abovementioned directory, a build may fail if a directory is
read-only. To fix that, add OPENSSL_NO_INSTALL_SSLDIRS
boolean flag which makes vcpkg skip install_ssldirs target:

  set(OPENSSL_NO_INSTALL_SSLDIRS ON)

Signed-off-by: Lev Stipakov <lev@openvpn.net>

- #### What does your PR fix?  
  See commit message above.

  Windows triplets are affected. No, haven't updated CI baseline.

  Yes

  Yes

### DIFF
--- a/ports/openssl/vcpkg.json
+++ b/ports/openssl/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "openssl",
   "version-string": "1.1.1l",
-  "port-version": 4,
+  "port-version": 5,
   "description": "OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols. It is also a general-purpose cryptography library.",
   "homepage": "https://www.openssl.org"
 }
--- a/ports/openssl/vcpkg.json
+++ b/ports/openssl/vcpkg.json
@@ -1,7 +1,7 @@
 {
   "name": "openssl",
   "version-string": "1.1.1l",
-  "port-version": 3,
+  "port-version": 4,
   "description": "OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols. It is also a general-purpose cryptography library.",
   "homepage": "https://www.openssl.org"
 }
--- a/ports/openssl/windows/portfile.cmake
+++ b/ports/openssl/windows/portfile.cmake
@@ -52,6 +52,11 @@ set(OPENSSL_MAKEFILE "makefile")
 file(REMOVE_RECURSE "${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel"
                     "${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-dbg")
 
+set(OPENSSL_INSTALL_TARGETS install_sw)
+if(NOT DEFINED OPENSSL_NO_INSTALL_SSLDIRS)
+    set(OPENSSL_INSTALL_TARGETS ${OPENSSL_INSTALL_TARGETS} install_ssldirs)
+endif()
+
 if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
 
     # Copy openssl sources.
@@ -63,11 +68,15 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
     message(STATUS "Copying openssl release source files... done")
     set(SOURCE_PATH_RELEASE "${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-rel")
 
-    set(OPENSSLDIR_RELEASE ${CURRENT_PACKAGES_DIR})
+    if (DEFINED OPENSSL_OPENSSLDIR)
+        set(OPENSSLDIR_RELEASE ${OPENSSL_OPENSSLDIR})
+    else()
+        set(OPENSSLDIR_RELEASE ${CURRENT_PACKAGES_DIR})
+    endif()
 
     message(STATUS "Configure ${TARGET_TRIPLET}-rel")
     vcpkg_execute_required_process(
-        COMMAND ${CONFIGURE_COMMAND} ${OPENSSL_ARCH} "--prefix=${OPENSSLDIR_RELEASE}" "--openssldir=${OPENSSLDIR_RELEASE}" -FS
+        COMMAND ${CONFIGURE_COMMAND} ${OPENSSL_ARCH} "--prefix=${CURRENT_PACKAGES_DIR}" "--openssldir=${OPENSSLDIR_RELEASE}" -FS
         WORKING_DIRECTORY ${SOURCE_PATH_RELEASE}
         LOGNAME configure-perl-${TARGET_TRIPLET}-rel
     )
@@ -84,7 +93,7 @@ if(NOT DEFINED VCPKG_BUILD_TYPE OR VCPKG_BUILD_TYPE STREQUAL "release")
         ERROR_FILE ${CURRENT_BUILDTREES_DIR}/build-${TARGET_TRIPLET}-rel-0-err.log
     )
     vcpkg_execute_required_process(
-        COMMAND nmake -f ${OPENSSL_MAKEFILE} install_sw install_ssldirs
+        COMMAND nmake -f ${OPENSSL_MAKEFILE} ${OPENSSL_INSTALL_TARGETS}
         WORKING_DIRECTORY ${SOURCE_PATH_RELEASE}
         LOGNAME build-${TARGET_TRIPLET}-rel-1)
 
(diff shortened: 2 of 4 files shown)


Neumann-A:
This PR is invalid. Use a port overlay for this case? Other question: Is there a absolute path in `openssl.cnf` which should be relativ?

lstipakov:
> This PR is invalid

I am sorry, could you be more specific?

> Use a port overlay for this case?

I think the current behavior, when openssl.cnf is hardcoded to a vcpkg path on a build machine, is wrong. How it is supposed to work when software is used on other machines? I can surely use the port overlay, but I believe this patch will benefit many users of this port. 

> Other question: Is there a absolute path in `openssl.cnf` which should be relativ?

I checked openssl.cnf which comes from openssl and I don't see any absolute paths.

Neumann-A:
> I am sorry, could you be more specific?

By setting `set(OPENSSL_OPENSSLDIR "c:\\Program Files\\SSL")` you seem to be wanting an empty openssl port with an external ssl library. This makes this PR invalid.

> searching for openssl config in build directory on build machine doesn't make much sense.

the search should happen within the relativ install try of vcpkg or alongside the application you plan to deploy. Not some random hardcoded `c:\\Program Files\\SSL`

cron2:
Hi,

On Fri, Nov 19, 2021 at 08:22:33AM -0800, Alexander Neumann wrote:
> > I am sorry, could you be more specific?
> 
> By setting `set(OPENSSL_OPENSSLDIR "c:\\Program Files\\SSL")` you seem to be wanting an empty openssl port with an external ssl library. This makes this PR invalid.

The goal is to make the OpenSSL *config* directory point to a location
where it can then be used on the target machine for an openssl.cnf.

In our case, this would be part of the OpenVPN tree.

Right now, the OpenSSL libraries built search their openssl.cnf in 
the directory that was used to build OpenSSL on the build machine - which
is not useful in any way.

> > searching for openssl config in build directory on build machine doesn't make much sense.
> 
> the search should happen within the relativ install try of vcpkg or alongside the application you plan to deploy. Not some random hardcoded `c:\\Program Files\\SSL`

Your current port has it hardcoded in the build directory, which is
what I'd call "a random hardcoded directory".

This PR makes the target location more flexible.

(For us, this would be c:\\program files\openvpn\ssl or something
under our control)

gert


-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

Neumann-A:
> Your current port has it hardcoded in the build directory, which is
what I'd call "a random hardcoded directory".
> This PR makes the target location more flexible.

The PR should focus on making the search location for `openssl.cnf` relativ to the executable/dll and not focus on making it more "flexible" by encoding a different hardcoded path ;)

lstipakov:
If I read the code right, this PR sets search location for openssl.cnf to whatever value user has passed via OPENSSL_OPENSSLDIR. No?

Lähetetty iPhonesta

lstipakov:
I have added a commit to also customize ENGINESDIR. I hope commit message explains rationale reasonably.

cron2:
Hi,

On Fri, Nov 19, 2021 at 12:39:40PM -0800, Alexander Neumann wrote:
> > Your current port has it hardcoded in the build directory, which is
> what I'd call "a random hardcoded directory".
> > This PR makes the target location more flexible.
> 
> The PR should focus on making the search location for `openssl.cnf` relativ to the executable/dll and not focus on making it more "flexible" by encoding a different hardcoded path ;)

This is not something OpenSSL supports "as is", so would be a much
more intrusive patch into OpenSSL code.

I see this PR as an improvement - from "path is not usable for anything"
to "possibly useful in the context of the project building this port"
(so it's not "arbitrarily hardcoded" but "the calling project can specify
what will be compiled-in").

Of course, relative to the final install directory, or even using
windows config mechanisms (%USERPROFILE%), would be even better - but
much more work.  Can we agree on doing a slight improvement as a first
step?

gert

-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

Neumann-A:
What about?
```
    if ((t = getenv("OPENSSL_CONF")) != NULL)
        return OPENSSL_strdup(t);
```
so setting the environment variable `OPENSSL_CONF` should achieve the same as this PR. It is also probably that setting OPENSSLDIR to some relativ path would also work.

cron2:
Hi,

On Mon, Nov 22, 2021 at 10:39:25AM -0800, Alexander Neumann wrote:
> What about?
> ```
>     if ((t = getenv("OPENSSL_CONF")) != NULL)
>         return OPENSSL_strdup(t);
> ```
> so setting the environment variable `OPENSSL_CONF` should achieve the 
> same as this PR. 

But the port would still try to access a default location that does not
make sense for anyone - and it would require that all packages *using*
the port would install mechanisms to setup environment variables correctly,
which is not straightforward for a windows service, for example...

> It is also probably that setting OPENSSLDIR to some relativ path would also work. 

Provided the current working directory is set reasonably, which is 
unlikely to be guaranteed in every case - possibly creating a security
issue again.

gert

-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

Neumann-A:
> But the port would still try to access a default location that does not
make sense for anyone 

Why would it try to access the defaults? All defaults can be set via environment variable (SSL_CERT_DIR; SSL_CERT_FILE).
I don't see how assuming `c:\\Program Files\\SSL` as the default is more correct than CWD.

lstipakov:
But C:\Program Files\SSL is not hardcoded anywhere, it is up to port users to define this path via overlay triplet. As a port user I could use something like C:\Program Files\My Software\SSL to make sure unprivileged users won’t be able to alter the behavior of my software which uses OpenSSL port. 

Lähetetty iPhonesta

Neumann-A:
> C:\Program Files\My Software\SSL to make sure unprivileged users won’t be able to alter the behavior of my software which uses OpenSSL port.

And that is where you are wrong. Setting the environment variables will always overwrite your setting. Also what happens if the software is installed to a custom path instead of the default one?

cron2:
Hi,

On Mon, Nov 22, 2021 at 11:00:03AM -0800, Alexander Neumann wrote:
> > But the port would still try to access a default location that does not
> > make sense for anyone 
> 
> Why would it try to access the defaults? 

Because that's what OpenSSL does - it tries to read the config file from
the compiled in default location.  Not my code, not my decisions, it's
just the way that it is.

> All defaults can be set via environment variable (SSL_CERT_DIR; SSL_CERT_FILE).

Yes, this I already acknowledged.  But this takes care and extra code
to get right, especially when running as a windows service.

> I don't see how assuming `c:\\Program Files\\SSL` as the default is more correct than CWD. 

CWD is never correct for security sensitive files.

Note that I'm not talking about "c:\\program files\\ssl" but specifically
about "c:\\program files\\openvpn\\ssl" which is exactly where we need
this to be - and since the PR adds this as a compile-time option, every
other program can have this wherever that other program needs it to be.

vcpkg stuff is not installed independently of the program it's compiled
with, so it needs to work for the software reloying on the port - no?

gert
-- 
"If was one thing all people took for granted, was conviction that if you 
 feed honest figures into a computer, honest figures come out. Never doubted 
 it myself till I met a computer with a sense of humor."
                             Robert A. Heinlein, The Moon is a Harsh Mistress

Gert Doering - Munich, Germany                             ***@***.***

selvanair:
> > C:\Program Files\My Software\SSL to make sure unprivileged users won’t be able to alter the behavior of my software which uses OpenSSL port.
> 
> And that is where you are wrong. Setting the environment variables will always overwrite your setting. Also what happens if the software is installed to a custom path instead of the default one?

Sounds like the objective of setting the default config path to a read-only location is misunderstood. When the default path points to a location accessible to anyone, user A running a program linked to OpenSSL can be attacked by user B by adding  a config file and that could force-load malicious dlls into the process. Unless user A ensures that an env var named `OPENSSL_CONF` is set to point to a location that other users cannot write to. Not all applications do this, its beyond a casual user to do this, and this has been the subject of some CVE's in the past. Someone even went ahead and wrote a script to check OpenSSL libraries with vulnerable `openssldir` (https://github.com/mirchr/openssldir_check)

In short, setting `openssldir` to a location potentially writable by "users" is wrong. For packaging purposes you can relocate it to any convenient location, but the path built-in to the library must remain read-only for unprivileged users on the target machine.

Of course, the user can over-ride the hard-coded path using env vars, and that is perfectly fine and safe. Its unfortunate that OpenSSL developers decided to use a hard-coded path in the first place, instead of defaulting to the location of the dll or refusing to load the config without an explicit instruction with an absolute path or some such. But that's beyond our control.

Setting the default location to a read-only  directory like "C:\Windows\system32\ssl" protects all users of the library from such attacks. This PR does exactly that by making the openssldir location configurable during build.

PR #21587, opened 2021-11-22
[boinc] Fix crosscompilation on linux for arm/arm64/armv6

AenBleidd:
Signed-off-by: Vitalii Koshura <lestat.de.lionkur@gmail.com>

### DIFF
--- a/ports/boinc/portfile.cmake
+++ b/ports/boinc/portfile.cmake
@@ -14,6 +14,14 @@ vcpkg_from_github(
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
 
 if(VCPKG_TARGET_IS_LINUX)
+    if(${VCPKG_TARGET_ARCHITECTURE} MATCHES "arm64")
+        set(OPTIONS --host=aarch64-linux)
+    elseif(${VCPKG_TARGET_ARCHITECTURE} MATCHES "armv6")
+        set(OPTIONS --host=armv6-linux)
+    elseif (${VCPKG_TARGET_ARCHITECTURE} MATCHES "arm")
+        set(OPTIONS --host=arm-linux)
+    endif()
+
     vcpkg_configure_make(
         SOURCE_PATH ${SOURCE_PATH}
         AUTOCONFIG
--- a/ports/boinc/vcpkg.json
+++ b/ports/boinc/vcpkg.json
@@ -1,7 +1,8 @@
 {
   "name": "boinc",
   "version": "7.18.1",
-  "port-version": 2,
+  "port-version": 3,
+  "maintainers": "Vitalii Koshura <lestat.de.lionkur@gmail.com>",
   "description": "Open-source software for volunteer computing and grid computing.",
   "homepage": "https://boinc.berkeley.edu/",
   "supports": "!(windows & arm) & !uwp",
--- a/versions/b-/boinc.json
+++ b/versions/b-/boinc.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "17dafe24f900d46baab70623c23b515893040b3c",
+      "version": "7.18.1",
+      "port-version": 3
+    },
+    {
       "git-tree": "ee6667eb5cf5a3fee75fd35de998ad322c5999ff",
       "version": "7.18.1",
       "port-version": 2
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -510,7 +510,7 @@
     },
     "boinc": {
       "baseline": "7.18.1",
-      "port-version": 2
+      "port-version": 3
     },
     "bond": {
       "baseline": "9.0.3",


ports/boinc/portfile.cmake
```diff
@@ -14,6 +14,14 @@ vcpkg_from_github(
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
 
 if(VCPKG_TARGET_IS_LINUX)
+    if (${VCPKG_TARGET_ARCHITECTURE} MATCHES "arm")
+        set(OPTIONS --host=arm-linux)
+    elseif(${VCPKG_TARGET_ARCHITECTURE} MATCHES "arm64")
+        set(OPTIONS --host=aarch64-linux)
+    elseif(${VCPKG_TARGET_ARCHITECTURE} MATCHES "armv6")
+        set(OPTIONS --host=armv6-linux)
+    endif()
```
JackBoosY, line 23: This should be processed by `vcpkg_configure_make`, can you double confirm that?

JackBoosY (changes requested):

ports/boinc/portfile.cmake
AenBleidd: Without this parameter 'configure' scripts fails to detect correct environment. So this is an essential change for this function. Tested with custom port here: https://github.com/BOINC/boinc-apps/pull/26

JackBoosY: Here:
https://github.com/microsoft/vcpkg/blob/653912591af709c162fba2d4000a6b3362c8d3c9/scripts/cmake/vcpkg_configure_make.cmake#L108-L143

vcpkg_configure_make will detect the host arch to suitable the configure options by adding option `DETERMINE_BUILD_TRIPLET`. So in the correct triplet it should have no issues. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_configure_make.md#determine_build_triplet
Can you double confirm that?
If not, you should use option `BUILD_TRIPLET` instead of this. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/vcpkg_configure_make.md#build_triplet

AenBleidd: I believe here's the some kind of confusion.
Function you mentioned correctly recognise host_cpu, but when I build 'arm' on 'amd64', I need to pass to 'configure' script not the host cpu ('amd64' in this case) but target cpu ('arm-linux').
So this is not an issue with vcpkg, this might be an issue in our scripts (wrong variable naming, probably?). I'll investigate this in BOINC codebase, but since next release is not planned for a few next months, this fix for this port will be fine (since I'm a maintainer for both of these products).

JackBoosY: cc @Neumann-A

ports/boinc/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "boinc",
   "version": "7.18.1",
-  "port-version": 2,
+  "port-version": 3,
```
JackBoosY, line 4: Also, since you are the maintainer, can you please add `maintainers` here? 
See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/manifest-files.md#maintainers

JackBoosY (changes requested):

ports/boinc/portfile.cmake
JackBoosY: I think you should use `--target` instead of `--host` since you are using cross build here?

AenBleidd: According to this ( https://gcc.gnu.org/onlinedocs/gccint/Configure-Terms.html ) I need to specify only `--host` since I'm using the correct compiler that builds for specific architecture.
Also, we already use exactly the `--host` parameter: https://github.com/BOINC/boinc/blob/06cc5302991217bef89ba60404e584bd94924d2a/android/build_boinc_arm.sh#L76
This is an example of Android script, but the logic is the same.

ports/boinc/vcpkg.json
AenBleidd: Fixed

ports/boinc/portfile.cmake
JackBoosY: >  need to specify only --host since I'm using the correct compiler that builds for specific architecture.

Therefore, I think you should create a custom triplet file and use it in the build command.
Modifying here will cause this "special" configuration to be used in all triplets, so I think it is inappropriate.

JackBoosY: I didn't understand which `specific architecture` mean.
`arm64-linux` or `arm-linux` uses gcc to build by default. When building `arm/arm64-linux` in `x86/x64-linux`, `--host` should be `x86/x64`, and `--target` should be `arm/arm64`.
Right?

AenBleidd: This is an error log with debug output `before` this change: https://github.com/BOINC/boinc-apps/runs/4280686170?check_suite_focus=true

Lines of interest:
```
    Command failed: /usr/bin/bash -c "V=1 ./../src/18_-7.18.1-1910b9f200.clean/configure  \"--disable-server\" \"--disable-client\" \"--disable-manager\" \"--disable-silent-rules\" \"--verbose\" \"--disable-shared\" \"--enable-static\" \"--prefix=/home/runner/work/boinc-apps/boinc-apps/build/boinc-autodock-vina/armvfpv3-linux-static/vcpkg_installed/armvfpv3-linux-static\""
```
As you can see above, `--host` parameter was not passed
And the hint from the error log:
```
checking whether we are cross compiling... configure: error: in `/home/runner/work/boinc-apps/boinc-apps/vcpkg/buildtrees/boinc/armvfpv3-linux-static-rel':
configure: error: cannot run C compiled programs.
If you meant to cross compile, use `--host'.
```

AenBleidd: > I didn't understand what `specific architecture` mean.
> `arm64-linux` or `arm-linux` uses gcc to build by default. When building `arm/arm64-linux` in `x86/x64-linux`, `--host` should be `x86/x64`, and `--target` should be `arm/arm64`.
> Right?

No.
`host` is the target architecture
`target` is used when you build toolchains.
Here's a good answer: https://stackoverflow.com/a/15901574/4210508

JackBoosY: Can you please add `DETERMINE_BUILD_TRIPLET` to `vcpkg_configure_make` and try again?

AenBleidd: It doesn't work: https://github.com/BOINC/boinc-apps/runs/4283961130?check_suite_focus=true

JackBoosY: @Neumann-A What do you think about?

AenBleidd: @JackBoosY, could you please remove 'requires:authors-response' label? Technically I answered your questions, so you are not waiting for me ^^

dg0yt: After resolving the initial confusion about autotools `--host`, the question remains where armv6 handling should be done, in port boinc, or here:
https://github.com/microsoft/vcpkg/blob/30465138ef7facf1d4c1aa8a06dd62ceb71cc2eb/scripts/cmake/vcpkg_configure_make.cmake#L131-L143

Note that `vcpkg_configure_make` already has case-insensitive handling for everything but armv6.

ports/boinc/portfile.cmake
```diff
@@ -14,6 +14,14 @@ vcpkg_from_github(
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/CMakeLists.txt DESTINATION ${SOURCE_PATH})
 
 if(VCPKG_TARGET_IS_LINUX)
+    if (${VCPKG_TARGET_ARCHITECTURE} MATCHES "arm")
```
dg0yt, line 17: This line will also match `arm64` and `armv6`, so the remaining branches are unreachable.

ports/boinc/portfile.cmake
AenBleidd: It's handled in BOINC code

ports/boinc/portfile.cmake
AenBleidd: Fixed. @dg0yt, thanks for this finding!

ports/boinc/portfile.cmake
dg0yt: If the special case is only "armv6-linux in BOINC", why not override the autotools host triplet (i.e. pass `BUILD_TRIPLET`) just for armv6-linux?

AenBleidd: Because it's not only for armv6-linux. Other were not working either.
Here's an example: https://github.com/BOINC/boinc-apps/runs/4280686170?check_suite_focus=true

dg0yt: I see. Basically, cross-compiling support for linux on linux seems to be fairly unsupported in `vcpkg_configure_make`.
- `z_vcpkg_determine_autotools_target_cpu` is called only for `VCPKG_TARGET_IS_WINDOWS`.
- `arg_DETERMINE_BUILD_TRIPLET` is only checked for `VCPKG_TARGET_IS_WINDOWS` and `VCPKG_TARGET_IS_OSX`.

This is what needs to be fixed.

AenBleidd: This is out-of-scope of this PR, and looks like a bigger ticket that will require extensive testing.

dg0yt: > This is out-of-scope of this PR, and looks like a bigger ticket that will require extensive testing.

Fair enough.
But you should still prefer setting `BUILD_TRIPLET` over adding to `OPTIONS` in this case. This avoids eventually having two `--host` arguments to `configure`.
Or simply set `VCPKG_MAKE_BUILD_TRIPLET` in the arm* toolchain files.

AenBleidd: I don't understand what you are talking about. When I build this port for some particular architecture, I'm using a triplet that specifies VCPKG_TARGET_ARCHITECTURE. And this target is not used to pass -host parameter correctly.
Maybe you can give me an example of triplet and how it should be called? Maybe I'm just doing smth wrong while building this port.

dg0yt: > prefer setting `BUILD_TRIPLET` over adding to `OPTIONS` in this case. This avoids eventually having two `--host` arguments to `configure`. 

This refers to `ports/boinc/portfile.cmake`.

> Or simply set `VCPKG_MAKE_BUILD_TRIPLET` in the arm* toolchain files.

I meant tor refer vcpkg triplet files, not toolchain files. Basically, next to 
~~~
set(VCPKG_TARGET_ARCHITECTURE armv6)
~~~
add 
~~~
set(VCPKG_MAKE_BUILD_TRIPLET armv6-linux)
~~~

AenBleidd: I tried this now and it fails:
```
    Command failed: /usr/bin/bash -c "V=1 ./../src/18_-7.18.1-1910b9f200.clean/configure aarch64-linux \"--disable-server\" \"--disable-client\" \"--disable-manager\" \"--disable-silent-rules\" \"--verbose\" \"--disable-shared\" \"--enable-static\" \"--prefix=/home/runner/work/boinc-apps/boinc-apps/build/boinc-autodock-vina/arm64-linux-static/vcpkg_installed/arm64-linux-static\""
```
Should I use
```
set(VCPKG_MAKE_BUILD_TRIPLET -host=aarch64-linux)
```
instead of
```
```
set(VCPKG_MAKE_BUILD_TRIPLET aarch64-linux)
```
?
If yes - then it looks weird for me

dg0yt: Ah, indeed it takes the full option -- it goes directly to the `configure` call if set.

AenBleidd: ok. looks like full option is working for me.
Thanks for your help @dg0yt.
@JackBoosY, looks like this PR is not needed anymore

PR #21653, opened 2021-11-24
[poly2tri] Fix consuming the library due to incorrectly encoded header

Osyotr:
`sweep.h` was in ANSI codepage and some characters were illegal in utf-8

Actual build error when building `assimp` 5.1.1:
```
sweep.h(1): error C2220: the following warning is treated as an error
sweep.h(1): warning C4828: The file contains a character starting at offset 0x773 that is illegal in the current source character set (codepage 65001).
sweep.h(1): warning C4828: The file contains a character starting at offset 0x776 that is illegal in the current source character set (codepage 65001).
```

Tested build with x86-windows and x64-linux.


LilyWangLL (approved):

BillyONeal:
Since the replacement also contains mojibake would it be a good idea to remove the name entirely (leaving 'Thomas' and the email address?)

BillyONeal:
image

I sent the original author this message and want to give them a bit of time to respond before landing this. If we get no response in 1.5 weeks we should merge this as is.

BillyONeal:
image

I got a reply OK-ing removing the surname.

BillyONeal (approved):

Osyotr:
@BillyONeal about that mojibake, I think it was misinterpreted by github. New line in diff looks fine if opened as UTF8. Still, I prefer your change, thanks!

BillyONeal:
> @BillyONeal about that mojibake, I think it was misinterpreted by github. New line in diff looks fine if opened as UTF8. Still, I prefer your change, thanks!

Right, but compilers are likely to not assume UTF-8 :/

BillyONeal:
Thanks for the fix!

PR #21689, opened 2021-11-27
[open62541] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/open62541/portfile.cmake
+++ b/ports/open62541/portfile.cmake
@@ -37,4 +37,11 @@ file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/share/open62541/tools")
 
+if(EXISTS "${CURRENT_PACKAGES_DIR}/include/open62541/types_generated_handling.h")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/open62541/types_generated_handling.h" "${SOURCE_PATH}" "")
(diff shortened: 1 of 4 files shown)


ports/open62541/portfile.cmake
```diff
@@ -37,4 +37,11 @@ file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/share/open62541/tools")
 
+if(EXISTS "${CURRENT_PACKAGES_DIR}/include/open62541/types_generated_handling.h")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/open62541/types_generated_handling.h" "${SOURCE_PATH}" "")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/open62541/statuscodes.h" "${SOURCE_PATH}" "")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/open62541/nodeids.h" "${SOURCE_PATH}" "")
+    vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/open62541/types_generated.h" "${SOURCE_PATH}" "")
+endif()
```
JackBoosY, line 45: I found that this is only for the comments to be modified, it does not affect the user's use, so I don't think we need to add additional code to modify it.

We welcome the addition of detection for absolute paths to vcpkg, but is it possible to add a skip option to this function to avoid this modification?

JackBoosY (changes requested):

ports/open62541/portfile.cmake
autoantwort: Maybe we can make the parser smarter to detect comments

autoantwort: Ok this is implemented now

PR #21704, opened 2021-11-27
[libheif] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/libheif/portfile.cmake
+++ b/ports/libheif/portfile.cmake
@@ -18,8 +18,11 @@ vcpkg_cmake_install()
 vcpkg_copy_pdbs()
 
 vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/libheif/)
-vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/lib/pkgconfig/libheif.pc" "Requires.private" "Libs.private")
-vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig/libheif.pc" "Requires.private" "Libs.private")
+# libheif's pc file assumes libstdc++, which isn't always true.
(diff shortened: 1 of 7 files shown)


ports/libheif/portfile.cmake
```diff
@@ -18,6 +18,8 @@ vcpkg_cmake_install()
 vcpkg_copy_pdbs()
 
 vcpkg_cmake_config_fixup(CONFIG_PATH lib/cmake/libheif/)
+# dependency libde265 has no pkgconfig file, so we have to remove this pkgconfig files
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/pkgconfig" "${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig")
```
JackBoosY, line 22: Can we move `Requires.private` to `Libs.private` instead?

JackBoosY (changes requested):

ports/libheif/portfile.cmake
autoantwort: Yeah that works. It also seems to be that the libs are only used privately and not in a public interface.

JackBoosY (approved):

ras0219-msft (approved):
>Can we move Requires.private to Libs.private instead?

@JackBoosY This is _never_ the correct thing to do unless upstream has made a grave mistake. `Requires.private` means "There should exist a .pc file with this name", `Libs.private` means "There should exist a library with this name -- also this entry is suitable for being added to a linker command line". These are fundamentally different things and it never makes sense to try to use one as the other.

Additionally, private libs are still critical if users are static linking (such as in our default Linux triplets).

I've resolved this issue by adding installation of the .pc file to `libde256`.

Thanks for the PR!

PR #21715, opened 2021-11-28
[fluidsynth] no absolute paths

autoantwort:
For #20469


autoantwort:
arm64 fails with 
```
CMake Error at CMakeLists.txt:13 (project):
  Generator

    Ninja

  does not support platform specification, but platform

    Win32

  was specified.
```
I have no clue ...

BillyONeal (approved):

PR #21722, opened 2021-11-28
[sdl2pp] no absolute paths

autoantwort:
For #20469


ras0219-msft (approved):
Similar to the other PR, I added .pc support to the expected dependencies.

LGTM, thanks for the PR!

PR #21723, opened 2021-11-28
[pdal] no absolute paths

autoantwort:
For #20469

### DIFF
--- a/ports/pdal/portfile.cmake
+++ b/ports/pdal/portfile.cmake
@@ -47,6 +47,11 @@ vcpkg_install_cmake(ADD_BIN_TO_PATH)
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/pdal/cmake)
 vcpkg_copy_pdbs()
 
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/pkgconfig" "${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig") # dependency gdal has no pkgconfig files
+
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/pdal/Dimension.hpp" "${SOURCE_PATH}" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/pdal/pdal_features.hpp" "#define PDAL_PLUGIN_INSTALL_PATH \"${CURRENT_PACKAGES_DIR}/bin\"" "")
+
 # Install PDAL executable
 file(GLOB _pdal_apps ${CURRENT_PACKAGES_DIR}/bin/*.exe)
 file(COPY ${_pdal_apps} DESTINATION ${CURRENT_PACKAGES_DIR}/tools/pdal)
--- a/ports/pdal/usage
+++ b/ports/pdal/usage
@@ -1,0 +1,14 @@
+The package pdal provides CMake targets:
+
+    find_package(PDAL CONFIG REQUIRED)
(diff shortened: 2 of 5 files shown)


autoantwort:
Also depends on #20439

dg0yt:
Don't modernize outdated pdal too much. There is #21275.

autoantwort:
Ah sorry, I only changed vcpkg_configure_cmake to vcpkg_cmake_configure etc. and added some quotes. But I can revert this changes

ports/pdal/portfile.cmake
```diff
@@ -47,6 +47,11 @@ vcpkg_install_cmake(ADD_BIN_TO_PATH)
 vcpkg_fixup_cmake_targets(CONFIG_PATH lib/pdal/cmake)
 vcpkg_copy_pdbs()
 
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib/pkgconfig" "${CURRENT_PACKAGES_DIR}/debug/lib/pkgconfig") # dependency gdal has no pkgconfig files
+
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/pdal/Dimension.hpp" "${SOURCE_PATH}" "")
+vcpkg_replace_string("${CURRENT_PACKAGES_DIR}/include/pdal/pdal_features.hpp" "#define PDAL_PLUGIN_INSTALL_PATH \"${CURRENT_PACKAGES_DIR}/bin\"" "")
```
dg0yt, line 53: It is probaly safe to define PDAL_PLUGIN_INSTALL_PATH as `"."`.

ports/pdal/usage
```diff
@@ -0,0 +1,14 @@
+The package pdal provides CMake targets:
+
+    find_package(PDAL CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE pdalcpp pdal_util pdal_arbiter pdal_kazhdan)
+
+Add the following cmake code to your CMakeLists.txt to create definitions for DCMTK
```
dg0yt, line 6: It is PDAL, not DCMTK.

ports/pdal/usage
```diff
@@ -0,0 +1,14 @@
...
+    find_package(PDAL CONFIG REQUIRED)
+    target_link_libraries(main PRIVATE pdalcpp pdal_util pdal_arbiter pdal_kazhdan)
+
+Add the following cmake code to your CMakeLists.txt to create definitions for DCMTK
+    if ("${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}" STREQUAL "/")
+        message(WARNING "No vcpkg install dir found, PDAL definitions are not set")
+    else()
+        target_compile_definitions(main
+            PUBLIC
+                "PDAL_PLUGIN_INSTALL_PATH=${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/bin"
```
dg0yt, line 12: I doubt that `bin` is the right path for linux or osx. I would prefer to not add non-portable instructions.

ports/pdal/usage
dg0yt: In addition, I think it would be better to handle PDAL_PLUGIN_INSTALL_PATH in the config files.

dg0yt (changes requested):
How about putting this PR in draft state? I will take it as a check list for the PDAL update.

autoantwort:
> It is probaly safe to define PDAL_PLUGIN_INSTALL_PATH as ".".

Ok good, then the `usage` file is not needed.

PR #21725, opened 2021-11-28
[realsense2] add missing libusb dep on macOS

autoantwort:
For #20469


ports/realsense2/vcpkg.json
```diff
@@ -1,13 +1,13 @@
...
   "version": "2.49.0",
-  "port-version": 1,
+  "port-version": 2,
   "description": "Intel® RealSense™ SDK 2.0 is a cross-platform library for Intel® RealSense™ depth cameras (D400 series and the SR300).",
   "homepage": "https://github.com/IntelRealSense/librealsense",
   "dependencies": [
     {
       "name": "libusb",
-      "platform": "linux"
+      "platform": "!windows"
```
JonLiu1993, line 10: @autoantwort ,Thanks for your pr, my investigation found that libusb port supports " !uwp ", why is "! windows" could you clarify?

JonLiu1993 (changes requested):

ports/realsense2/vcpkg.json
autoantwort: Because this dependency is only needed on non windows platforms.

JonLiu1993 (approved):

ras0219-msft:
LGTM, thanks!

PR #21783, opened 2021-12-01
[vcpkg] Add special functions for wrappers

dg0yt:
This PR complements #21641 by an improved way for library lookup in cmake wrappers. If accepted, existing wrapper should be update for unified behaviour.

- #### What does your PR fix?
  ~The vcpkg cmake toolchain respects `VCPKG_PREFER_SYSTEM_LIBS` when modifying `CMAKE_PREFIX_PATH`, `CMAKE_LIBRARY_PATH` and `CMAKE_FIND_ROOT_PATH`.
  However, it is hard to respect this preference correctly in wrappers when looking specifically for debug or release variants of libraries. Current wrappers provide explicit paths and so completely igore `VCPKG_PREFER_SYSTEM_LIBS`. This is inconsistent with behaviour for ports which do not need wrappers.~
  In addition, it is difficult to deal with the fact that custom triplets may set a specific `VCPKG_BUILD_TYPE` for some ports.

  This PR adds two high level functions which can be used in wrappers before and after `_find_package`. Two low level functions are used instead of plain `find_library` in order to consistently respect  `VCPKG_PREFER_SYSTEM_LIBS`.

  As an example and test case, port gdal's wrapper is updated to use these functions.

  ~Limitations: With `VCPKG_PREFER_SYSTEM_LIBS` activated, variations in library names may still lead to inconsistencies (e.g. release variant from system, debug variant from vcpkg). However, the user can directly initialize the cache variables to system variants if needed.)~

  all, no

  yes

  yes


scripts/buildsystems/vcpkg.cmake
```diff
@@ -198,6 +198,42 @@ function(z_vcpkg_set_powershell_path)
...
+
+Find the variant of a library to be used in debug builds.
+This function is for vcpkg cmake wrappers.
+It ensures that <vcpkg>/debug/lib is searched before <vcpkg>/lib,
+regardless of current build type.
+#]===]
+function(z_vcpkg_find_library_debug)
+    foreach(var CMAKE_PREFIX_PATH CMAKE_LIBRARY_PATH CMAKE_FIND_ROOT_PATH)
+        list(FIND ${var} "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}" index)
+        math(EXPR index "${index} - 1")
```
dg0yt, line 230: Same here.

Neumann-A:
Is there a reason you use the complicated way instead of doing two `find_library(${ARGN})` calls with the first one being with `NO_DEFAULT_PATHS` and only the vcpkg path as `PATHS`?

dg0yt:
> Is there a reason you use the complicated way instead of doing two `find_library(${ARGN})` calls with the first one being with `NO_DEFAULT_PATHS` and only the vcpkg path as `PATHS`?

I hoped to explain it in the top post. Basically, I want to put the responsibility for correct vpckg library lookup into a single place, instead of scattering it over all wrappers.

Again taking the gdal example. We come from:
~~~
find_library(GDAL_LIBRARY_DEBUG
    NAMES gdal_d gdal_i_d gdal
    NAMES_PER_DIR
    PATHS "${CMAKE_CURRENT_LIST_DIR}/../../debug/lib"
    NO_DEFAULT_PATH)
~~~
Problem: Doesn't systematically respect VCPKG_PREFER_SYSTEM_LIBS. Doesn't systematically find the vcpkg lib when gdal is built only for release. (There will be another `find_library` call from `_find_package`.)

You propose (and I add `gdal_i` to cover all release lib names for gdal):
~~~
find_library(GDAL_LIBRARY_DEBUG
    NAMES gdal_d gdal_i_d gdal_i gdal
    NAMES_PER_DIR
    PATHS "${CMAKE_CURRENT_LIST_DIR}/../../debug/lib" "${CMAKE_CURRENT_LIST_DIR}/../../lib"
    NO_DEFAULT_PATH)
find_library(GDAL_LIBRARY_DEBUG
    NAMES gdal_d gdal_i_d gdal_i gdal
    NAMES_PER_DIR)
~~~
Problem: Doesn't systematically respect VCPKG_PREFER_SYSTEM_LIBS. Verbose syntax.

I propose:
~~~
z_vcpkg_find_library_debug(GDAL_LIBRARY_DEBUG
    NAMES gdal_d gdal_i_d gdal_i gdal
    NAMES_PER_DIR)
~~~
Problem: Needs helper functions. (This PR.)

... and everything duplicated for release, and done in every wrapper.

Neumann-A:
> You propose

No i mean that you do two `find_library` calls in `z_vcpkg_find_library_<config>` instead of manipulating the lists.

dg0yt:
> > You propose
> 
> No i mean that you do two `find_library` calls in `z_vcpkg_find_library_<config>` instead of manipulating the lists.

Okay.

This leaves the problem that it doesn't respect `VCPKG_PREFER_SYSTEM_LIBS`. 
Guarding the first call with `if(VCPKG_PREFER_SYSTEM_LIBS)` doesn't help, because then we fall back to the default search path. The default search path contains the debug-release search order for the consuming project, not for the desired library. So it is wrong either for release or for debug.

This search order is the problem which is reliably addressed by *locally* inserting the configuration-specific item before the first occurence of the wrong item. This construction doesn't need particular parameters to `find_library`, allowing to transparently forward all arguments to `find_library`.

Neumann-A:
> Guarding the first call with if(VCPKG_PREFER_SYSTEM_LIBS) doesn't help, because then we fall back to the default search path.

I would say people should use empty overlays in this case since it is the only way to not screw up the binary cache ...

dg0yt:
> I would say people should use empty overlays in this case since it is the only way to not screw up the binary cache ...

I tend to agree. Mixing system libraries with vcpkg invites undefined behaviour at runtime. But `VCPKG_PREFER_SYSTEM_LIBS` was merged nevertheless.
However, we can get rid of `VCPKG_PREFER_SYSTEM_LIBS` later, without changing the interface which is added in this PR.

Cheney-W:
@dg0yt Could you please handle the conflict?

dg0yt:
> @dg0yt Could you please handle the conflict?

Done. But this is a PR for discussion. Merge conflicts can be resolved after discussion.

Added two more functions. 
The test port is gdal because it does have some dependening ports. Note that the even the curent version of `FindGDAL.cmake` uses `GDAL_LIBRARY` as a cache variable (filepath) and doesn't use `select_library_configurations`.

ras0219-msft:
> Mixing system libraries with vcpkg invites undefined behaviour at runtime. But VCPKG_PREFER_SYSTEM_LIBS was merged nevertheless.

I think this (and linked issue) has exposed fundamental problems with VCPKG_PREFER_SYSTEM_LIBS. This suggests to me that we should instead deprecate it, emit a warning, and instruct users that they should use empty overlay ports to correctly stub out system libraries.

JackBoosY:
Ping @ras0219-msft for review again.

Cheney-W:
Pinging @dg0yt for response. Is work still being done for this PR?

Cheney-W:
@dg0yt Could you please handle the conflicts? Thank you!

dg0yt:
I tend to withdraw the propose special functions.
However, the deprecation of `VCPKG_PREFER_SYSTEM_LIBS` could be merged via a separate PR.

ras0219-msft:
If we consider defining `VCPKG_BUILD_TYPE` to `release` and then building in debug to be "UB", are these special functions still required? My understanding of the status quo (calling `find_library()` twice) is that the release library will be found and correctly used but the debug library will be one of {system copy, not found}. Then using that debug library is ambiguous and undefined for customers.

Adding these special functions helps to make this illegal scenario less ambiguous, but I'm not sure they're worth their weight in that scenario since there are no longer other policies they need to implement and regularize.

I would like to merge the `VCPKG_PREFER_SYSTEM_LIBS` deprecations you've put into this PR however. Would you be willing to shrink this PR to just those or open a different one?

dg0yt:
> > I tend to withdraw the proposed special functions.
> > However, the deprecation of `VCPKG_PREFER_SYSTEM_LIBS` could be merged via a separate PR.



> I would like to merge the `VCPKG_PREFER_SYSTEM_LIBS` deprecations you've put into this PR however. Would you be willing to shrink this PR to just those or open a different one?

-> #26449

PR #21788, opened 2021-12-01
[pcl] Add feature apps, visualization, simulation, examples and add usage

### LINKED ISSUE #15130: [pcl] Visualization headers are not installed
When pcl libraries are installed, I am trying to compile some samples found in the tutorials, but I cannot compile because the pcl/visualization headers cannot be found.

They do not exist in vcpkg/installed/include directory. Maybe they should be installed and the required components compiled.

JackBoosY:
Add the following features:
- apps
- visualization
- examples
- simulation

Add usage file since the usage provided by vcpkg is incorrect.

Fixes #15130.

Already test all features on x86-windows.

### DIFF
--- a/ports/pcl/install-examples.patch
+++ b/ports/pcl/install-examples.patch
@@ -1,0 +1,13 @@
+diff --git a/cmake/pcl_targets.cmake b/cmake/pcl_targets.cmake
+index 91f2404..18f5a24 100644
+--- a/cmake/pcl_targets.cmake
++++ b/cmake/pcl_targets.cmake
+@@ -473,6 +473,8 @@ macro(PCL_ADD_EXAMPLE _name)
+   # add target to list of example targets created at the parent scope
+   list(APPEND PCL_EXAMPLES_ALL_TARGETS ${_name})
+   set(PCL_EXAMPLES_ALL_TARGETS "${PCL_EXAMPLES_ALL_TARGETS}" PARENT_SCOPE)
++
++  install(TARGETS ${_name} RUNTIME DESTINATION ${BIN_INSTALL_DIR})
+ endmacro()
+ 
+ ###############################################################################
--- a/ports/pcl/portfile.cmake
+++ b/ports/pcl/portfile.cmake
@@ -32,16 +32,19 @@ endif()
 
 vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
     FEATURES
-        openni2 WITH_OPENNI2
-        qt      WITH_QT
-        pcap    WITH_PCAP
-        cuda    WITH_CUDA
-        cuda    BUILD_CUDA
-        cuda    BUILD_GPU
-        tools   BUILD_tools
-        opengl  WITH_OPENGL
-        vtk     WITH_VTK
-        libusb  WITH_LIBUSB
+        openni2         WITH_OPENNI2
+        qt              WITH_QT
+        pcap            WITH_PCAP
+        cuda            WITH_CUDA
+        cuda            BUILD_CUDA
+        cuda            BUILD_GPU
+        tools           BUILD_tools
+        opengl          WITH_OPENGL
+        vtk             WITH_VTK
+        libusb          WITH_LIBUSB
+        visualization   BUILD_visualization
+        examples        BUILD_examples
+        apps            BUILD_apps
 )
 
 vcpkg_cmake_configure(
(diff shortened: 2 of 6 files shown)


ports/pcl/vcpkg.json
```diff
@@ -96,6 +116,18 @@
...
+        {
+          "name": "pcl",
+          "default-features": false,
+          "features": [
+            "vtk"
+          ]
+        }
+      ]
+    },
     "vtk": {
```
JackBoosY, line 131: This feature will undoubtedly not work. To use visualization, we not only need to set `WITH_VTK`, but also `BUILD_visualization`.
Due to historical compatibility issues, I am considering whether to interchange this feature with visualization?

LilyWangLL (approved):

ports/pcl/install-examples.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/cmake/pcl_targets.cmake b/cmake/pcl_targets.cmake
```
JackBoosY, line 1: Only generate these examples in CMAKE_BINARY_DIR is not enough.
They also need the dependency libraries to execute.

ports/pcl/vcpkg.json
BillyONeal: We can make this feature depend on `pcl[core,visualization]` right?

BillyONeal (dismissed):
Thanks for the fixes! Can you confirm the change on line 117 of porfile.cmake?

ports/pcl/portfile.cmake
```diff
@@ -103,12 +114,13 @@ vcpkg_fixup_pkgconfig()
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
 file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/share")
 
-if("tools" IN_LIST FEATURES) 
+if(BUILD_tools OR BUILD_apps OR BUILD_examples)
```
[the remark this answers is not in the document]
dg0yt: `vcpkg_check_features` also sets the variables. Read: side effect. It is already used in many ports, often indirectly via `configure_file`.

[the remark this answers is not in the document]
dg0yt: Documentation:
> `vcpkg_check_features() will: ... add -D<feature-var>=ON to <out-var>, and set <prefix>_<feature-var> to ON.`

JackBoosY:
> We can make this feature depend on pcl[core,visualization] right?

@BillyONeal I think the name of this feature cannot completely replace the corresponding function.
Many features in pcl rely on vtk, but vtk does not necessarily refer to these features.

Since this involves modification history, we can only keep this feature, but maybe it is more appropriate to set it as an alias for visualization?

ports/pcl/portfile.cmake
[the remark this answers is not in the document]
JackBoosY: Yeah, I didn't know that before. But it support that now.
See:
https://github.com/microsoft/vcpkg/blob/63e935d967a3410e26bf4a708efa39d8384d2bbb/scripts/cmake/vcpkg_check_features.cmake#L198-L204

https://github.com/microsoft/vcpkg/blob/63e935d967a3410e26bf4a708efa39d8384d2bbb/scripts/cmake/vcpkg_check_features.cmake#L214-L220

JackBoosY:
> > We can make this feature depend on pcl[core,visualization] right?
> 
> @BillyONeal I think the name of this feature cannot completely replace the corresponding function. Many features in pcl rely on vtk, but vtk does not necessarily refer to these features.
> 
> Since this involves modification history, we can only keep this feature, but maybe it is more appropriate to set it as an alias for visualization?

@here-mikelley @raahilsha-z @UnaNancyOwen @claudiofantacci @fran6co What do you think about this?

ports/pcl/portfile.cmake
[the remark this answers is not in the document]
strega-nil: Yeah, much to my chagrin, this got added.

JackBoosY:
If there is no objection, I will convert vtk to alias tomorrow.

ports/pcl/vcpkg.json
strega-nil-ms: so my understanding is that `vtk` is not especially useful without `visualization`, but it's not broken, so I'm not against just having it; @JackBoosY do you have a different opinion on what we should do?

JackBoosY: As I disccused with @BillyONeal, here are my suggestions:
1. Feature vtk should only mean enabling dependent vtk, and should not involve features.
2. Since it has been added a long time ago, we should take care of the dependencies of other ports and user settings.

If we want to deal with the above two points at the same time, the best way is to use vtk as an alias for visualization.

JackBoosY: Ping @strega-nil-ms for reply, if you agree with me, I'll update this PR.

BillyONeal:
Thanks for the feature fixes!

PR #21852, opened 2021-12-04
[qjsonwebtoken] New port for Qt/JWT support

kafeg:
New port for Qt/JWT support: https://github.com/juangburgos/QJsonWebToken


cenit:
shall you export unofficial targets for downstream library consumption?

kafeg:
> shall you export unofficial targets for downstream library consumption?

Sorry I'm not sure i understand you question correctly. The lib for now supports only static build without extra exports for dlls and like that, so in this PR i marked it as static only.

cenit:
i mean in cmakelists, export targets and create a port config… so that downstream projects can do a find_package(qjsonwebtoken) and a target_link_library(…)

kafeg:
It's a very tiny lib, includes already in vcpkg includes dir and it's enough just to link lib itself to the project.

I notified the lib maintainer and if he will would like to change it to support full cmake build i think he can do that.

ports/qjsonwebtoken/vcpkg.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "qjsonwebtoken",
+  "version": "0.1.0",
+  "description": "Qt C++ implementation of JSON Web Token (JWT)",
+  "supports": "!uwp",
```
JackBoosY, line 5:
```suggestion
  "supports": "!uwp",
  "license": "MIT",
```

ports/qjsonwebtoken/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+set(PROJECT_VERSION "0.1.0")
+project(qjsonwebtoken VERSION 0.1.0 LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+find_package(Qt5Core REQUIRED)
+
+add_library( qjsonwebtoken STATIC src/QJsonWebToken src/qjsonwebtoken.h src/qjsonwebtoken.cpp )
```
JackBoosY, line 11: I think the author wants it to be used as source code, so here we should use `INTERFACE` instead of `STATIC`, and we should export this target for downstream use. See
https://github.com/microsoft/vcpkg/blob/030cfaa24de9ea1bbf0a4d9c615ce7312ba77af1/ports/tinyxml/CMakeLists.txt#L25-L40

ports/qjsonwebtoken/portfile.cmake
```diff
@@ -0,0 +1,24 @@
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
+
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO juangburgos/QJsonWebToken
+    REF master
+    SHA512 20f30d264ca5f761fcba27ddd05ed022bfef47a0e584683750a6eda23e3db12cdb4e8c8f95699e805328a42588e15aff8e5abe4554b4d99eb6c596cfa32b922e
+    HEAD_REF master
+    #PATCHES 
+    #    allow-build-static.patch
```
JackBoosY, line 10:
```suggestion
```

ports/qjsonwebtoken/portfile.cmake
```diff
@@ -0,0 +1,24 @@
+vcpkg_check_linkage(ONLY_STATIC_LIBRARY)
```
JackBoosY, line 1:
```suggestion
```

JackBoosY (changes requested):

kafeg:
Thank you guys for comments, review and samples. Now I understand what i should to do and made changes.

ports/qjsonwebtoken/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+set(HEADERS "src/QJsonWebToken" "src/qjsonwebtoken.h")
+set(SOURCES "src/qjsonwebtoken.cpp")
+
+add_library(unofficial-qjsonwebtoken ${SOURCES} )
+set_target_properties(unofficial-qjsonwebtoken PROPERTIES OUTPUT_NAME qjsonwebtoken)
+target_link_libraries(unofficial-qjsonwebtoken Qt5::Core)
+
+include(GenerateExportHeader)
+generate_export_header(unofficial-qjsonwebtoken BASE_NAME qjsonwebtoken)
```
JackBoosY, line 19: Maybe we should provide all the headers instead of this.

ports/qjsonwebtoken/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+target_include_directories(unofficial-qjsonwebtoken 
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
```
JackBoosY, line 23:
```suggestion
```
Since it's a interface library, we may not need this.

JackBoosY (changes requested):

ports/qjsonwebtoken/CMakeLists.txt
kafeg: I provided other headers just the same way as in tinyxml with HEADERS env var which used below in `install(FILES ${HEADERS} ...`

JackBoosY: I still think it's incorrect.
We should only provide the headers which upstream provided.
Any reason to do that?

kafeg: Please check it now)

ports/qjsonwebtoken/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+find_package(Qt5Core REQUIRED)
+
+set(HEADERS "src/QJsonWebToken" "src/qjsonwebtoken.h")
+set(SOURCES "src/qjsonwebtoken.cpp")
+
+add_library(unofficial-qjsonwebtoken ${SOURCES} )
```
JackBoosY, line 14:
```suggestion
add_library(unofficial-qjsonwebtoken INTERFACE ${SOURCES} )
```

JackBoosY (changes requested):

ports/qjsonwebtoken/CMakeLists.txt
kafeg: Can we avoid this? This change breaks the lib build.

kafeg: Sorry i definitely don't understand sense of this. Now lib builds staticaly and has all headers in right places. So, what exatly wrong there?

JackBoosY:
@kafeg Can you please test the usage?

kafeg:
> @kafeg Can you please test the usage?

It doesn't work in my case unfortunately in this config.

Anyway, i found a problem in this lib which breaks parsing of JWT tokens generated by another implimentations so it's not so usable for now. I propose to close and hold this until that will be fixed. As a good replacement i found in vcpkg port jwt-cpp.

JackBoosY:
Okay, feel free to reopen this PR.

PR #21920, opened 2021-12-08
[decimal-for-cpp] update to 1.18

josuegomes:
Update decimal-for-cpp to version 1.18


ras0219-msft:
LGTM, thanks for the update!

Issue #17750, opened 2021-05-08
[liblzma] build failure with wasm32-emscripten


bmanga:
**Host Environment**
- OS: Windows
- Compiler: emcc 2.0.20

**To Reproduce**
`./vcpkg install liblzma:wasm32-emscripten`

**Failure logs** 
```
[...]

Issue #17312, opened 2021-04-15
[boost] Failed when build external libtorrent that uses boost library.


hksdpc250:
**Describe the bug**
```
D:\vcpkg\installed\arm64-windows-static\include\boost/multiprecision/cpp_int/intel_intrinsics.hpp(103,11): error C3861:
 '_addcarry_u32': identifier not found
```

**Environment**
- OS: windows x64 cross compile to arm64
- Compiler: visual studio 2019

**To Reproduce**
1. ./vcpkg install --triplet=arm64-windows boost openssl
2. build latest libtorrent manually

**Expected behavior**
build success

**Failure logs** 
```
D:\libtorrent-rasterbar-2.0.3\build>vcpkg --triplet=arm64-windows-static install boost openssl
...................................
D:\libtorrent-rasterbar-2.0.3\build>cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE="D:\vcpkg\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET="arm64-windows-static" -DBUILD_SHARED_LIBS=OFF -Dstatic_runtime=ON ..
-- The C compiler identification is MSVC 19.28.29913.0
-- The CXX compiler identification is MSVC 19.28.29913.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: D:/MicrosoftVisualStudio/2019/Community/VC/Tools/MSVC/14.28.29910/bin/Hostx64/arm64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: D:/MicrosoftVisualStudio/2019/Community/VC/Tools/MSVC/14.28.29910/bin/Hostx64/arm64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE
-- Performing Test HAVE_CXX_ATOMICS_WITHOUT_LIB
-- Performing Test HAVE_CXX_ATOMICS_WITHOUT_LIB - Success
-- Performing Test HAVE_CXX_ATOMICS64_WITHOUT_LIB
-- Performing Test HAVE_CXX_ATOMICS64_WITHOUT_LIB - Success
[...]

PR #21928, opened 2021-12-08
[boost] update to 1.78.0

### LINKED ISSUE #14850: [boost-mpi] make boost-python optional
****

The `boost-python` dependency of `boost-mpi` is only needed to build the python bindings. As a lot of persons don't need this and building the `python3` package takes quite long this sould be a feature.

**Proposed solution**
keep the python bindings but move them to a feature

Non

**Additional context**

### LINKED ISSUE #16507: [boost-fiber] New feature boost-fiber[numa]
Library name:
Boost fiber (maybe name it boost-fiber[numa] ?)
Library description:
This is the boost-fiber library compiled with some additional flags to support non-uniform memory addressing.
Source repository URL:
https://github.com/boostorg/fiber

I've looked at https://github.com/boostorg/fiber/blob/develop/build/Jamfile.v2 . Seems like we need some minor changes in https://github.com/microsoft/vcpkg/blob/master/ports/boost-fiber/portfile.cmake to pass the numa flag, but I don't know how to do it.

### LINKED ISSUE #20718: [boost-asio] Make OpenSSL optional
We want to use boost-wintls and not openssl. So we don't need the openssl dependency.  

**Proposed solution**
Make OpenSSL a feature

### LINKED ISSUE #21474: [boost] [boost-mpi] [boost-python] Boost does not build for iOS, and probably Android as well
****

The problem is about Boost.Python not being able to build for iOS, and probably Android as well. It brings in Python3 as a dependency and that can only be built for the host OS.

**Proposed solution**
The solution is to exclude iOS (and Android) in 3 places - the Boost main manifest, the Boost.MPI manifest, both for dependencies, and finally in the Boost.Python manifest for supported platforms. Here are the relevant changes:

Boost (line 144) and Boost.MPI (line 27) dependencies in vcpkg.json:
    {
      "name": "boost-python",
      "platform": "!uwp & !(arm & windows) & !emscripten & !ios & !android"
    },

Boost.Python supported platforms (line 7) in vcpkg.json:
  "supports": "!uwp & !(arm & windows) & !emscripten & !ios & !android",

Currently using a private vcpkg registry with the modified Boost manifests. This results in successful iOS builds using vcpkg.

The check could list supported platforms explicitly instead of excluding unsupported platforms as they are found. E.g.:
"windows & !arm & !uwp | linux | osx"

**Additional context**
I have an open issue with the details of how Python3 fails to build for iOS. I doubt it's of much use since Python3 shouldn't be built for iOS in the first place, but here it is for reference: https://github.com/microsoft/vcpkg/issues/20214

A similar fix can be applied to the Python3 vcpkg.json manifest (line 6) as well:
  "supports": "!(arm | uwp | emscripten | ios | android)",

I notice the ARM condition here is not only about Windows, no idea how does that affect modern Apple Mac hardware since the new Macs now use ARM MacOS. It's a separate issue from my problem, but whoever is addressing may want to revisit it.

### LINKED ISSUE #21921: [boost] update to 1.78.0
Library name: boost

New version number: 1.78

Other information that may be useful: https://www.boost.org/users/history/version_1_78_0.html

yurybura:
Updates Boost to 1.78.0

Fixes #21921
Fixes #21474
Fixes #20718
Fixes #16507
Fixes #14850


dg0yt:
I recently learned that even in classic mode, vcpkg supports `>=` versioning in dependencies. Maybe this can be applied at least to `boost-modular-build-helper` dependencies to reduce the number of issues due to incomplete update attempts?
OTOH I wonder if the build helper could avoid hardcoding the boost version.

yurybura:
> I recently learned that even in classic mode, vcpkg supports `>=` versioning in dependencies. Maybe this can be applied at least to `boost-modular-build-helper` dependencies to reduce the number of issues due to incomplete update attempts? OTOH I wonder if the build helper could avoid hardcoding the boost version.

@dg0yt Can you give me an example? I'm not sure if I understood you correctly.

Kojoley:
@yurybura stacktrace and fiber regressions most likely due to https://github.com/bfgroup/b2/issues/104, try to include  https://github.com/bfgroup/b2/pull/113 patch.

Kojoley:
I see that uwp builds fail in boost-atomic but I don't understand how to obtain the actual build log that contains the error message or any useful diagnostic info about the failure.

yurybura:
> I see that uwp builds fail in boost-atomic but I don't understand how to obtain the actual build log that contains the error message or any useful diagnostic info about the failure.

@Kojoley  You can download logs here:
https://dev.azure.com/vcpkg/public/_build/results?buildId=64530&view=artifacts&pathAsName=false&type=publishedArtifacts
image

Kojoley:
Boost.Atomic cannot be built for UWP at least since Boost 1.74, I don't know why CI started failing only now. Filed the bug upstream https://github.com/boostorg/atomic/issues/54.

I guess you can safely add `"supports": "!uwp",` to `ports/boost-atomic/vcpkg.json` to fix the CI?

Kojoley:
Boost.Atomic cannot be built for UWP, most likely was broken by https://github.com/boostorg/atomic/commit/805e309a29e7bbbac01126884f368a1e56d669d3. Filed the bug upstream https://github.com/boostorg/atomic/issues/54.

I guess you can add `"supports": "!uwp",` to `ports/boost-atomic/vcpkg.json`?

Kojoley:
mapnik on arm64 fails due to boost-multiprecision, it looks like https://github.com/microsoft/vcpkg/issues/17312 was not actually fixed?
```
D:\installed\arm64-windows\include\boost/multiprecision/cpp_int/intel_intrinsics.hpp(118): error C3861: '_addcarry_u32': identifier not found
D:\installed\arm64-windows\include\boost/multiprecision/cpp_int/intel_intrinsics.hpp(123): error C3861: '_subborrow_u32': identifier not found
```

Kojoley:
> Boost.Atomic cannot be built for UWP, most likely was broken by [boostorg/atomic@805e309](https://github.com/boostorg/atomic/commit/805e309a29e7bbbac01126884f368a1e56d669d3). Filed the bug upstream [boostorg/atomic#54](https://github.com/boostorg/atomic/issues/54).
> 
> I guess you can add `"supports": "!uwp",` to `ports/boost-atomic/vcpkg.json`?

There is now a patch upstream that supposed to fix the issue https://github.com/boostorg/atomic/commit/bf08155d16ea8c359d1bfab6bac684cdfada4a5b

Kojoley:
multiprecision arm64 upstream patch https://github.com/boostorg/multiprecision/pull/406

yurybura:
@PhoebeHui FYI: Most of features have been tested with CI (see scripts/test_ports/vcpkg-ci-boost/vcpkg.json).

PhoebeHui:
> @PhoebeHui FYI: Most of features have been tested with CI (see scripts/test_ports/vcpkg-ci-boost/vcpkg.json).

I noticed that.

yurybura:
@Kojoley Could you review my changes? Thanks in advanced.

PhoebeHui (approved):
LGTM! thanks for your contribution!

ports/boost-fiber/fix-build-numa.patch
```diff
@@ -0,0 +1,15 @@
...
+
+diff --git a/build/Jamfile.v2 b/build/Jamfile.v2
+index 3efdb96..8a6b6dd 100644
+--- a/build/Jamfile.v2
++++ b/build/Jamfile.v2
+@@ -130,6 +130,7 @@ lib boost_fiber_numa
+       numa/algo/work_stealing.cpp
+     : 
+     <conditional>@numa
++    <library>boost_fiber
```
Kojoley, line 12: I don't see this patch upstream. The examples are explicitly linking to both libraries manually, but I agree that that autolinking a dependency is more user-fiendly and there seems to be no usages of boost_fiber_numa without boost_fiber.

https://github.com/boostorg/fiber/blob/5263ac1fe9385eaab24c8b297f9875cfaa482e7d/examples/numa/Jamfile.v2#L19-L20

Kojoley (approved):
Everything else looks good to me.

ports/boost-fiber/fix-build-numa.patch
yurybura: Without this patch library boost_fiber_numa cannot link on Windows:
```
   Creating library D:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a\boost_fiber_numa-vc142-mt-gd-x64-1_78.lib and object D:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a\boost_fiber_numa-vc142-mt-gd-x64-1_78.exp
work_stealing.obj : error LNK2019: unresolved external symbol "public: static class boost::fibers::context * __cdecl boost::fibers::context::active(void)" (?active@context@fibers@boost@@SAPEAV123@XZ) referenced in function "public: virtual class boost::fibers::context * __cdecl boost::fibers::numa::algo::work_stealing::pick_next(void)" (?pick_next@work_stealing@algo@numa@fibers@boost@@UEAAPEAVcontext@45@XZ)
work_stealing.obj : error LNK2019: unresolved external symbol "public: virtual __cdecl boost::fibers::context::~context(void)" (??1context@fibers@boost@@UEAA@XZ) referenced in function "public: virtual void * __cdecl boost::fibers::context::`vector deleting destructor'(unsigned int)" (??_Econtext@fibers@boost@@UEAAPEAXI@Z)
work_stealing.obj : error LNK2019: unresolved external symbol "public: void __cdecl boost::fibers::context::detach(void)" (?detach@context@fibers@boost@@QEAAXXZ) referenced in function "public: virtual void __cdecl boost::fibers::numa::algo::work_stealing::awakened(class boost::fibers::context *)" (?awakened@work_stealing@algo@numa@fibers@boost@@UEAAXPEAVcontext@45@@Z)
work_stealing.obj : error LNK2019: unresolved external symbol "public: void __cdecl boost::fibers::context::attach(class boost::fibers::context *)" (?attach@context@fibers@boost@@QEAAXPEAV123@@Z) referenced in function "public: virtual class boost::fibers::context * __cdecl boost::fibers::numa::algo::work_stealing::pick_next(void)" (?pick_next@work_stealing@algo@numa@fibers@boost@@UEAAPEAVcontext@45@XZ)
D:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a\boost_fiber_numa-vc142-mt-gd-x64-1_78.dll : fatal error LNK1120: 4 unresolved externals
...removing D:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a\boost_fiber_numa-vc142-mt-gd-x64-1_78.lib
...removing D:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a\boost_fiber_numa-vc142-mt-gd-x64-1_78.pdb
...skipped <pD:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\stage\lib>boost_fiber_numa-vc142-mt-gd-x64-1_78.dll for lack of <pD:\vcpkg\buildtrees\boost-fiber\x64-windows-dbg\boost\build\1cca5711dea309f44ef176a2abd1842a>boost_fiber_numa-vc142-mt-gd-x64-1_78.dll...
...failed updating 1 target...
```

Kojoley: That's indeed the case, reproducible with `b2 libs/fiber/build//boost_fiber_numa numa=on`, could you please file a bug upstream?

yurybura: Reported issue: https://github.com/boostorg/fiber/issues/288

ports/boost-modular-build-helper/boost-modular-build.cmake
```diff
@@ -79,6 +79,19 @@ function(boost_modular_build)
...
+        # Find Python2 in the current installed directory
+        file(GLOB python2_include_dir "${CURRENT_INSTALLED_DIR}/include/python2.*")
+        string(REGEX REPLACE ".*python([0-9\.]+).*" "\\1" python2_version "${python2_include_dir}")
+        string(REPLACE "." "" PYTHON_VERSION_TAG "${python2_version}")
+    endif()
+    if("python3" IN_LIST FEATURES)
+        # Find Python3 in the current installed directory
+        file(GLOB python3_include_dir "${CURRENT_INSTALLED_DIR}/include/python3.*")
+        string(REGEX REPLACE ".*python([0-9\.]+).*" "\\1" python3_version "${python3_include_dir}")
+        string(REPLACE "." "" PYTHON_VERSION_TAG "${python3_version}")
```
strega-nil, line 92: Are both python2 and python3 supported at the same time? This seems to write to the same variable, but elsewhere it looks like both at the same time are supported.

strega-nil:
I need confirmation on that one thing; otherwise, LGTM!

strega-nil (changes requested):
grr, clicked wrong button

ports/boost-modular-build-helper/boost-modular-build.cmake
yurybura: This part of code required to build `boost-mpi[python2]` or `boost-mpi[python3]`.
In preparing this PR, I had two problems.
First problem: `boost-mpi` doesn't build with both python versions at the same time.
Second problem: I didn't find a solution how to explicitly define both boost-python libraries (boost_python27-vc140.lib and boost_python310-vc140.lib) in the user-config.jam.

However `boost-python` itself supports both versions at the same time (`boost-python[python2,python3]`).

BillyONeal:
I'm concerned that this seems to be reverting https://github.com/microsoft/vcpkg/pull/14102 -- @stoperro can you check this and see that it won't break you?

ports/boost-beast/vcpkg.json
```diff
@@ -1,18 +1,16 @@
...
+  "version": "1.78.0",
   "description": "Boost beast module",
   "homepage": "https://github.com/boostorg/beast",
   "supports": "!emscripten",
   "dependencies": [
-    {
-      "name": "boost-asio",
-      "platform": "!emscripten"
-    },
+    "boost-asio",
```
BillyONeal, line 8: This is what looks like https://github.com/microsoft/vcpkg/pull/14102 revert

ports/boost-log/vcpkg.json
```diff
@@ -1,16 +1,13 @@
...
   "homepage": "https://github.com/boostorg/log",
   "supports": "!uwp & !emscripten",
   "dependencies": [
     "boost-align",
     "boost-array",
-    {
-      "name": "boost-asio",
-      "platform": "!emscripten"
-    },
+    "boost-asio",
```
BillyONeal, line 10: Ditto

ports/boost-modular-build-helper/boost-modular-build.cmake
BillyONeal: Given that this would create a new violation of https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-control-alternatives-in-published-interfaces I think we should always choose one python version here (probably 3).

BillyONeal (dismissed):
I would like to see the features-as-alternatives removed (choose Python 3 if upstream can't handle both) and want to give emscripten folks some time to confirm success.

Thank you so much for the update!

BillyONeal:
I confirmed that rerunning generate-ports.ps1 results in no changes

ports/boost-beast/vcpkg.json
yurybura: boost-asio is header only library. I think it shouldn't be excluded for emscripten.

yurybura: > I would like to see the features-as-alternatives removed (choose Python 3 if upstream can't handle both) and want to give emscripten folks some time to confirm success.
> 
> Thank you so much for the update!

@BillyONeal 
I think there are no alternatives in published interfaces.

Example 1: 
boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
boost-mpi[python2] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python27-vc140.lib)
boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
boost-mpi[python2,python3] - Build failed.

Example 2: 
boost-python[core] - Build failed (at least one Python version must be specified).
boost-python[python2] - Ok (produces 1 library: boost_python27-vc140.lib)
boost-python[python3] - Ok (produces 1 library: boost_python310-vc140.lib)
boost-python[python2,python3] - Ok (produces 2 libraries: boost_python27-vc140.lib +boost_python310-vc140.lib)

Most of users needs only Boost+Python3 to build. Some users need only Boost+Python2. And some users (like me) needs both versions to build. I hope boost-mpi[python2,python3] will be fixed later.

yurybura: P.S. Actually, I need boost-python27, boost-python37, boost-python38, boost-python39 and boost-python310 ...
Boost.Python supports this, but vcpkg does not support different Python versions.

yurybura:
> I would like to see the features-as-alternatives removed (choose Python 3 if upstream can't handle both) and want to give emscripten folks some time to confirm success.
> 
> Thank you so much for the update!

@BillyONeal
I think there are no alternatives in published interfaces.

Example 1:
boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
boost-mpi[python2] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python27-vc140.lib)
boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
boost-mpi[python2,python3] - Build failed.

Example 2:
boost-python[core] - Build failed (at least one Python version must be specified).
boost-python[python2] - Ok (produces 1 library: boost_python27-vc140.lib)
boost-python[python3] - Ok (produces 1 library: boost_python310-vc140.lib)
boost-python[python2,python3] - Ok (produces 2 libraries: boost_python27-vc140.lib +boost_python310-vc140.lib)

Most of users needs only Boost+Python3 to build. Some users need only Boost+Python2. And some users (like me) needs both versions to build. I hope boost-mpi[python2,python3] will be fixed later.

yurybura:
P.S. Actually, I need boost-python27, boost-python37, boost-python38, boost-python39 and boost-python310 ...
Boost.Python supports this, but vcpkg does not support different Python versions.

ports/boost-beast/vcpkg.json
stoperro: Hmm, I think I indeed had a reason why I disabled boost-asio it for Emscripten, maybe some of its dependencies didn't compile...
I will check soon the status and if it's still needed, as I didn't have to recompile it for a while.

yurybura: @stoperro openssl was removed from asio default dependencies.

BillyONeal:
>@BillyONeal
>I think there are no alternatives in published interfaces.
>Example 1:
>boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
>boost-mpi[python2] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python27-vc140.lib)
>boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
>boost-mpi[python2,python3] - Build failed.

This is exactly an alternative. Only one of python2 or python3 can be selected. The rule that matters here is that if `port[a]` installs, and `port[b]` installs, then `port[a,b]` must also install (since this is generally what happens when transitive dependencies are involved).

As a result I think one feature or the other must be removed. (Of course I prefer keeping the python3 one)

yurybura:
> > @BillyONeal
> > I think there are no alternatives in published interfaces.
> > Example 1:
> > boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
> > boost-mpi[python2] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python27-vc140.lib)
> > boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
> > boost-mpi[python2,python3] - Build failed.
> 
> This is exactly an alternative. Only one of python2 or python3 can be selected. The rule that matters here is that if `port[a]` installs, and `port[b]` installs, then `port[a,b]` must also install (since this is generally what happens when transitive dependencies are involved).
> 
> As a result I think one feature or the other must be removed. (Of course I prefer keeping the python3 one)

Could I keep this unchanged and report a new issue: failed to build boost-mpi[python2,python3]?
I don't want to dalay this PR due to this issue.
I still try to explain that is a port issue, not an alternative.

ports/boost-beast/vcpkg.json
stoperro: In the past I had also different issues, but now indeed after removal of openssl, the boost-asio build works for me in this PR :)

I see #17750 issue but that is of course unrelated to this PR, and with hack mentioned there I can now build Boost for latest Emscripten on this PR (`vcpkg install boost:wasm32-emscripten --binarysource=clear`).

yurybura:
> > > @BillyONeal
> > > I think there are no alternatives in published interfaces.
> > > Example 1:
> > > boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
> > > boost-mpi[python2] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python27-vc140.lib)
> > > boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
> > > boost-mpi[python2,python3] - Build failed.
> > 
> > 
> > This is exactly an alternative. Only one of python2 or python3 can be selected. The rule that matters here is that if `port[a]` installs, and `port[b]` installs, then `port[a,b]` must also install (since this is generally what happens when transitive dependencies are involved).
> > As a result I think one feature or the other must be removed. (Of course I prefer keeping the python3 one)
> 
> Could I keep this unchanged and report a new issue: failed to build boost-mpi[python2,python3]? I don't want to dalay this PR due to this issue. I still try to explain that is a port issue, not an alternative.

@BillyONeal Could you answer to my question? Related Boost.Mpi issues: https://github.com/boostorg/mpi/issues/112 and https://github.com/boostorg/mpi/issues/120

ports/boost-modular-build-helper/boost-modular-build.cmake
yurybura: I think this is not alternative, this is Boost.Mpi issue: boostorg/mpi#112

BillyONeal:
> > Could I keep this unchanged and report a new issue: failed to build boost-mpi[python2,python3]? I don't want to dalay this PR due to this issue. I still try to explain that is a port issue, not an alternative.
> 
> @BillyONeal Could you answer to my question? Related Boost.Mpi issues: [boostorg/mpi#112](https://github.com/boostorg/mpi/issues/112) and [boostorg/mpi#120](https://github.com/boostorg/mpi/issues/120)

If I understand correctly this is PR is adding the problem and adding it at all creates the backcompat concern; I would be happier to remove the python2 feature and create an issue saying to add python2 back if/when upstream resolves that issue

yurybura:
> > > Could I keep this unchanged and report a new issue: failed to build boost-mpi[python2,python3]? I don't want to dalay this PR due to this issue. I still try to explain that is a port issue, not an alternative.
> > 
> > 
> > @BillyONeal Could you answer to my question? Related Boost.Mpi issues: [boostorg/mpi#112](https://github.com/boostorg/mpi/issues/112) and [boostorg/mpi#120](https://github.com/boostorg/mpi/issues/120)
> 
> If I understand correctly this is PR is adding the problem and adding it at all creates the backcompat concern; I would be happier to remove the python2 feature and create an issue saying to add python2 back if/when upstream resolves that issue

@BillyONeal Thank you for your opinion. The last my comment related to this question.

If I remove python2 we will have: 
boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.lib)
boost-mpi[python2] - N/A
boost-mpi[python3] - Ok (produces 2 libraries: boost_mpi-vc140.lib + boost_mpi_python310-vc140.lib)
boost-mpi[python2,python3] - N/A.

I don't want to remove `boost-mpi[python2]` because it works fine. But if I haven't other choice I will remove it).

BillyONeal:
> > If I understand correctly this is PR is adding the problem and adding it at all creates the backcompat concern; I would be happier to remove the python2 feature and create an issue saying to add python2 back if/when upstream resolves that issue
> 
> @BillyONeal Thank you for your opinion. The last my comment related to this question.
> 
> If I remove python2 we will have: boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.dll) boost-mpi[python2] - N/A boost-mpi[python3] - Ok (produces 3 libraries: boost_mpi-vc140.dll + boost_mpi_python310-vc140.dll + mpi.pyd) boost-mpi[python2,python3] - N/A.
> 
> I don't want to remove `boost-mpi[python2]` because it works fine. But if I have no other choice, I will remove it)

To be clear, the reason for the concern here is not the immediate customer who might want both, it's that as soon as there is another port that depends on one or the other transitively, that explodes in customers' face that have never heard of `boost-mpi` and just wanted `google-wants-python2` and `aws-wants-python3` in the same program. That experience is bad for the user because it explodes for something they've never heard of and can't resolve. Moreover, once something is listed in the catalog we functionally can't ever remove it, so if upstream chooses to never fix it, we're stuck with unhappy customers. That's why this is an alternative as far as vcpkg cares; that's what transitive dependencies see it as. We want customers to expect that different parts of the catalog work together, and have been suggesting overlays for folks who want to change the situation. For example here you could leave all the code for the python2 feature in `portfile.cmake` and just drop it from `vcpkg.json` for now so it'd be super easy to turn back on.

I asked @strega-nil to confirm.

Kojoley:
Python 2 EOL was 2 years ago, is it really worth to support it?

yurybura:
> Python 2 EOL was 2 years ago, is it really worth to support it?

Boost.Python still support it.

Kojoley:
> > Python 2 EOL was 2 years ago, is it really worth to support it?
> 
> Boost.Python still supports it.

Boost.Python is in lethargic state for half a decade, adding Python 2 support in vcpkg makes it even harder for Boost to justify intentionally dropping the support.

yurybura:
> Boost.Python is in lethargic state for half a decade, adding Python 2 support in vcpkg makes it even harder for Boost to justify intentionally dropping the support.

IMHO. Python 2 support has been already implemented. After removing Python2 from Boost.Python or removing Boost.Python from Boost, we can remove it in vcpkg.

strega-nil-ms:
> > > If I understand correctly this is PR is adding the problem and adding it at all creates the backcompat concern; I would be happier to remove the python2 feature and create an issue saying to add python2 back if/when upstream resolves that issue
> > 
> > 
> > @BillyONeal Thank you for your opinion. The last my comment related to this question.
> > If I remove python2 we will have: boost-mpi[core] - Ok (produces 1 library: boost_mpi-vc140.dll) boost-mpi[python2] - N/A boost-mpi[python3] - Ok (produces 3 libraries: boost_mpi-vc140.dll + boost_mpi_python310-vc140.dll + mpi.pyd) boost-mpi[python2,python3] - N/A.
> > I don't want to remove `boost-mpi[python2]` because it works fine. But if I have no other choice, I will remove it)
> 
> To be clear, the reason for the concern here is not the immediate customer who might want both, it's that as soon as there is another port that depends on one or the other transitively, that explodes in customers' face that have never heard of `boost-mpi` and just wanted `google-wants-python2` and `aws-wants-python3` in the same program. That experience is bad for the user because it explodes for something they've never heard of and can't resolve. Moreover, once something is listed in the catalog we functionally can't ever remove it, so if upstream chooses to never fix it, we're stuck with unhappy customers. That's why this is an alternative as far as vcpkg cares; that's what transitive dependencies see it as. We want customers to expect that different parts of the catalog work together, and have been suggesting overlays for folks who want to change the situation. For example here you could leave all the code for the python2 feature in `portfile.cmake` and just drop it from `vcpkg.json` for now so it'd be super easy to turn back on.
> 
> I asked @strega-nil to confirm.

For this specific PR, I agree with Billy; just leave the support in the portfile and remove it from the mainline, since python2 is EOL; if someone wants python2 support, they can use overlays.

strega-nil-ms (changes requested):
Please remove python2 support from the vcpkg.json; if someone wants the EOL version of python, they can use an overlay.

yurybura:
> Please remove python2 support from the vcpkg.json; if someone wants the EOL version of python, they can use an overlay.

Done.

BillyONeal (approved):

BillyONeal:
Thanks!

PhoebeHui (approved):

strega-nil-ms:
Thanks so much @yurybura :)

PR #21951, opened 2021-12-09
docs: CMake toolchain needs `vcpkg integrate install` first

jktjkt:
Perhaps this is obvious to someone who has been using this tool for a long time, but as a newcomer, I read all these sections as an either-or: if you use MSBuild, do this, and if you use CMake, you're good to go just using the `.toolchain` file. Well, that's not the case, CMake's `find_path` et al won't find the *installed* packages, they also have to be *integrated* like this.

- #### What does your PR fix?  
    Instructions on how to use `vcpkg` with CMake were incomplete.

    Not applicable.

  Yup.

  Not applicable.

### DIFF
--- a/docs/examples/installing-and-using-packages.md
+++ b/docs/examples/installing-and-using-packages.md
@@ -90,7 +90,7 @@ To remove the integration for your user, you can use `.\vcpkg integrate remove`.
 #### CMake (Toolchain File)
 
 The best way to use installed libraries with cmake is via the toolchain file `scripts\buildsystems\vcpkg.cmake`. To use this file, you simply need to add it onto your CMake command line as:  
-`-DCMAKE_TOOLCHAIN_FILE=D:\src\vcpkg\scripts\buildsystems\vcpkg.cmake`.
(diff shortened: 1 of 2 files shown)


autoantwort:
Afaik this is not needed

docs/examples/installing-and-using-packages.md
```diff
@@ -90,7 +90,7 @@ To remove the integration for your user, you can use `.\vcpkg integrate remove`.
 #### CMake (Toolchain File)
 
 The best way to use installed libraries with cmake is via the toolchain file `scripts\buildsystems\vcpkg.cmake`. To use this file, you simply need to add it onto your CMake command line as:  
-`-DCMAKE_TOOLCHAIN_FILE=D:\src\vcpkg\scripts\buildsystems\vcpkg.cmake`.
+`-DCMAKE_TOOLCHAIN_FILE=D:\src\vcpkg\scripts\buildsystems\vcpkg.cmake`.  Be sure to run `vcpkg integrate install` first.
```
Neumann-A, line 93: That is incorrect. Integration is not needed. instead make sure you really set it the first time you run configure......

PR #21955, opened 2021-12-10
[backward] new port

csw:
**Describe the pull request**

This adds a port for the [backward](https://github.com/bombela/backward-cpp) library for printing nice stack traces.

- #### What does your PR fix?  
  Fixes this port not existing.

  All, according to the project page. No.

  I think so.

  Yes

### DIFF
--- a/ports/backward/portfile.cmake
+++ b/ports/backward/portfile.cmake
@@ -1,0 +1,23 @@
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO bombela/backward-cpp
+  REF 3bb9240cb15459768adb3e7d963a20e1523a6294 # v1.6
+  SHA512 5f3998202e87e43ae60d3b3de70e8d48284284135183a85aa59aad25c77eea1f6664b37b65131b383270dfa51cfe205ea5b1707b922015ed88771073ad8e79e4
+  HEAD_REF master
+  )
+
+vcpkg_cmake_configure(
+    SOURCE_PATH "${SOURCE_PATH}"
+)
+
+vcpkg_cmake_install()
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/backward)
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
+
+# # Handle copyright
+file(INSTALL "${SOURCE_PATH}/LICENSE.txt" DESTINATION "${CURRENT_PACKAGES_DIR}/share/backward" RENAME copyright)
--- a/ports/backward/vcpkg.json
+++ b/ports/backward/vcpkg.json
@@ -1,0 +1,17 @@
+{
+  "name": "backward",
+  "version": "1.6",
+  "description": "A beautiful stack trace pretty printer for C++",
+  "homepage": "https://github.com/bombela/backward-cpp",
+  "supports": "linux | osx | android | emscripten | (windows & (x86 | x64) & !uwp)",
+  "dependencies": [
+    {
+      "name": "vcpkg-cmake",
+      "host": true
+    },
+    {
+      "name": "vcpkg-cmake-config",
+      "host": true
+    }
+  ]
+}
--- a/versions/b-/backward.json
+++ b/versions/b-/backward.json
@@ -1,0 +1,9 @@
+{
+  "versions": [
+    {
+      "git-tree": "7b6f97f995696b1192d9904a67c72a1a843d6bdc",
+      "version": "1.6",
+      "port-version": 0
+    }
+  ]
+}
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -392,6 +392,10 @@
       "baseline": "2.0.0.1",
       "port-version": 0
     },
+    "backward": {
+      "baseline": "1.6",
+      "port-version": 0
+    },
     "basisu": {
       "baseline": "1.11",
       "port-version": 5


JonLiu1993:
@csw ,Thanks for your pr, please sign CLA first

ports/backward/portfile.cmake
```diff
@@ -0,0 +1,23 @@
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO bombela/backward-cpp
+  REF 3bb9240cb15459768adb3e7d963a20e1523a6294 # v1.6
+  SHA512 5f3998202e87e43ae60d3b3de70e8d48284284135183a85aa59aad25c77eea1f6664b37b65131b383270dfa51cfe205ea5b1707b922015ed88771073ad8e79e4
+  HEAD_REF master
+  )
+
+vcpkg_cmake_configure(
+    SOURCE_PATH ${SOURCE_PATH}
```
JonLiu1993, line 10:
```suggestion
    SOURCE_PATH "${SOURCE_PATH}"
```

ports/backward/portfile.cmake
```diff
@@ -0,0 +1,23 @@
...
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/backward)
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
+
+# # Handle copyright
+file(INSTALL ${SOURCE_PATH}/LICENSE.txt DESTINATION ${CURRENT_PACKAGES_DIR}/share/backward RENAME copyright)
```
JonLiu1993, line 23:
```suggestion
file(INSTALL "${SOURCE_PATH}/LICENSE.txt" DESTINATION "${CURRENT_PACKAGES_DIR}/share/backward" RENAME copyright)
```

ports/backward/portfile.cmake
```diff
@@ -0,0 +1,23 @@
...
+)
+
+vcpkg_cmake_install()
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug/include")
+
+vcpkg_cmake_config_fixup(CONFIG_PATH lib/backward)
+
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/debug")
+file(REMOVE_RECURSE "${CURRENT_PACKAGES_DIR}/lib")
```
JonLiu1993, line 20: @csw ,Is “backwards” a header-only library? Can you clarify why you want to delete the lib folder and Debug folder

JonLiu1993 (changes requested):

ports/backward/portfile.cmake
csw: It is header-only, yes. (There’s one .cpp file but it’s optional and meant to be copied into a project if desired, to run a static initializer.) The lib and debug folders were empty and vcpkg gave warnings about that; I didn’t see how to prevent them from being created, so I copied in the suggested code for deleting them.

JackBoosY: According to the upstream code https://github.com/bombela/backward-cpp/blob/master/CMakeLists.txt#L86:
```cmake
add_library(backward ${libtype} backward.cpp)
```
I think this port is not a header only library.
However, it doesn't install the generated library.

csw: It's an unusual case, as described [in the README](https://github.com/bombela/backward-cpp#install-backwardcpp); the .cpp file doesn't implement anything declared in the headers, it only registers signal handlers at static initialization time. That could be convenient in some cases, but IMO it's not really suitable as default behavior; I've only ever used it header-only and registered the signal handlers myself. (See the note in there about the automatic signal handler setup interacting badly with Folly, for instance.) So I think the upstream behavior of not including that in the CMake library target `Backward::Backward`, and having that just include the header and link with the necessary external libraries, is the right thing for most users. We could add an optional vcpkg feature to make it actually link with backward.cpp, if desired, which would follow upstream's header-only-by-default behavior.

ports/backward/portfile.cmake
csw: Thanks, I've fixed these. I took them verbatim from the [template](https://github.com/microsoft/vcpkg/blob/5c54cc06554e450829d72013c4b9e4baae41529a/scripts/templates/portfile.in.cmake) used by `vcpkg create`; should I open a PR to add the quotes there?

csw:
Also, according to upstream this is meant to build on all platforms, but I see that CI is failing for some exotic Windows platforms. It looks like the pipelines are capturing the actual failure logs [as artifacts](https://dev.azure.com/vcpkg/public/_build/results?buildId=64438&view=artifacts&pathAsName=false&type=publishedArtifacts), but I can't seem to view them, and I'm not in a position to try building things for Windows myself. Should I update the CI baseline to show that it's expected to fail on arm64_windows, arm_uwp, and x64_uwp, or try to investigate this further?

dg0yt:
You can download the failure log artifacts. If you move the mouse pointer over a line, there is a three-dot menu on the right which has the download action.

csw:
Oh, I see, thanks. OK, apparently the code only anticipates Windows running on x86 or x64, and uses API calls like RtlCaptureContext that are apparently not available on UWP; I'll mark those triplets as unsupported.

ports/backward/portfile.cmake
JonLiu1993: @csw ,No ,we have dedicated colleagues in charge of these

JonLiu1993:
@csw ,Have you investigated these error logs?
https://dev.azure.com/vcpkg/public/_build/results?buildId=64438&view=artifacts&pathAsName=false&type=publishedArtifacts

csw:
@JonLiu1993 yes, those were how I figured out that the library doesn't support Windows on ARM or UWP. I'll list those platforms as unsupported.

ports/backward/portfile.cmake
csw: @JackBoosY thoughts on this? I wouldn't be able to use this port myself if it registered these signal handlers automatically, and it'd be both surprising behavior for a library and a significant departure from the way it works upstream. So I think it's best to leave it as is.

scripts/ci.baseline.txt
```diff
@@ -52,6 +52,9 @@ azure-storage-cpp:arm64-windows=fail
 
 aubio:arm-uwp=fail
 aubio:x64-uwp=fail
+backward:arm64-windows=fail
+backward:arm-uwp=fail
+backward:x64-uwp=fail
```
JonLiu1993, line 57: @csw ,Thanks for your in-depth investigation. If these triplet are not supported by upstream, we should use the ‘supports’ field in vcpkg.json to indicate that it should not be added here

JonLiu1993 (changes requested):

scripts/ci.baseline.txt
csw: OK, will do. Should I leave these entries in the CI baseline file as well?

JonLiu1993: No, we only need to indicate it in vcpkg.json

ports/backward/portfile.cmake
JackBoosY: I think we should first confirm this with upstream before deciding whether to modify this place.
Can you please open an issue about that?

Thanks.

JonLiu1993: @csw ,Can you please open an issue about that?

JonLiu1993:
@csw ,Can you consider this suggestion？
https://github.com/microsoft/vcpkg/pull/21955#discussion_r771181270

JonLiu1993:
@csw ,Ping for response!

JonLiu1993:
Draft this PR since there is no response.

JonLiu1993:

PR #21967, opened 2021-12-10
[qt5-websockets] Option to disable qt5-decalarative dependency

kafeg:
[qt5-websockets] Option to disable qt5-decalarative dependency

### DIFF
--- a/ports/qt5-websockets/vcpkg.json
+++ b/ports/qt5-websockets/vcpkg.json
@@ -7,7 +7,17 @@
     {
       "name": "qt5-base",
       "default-features": false
-    },
-    "qt5-declarative"
-  ]
+    }
+  ],
+  "default-features": [
+    "declarative"
+  ],
+  "features": {
+    "declarative": {
+      "description": "Build qt5-websockets with qt5-declarative support and provide data-types for QML",
+      "dependencies": [
+        "qt5-declarative"
+      ]
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


Neumann-A:
The way it is currently implemented breaks binary caching. You need to pass an option to qmake to toogle/enforce the correct behavior. (but don't ask me which one it is)

kafeg:
Everything what i found is condition in `buildtrees\qt5-websockets\src\5.15.2-d3bc637515.clean\src\src.pro` file :
```
TEMPLATE = subdirs
CONFIG += ordered

qtConfig(textcodec) {
    SUBDIRS += websockets
    qtHaveModule(quick): SUBDIRS += imports
}
```
It builds just a plugin for QML if Qt:Quick module available. And if not then it will not build a plugin.

But not in all cases `Qt:Quick` and all it deps needed to build app project.

So i just propose to make that dependency optional but enabled by default for all and if developer has a special case he can disable it for his risk.

Osyotr:
I think `qt5-websockets/src/src.pro` should be patched (replace qtHaveModule with some other condition and set this condition from vcpkg).

JackBoosY:

Neumann-A:
> is not so good in my opinion

not patching it is breaking binary caching and thus introduces inconsistencies. Meaning building a/b is not the same as building b/a. This is something vcpkg can not have. Patching is required in this case. (there is a reason i introduced such features in Qt6 and not Qt5.)

kafeg:
Ok, then i will just decline this. Will think about patch later. Thank you.

PR #22012, opened 2021-12-13
[qwt|vcpkg-qmake] Update and switch to Qt6

Neumann-A:
(other related PR which was revert #21998; reopen after this has been solved.)


ports/qtbase/vcpkg.json
```diff
@@ -16,6 +16,14 @@
       "host": true,
       "default-features": false
     },
+    {
+      "name": "qtbase",
```
Osyotr, line 20: I'm curious, how does this work? Does this just make `doubleconversion` feature mandatory on windows?

Neumann-A: Yes it does.

JackBoosY:
Ready for review now?

ports/qwt/portfile.cmake
```diff
@@ -1,38 +1,33 @@
...
-        fix-dynamic-static.patch
+    URL "https://git.code.sf.net/p/qwt/git"
+    REF "a9ac6b28ee990f5d51ea36523057a5af54875e2e"
+    PATCHES 
+        config.patch
+        fix_dll_install.patch
 )
 
-vcpkg_configure_qmake(
+string(COMPARE EQUAL  "${VCPKG_LIBRARY_LINKAGE}" "dynamic" IS_STATIC)
```
JackBoosY, line 10: Can you double confirm this? It seems this is not correct.

JackBoosY (changes requested):

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
```diff
@@ -0,0 +1,87 @@
...
+    else()
+        set(items release)
+    endif()
+    foreach(build_type IN ITEMS ${items})
+        set(current_installed_prefix "${CURRENT_INSTALLED_DIR}${path_suffix_${build_type}}")
+
+        vcpkg_add_to_path(PREPEND "${current_installed_prefix}/lib" "${current_installed_prefix}/bin")
+
+        # We set LD_LIBRARY_PATH ENV variable to allow executing Qt tools (rcc,...) even with dynamic linking
+        if(CMAKE_HOST_UNIX)
```
JackBoosY, line 73: Can we use:
```suggestion
        if(NOT VCPKG_HOST_IS_WINDOWS)
```

ports/vcpkg-qmake/vcpkg_qmake_configure.cmake
```diff
@@ -119,26 +132,37 @@ function(vcpkg_qmake_configure)
...
         file(REMOVE_RECURSE "${CURRENT_BUILDTREES_DIR}/${config_triplet}")
 
+        set(qmake_comp_flags "")
+        # Note sure about these. VCPKG_QMAKE_OPTIONS offers a way to opt out of these. (earlier values being overwritten by later values; = set +=append *=append unique -=remove)
+        macro(qmake_add_flags qmake_var operation flags)
+            string(STRIP "${flags}" striped_flags)
+            if(striped_flags)
+                vcpkg_list(APPEND qmake_comp_flags "${qmake_var}${operation}${striped_flags}")
+            endif()
+        endmacro()
```
JackBoosY, line 142: Can we move this out of this function?

ports/vcpkg-qmake/vcpkg_qmake_fix_makefiles.cmake
```diff
@@ -0,0 +1,31 @@
...
+function(vcpkg_qmake_fix_makefiles BUILD_DIR)
+    #Fix the installation location
+    file(TO_NATIVE_PATH "${CURRENT_INSTALLED_DIR}" NATIVE_INSTALLED_DIR)
+    file(TO_NATIVE_PATH "${CURRENT_PACKAGES_DIR}" NATIVE_PACKAGES_DIR)
+    
+    if(WIN32)
+        string(SUBSTRING "${NATIVE_INSTALLED_DIR}" 2 -1 INSTALLED_DIR_WITHOUT_DRIVE)
+        string(SUBSTRING "${NATIVE_PACKAGES_DIR}" 2 -1 PACKAGES_DIR_WITHOUT_DRIVE)
+        string(SUBSTRING "${NATIVE_INSTALLED_DIR}" 0 2 INSTALLED_DRIVE)
+        string(SUBSTRING "${NATIVE_PACKAGES_DIR}" 0 2 PACKAGES_DRIVE)
```
JackBoosY, line 11: Ahh magic numbers, can you please add some comments here?

ports/vcpkg-qmake/vcpkg_qmake_install.cmake
```diff
@@ -0,0 +1,46 @@
...
+## Parameters:
+See [`vcpkg_qmake_build()`](vcpkg_qmake_build.md).
+
+## Notes:
+This command transparently forwards to [`vcpkg_qmake_build()`](vcpkg_qmake_build.md).
+and appends the 'install' target
+
+#]===]
+
+function(vcpkg_qmake_fix_prl PACKAGE_DIR PRL_FILES)
```
JackBoosY, line 20: If we need to provide this function externally, we need to use `cmake_parse_arguments`, otherwize it should have prefix `z_`.

JackBoosY (changes requested):

ports/vcpkg-qmake/vcpkg_qmake_fix_makefiles.cmake
Neumann-A: The comments are the variable names. There is nothing to add.

ports/vcpkg-qmake/vcpkg_qmake_configure.cmake
Neumann-A: why? macros are local to function scope and I don't think it will be needed elsewhere.

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
Neumann-A: done

ports/vcpkg-qmake/vcpkg_qmake_install.cmake
Neumann-A: added z_ prefix

JackBoosY (dismissed):

MehdiChinoune:
any progress?

JackBoosY:
@strega-nil-ms Any results?

JackBoosY:
cc @BillyONeal

ports/qtbase/vcpkg.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "qtbase",
   "version": "6.3.0",
-  "port-version": 2,
+  "port-version": 3,
```
BillyONeal, line 4: Whoops, I fail at merge conflicts.

Neumann-A:
@BillyONeal: Why did you try to solve it? I mean the PR is still tagged `depends:vcpkg-team`. this PR probably still needs other changes from https://github.com/Neumann-A/my-vcpkg-ports/tree/master/vcpkg-qmake or https://github.com/Neumann-A/my-vcpkg-ports/tree/master/qwt

BillyONeal:
> @BillyONeal: Why did you try to solve it?

I got pinged by @MehdiChinoune 's comment, saw PR with merge conflicts, so tried to fix. Feel free to force push them off.

JackBoosY:
Remove label vcpkg-team, please ping me if this PR is ready for review or needs some help.

Neumann-A:
@JackBoosY: This PR just needs a decision on how to transition qwt to qt6.

LilyWangLL (approved):

LilyWangLL (dismissed):

Neumann-A:
Should the qwt-qt6 port be removed ?

BillyONeal:
> Should the qwt-qt6 port be removed ?

Hmmmm... I didn't realize there was already a split for this port. Does it make sense to leave this one as qt5 (and consider it morally part of qt5) and instead change callers to name `qwt-qt6`?

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
```diff
@@ -0,0 +1,81 @@
...
+    # Make sure that the linker finds the libraries used
+    vcpkg_backup_env_variables(VARS PATH LD_LIBRARY_PATH)
+
+    if(CMAKE_HOST_WIN32)
+        if (VCPKG_QMAKE_USE_NMAKE)
+            find_program(NMAKE nmake)
+            set(invoke_command "${NMAKE}")
+            get_filename_component(nmake_exe_path "${NMAKE}" DIRECTORY)
+            vcpkg_host_path_list(APPEND ENV{PATH} "${nmake_exe_path}")
+        else()
```
BillyONeal, line 39:
```suggestion
            vcpkg_host_path_list(APPEND ENV{PATH} "${nmake_exe_path}")
            set(ENV{CL} "$ENV{CL} /MP${VCPKG_CONCURRENCY}")
        else()
```

This line is in `vcpkg_build_qmake`, do you know why it was removed here?

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
```diff
@@ -0,0 +1,81 @@
...
+        vcpkg_list(SET targets ${targets_${build_type}} ${arg_TARGETS})
+        if(NOT arg_SKIP_MAKEFILES)
+            z_run_jom_build("${invoke_command}" qmake_all makefiles "${short_name_${build_type}}")
+            z_vcpkg_qmake_fix_makefiles("${CURRENT_BUILDTREES_DIR}/${TARGET_TRIPLET}-${short_name_${build_type}}")
+        endif()
+        z_run_jom_build("${invoke_command}" "${targets}" "${arg_BUILD_LOGNAME}" "${short_name_${build_type}}")
+
+        vcpkg_restore_env_variables(VARS PATH LD_LIBRARY_PATH)
+    endforeach()
+endfunction()
```
BillyONeal, line 81:
```suggestion
    endforeach()
    vcpkg_restore_env_variables(VARS CL _CL_)
endfunction()
```

ports/vcpkg-qmake/vcpkg_qmake_install.cmake
```diff
@@ -0,0 +1,46 @@
...
+
+## Notes:
+This command transparently forwards to [`vcpkg_qmake_build()`](vcpkg_qmake_build.md).
+and appends the 'install' target
+
+#]===]
+
+function(z_vcpkg_qmake_fix_prl PACKAGE_DIR PRL_FILES)
+        file(TO_CMAKE_PATH "${PACKAGE_DIR}/lib" CMAKE_LIB_PATH)
+        file(TO_CMAKE_PATH "${PACKAGE_DIR}/include/qt5" CMAKE_INCLUDE_PATH)
```
BillyONeal, line 22: Why qt5 here?

ports/vcpkg-qmake/z_vcpkg_qmake_fix_makefiles.cmake
```diff
@@ -0,0 +1,31 @@
...
+    else()
+        set(INSTALLED_DRIVE)
+        set(PACKAGES_DRIVE)
+        set(INSTALLED_DIR_WITHOUT_DRIVE ${NATIVE_INSTALLED_DIR})
+        set(PACKAGES_DIR_WITHOUT_DRIVE ${NATIVE_PACKAGES_DIR})
+    endif()
+
+    file(TO_NATIVE_PATH "${CURRENT_INSTALLED_DIR}" NATIVE_INSTALLED_DIR)
+    file(TO_NATIVE_PATH "${CURRENT_PACKAGES_DIR}" NATIVE_PACKAGES_DIR)
+
```
BillyONeal, line 21:
```suggestion
```

Duplicates lines 3 and 4?

ports/vcpkg-qmake/z_vcpkg_qmake_fix_makefiles.cmake
```diff
@@ -0,0 +1,31 @@
...
+    file(TO_NATIVE_PATH "${CURRENT_INSTALLED_DIR}" NATIVE_INSTALLED_DIR)
+    file(TO_NATIVE_PATH "${CURRENT_PACKAGES_DIR}" NATIVE_PACKAGES_DIR)
+    
+    file(GLOB_RECURSE MAKEFILES "${BUILD_DIR}/**Makefile**")
+
+    foreach(MAKEFILE ${MAKEFILES})
+        file(READ "${MAKEFILE}" _contents)
+        #Set the correct install directory to packages
+        string(REPLACE "${INSTALLED_DRIVE}$(INSTALL_ROOT)${INSTALLED_DIR_WITHOUT_DRIVE}" "${PACKAGES_DRIVE}$(INSTALL_ROOT)${PACKAGES_DIR_WITHOUT_DRIVE}" _contents "${_contents}")
+        file(WRITE "${MAKEFILE}" "${_contents}")
```
BillyONeal, line 28:
```suggestion
        #Set the correct install directory to packages
        vcpkg_replace_string("${MAKEFILE}"
            "${INSTALLED_DRIVE}$(INSTALL_ROOT)${INSTALLED_DIR_WITHOUT_DRIVE}"
            "${PACKAGES_DRIVE}$(INSTALL_ROOT)${PACKAGES_DIR_WITHOUT_DRIVE}")
```

This whole operation is a bit suspect to me given that it still leaves absolute paths being installed?

ports/qwt/portfile.cmake
```diff
@@ -1,38 +1,34 @@
...
     SOURCE_PATH "${SOURCE_PATH}"
-    OPTIONS
-        CONFIG+=${VCPKG_LIBRARY_LINKAGE}
+    QMAKE_OPTIONS
+        ${OPTIONS}
+        "CONFIG-=debug_and_release"
+        "CONFIG+=create_prl"
+        "CONFIG+=link_prl"
+        #QWT_CONFIG += QwtDesigner
+        #QWT_CONFIG += QwtPkgConfig
```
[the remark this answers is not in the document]
Neumann-A: `QWT_CONFIG += QwtDesigner` builds the QtDesigner plugin. Could be activated via introducing a feature in the port. (But let me recheck the deps for it. I think the qttools dep can go away without this set.)

`QWT_CONFIG += QwtPkgConfig` is deactivated since Qt6 does not yet have pc files so Qwt cannot build pc files itself. (hmm probably needs a recheck if Qt6 can now install pc files. There was an issue on Jira about it but that is a different PR to enable pkg-config [probably with Qt6.4].)

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
Neumann-A: because of https://github.com/microsoft/vcpkg/blob/90b5fb836cda4eba4569a123bca63b957f55232b/scripts/toolchains/windows.cmake#L54

maybe the `${VCPKG_CONCURRENCY}` setting should be correctly moved into the toolchain

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
Neumann-A: connected to https://github.com/microsoft/vcpkg/pull/22012/files#r1024041827

ports/vcpkg-qmake/vcpkg_qmake_install.cmake
Neumann-A: good catch needs to be Qt6. Since there is no qmake consumer of qwt i didn't notice it.

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
```diff
@@ -0,0 +1,81 @@
+#[===[.md:
+# vcpkg_qmake_build
+
+Build a qmake-based project, previously configured using vcpkg_configure_qmake.
```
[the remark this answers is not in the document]
Neumann-A: wait the docs belong somewhere else now.,

Neumann-A:
relevant for the qwt-qt6 discussion: #21825 #20921.
@BillyONeal Did you just forgot to remove the v db entry for qwt-qt6 or was that on purpose?

ports/qwt/portfile.cmake
```diff
@@ -1,38 +1,34 @@
-vcpkg_from_sourceforge(
+vcpkg_from_git(
```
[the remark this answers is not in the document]
BillyONeal: >Qwt had a few trivial port important bugfixes since the release over a year ago.

OK, I just wanted to make sure it wasn't seen as a 'style change'. If the content is meaningfully different then that's OK.

>But I actually should bump the version. Can I use 6.2.0-20220616 as a version? (Ref should be updated to 06d6822b595b70c9fd567a4fe0d835759bf271fe)

Yes, that should work, but I don't think it has the right semver connotation since a -dashed thing means 'prerelease'. This seems closer to a +plused thing?

```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>
```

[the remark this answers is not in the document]
Neumann-A: does the version resolution inside vcpkg even care about - and + being correct ?

ports/qwt/portfile.cmake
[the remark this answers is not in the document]
Neumann-A: rechecked qtbase. pc files are only provided if it is a shared build & !windows. So no triplet currently generates pc files for qt6

ports/vcpkg-qmake/vcpkg_qmake_build.cmake
BillyONeal: Makes sense, thank you!

ports/qwt/portfile.cmake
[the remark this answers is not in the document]
BillyONeal: > does the version resolution inside vcpkg even care about - and + being correct ?

We will sort dashed things before undashed things for purposes of minimum version constraints, yes.

BillyONeal (dismissed):
I'm not confident in the qt specific stuff here so I've given the other maintainers notice that I intend to merge this if nobody objects today.

ports/qwt/portfile.cmake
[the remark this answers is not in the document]
BillyONeal: Sorry, to clarify, the question isn't "why commented vs not commented", but "why commented vs not existing".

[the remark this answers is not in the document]
Neumann-A: So I am reminded that those options exist if they are ever needed ;). But I'll simply remove them. The `QWT_CONFIG+=QwtDll` and the config patch should fullfil the same purpose.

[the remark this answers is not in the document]
BillyONeal: haha OK I was willing to merge as is, but appreciate checking in less commented out code :). Thanks!

BillyONeal:
Thanks!

mitza-oci:
What was the resolution of https://github.com/microsoft/vcpkg/pull/20921#issuecomment-985276177 relative to this PR?  In particular, since vcpkg continues to support qt5 as separate port, what should we do in an application that's using qt5 and qwt?

BillyONeal:
> What was the resolution of [#20921 (comment)](https://github.com/microsoft/vcpkg/pull/20921#issuecomment-985276177) relative to this PR? In particular, since vcpkg continues to support qt5 as separate port, what should we do in an application that's using qt5 and qwt?

You would need to use overlays. We accepted the switch to qt6 because current versions of qt5 are no longer available from the qt project to open source customers.

mitza-oci:
I'm not sure I follow - Qt5 source is available, see https://wiki.qt.io/Building_Qt_5_from_Git#Getting_the_source_code

Since qt5 and qt6 are different ports, could there be a feature in qwt to select which one to use?  A major advantage to using vcpkg instead of system package managers is that this decision is local to the application.  We don't need to make a decision that every application on the system needs the qt5 version of libqwt, but this one does.

BillyONeal:
>I'm not sure I follow - Qt5 source is available, see https://wiki.qt.io/Building_Qt_5_from_Git#Getting_the_source_code

There was more than a month where the current Qt5 was not available, although I see they may have just fixed it 8 days ago according to https://code.qt.io/cgit/qt/qt5.git/log/?h=5.15

Notably, 5.15 will be the last qt5 release and open source customers no longer receive LTS ( https://www.qt.io/blog/qt-offering-changes-2020 ), meaning open source consumption support by the qt project ended 2021-05-26.

Between the unavailability (at the time) of current qt5 sources, the public statements by the qt project that such use is no longer in support, and our general policy that the canonical version available in the catalog is that selected by upstream project owners, switching vcpkg's canonical version to qt6 was long past due.

>Since qt5 and qt6 are different ports, could there be a feature in qwt to select which one to use? 

Features are additive and cannot be used to control "alternatives" like this. (For example, it is always assumed that enabling a feature is 'safe' which is not true when only one may be selected)

That is, if `port[a]` works, and `port[b]` works, `port[a,b]` must work. See https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#do-not-use-features-to-implement-alternatives

>A major advantage to using vcpkg instead of system package managers is that this decision is local to the application

You can still do that. You just need to do it with an overlay port.

>We don't need to make a decision that every application on the system needs the qt5 version of libqwt, but this one does.

vcpkg needs to make a decision on which versions of dependencies we select in the curated catalog, and test. We are only testing one outcome of what `vcpkg install qwt` does.

mitza-oci:
Thanks for the additional info, it's certainly helpful.  I'm not all that tuned-in to what Qt upstream has changed in their policies.

In terms of how vcpkg handles this, it seems like the ideal state would be to have a single `qt` port and not both `qt5` and `qt` (the latter being v6).  However, it's reasonable that for a large framework like Qt there will be some downstream users that need to stay on v5 during a transition period, and that's what we see now realized in vcpkg (although it's not the ideal state).  In this case would a separate port `qwt-qt5` be accepted?  It would need to use different directory names in `include` and `share` and a different library name.  Downstream projects would need to add some CMake code to adapt.

PR #22038, opened 2021-12-15
[hiredis] Fix static build

kafeg:
Hey guys. This PR https://github.com/microsoft/vcpkg/pull/21914/files breaked static build on Linux and i think on macOS too.

Catched it after updating my vcpkg to the latest version. So build for me looks like this now:

image

And if you will check the patch `support-static-in-win.patch` then you can see this patch required on all platforms because by default in `hiredis` present `add_library(... SHARED ...)` which changed by vcpkg patch.

ldd of my app after build with latest vcpkg:
image


BillyONeal:
Thanks for the fix!

plutonea:
Гавно, переделывай

kafeg:
7D7BA60D-75DB-4EBC-9796-F93F7AAFA9CE

Let me share this hero 😁

PR #22082, opened 2021-12-18
[grpc] Turn on codegen feature automatically when needed

### LINKED ISSUE #21514: [GRPC] x64-windows-static does not build grpc++_reflection.lib
**Problem Description** 
When GRPC **x64-windows-static** is installed from VCPkg, it does not build several libraries that are built under the default `x64-windows` configuration including: `grpc++_reflection.lib`, `grpcpp_channelz.lib`

Release and debug copies of these exist on my system in the non-static build:
`C:\local\vcpkg\installed\x64-windows\lib\`
but **not** in the static build:
`C:\local\vcpkg\installed\x64-windows-static\lib\`

Subsequently, when CMake is invoked with the flag `-DVCPKG_TARGET_TRIPLET=x64-windows-static` on a project that uses GRPC++_Reflection, including on the included RouteGuide example project, CMake fails with error (below). 

- Comparing the two folders I see several missing libraries but not sure what other GRPC functionality might be impacted. 
- Also `x86-windows` - someone reported this in April in an unanswered Stackoverflow. [Does grpc++_reflection library come with vcpkg install grpc for x86-windows?](https://stackoverflow.com/questions/67054872/does-grpc-reflection-library-come-with-vcpkg-install-grpc-for-x86-windows)
- [edit] Looks like the similar x86 issue was addressed here: https://github.com/microsoft/vcpkg/issues/17247 but they don't mention 'static' configurations.

**Host Environment**
- OS: Windows 10 Pro
- Compiler: MSVC 19.28.29337.0

**To Reproduce**
First:
```
vcpkg install grpc:x64-windows-static
vcpkg install protobuf protobuf:x64-windows-static
vcpkg install --recurse protobuf[zlib] protobuf[zlib]:x64-windows-static
vcpkg integrate install
```
Next: Build example project and watch failure
```
cd vcpkg\buildtrees\grpc\src\2180080eb4-87c05d756b.clean\examples\cpp\route_guide
cmake -DCMAKE_TOOLCHAIN_FILE=C:/local/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static .
```

Optional Step:
I tried before and after adding this to the top of the RouteGuide project's CMakeLists.txt, replacing the exising `cmake_minimum_required` line. 
```
# CMAKE_MSVC_RUNTIME_LIBRARY flag requires CMAKE >= v3.15
cmake_minimum_required(VERSION 3.15)
cmake_policy(SET CMP0091 NEW)
set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```

**Failure logs** 
Repeated many times: 
```
CMake Error at C:/local/vcpkg/scripts/buildsystems/vcpkg.cmake:540 (_add_executable):
  Target "route_guide_client" links to target "gRPC::grpc++_reflection" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
  CMakeLists.txt:72 (add_executable)
```
(full console output attached)
[grpc++_reflection_not_found.txt](https://github.com/microsoft/vcpkg/files/7563634/grpc%2B%2B_reflection_not_found.txt)

**Additional context**
We're building a new GRPC client app to consume the libraries we develop and want to build everything fully statically-linked and `/MT`. 

```
> vcpkg list
abseil:x64-windows                                 20210324.2#1     an open-source collection designed to augment th...
abseil:x64-windows-static                          20210324.2#1     an open-source collection designed to augment th...
c-ares:x64-windows                                 1.17.2#1         A C library for asynchronous DNS requests
c-ares:x64-windows-static                          1.17.2#1         A C library for asynchronous DNS requests
grpc:x64-windows                                   1.41.0           An RPC library and framework
grpc:x64-windows-static                            1.41.0           An RPC library and framework
grpc[codegen]:x64-windows                                           Build code generator machinery
openssl:x64-windows                                1.1.1l#3         OpenSSL is an open source project that provides ...
openssl:x64-windows-static                         1.1.1l#3         OpenSSL is an open source project that provides ...
protobuf:x64-windows                               3.18.0           Protocol Buffers - Google's data interchange format
protobuf:x64-windows-static                        3.18.0           Protocol Buffers - Google's data interchange format
protobuf:x86-windows                               3.18.0           Protocol Buffers - Google's data interchange format
protobuf[zlib]:x64-windows-static                                   ZLib based features like Gzip streams
protobuf[zlib]:x86-windows                                          ZLib based features like Gzip streams
re2:x64-windows                                    2021-11-01       RE2 is a fast, safe, thread-friendly alternative...
re2:x64-windows-static                             2021-11-01       RE2 is a fast, safe, thread-friendly alternative...
upb:x64-windows                                    2021-10-19       μpb (often written 'upb') is a small protobuf i...
upb:x64-windows-static                             2021-10-19       μpb (often written 'upb') is a small protobuf i...
vcpkg-cmake-config:x64-windows                     2021-11-01
vcpkg-cmake:x64-windows                            2021-09-13
zlib:x64-windows                                   1.2.11#13        A compression library
zlib:x64-windows-static                            1.2.11#13        A compression library
zlib:x86-windows                                   1.2.11#13        A compression library
```

ykadowak:
This PR changes to turn on codegen feature when it is not an actual cross compiling like `Host: x64-windows, Target: x64-windows-static`.

- #### What does your PR fix?  
  Fixes #21514 

  Yes

  Yes

### DIFF
--- a/ports/grpc/portfile.cmake
+++ b/ports/grpc/portfile.cmake
@@ -23,8 +23,18 @@ vcpkg_from_github(
         00014-pkgconfig-upbdefs.patch
 )
 
-if(NOT TARGET_TRIPLET STREQUAL HOST_TRIPLET)
-    vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/grpc")
+if(VCPKG_CROSSCOMPILING)
+    if(NOT VCPKG_CMAKE_SYSTEM_NAME 
+       AND (NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm") 
+       AND (NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64"))
+        # If both are windows on x86-64, turn on codegen feature automatically since it's not a crosscompiling.
+        # This is required because VCPKG_CROSSCOMPILING only checks if the triplets are different
+        # and does not care about the situations like HOST:x64-windows and TARGET:x64-windows-static
+        list(APPEND FEATURES codegen)
+    else()
+        # Add the path of generated plugins for host OS
+        vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/grpc")
+    endif()
 endif()
 
 string(COMPARE EQUAL "${VCPKG_CRT_LINKAGE}" "static" gRPC_MSVC_STATIC_RUNTIME)
--- a/ports/grpc/vcpkg.json
+++ b/ports/grpc/vcpkg.json
@@ -1,6 +1,7 @@
 {
   "name": "grpc",
   "version-semver": "1.41.0",
+  "port-version": 1,
   "description": "An RPC library and framework",
   "homepage": "https://github.com/grpc/grpc",
   "dependencies": [
--- a/versions/baseline.json
+++ b/versions/baseline.json
@@ -2578,7 +2578,7 @@
     },
     "grpc": {
       "baseline": "1.41.0",
-      "port-version": 0
+      "port-version": 1
     },
     "grppi": {
       "baseline": "0.4.0",
--- a/versions/g-/grpc.json
+++ b/versions/g-/grpc.json
@@ -1,6 +1,11 @@
 {
   "versions": [
     {
+      "git-tree": "e59d0419c509ef15306cfd3c8ae9b390edf94a2f",
+      "version-semver": "1.41.0",
+      "port-version": 1
+    },
+    {
       "git-tree": "5e3795bf43afe5243e47e18ac66a28bc722207f5",
       "version-semver": "1.41.0",
       "port-version": 0


ports/grpc/portfile.cmake
```diff
@@ -23,8 +23,18 @@ vcpkg_from_github(
...
+       AND (NOT VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64"))
+        # If both are windows on x86-64, turn on codegen feature automatically since it's not a crosscompiling.
+        # This is required because VCPKG_CROSSCOMPILING only checks if the triplets are different
+        # and does not care about the situations like HOST:x64-windows and TARGET:x64-windows-static
+        list(APPEND FEATURES codegen)
+    else()
+        # Add the path of generated plugins for host OS
+        vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/grpc")
+    endif()
 endif()
```
JackBoosY, line 38: The depedendcy has feature grpc when crossbuilding, so we don't need to modify here.
```

**Review by JackBoosY — CHANGES_REQUESTED** [review]

File: ports/grpc/portfile.cmake (hunk 1 above)
Comment (yusuke-kadowaki) [reply ports/grpc/portfile.cmake:38]: Is it possible for you to explain more specifically?
I could not get if you are saying there is another way to fix this or this issue does not have to be fixed.

Comment (JackBoosY) [reply ports/grpc/portfile.cmake:38]: Please note that the keyword `host` only applies to the build process, not to the use process.
If the users needs to use codegen in their custom project, the users must add the `codegen` feature during installation.

Comment (Sand2Silicon) [reply ports/grpc/portfile.cmake:38]: JackBoosY, I don't think I agree with that. 
1) The default triplet x64-windows includes the 'codegen' feature automatically, where the 'x64-windows-static' does not. It seems "codegen" is required just to compile the GRPC examples and anything based on them, but if you don't have it, the compiler error message is about Reflection. It takes a lot of digging into GRPC and vcpkg sources to realize that codegen is what provide Reflection.

Comment (JackBoosY) [reply ports/grpc/portfile.cmake:38]: If the host tool is only used at build time, it should be marked "host", otherwise it should be a default feature.

Comment (yusuke-kadowaki) [reply ports/grpc/portfile.cmake:38]: @JackBoosY 
>Please note that the keyword host only applies to the build process, not to the use process.
If the users needs to use codegen in their custom project, the users must add the codegen feature during installation.

What do you mean by the `build process`, `use process`, and `installation`?
For example, what is `./vcpkg.exe install grpc:x64-windows-static` in one of those?

In most cases, the `codegen` feature should be on because, as @Sand2Silicon mentioned, most programs using grpc is dependant on those features. Only in some special cross compiling cases, you need to turn it off (see https://github.com/grpc/grpc/issues/16856). So I think it's more reasonable to make it default and turn it off only for those cases. OR we can at least turn it on for obvious cases like my fix.

File: ports/grpc/portfile.cmake [hunk 2]
```diff
@@ -23,8 +23,18 @@ vcpkg_from_github(
         00014-pkgconfig-upbdefs.patch
 )
 
-if(NOT TARGET_TRIPLET STREQUAL HOST_TRIPLET)
-    vcpkg_add_to_path(PREPEND "${CURRENT_HOST_INSTALLED_DIR}/tools/grpc")
+if(VCPKG_CROSSCOMPILING)
+    if(NOT VCPKG_CMAKE_SYSTEM_NAME
```
JackBoosY, line 27:
```suggestion
    if( VCPKG_TARGET_IS_WINDOWS
```

JackBoosY (changes requested):

ports/grpc/portfile.cmake
JackBoosY: 1. For building this port in a cross-triplet, if the build process needs to use this tool, we must add this feature as a host dependency in vcpkg.json.
2. For custom projects that need to use this tool, the user should use the feature name to declare and build it, instead of enabling it by default in portfile.cmake.

JackBoosY:
Draft this PR since there is no response for a long time.

JackBoosY:

PR #22106, opened 2021-12-20
[LuaBridge] update to 3.x

### LINKED ISSUE #22097: [LuaBridge] update to 3.x
Library name: LuaBridge

New version number: 3.x

LuaBridge 3.x is also a header-only library, which can be easily introduced into vcpkg.
https://github.com/kunitoki/LuaBridge3

JonLiu1993:
Fix #22097 
Update  LuaBridge to the latest version 3.0-beta1

Note: no feature need to test


JackBoosY:
This is not the official repo and we shouldn't accept this.

